PR #182, opened 2014-12-12
Move to Github organization

vhda:
I am aware that we are still not using github's organization, so I'm basically using this PR as a way of asking for status information this migration.
In issue #101 we arrived at the conclusion that we should keep "Exuberant Ctags" name.

PR #183, opened 2014-12-12
Add a `dist` make target to create a tarball

b4n:
One possible approach for easily creating a tarball.  It creates tarballs of the Git contents and names it after the current git tag or revision.
See #182.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -351,7 +351,8 @@ check: $(CTAGS_TEST)
 #
 dist:
 	git rev-parse  # this will bail out if not in a Git repository
-	rev=$$(git tag --points-at HEAD); \
+	rev=$(REV); \
+	test -n "$$rev" || rev=$$(git tag --points-at HEAD); \
 	test -n "$$rev" || rev=$$(git rev-parse --short HEAD); \
 	name="ctags-$$rev"; \
 	archive="$$name.tar.gz"; \
--- a/Makefile.in
+++ b/Makefile.in
@@ -345,4 +345,16 @@ include $(srcdir)/testing.mak
 check: $(CTAGS_TEST)
 	$(MAKE) -f testing.mak test.units
 
+#
+# dist
+# Creates a tarball named after the current Git tag or revision
+#
+dist:
+	git rev-parse  # this will bail out if not in a Git repository
+	rev=$$(git tag --points-at HEAD); \
+	test -n "$$rev" || rev=$$(git rev-parse --short HEAD); \
+	name="ctags-$$rev"; \
+	git archive --prefix="$$name/" --format tar "$$rev" | \
+		gzip -9 > "$$name.tar.gz"
+
 # vi:set tabstop=8:
--- a/Makefile.in
+++ b/Makefile.in
@@ -349,7 +349,7 @@ check: $(CTAGS_TEST)
 # dist
 # Creates a tarball named after the current Git tag or revision
 #
-dist:
+dist: check
 	git rev-parse  # this will bail out if not in a Git repository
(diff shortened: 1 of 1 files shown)


masatake:
I tried and found that configure is not included in  the tar.gz file generated from dist target.
We may have to write our own version based on maintainer.mak and @b4n's patch.

b4n:
> I tried and found that configure is not included in the tar.gz file generated from dist target.

Indeed, because it's a generated file (not tracked in the repository).

If we want a tarball that doesn't require autoconf & friends to be used (e.g. to distribute *configure* and all the extra files like *missing*, *mkinstalldirs*, *install-sh*, *depcomp* and whatnot) we probably should use the Autotools facility for that.  However, I'm not sure if it's possible without using Automake -- and apparently it's hard to use Automake to build a software named `ctags` because it has builtin support for generating CTags files and the likes which conflict with it.

However, how relevant is it to distribute the various files generated by autoconf & friends out of *configure.ac*?  The advantage is allows someone to use the tarball without having autoconf, but how actually do that?  Linux distributions for one thing don't generally care, and often re-generate those files anyway (e.g. Debian always (?) does with *dh-autoreconf*).  So yeah, one less dependency is nice, but do someone caring to build from sources really care about the autoconf dependency?

nenadalm:
according to the help: https://github.com/fishman/ctags/blob/cfffbcf2c574fa391672a7dc5443d76735e049c3/maintainer.mak#L208

I've tried to merge this into my fork and then I run:
```shell
$ autoreconf
$ ./configure
$ make
$ make dist
$ cp ctags-da56977.tar.gz /tmp/ctags/
$ cd /tmp/ctags
$ tar -xf ctags-da56977.tar.gz
$ cd ctags-da56977
$ make -f maintainer.mak release-rpm-2.5
```

and it told me:
```
make: *** No rule to make target `release-rpm-2.5'.  Stop.
```

(the same when I try to run the last command from the repository folder)

it's probably not related with this PR - wasn't sure where sure I put it

b4n:
@nenadalm the tarball generated by that PR doesn't include the configure script and Makefiles, you need to generate them like you do from the Git repository (`autoreconf -vfi && ./configure`).

nenadalm:
inside of the extracted tarball
```
$ autoreconf -vfi
$ ./configure
$ make -f maintainer.mak release-rpm-2.5
```
the last command ends with the error agan:
```
make: *** No rule to make target `release-rpm-2.5'.  Stop.
```

nenadalm:
maybe I need different version of make in case it works for others?

```shell
$ make -v
GNU Make 3.82
Built for x86_64-redhat-linux-gnu
Copyright (C) 2010  Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

b4n:
Doesn't work on the Git repository either.  I have no idea why nor how it's supposed to work, but it's not the tarball's fault.

nenadalm:
ok, so I'll move it to another issue. thanks

masatake:
We should not make a release till #182 is solved.

steveno:
For the record, #182 was closed.

mike-burns:
I added `dist` in #676, among other targets.

masatake:
The situation is much changed; automake is introduced.
We have reliable `make dist` now. So I would like to close this.
 @mike-burns gets the kudos for it.

There is still a critical issue. Where we should define the version number: main/ctags.h, configure.ac or/and our git repository with "git tag".

PR #204, opened 2015-01-14
c++, d: Parse enum base type

ntrel:


b4n:
New units should have the `.d` suffix, the `.t` one is for the ones imported from the (old) *Test* directory [[1]](/fishman/ctags/blob/master/docs/f-units.rst#note-for-importing-a-test-case-from-test-directory).

b4n:
Otherwise LGTM

ntrel:
> New units should have the .d suffix

OK, renamed to `cxx11enum.cpp.d`.

PR #206, opened 2015-01-15
Import Rust parser from Geany

b4n:
This is an alternative version of #83 importing the character-based parser @SiegeLord wrote for Geany.

As CTags doesn't have the equivalent of Geany's `varType` (`typeRef` is similar but not identical as discussed somewhere already) I commented-out this feature for now.  A nice way to fix this is welcome.

This PR imports all commits from Geany to reproduce the history clean, only altering those so they match CTags format (basically, the test results format).

@SiegeLord: are you OK with your authorship info ending up in this repository? (and my small amendments)
@masatake @Passw the types reported by this parser are slightly different than those of the previous regex-based optlib entry, are the changes OK with you?  (see 141584913fe0eaf21c170c4b1b61bf06d090274b)


SiegeLord:
I'm fine with this. I'm a bit confused where my future contributions should go though... just here? Just Geany? Both?

b4n:
> I'm a bit confused where my future contributions should go though... just here? Just Geany? Both?

Ideally the future would be this becoming official CTags and Geany would use it, so contributing here would be all needed and Geany would pull it in in one way or another.

In practice right now it's not perfect, but as I am willing to try and keep Geany and CTags as synched as possible, you can contribute to whichever end you prefer and I (or someone else) will make sure it is kept in sync.

b4n:
(I just fixed commit 5e1c4bbcb76d7615e1a1fa8b57cae13d7f8d6c85 that missed the `args.ctags` files, hence breaking the tests)

ffes:
I personally have no knowledge of Rust, but it looks good to me.
After you've merged this, I will update the win32/*vcxproj* files.

ffes:
When you merge this, I assume you will close #83 as well

masatake:
> @masatake @Passw the types reported by this parser are slightly different than those of the previous regex-based optlib entry, are the changes OK with you? (see 1415849)

Sure, it is o.k.

rust.c
```diff
@@ -0,0 +1,984 @@
...
+	K_STATIC,
+	K_MACRO,
+	K_FIELD,
+	K_VARIANT,
+	K_METHOD,
+	K_NONE
+} RustKind;
+
+static kindOption rustKinds[] = {
+	{TRUE, 'n', "namespace", "module"},
```
b4n: @SiegeLord as CTags doesn't have the naming limitation Geany has, do you want this to be renamed to "module" here?  (and same question for e.g. "class")

b4n:
@ffes 
> After you've merged this, I will update the win32/*vcxproj* files.

If you want you can prepare a commit and I can add it to this PR.

rust.c
SiegeLord: Whatever makes it easier I suppose... if I recall correctly these strings are not user-visible anyway?

ffes:
@b4n I never added something to someone else's PR. What is the best to get the commit to you? Should I just mail you a ``git format-patch`` file? Or is there another/better way?

b4n:
@ffes 
> What is the best to get the commit to you? Should I just mail you a git format-patch file? Or is there another/better way?

You can do that, or you can pull my branch, add a commit, push it somewhere (e.g. your GH) and give me the URL so I can fetch it (just like a manual merge -- but without the merge commit as the branch would be fast-forward).

rust.c
b4n: > Whatever makes it easier I suppose...

For Geany compatibility keeping the current values is a bit easier, but if they don't change over time (and they shouldn't) it's not very much of a concern as it wouldn't affect a diff.

>  if I recall correctly these strings are not user-visible anyway?

In Geany they aren't really, but in CTags they end up in the tag file as the tag kind.  How it is used will depend on the application I guess, but I guess they might end up being the label used to represent their kind, or they might get mapped back to known kinds (kind of like in Geany, and as I plan to do in Geany at some point so that we don't have to alter them at the parser level).
So… I guess the best is to have meaningful names, but it's probably not so important either.

SiegeLord: I see. In that case the good names are 'module' and 'implementation' I think.

b4n:
I fixed the kind names as discussed with @SiegeLord, all should be good now.

b4n:
Err wait I forgot to edit the tests

b4n:
ok, *now* it should be ok, sorry

masatake:
fuzz and noise tests are passed.

b4n:
Thanks everyone, merged :)

PR #207, opened 2015-01-20
D: Parse function & aggregate declaration templates

ntrel:
I.e. class, struct, interface, union.


b4n:
So, LGTM

masatake:
You removed TAG_FUNCTION_TEMPLATE. Is it really o.k.?
Don't you need a separate kind of function template?

ntrel:
The code to parse DECL_FUNCTION_TEMPLATE was never enabled: https://github.com/fishman/ctags/pull/207/files#diff-9c07e5e602aac928d6d779592925a132L2504. So it's not really removing anything. In D there are function, class, struct, union, enum and alias template declarations, I really don't think it's worth having separate tag types for each of them.

masatake:
I see. Thank you.

PR #208, opened 2015-01-20
verilog: Generate tags for `genvar` statements

b4n:
DISCLAIMER: I have no clue about Verilog, so this requires reviewing.

This is based on the work of a (mostly anonymous) Geany user from some time ago.

Also, we may need to write a test of our own, because I grabbed this one from SO, and while no one will ever care this might violate their copyright.

@vhda could you check this out and comment?

### DIFF
--- a/Units/verilog-basic.d/expected.tags
+++ b/Units/verilog-basic.d/expected.tags
@@ -12,6 +12,7 @@ STATE7	input.v	/^          STATE7 = 4'h8;$/;"	c	module:mod
 a	input.v	/^input a,b, c, d ;$/;"	p	module:mod
 a	input.v	/^wire a,b,c,d,e;$/;"	n	module:mod
 add	input.v	/^task add;$/;"	t	module:mod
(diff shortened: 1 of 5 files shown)


vhda:
Variables of type "genvar" are treated as integers, so I would prefer to have it in K_REGISTER group.

b4n:
@vhda fixed.  If you want to add a new (copyright-concerns-free) test or anything, feel free to steal my changes, ammend, copy or throw them away.

vhda:
That would probably be a good idea.
Will you allow me one day or two to take a look at that? IRL is getting in the way...

b4n:
> Will you allow me one day or two to take a look at that? IRL is getting in the way...

No problem, take the time you need ^^

vhda:
Could you please cherry-pick the commit from my `verilog/genvar` branch?
I've expanded the Verilog and SystemVerilog basic test units to include an usage of `genvar`, so you can remove the test unit from your commit. If you agree, that is :)

b4n:
@vhda done, I added your commit and ammended mine to remove the test.

Issue #211, opened 2015-01-22
Strange character as tag in Units/bug1777340.java.t


ffes:
I accidentally noticed that when you look at ``expected.tags`` in ``bug1777340.java.t`` there is a strange result at line 3. The tag is ``Ó`` but this should be an ``n``.

https://github.com/fishman/ctags/blob/master/Units/bug1777340.java.t/expected.tags#L3

b4n:
Indeed.  After some extensive digging I found out where this `0xd3` character comes from: it's `STRING_SYMBOL` from *read.h*.

The actual bug is a combination of 2 things:

1. there is a special case for Java allowing characters out of ASCII in some identifiers (through `isHighChar()`).  As `STRING_SYMBOL` returned by `cppGetc()` is considered a "high character" the string gets recognized as a name and it's character representation is used, `"\x3d"`, which happens to be `Ó` in ISO-8859 character sets.
This part IMO should be fixed by changing the value of `STRING_SYMBOL` to be above `0xff` so it's not a valid character, and make `isHighChar()` only recognize possible high-end characters (`<= 0xff`).  I'll prepare a PR for this.

2. the code actually handling the construct doesn't seem to recognize the `default "..."` construct and thus uses an inappropriate token for creating the tag.

cweagans:
Sorry for the noise. Just switching the "Release blocker" tag for the "Initial release" milestone so that everything is tracked in one place.

PR #212, opened 2015-01-22
c#, java: Don't mistake string and character constants for high bytes

b4n:
Change the value of `STRING_SYMBOL` and `CHAR_SYMBOL` for them not to conflict with legitimate high bytes.

Alter C# and Java-specific checks allowing for high characters to exclude any non-byte value (anything above 0xFF).

Note: the test *bug1777340.java.t* still isn't correct.

Part of #211.


ntrel:
Looks good.

PR #213, opened 2015-01-26
core: Add development area pseudo tag

vhda:
* Revert original to using original URL pseudo tag and add new development tag
  pointing to github area.
* Update copyright date.

### DIFF
--- a/ctags.h
+++ b/ctags.h
@@ -18,8 +18,9 @@
 # define PROGRAM_VERSION "Development"
 #endif
 #define PROGRAM_NAME      "Exuberant Ctags"
-#define PROGRAM_URL       "https://github.com/fishman/ctags"
-#define PROGRAM_COPYRIGHT "Copyright (C) 1996-2009"
+#define PROGRAM_URL       "http://ctags.sourceforge.net"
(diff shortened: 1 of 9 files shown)


Pyrohh:
Should the copyright date not be 2015?

vhda:
Thanks for spotting that.
The change was originally created still last year and I did not give the necessary attention to the diff while splitting the changes into different PRs.

Pyrohh:
LGTM

masatake:
I think we don't need development area pseudo tag anymore.
Can I close this?

ffes:
I also think it is OK to close this one. If not @vhda please reopen.

vhda:
That's fine by me.
But please review that file, I think we need to discuss all those values.

masatake:
@vhda, now we set our own value to PROGRAM_URL, https://github.com/universal-ctags/ctags.
(This can be changed to http://ctags.io.)

For a while, I think it is enough for us. The most concerned thing was that a bug report about our implementation went to sf.net. This was really bad thing.  Now we don't have to care about it. The user can reach us.

vhda:
:thumbsup:

PR #214, opened 2015-01-26
Remove CVS keywords

Pyrohh:
I used [this page](http://www.badgertronics.com/writings/cvs/keywords.html) as a point of reference.


masatake:
Thank you.

PR #215, opened 2015-01-26
Remove mkinstalldirs

Pyrohh:
From the [mkinstalldirs manual](https://www.gnu.org/software/gettext/manual/html_node/mkinstalldirs.html):

"With earlier versions of GNU gettext, you needed to add the GNU
mkinstalldirs script to your distribution. This is not needed any more.
You can remove it if you not also using an automake version older than
automake 1.9."

[automake 1.9 was released in 2004](http://ftp.gnu.org/gnu/automake/), so (hopefully) almost noone will be affected by this.


Pyrohh:
This causes `make install` to fail for me. I'll look into what else is needed for its removal.

b4n:
That's because of https://github.com/fishman/ctags/blob/master/Makefile.in#L191

b4n:
Also, CTags doesn't use Automake, only Autoconf.

Pyrohh:
Closing this for now.

Issue #216, opened 2015-01-27
incorrect detection of private functions/methods in python


gotyaoi:
The current chunk of code that controls whether a function is public or private is

    /* If a function starts with __, we mark it as file scope.
    * FIXME: What is the proper way to signal such attributes?
    * TODO: What does functions/classes starting with _ and __ mean in python?
    */
    if (strncmp (vStringValue (function), "__", 2) == 0 &&
        strcmp (vStringValue (function), "__init__") != 0)
    {
        tag.extensionFields.access = "private";
        tag.isFileScope = TRUE;
    }
    else
    {
        tag.extensionFields.access = "public";
    }

This is missing a few cases.

* Any name that begins with a single underscore should be considered private. The interpreter does not enforce this, but it's a useful convention.
* Any name inside a class that begins with two underscores but does not end with two underscores undergoes name mangling to _classname__name, to avoid conflicts with child classes using the same name. Double underscore should also be considered private.
* Finally, names that begin two underscores and end with two underscores are "magic". They're used by the interpreter to implement various functionality. They should not be considered private.

The current implementation appears to mark any function/method that begins with two underscores but is not `__init__` as private. This leaves out names beginning with one underscore, and includes names that start with two underscores but also end with two underscores.

`makeClassTag` and `makeVariableTag` do not include any logic to set the access extension field at all right now, as far as I can see.

b4n:
This is already fixed in Geany, we need to import the changes here.  See https://github.com/geany/geany/commit/dd7c20a19470f588eaea6196d864142458d3d5ee for the commit relevant to method access.

gotyaoi:
Ah, very cool. Though, it looks like it's only being called for functions. Do you think it makes sense for it to be applied to variables and classes as well?

b4n:
Good remark.  For member variables, sure.  For classes I'm not sure what the Python guidelines say, but possibly yeah.

gotyaoi:
I think the single leading underscore can be applied to pretty much any name. In a package/module it prevents `from name import *` from grabbing it, which includes classes and module level variables beginning with an underscore.

b4n:
> I think the single leading underscore can be applied to pretty much any name. In a package/module it prevents `from name import *` from grabbing it, which includes classes and module level variables beginning with an underscore.

Indeed, a leading underscore at the module level prevents automatic import, and seem to work with anything (classes, functions, constants).

However, double underscore don't lead to any name mangling, so it's just "protected" at the module level, not private.

[...]

PR #217, opened 2015-01-28
Import Python changes from Geany + fixes and additional improvements

b4n:
Import Geany changes to the Python parser.  Also fix a bug in a Geany change, and further improve access reporting.

Fixes part of #216 (and various other bugs, see commits).


b4n:
I have an additional commit that doesn't come from Geany but improves the access reporting further, and fixes #216: 22c4cbe47e4985aefede1dee9dbeca382ebd5105.  I'm not sure if I should add it here or PR it afterward (as it doesn't fit the title to the PR here).  Anyhow if this gets merged I can make another PR then :)

fishman:
haven't checked the last two patches yet, but these patches don't really have anything to do with each others besides the fact that theyre from geany. therefore, personally, i don't think adding the patch from #216 is a problem. If the pull request title is a problem just change it or add the patches to the new one. I guess if this was a mailing list it would be a v2 patchset anyway, so I'm not sure what standard procedure we want to follow on github, yet.

b4n:
> haven't checked the last two patches yet, but these patches don't really have anything to do with each others besides the fact that theyre from geany. therefore, personally, i don't think adding the patch from #216 is a problem.

OK, fair enough indeed.  So, I added it here.

b4n:
You guys are OK with me (or anyone, really ^^) merging this?

vhda:
Just one question: is it valid to have empty extended attributes (`file:` in this case)?

b4n:
> Just one question: is it valid to have empty extended attributes (`file:` in this case)?

yes.  `file:` is generated by `entry->isFileScope` and means "this has file scope".  All parsers that set this field will generate an empty `file:` field (see for example `git grep -P '\tfile:' Units/`).

This said, and as mentioned in the commit https://github.com/b4n/fishman-ctags/commit/c16f0b819e1f2302e20f3d3f14d85af5bc6c9e1e#diff-e9dffeb0a60dddb2b51c1d8b6b15c3c9R198 I'm not certain setting `isFileScope` is useful/the thing to do.

fishman:
merge is fine imho

PR #220, opened 2015-02-01
Add simple regex parser for Jade mixins

leourbina:
Straightforward definition of a regex-based parser for Jade mixin definitions.  Tested on my local machine and it works as expected. @fishman

### DIFF
--- a/jade.c
+++ b/jade.c
@@ -1,0 +1,18 @@
+#include "general.h"
+#include "parse.h"
+
+static void installJadeParser (const langType language)
+{
+  addTagRegex(language, "^mixin[ \t]*([a-sA-Z0-9_]+)", "\\1", "d,definition", NULL);
+}
+
+
+extern parserDefinition* JadeParser (void)
+{
+  static const char* extensions[] = { "jade", NULL };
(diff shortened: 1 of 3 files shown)


masatake:
Thank you for contributing.
We need test cases for merging.
See docs/units.rst.

See also docs/optlib.rst. The directory structure of optlib is not stable. However, optlib can be an option to implement what you want.

leourbina:
Hi Masatake, I will add tests shortly.

masatake:
Hey, any progress?
If you have any question, feel free to ask me.

jade.c
```diff
@@ -0,0 +1,18 @@
+#include "general.h"
+#include "parse.h"
+
+static void installJadeParser (const langType language)
+{
+  addTagRegex(language, "^mixin[ \t]*([a-sA-Z0-9_]+)", "\\1", "d,definition", NULL);
```
sergeyromanov: The character class in the regex should probably be "[a-zA-Z0-9_]".

masatake:
@sergeyromanov, do you use/know Jade?

PR #221, opened 2015-02-01
C/C++ improvements

Corax26:
- Support for C++11 enum classes (by treating them as enums)
- Ignore #undef directives (instead of generating a tag)

### DIFF
--- a/Units/cxx11enum.cpp.d/expected.tags
+++ b/Units/cxx11enum.cpp.d/expected.tags
@@ -11,6 +11,14 @@ C_a	input.cpp	/^enum C:unsigned int {C_a, C_b, C_c};$/;"	e	enum:C	file:
 C_b	input.cpp	/^enum C:unsigned int {C_a, C_b, C_c};$/;"	e	enum:C	file:
 C_c	input.cpp	/^enum C:unsigned int {C_a, C_b, C_c};$/;"	e	enum:C	file:
 D	input.cpp	/^  enum D:int {a, b, c};$/;"	g	class:Foo	file:
+E1	input.cpp	/^enum class E1:short {E1_a, E1_b, E1_c};$/;"	g	file:
+E1_a	input.cpp	/^enum class E1:short {E1_a, E1_b, E1_c};$/;"	e	enum:E1	file:
+E1_b	input.cpp	/^enum class E1:short {E1_a, E1_b, E1_c};$/;"	e	enum:E1	file:
+E1_c	input.cpp	/^enum class E1:short {E1_a, E1_b, E1_c};$/;"	e	enum:E1	file:
+E2	input.cpp	/^  enum class E2:short {E2_a, E2_b, E2_c};$/;"	g	class:Foo	file:
+E2_a	input.cpp	/^  enum class E2:short {E2_a, E2_b, E2_c};$/;"	e	enum:Foo::E2	file:
+E2_b	input.cpp	/^  enum class E2:short {E2_a, E2_b, E2_c};$/;"	e	enum:Foo::E2	file:
+E2_c	input.cpp	/^  enum class E2:short {E2_a, E2_b, E2_c};$/;"	e	enum:Foo::E2	file:
 Foo	input.cpp	/^class Foo {$/;"	c	file:
 a	input.cpp	/^  enum D:int {a, b, c};$/;"	e	enum:Foo::D	file:
 b	input.cpp	/^  enum D:int {a, b, c};$/;"	e	enum:Foo::D	file:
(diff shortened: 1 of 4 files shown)


masatake:
undef should not be removed. In some case it is informative.

If you are not satisfied with "ctags foo.c | grep -v #undef", you should introduce something command line option to conotrol whether undef'ed names are tagged or not.

e.g.

   $ ctags --c-option={record-undef:no}

b4n:
> you should introduce something command line option to conotrol whether undef'ed names are tagged or not.
>
>e.g.
>
>$ ctags --c-option={record-undef:no}

Or simply change the kind name for `#undef` directives?  Though, this would possibly be deemed an incompatible change and not desirable.

ntrel:
There appears to also be `enum struct`. Geany supports both: [commit](https://github.com/ntrel/geany/commit/6c7f69). This pull may affect other languages besides C++, not sure if it's a problem though.

Corax26:
@masatake @b4n For the life of me, I can't see how a directive meant to *undefine* an identifier should generate a tag: either the macro is `#define`'d somewhere else (generating a tag), or it is `#define`'d in some system header and we want to get rid of it. Anyway, if some people use this "feature" I agree it must not be removed, so introducing an option seems to be reasonable (I'm not going to pipe ctags into grep each time I update the tags file...). Adding a kind doesn't make sense to me, the identifier kind is "macro", not define or undef.

@ntrel This is completely correct, I didn't know it but the C++11 standard allows `enum struct` as a synonym of `enum class` (see [here](http://en.cppreference.com/w/cpp/language/enum)). The commit you mention looks perfect (checking the language is also a good idea), could it be cherry-picked from Geany's repo instead of mine?

ntrel:
I don't know if git cherry-pick can understand Geany's repo, it's structure is quite different. But patching that commit to ctags and adding your tests would be good IMO.

b4n:
> I don't know if git cherry-pick can understand Geany's repo, it's structure is quite different.

`git format-patch --relative=tagmanager/ctags` does the trick.  However, Geany's and CTags have diverged a bit here, so it required a few additional tuning of the patch.  Anyway, it's here: 52f09c44690a59aa8b8e6df5e4ebe9df0aae9186.  Tests should indeed be added too.

b4n:
> @b4n For the life of me, I can't see how a directive meant to undefine an identifier should generate a tag

I'm not sure either, I just meant to suggest a simple solution that could please everyone.

> Adding a kind doesn't make sense to me, the identifier kind is "macro", not define or undef.

Well, OK, but OTOH it might possibly be useful to distinguish macro introduction from removal.  Having the distinction could e.g. be used by a completion tool not to suggest macros that were undefined at a specific position.
I'm just suggesting, I don't mind much either way.

Corax26:
Thanks @b4n for importing Geany's commit.

I think this pull request should be closed, I'll create 2 new separate pull requests, one for the enum class/struct and another for the `#undef` thing.

About `#undef`, would it be reasonable to create an option and disable tag creation by default? I don't think many people use this feature, and it can be quite annoying (typically when you undef MSVC's stupid macros like `ERROR` and then use `ERROR` as an enumerator).

masatake:
Like --if0=yes|no, --under=yes|no may be better than --c-option={record-undef:no}.

I want to enable the tag by default(--undef=yes). I don't want to change the behavior of ctags.
~/.ctags will help you.

Corax26:
Oh right I forgot about ~/.ctags ^^

About --undef, I agree it is shorter than a --c-option=..., but isn't it a bit overkill to introduce a new full-blown option just for it instead of using the existing infrastructure?

masatake:
@Corax26,  language specific option facility is interesting but it is not implemented yet:-P No code for it here.
It is long time my idea. I wanted to implement  the language specific option facility but I didn't because designing its syntax is not easy for me.

I think --undef is o.k. because

1. --if0 is already there. This means c preprocessor related option can be top level option, 
2. handling #undef is not c language parser specific. Ignoring #undef has impact at least C and C++, and
3. we can change it later. Our code is not released yet!

Another approach is introduce --cpp option name space(prefix).
So what you want can be invoked with --cpp-undef=no.
In this approach we have to prepare alternative name --cpp-if0 for --if0.

I think --undef is acceptable.

Corax26:
Oh I see, I thought your initial suggestion using `--c-option` used some existing structure ^^

Given that `--if0` already exists, I agree that `--undef` makes sense. We could add a prefix indeed, but for now let's keep it simple (and keep it backwards compatible by not modifying an existing option) :)

I'll open the PR for the enum class/struct momentarily, and I'm going to see how `--if0` works to do the same for undef.

PR #225, opened 2015-02-08
Add clojure support

nenadalm:
Hello. Just starting with clojure and couldn't find some sort of tags supporting namespaces.

I insipred by php.c and lisp.c.

Also not very familiar with c so feel free to check and comment what I've done.


vhda:
In my personal opinion, using `strncmp` would make the code more readable. The cost for performance should not be that big, right?
Also, take `EOF` into consideration in your loops.
But wait for more feedback before changing anything :)

clojure.c
```diff
@@ -0,0 +1,163 @@
+/**
+ *   Copyright (c) 2015, Miloslav Nenadál <nenadalm@gmail.com>
+ *
+ *   This source code is released for free distribution under the terms of the
+ *   GNU General Public License.
+ *
+ *   This module contains code for generating tags for the PHP scripting
```
b4n: PHP? :)

clojure.c
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+static void skipSpaces (const unsigned char **p)
+{
+	while (**p != '\0' && !isspace ((int) **p))
+		*p = *p + 1;
+	while (isspace ((int) **p))
+		*p = *p + 1;
+}
```
b4n: The function look a bit misnamed as IIUC it first skips to a space, and then skips spaces

clojure.c
```diff
@@ -0,0 +1,163 @@
...
+				makeNamespaceTag (name, p);
+			}
+			else if (isFunction (p))
+			{
+				skipSpaces (&p);
+				makeFunctionTag (name, p);
+			}
+		}
+	}
+	vStringDelete (name);
```
b4n: you also need to free the `CurrentNamespace`

clojure.c
```diff
@@ -0,0 +1,163 @@
...
+static vString *CurrentNamespace;
+
+static int isNamespace (const unsigned char *strp)
+{
+	return strp[1] == 'n' && strp[2] == 's';
+}
+
+static int isFunction (const unsigned char *strp)
+{
+	return strp[1] == 'd' && strp[2] == 'e' && strp[3] == 'f' && strp[4] == 'n';
```
b4n: shouldn't you make sure the character next to "defn" is either a space or the end of the line?  kinda like you do in `isQuote()` below.
Same would apply to `isNamespace()` above.

Units/clojure.d/input.clj
```diff
@@ -0,0 +1,10 @@
+(ns app.controller)
+
+(defn empty-fn [])
+
+(defn function-with-body []
+    (println "body"))
+
+'(defn quoted-function [])
+(quote quoted-function2 [])
```
b4n: shouldn't `quoted-function` and `quoted-function2` appear in the tags?  (just assuming, I actually can't read clojure :)

clojure.c
```diff
@@ -0,0 +1,163 @@
...
+{
+	vString *name = vStringNew ();
+	const unsigned char *p;
+	CurrentNamespace = vStringNew ();
+
+	while ((p = fileReadLine ()) != NULL)
+	{
+		vStringClear (name);
+
+		if (*p == '(')
```
b4n: you only match when the parenthesis is at column 0, shouldn't you accept whites space before it?

source.mak
```diff
@@ -73,7 +74,7 @@ SOURCES = \
 
 ENVIRONMENT_HEADERS = e_msoft.h
 
-ENVIRONMENT_SOURCES = 
+ENVIRONMENT_SOURCES =
```
b4n: this change probably shouldn't be included (I imagine your editor stripped trainling spaces, and it's not so bad, but it's a little irrelevant).  but maybe we don't care

b4n:
> In my personal opinion, using `strncmp` would make the code more readable. The cost for performance should not be that big, right?

Agreed, it might be better.

> Also, take `EOF` into consideration in your loops.

Hum, AFAICT all loops properly check for the string's `\0`, and the line reading loop properly checks for `NULL`, so it looks alright on this regard to me

clojure.c
nenadalm: renamed: https://github.com/nenadalm/ctags/commit/3bdd9e84943f253717e8919709dba8ae565b4480#diff-088fe33f54773705f51613eed29926b4L110

clojure.c
nenadalm: done https://github.com/nenadalm/ctags/commit/3bdd9e84943f253717e8919709dba8ae565b4480#diff-088fe33f54773705f51613eed29926b4R142

Units/clojure.d/input.clj
nenadalm: quote (and it's syntax sugar ') prevents the code from evaluating: http://clojure.org/special_forms#quote so this shouldn't be inside the tags

nenadalm:
thanks for reviews. now all comments should be fixed

Units/clojure.d/input.clj
b4n: Oh, ok, my bad ^^

nenadalm:
> hum, shuoldn't it be strncmp (...) == 0 here?
yes it should. fixed. thanks

clojure.c
```diff
@@ -0,0 +1,163 @@
...
+	const unsigned char *p;
+
+	if (*dbp == '\'')
+		dbp++;
+	else if (*dbp == '(' && isQuote (dbp))
+	{
+		dbp += 7;
+		while (isspace (*dbp))
+			dbp++;
+	}
```
b4n: Hum… wait, what is this supposed to match?  I mean, stop me if I miss something, but from `findClojureTags()` loop, you only match with `isNamespace()` (`(ns `) or `isFunction()` (`(defn `).  Then, in either case you skip the word (`ns` or `defn`) and the following whitespaces, and read a name with `functionName()`.  `functionName()` will match [`^(?:'|\(quote +)([^() \t]+)`](http://regexper.com/#%2F^%28%3F%3A%27|\\%28quote%20%2B%29%28%5B^%28%29%20\t%5D%2B%29%2F) -- e.g. a function name prefixed by either nothing, `'` or `(quote `.

So, IIUC it allows to match constructs like `(ns 'somename)` or `(ns (quote someothername))`.  If that's it, there should be a test for them :)

masatake:
make fuzz LANGUAGES=Clojure VG=1 => passed!
I'm running noise test now.

clojure.c
nenadalm: in clojure, you can define function ``hello`` like this ``(defn hello [])``
or create list with values ``defn, hello, []`` like this ``'(defn hello [])`` which is equivalent to ``(quote (defn hello []))``

so it's true quote check in the code is kind of useless

this PR only adds tags like high level function definitions with their namespaces
(can add nested functions by accident or can skip function if it's not written the way I never saw it like
```clojure
(
defn hello[])
```

So I could probably remove the quote thing for now.

```clojure
(ns 'somename)
```
will cast exception
``ClassCastException clojure.lang.PersistentList cannot be cast to clojure.lang.Symbol  clojure.lang.RT$1.invoke (RT.java:237)``

nenadalm:
I do plan to write it later like it is for php: https://github.com/fishman/ctags/blob/master/php.c so it'll understand more stuff. Wanted to get familiar with ctags, c first and try to support namespaces in vim. But in case it would be problem I can leave it in my fork for now and send PR later. (I'll remove quote thing and fix the fuzz meanwhile...)

nenadalm:
oh - fuzz is working. I did misunderstand somehow :)

wouldn't it be better to add fuzz to travis? so that is checked each time automatically @masatake?

clojure.c
b4n: OK.  So, why did you keep the support for `'` prefix then? :)

b4n:
> I do plan to write it later like it is for php: https://github.com/fishman/ctags/blob/master/php.c so it'll understand more stuff. Wanted to get familiar with ctags, c first and try to support namespaces in vim. But in case it would be problem I can leave it in my fork for now and send PR later.

No it's fine you can always improve it later that's not a problem.  I'm just trying to strip the unnecessary/incorrect things that's all :)

masatake:
$ make noise LANGUAGES=Clojure VG=1  => passed

@nenadaim, yes, fuzz target can be run in travis. noise target takes too long time to be completed.
About fuzz target in travis, a PR is welcome:)

masatake:
@b4n, could you merge this if you are o.k?

masatake:
(off topic)
It looks interesting, doesn't it?
https://github.com/masatake/ctags/commit/144cec9337b4e25f587f0aa27ff430fe97d171e5

masatake:
I merged your changes manually; before merging I unified your 3 commits into 1.
@nenadalm, thank you for your PR.
The reviewers, thank you.

PR #231, opened 2015-02-08
Add a new --undef command-line option

Corax26:
This option allows to disable the generation of macro tags from #undef
directives, see PR #221 for the rationale. The default behaviour is
unchanged, i.e. a tag is still generated.

The help and man have been updated accordingly, and I added a small unit
to check that --undef=no disables the generation of a tag. The
directives.c unit already checks that a tag is generated when no option
is passed (which is equivalent to --undef=yes).


options.c
```diff
@@ -259,7 +260,7 @@ static optionDescription LongOptionDescription [] = {
  {1,"  --help"},
  {1,"       Print this option summary."},
  {1,"  --if0=[yes|no]"},
- {1,"       Should C code within #if 0 conditional branches be parsed [no]?"},
+ {1,"       Should C/C++ code within #if 0 conditional branches be parsed [no]?"},
```
b4n: I'd rather remove the language detail like you did in `--undef` description than refer to "C/C++" (which is probably the greatest lie about C and C++).  Moreover, I think there are other languages that are affected (D? Verilog?), too :)

b4n:
I didn't test, but code LGTM (apart from my remark on wording).

Corax26:
Regarding the wording, it looks like CPP is only really used in C and C++: http://en.wikipedia.org/wiki/C_preprocessor (Verilog definitely doesn't use it, neither does D). I agree that "C/C++" is quite a bad way to refer to both C and C++ (often coming from people thinking that C++ is just an overlay over C....), I just used for conciseness, maybe the easiest solution would be just "Should code within..."?

b4n:
> Regarding the wording, it looks like CPP is only really used in C and C++ […] (Verilog definitely doesn't use it, neither does D)

Verilog indeed, I just always get confused by the Verilog parser including `get.h` sorry.  For D (and other `c.c`. languages like Vera), OK, I don't know.

> […] maybe the easiest solution would be just "Should code within..."?

Agreed, that's what I was trying to suggest :)

masatake:
@Corax26, could you write your change in docs/news.rst, too?
Thanks you.

Corax26:
@b4n I've amended the commit removing "C/C++" :)
@masatake I've pushed another commit modifying news.rst, the new option didn't fit anywhere so I created a new section (miscellaneous new options).

masatake:
Thank you.

vhda:
@b4n Verilog parser only includes `get.h` to use `skipOverCComment()`.
The concept of pre-processing exists in Verilog, but it has some differences:
1. Uses `` ` `` instead of `#`.
2. Verilog only has `` `ifdef ``, which checks for the **existence** of the directive; it does not process expressions.

PR #232, opened 2015-02-09
Travis update

nenadalm:
run make noise for working languages to make sure they'll not break.

this pr was sent based on: https://github.com/fishman/ctags/pull/225#issuecomment-73458206

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,5 @@ script:
   - ./configure
   - make -j2
   - make check TRAVIS=1
+  - make noise VG=1
   - if [ $CC = 'gcc' ]; then make distclean; make -f mk_mingw.mak CC=i686-w64-mingw32-gcc; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,5 @@ script:
   - ./configure
   - make -j2
   - make check TRAVIS=1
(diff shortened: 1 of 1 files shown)


b4n:
While testing stuff is good, isn't running `noise` on all languages under Valrgind gonna take like forever?  I don't think that if we need to wait several hours (if not more) to get the Travis results it's going to make us so much good…  maybe there could be something separate?  or at least only run it for languages that were affected since `master` or something?

b4n:
> this pr was sent based on: https://github.com/fishman/ctags/pull/225#issuecomment-73458206

That comment was referring to `make fuzz`, not `make noise` actually:

> @nenadalm, yes, fuzz target can be run in travis. noise target takes too long time to be completed.
About fuzz target in travis, a PR is welcome:)

b4n:
> https://travis-ci.org/nenadalm/ctags/builds/50097932

I'm willing to bet that failure is because Valgrind is not installed on the Travis VM :)  Exit code 127 generally means "command not found" -- and as you don't explicitly install Valgrind, I doubt it's there.

nenadalm:
oh... so make noise should't pass?

b4n:
> oh... so make noise should't pass?

oh it should (if you have Valgrind installed, as you run it with `VG=1`), but it's gonna take some time :) (count hours at the very least)

masatake:
As I wrote before somewhere, the noise target is too expensive too to be run inside travis.

nenadalm:
ok closing

PR #235, opened 2015-02-10
Fix filename references to files in docs/

marcows:
The "f-" prefix has been discarded in git commit
f439b71b949adadba4c83df7ae033a8f8a7015ec on 2014-12-23.


vhda:
Nice catch! It appears I've left those behind.

masatake:
@vhda, please merge if you are o.k.

PR #236, opened 2015-02-10
python: do not ignore the character after a skipped string

marcows:
This fixes missing or too much entries in the tags file.


b4n:
Looks great, thanks!

PR #237, opened 2015-02-11
Ada noise fixes

b4n:
Fix two bugs in the Ada parser found by `make noise`.


masatake:
Thank you. Noise target is really useful:-P

b4n:
> Noise target is really useful:-P

Yep, just a bit slow ^^

masatake:
As far as I can remember ctags in geany utilizes memory stream as input.
If e-ctags@github has it, we can make the noise target much faster.

PR #238, opened 2015-02-11
units: Exit with an error if `fuzz` failed

b4n:
This is important for scripting, and especially for Make integration so
failures are recognized.


b4n:
If we want not to error-out on the first failure we can avoid `break`ing out of the loop, but while it might be interesting for manual run (so we see all errors, not only the first), it should e.g. not be the default for automated setups like Travis at least, where we basically only want to know whether everything is good.

PR #239, opened 2015-02-11
beta: Fix a below-bounds buffer access

b4n:
Found by Miloslav Nenadál (@nenadalm) running `make fuzz VG=1`.


masatake:
Thank you very much.

PR #241, opened 2015-02-21
Fixed MSVC 2013 warnings, small changes in EXENDING.html

bmateusz:
Found some typo in the example code. I also added a hint for debugging,
I would found it useful.

### DIFF
--- a/EXTENDING.html
+++ b/EXTENDING.html
@@ -166,6 +166,12 @@ file, so that your new module will be compiled into the program.
 </p>
 
 <p>
+In case you have some problems run <code>ctags --verbose</code> to see if the
+extensions or patterns you defined for your language conflict with other
+languages.
+</p>
+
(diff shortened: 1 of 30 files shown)


masatake:
Thank you for patches. However, could you splits this PR to smaller PRs for each issues?
So reviewing becomes easier. 
I would like to merge  14d218c and b85937e first. 1c3a41b deals many topics. I would like you to split 1c3a41b into smaller patch sets.

Thank you again.

ffes:
Thanks for the patches! But it is not the preferred method to fix 3 completely unrelated problems with one PR. Could you please split them in three PRs, so we can review them separately?

ffes:
I see @masatake already gave a similar answer ;-)

bmateusz:
@masatake @ffes Ok I create separate PRs.

Issue #234, opened 2015-02-10
Merging patch sf.patch:77


Corax26:
[Patch #77](http://sourceforge.net/p/ctags/patches/77/) brings *major* bugfixes to c.c (mainly regarding local variables), both for C and C++, for this reason merging it should be considered high priority. I merged it about one year ago in my own repo and I haven't encountered any issue so far (at least no new issue).

However, the problem is that the changeset is rather big, adding some logic and heuristics to several parts of the parser. My question would be: is it acceptable to directly merge the patch (in a single commit), or is it necessary to split it in several commits?

Note that the author provided a test file (see [the linked archive](http://sourceforge.net/p/ctags/patches/_discuss/thread/2a3a8916/6741/attachment/ctags_cons_patch-20120531.tgz)), so at least it will be easy to add a test unit.

steveno:
Both of the links you give result in a 404 for me.

kopischke:
@steveno:

> The sourceforge.net website is temporarily in static offline mode.
Only a very limited set of project pages are available until the main website returns to service.

… should be temporary :).

Corax26:
Yep: "The sourceforge.net website is temporarily in static offline mode.", and also:
<blockquote class="twitter-tweet" lang="fr"><p>There&#39;s instability and frequent 500 errors on the <a href="https://twitter.com/hashtag/SourceForge?src=hash">#SourceForge</a> site currently. We&#39;re working on getting this fixed ASAP.</p>&mdash; SF.net Operations (@sfnet_ops) <a href="https://twitter.com/sfnet_ops/status/565180532535730178">10 Février 2015</a></blockquote>
<script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

masatake:
@Corax26, thank you for trakcing the bug and change.

I have two comments.

1. when submitting and merging changes, please, make their origin and author clear in commit log.
   See http://sourceforge.net/p/ctags/patches/1/
   We cannot pick this because its author is anonymous.

2. .b test
    You may know write how to write a test of Units.
    Could you write a test case for the bug, first?
    We can pick the existing fix (or fix from scratch) later.
    Could you use .b as suffix fot the test directory?
    With the suffix, travis may ignore the error.
    See docs/units.rst.
    
Thank you again.

fishman:
seems to be this guy https://github.com/lyokha

Corax26:
[...]

PR #242, opened 2015-02-22
Add a unit based on patch sf.patch:77's  test.cpp

Corax26:
Patch #77 fixes an important number of issues with the way variables
(local and global) and prototypes are parsed in C and above all C++.

This commit adds a new unit (marked as known bug) based on the test file
provided with patch #77, removing some cases which weren't valid C++ and
duplicating all the identifiers as global variables, on top of local
variables.

The expected.tags file has been generated with a version of ctags where
patch #77 has been applied, with the following modifications:
- An issue with an invalid character in signatures has been worked
  around by the patch author @lyokha, see issue #234
- I removed the support for MS.Net, so that identifiers such as
  int x4(5 ^ 4); are effectively recognized as variables (it was
  just a matter of uncommenting a "case '^':" in the patch).

### DIFF
--- a/Units/variables-prototypes.cpp.b/args.ctags
+++ b/Units/variables-prototypes.cpp.b/args.ctags
@@ -1,0 +1,1 @@
+--c++-kinds=+pl --fields=+S
--- a/Units/variables-prototypes.cpp.b/expected.tags
+++ b/Units/variables-prototypes.cpp.b/expected.tags
@@ -1,0 +1,94 @@
+A1	input.cpp	/^int  A1 = i < 7 ? i : 7;$/;"	v
+A2	input.cpp	/^int  A2 = 8;$/;"	v
+A3	input.cpp	/^int  A3 = 9;$/;"	v
+A4	input.cpp	/^int  A4 = p->data;$/;"	v
+A5	input.cpp	/^int  A5 = 8;$/;"	v
+A6	input.cpp	/^int ( *A6 )( int, char ** ) = &main;$/;"	v
+A7	input.cpp	/^int  A7( short ( *p )( int ) );$/;"	p	file:	signature:( short ( *p )( int ) )
+B1	input.cpp	/^const char *B1( "uuu" );                \/\/variable$/;"	v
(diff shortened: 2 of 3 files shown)


masatake:
In my understanding `#77' refers an entry in patch or bug tracker in sourceforge.
Instead of using hash notation(#), could you use URL, sf.patch 77 or something differrent?
If you are using the hash notation here, github web interface translates it to a pull request or an issue on fishman/ctags.

BTW, you can split the test smaller ones. The benefit of smaller test cases is that we can introduce fixes for each of test cases incrementally.

Corax26:
Yeah I realized that #77 would confuse GitHub, I guess sf.patch 77 is a good alternative.

Regarding splitting test cases, in theory I would agree, but here apart from the 1st and 4th parts (a1-a5 and a8) everything is more or less related to the  "prototype vs variable initializer" issue so it isn't really worthwhile to split them. So what I could do is put a1-a5 in a separate unit `angle-bracket.c.b`and a8 in another unit `return-ignore.c.b` (something like that) and leave the rest where it is.

masatake:
I see. We can split the test case to smaller one when they are needed.
Could you fix commit log and  git push --force?

Corax26:
I fixed the commit message :)

masatake:
Thanks.

masatake:
(It is nice if I can move this PR to Issue.)

Corax26:
Indeed you're right, looking at STDERR.tmp shows that the parsing of args.ctags failed (too bad the error isn't treated differently from a diff error, since the unit is marked as known bug I couldn't see there was an issue just by looking at the output of `make units`).

expected.tags still looks OK to me, do you want me to commit the linebreak in args.ctags?

masatake:
@Corax26, thank you for checking.
It is not easy to detect the bug of test case itself:-(

> expected.tags still looks OK to me, do you want me to commit the linebreak in args.ctags?

Yes, I want.
Could you prepare the PR?

Corax26:
Sorry for the delay, forgot about it <.< Done: #304

PR #243, opened 2015-02-22
clojure: use singed char for c string for suppressing warning in compila...

masatake:
...tion

People who know well about standard C library, is this correct fix?
Should I make fileReadLine return char * value instead?

### DIFF
--- a/clojure.c
+++ b/clojure.c
@@ -28,24 +28,24 @@ static kindOption ClojureKinds[] = {
 
 static vString *CurrentNamespace;
 
-static int isNamespace (const unsigned char *strp)
+static int isNamespace (const char *strp)
 {
 	return strncmp (++strp, "ns", 2) == 0 && isspace (strp[2]);
 }
 
-static int isFunction (const unsigned char *strp)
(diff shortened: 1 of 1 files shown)


masatake:
@b4n, could you look at this?

b4n:
LGTM.

`char` vs. `unsigned char` is a bit subtle, especially because an implementation is actually allowed to have `char` itself as unsigned.  Anyway, most API taking a string expect a pointer to `char`, not `unsigned char`.  OTOH, most API taking single characters expect an `int` with the value of an `unsigned char`.

In theory, `unsigned char` makes more sense so you get meaningful values for bytes `> 0x7f` (e.g. above ASCII) -- having e.g. byte `0xff` as `255` looks more sensible than getting `-1`.
But in practice it doesn't matter so much, because not only as we see most APIs take `char`s, and APIs that take `int`s properly handle the conversion anyway, but also we generally don't deal with non-ASCII characters directly anyway.

So I guess that when we want to use *string.h* API it makes more sense to use `char`.

Issue #202, opened 2015-01-05
ctags might fail silently if /tmp is full


blueyed:
When /tmp is full, ctags might fail silently (outputting nothing) (ref: #201):

Using `strace` then it looks like this:

    [pid 18982] stat("/tmp/tags.cP8Y89", {st_mode=S_IFREG|0600, st_size=0, ...}) = 0
    [pid 18982] geteuid()                   = 1000
    [pid 18982] getegid()                   = 1000
    [pid 18982] getuid()                    = 1000
    [pid 18982] getgid()                    = 1000
    [pid 18982] access("/tmp/tags.cP8Y89", R_OK) = 0
    [pid 18982] open("/tmp/tags.cP8Y89", O_WRONLY|O_CREAT, 0666) = 3
    [pid 18982] dup2(3, 1)                  = 1
    [pid 18982] close(3)                    = 0
    [pid 18982] sched_getaffinity(0, 128, {f, 0, 0, 0}) = 32
    [pid 18982] open("/tmp/tags.cP8Y89", O_RDONLY) = 3
    [pid 18982] fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
    [pid 18982] fstat(3, {st_mode=S_IFREG|0600, st_size=0, ...}) = 0
    [pid 18982] getrlimit(RLIMIT_DATA, {rlim_cur=RLIM64_INFINITY, rlim_max=RLIM64_INFINITY}) = 0
    [pid 18982] getrlimit(RLIMIT_AS, {rlim_cur=RLIM64_INFINITY, rlim_max=RLIM64_INFINITY}) = 0
    [pid 18982] getrlimit(RLIMIT_RSS, {rlim_cur=RLIM64_INFINITY, rlim_max=RLIM64_INFINITY}) = 0
    [pid 18982] open("/proc/meminfo", O_RDONLY|O_CLOEXEC) = 4
    [pid 18982] fstat(4, {st_mode=S_IFREG|0444, st_size=0, ...}) = 0
    [pid 18982] mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7ff96fab8000
    [pid 18982] read(4, "MemTotal:        8056664 kB\nMemF"..., 1024) = 1024
    [pid 18982] close(4)                    = 0
    [pid 18982] munmap(0x7ff96fab8000, 4096) = 0
[...]

PR #244, opened 2015-02-22
Check file error flag in makeTagEntry function

bmateusz:
Possibly fixes issue: ctags might fail silently if /tmp is full #202


masatake:
Thank you.

PR #245, opened 2015-02-22
small changes in EXENDING.html

bmateusz:
Found some typo in the example code. I also added a hint for debugging,
I would found it useful.


masatake:
Thank you.

PR #247, opened 2015-02-23
check the file output error more frequently

masatake:
The original code introducing ferror doesn't check errors in endTagsFile, closeTagsFile and writePseudoTag.

### DIFF
--- a/entry.c
+++ b/entry.c
@@ -841,11 +841,11 @@ extern void makeTagEntry (const tagEntryInfo *const tag)
 		else
 			length = writeCtagsEntry (tag);
 
-		abort_if_ferror (TagFile.fp);
-
 		++TagFile.numTags.added;
 		rememberMaxLengths (strlen (tag->name), (size_t) length);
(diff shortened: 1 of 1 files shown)


entry.c
```diff
@@ -530,10 +538,16 @@ extern void closeTagFile (const boolean resize)
...
 		writeEtagsIncludes (TagFile.fp);
+	abort_if_ferror (TagFile.fp);
 	desiredSize = ftell (TagFile.fp);
 	fseek (TagFile.fp, 0L, SEEK_END);
 	size = ftell (TagFile.fp);
-	fclose (TagFile.fp);
+	if (fclose (TagFile.fp) != 0)
+	{
+		error (FATAL | PERROR, "cannot close tag file");
+		exit (1);
```
b4n: `error(FATAL)` will `exit` itself, so the `exit(1)` is dead code.  And if didn't, `abort_if_ferror()` should also `exit()` then.

PR #248, opened 2015-02-23
change vs2013 debug configuration

bmateusz:
Linker error: ada.obj : warning LNK4075: ignoring '/EDITANDCONTINUE' due
to '/SAFESEH' specification
So for debug mode I changed in Linker settings /SAFESEH to false.
In release mode it can stay true, since it does not create warnings.

Any comments are welcome, I mostly use gcc. All I can say the warning gone and you can still debug.
https://msdn.microsoft.com/en-us/library/9a89h429.aspx


ffes:
Thanks for the patch!

PR #249, opened 2015-02-23
vs2013 errno defined twice

bmateusz:
warning C4273: '_errno' : inconsistent dll linkage
...\12.0\vc\include\errno.h(31) : see previous definition of '_errno'

If _CRT_ERRNO_DEFINED is defined then do not define it again.


ffes:
Thanks for the patch!

PR #250, opened 2015-02-23
fix warnings in tg.c

bmateusz:
- big_table is unsigned short, so cast (float * char * char)
- 1.0 is double, but tg_big_create expects float, so change constant to 1.0f
- r = - ascore:
warning C4146: unary minus operator applied to unsigned type, result
still unsigned
so I changed it to r = -1 * ascore.


ffes:
LGTM

@masatake Since tg is your code, could you take care if this one?

masatake:
Thank you. I merged your patch manually. I reformated your commit log.

PR #251, opened 2015-02-23
refactor ada.c cmp function

bmateusz:
Do not call strlen 8 times, call only once.
Fix: ada.c(645): warning C4018: '<' : signed/unsigned mismatch

### DIFF
--- a/ada.c
+++ b/ada.c
@@ -623,6 +623,7 @@ static void movePos(int amount)
 static boolean cmp(const char *buf, int len, const char *match)
 {
   boolean status = FALSE;
+  int matchLen;
 
   /* if we are trying to match nothing, that is always true */
   if(match == NULL)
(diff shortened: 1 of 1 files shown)


vhda:
LGTM

ffes:
Again thanks for the patch

b4n:
why not rather make `len` unsigned and `matchLen` properly `size_t`?

in practice it probably won't ever cause a problem (as did the old code), but here it can theoretically be more dangerous, because if `strlen(match)` is greater than `INT_MAX` it'll probably flip around and result in a negative `matchLen`, leading to indexing below bounds (as the check `matchLen < len` will most likely succeed).
It's probably a bit tricky but I believe one could create a specially crafted file to trigger this (though, of course the large size of integers in modern computers makes this a lot less likely.

also, this file is riddled with implicit sign conversion with potential problematic effects, so if the goald is to fix theme all (which would be nice I guess) there are a lot of work ahead -- and you added one here, with the `size_t` assignation to an `int` :)

I know I'm a bit late (sorry), but I just want to point out that "fixing a warning" is not necessarily so great of an idea, unless you know exactly what it means and what are the implications of the fix.

masatake:
@b4n, thank you for the important comment.

> also, this file is riddled with implicit sign conversion with potential problematic effects
How do you do in geany?

b4n:
> How do you do in geany? 

Apparently Geany doesn't have the Ada parser (not sure why, maybe it didn't exist at the time someone  added Ada support, or nobody cared maybe), so we did nothing for this one.  But in general we avoid changing CTags parser for things that don't have an actual real-life/proven implication not to get a large diff for no good reason.

However, the limitation doesn't apply here (does it?), so if we do care about this what I'd suggest is "simply" fix everything to use the appropriate type (basically, `size_t` for length).  This is not trivial though, because e.g. *ada.c* somewhere uses `-1` as meaning "no set", so it requires something a little more subtle than just replacing some occurrences of `int` with `size_t`.  If we do care, I can take a look (GCC's `-Wconversion` and `-Wsign-conversion` help tracking those -- but beware, there are many).

This said, and while I think new code should avoid such issues, it's not a really big deal.  In most real-life situations this won't be the reason a problem occurs, and in some corner cases changing type signedness can lead to subtle behavioral changes -- that's why I think this has to be done carefully (if at all) and not just to "fix a warning".  So maybe we want to adopt a case-by-case policy for this, and not touch it unless there is a practical reason to.  Or we want to review them all carefully, and get this over with.

bmateusz:
@b4n Thanks for your comments.
My goal was to remove warnings in these PRs - as a first step.
I was careful with the changes to keep the old functionality and follow the convention of the files changed.

I agree with your point to change this, and we should change the logic behind magic constants like -1, STRING_LITERAL, CHAR_LITERAL

I will have a closer look at it. The other two constants I mentioned caused problems as well in Debug mode, since isspace(int chr) gives an assertion error if 0 < chr < 255 is false in vs2013.

These are small things, but can help readability.

PR #252, opened 2015-02-25
add coverity support to travis ci

fishman:


masatake:
Intersting.
How can we receive the result of coverity?

vhda:
We could add a status image to our README, as we have done for Travis.

masatake:
Great. Thank you.

PR #254, opened 2015-02-25
Set timeout value to 0 if timeout command is not available in a test environment

masatake:
(It seems that macosx at travis doesn't have timeout command.)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -367,7 +367,7 @@ check: units
 #
 # SHELL must be dash or bash.
 #
-fuzz: TIMEOUT := 1
+fuzz: TIMEOUT := $(shell which timeout > /dev/null 2>&1 && echo 1 || echo 0)
 fuzz: $(CTAGS_TEST)
 	@ \
 	c="misc/units fuzz \
@@ -392,7 +392,7 @@ noise: $(CTAGS_TEST)
 #
 # UNITS Target
 #
-units: TIMEOUT := 5
+units: TIMEOUT := $(shell which timeout > /dev/null 2>&1 && echo 5 || echo 0)
 units: $(CTAGS_TEST)
 	@ \
 	c="misc/units run \


masatake:
GNU sed is avaiable in macosx?
units is failed on macosx because sed on the platform does not know -r option. Maybe -r is available on GNU implementation.

vhda:
Don't make things harder than they need to :)
With the patch below `make units` still works correctly on my side.

diff --git a/misc/units b/misc/units
index ce44ad5..ed61f97 100755
--- a/misc/units
+++ b/misc/units
@@ -291,10 +291,10 @@ run_shrink ()
 }
 
 # filters out the directory prefix in a ctags input
-tags_basename_filter_regex='s%(^[^\t]+\t)([^/\t]+/)*%\1%'
+tags_basename_filter_regex='s%\(^[^\t]\+\t\)\([^/\t]\+/\)*%\1%'
 tags_basename_filter()
 {
-    sed -r "${tags_basename_filter_regex}"
+    sed "${tags_basename_filter_regex}"
 }
 
 run_record_cmdline ()
@@ -304,7 +304,7 @@ run_record_cmdline ()
 
     printf "%s \\\\\n| %s \\\\\n| %s\n"  \
        "${_CMDLINE}" \
-       "sed -r '${tags_basename_filter_regex}'" \
+       "sed '${tags_basename_filter_regex}'" \
        "${ffilter}" \
        > "${ocmdline}"
 }

b4n:
> Maybe -r is available on GNU implementation.

Maybe it is.  But as @vhda showed, it's easy enough to avoid using it (basically `-r` enables the extended mode which more or less means more characters are control characters by default and don't need escaping).

b4n:
Otherwise I guess the change here is OK (if we can depend on GNU make of course, but we already do AFAIK).  While running failing test under timeout is quite important, I guess not so much with the normal test suite -- and indeed, probably better than failing the test suite simply because we miss that command.

masatake:
@vhda, could you commit your change for sed -r?

PR #255, opened 2015-02-26
Use `timeout --version` to check for GNU timeout

ntrel:
Fixes `make units` on MSYS 1.0.
Note: `which timeout` succeeds even for MS timeout (c:\windows\system32\timeout.exe), which is not compatible.

(Follows on from #254).


masatake:
Thank you.

PR #256, opened 2015-02-26
units: Don't use extended regular expressions in sed

vhda:


b4n:
LGTM

vhda:
Let's wait to see if the second Travis run fares better than the first...

b4n:
@vhda apparently it does not… :/

b4n:
@vhda Maybe the `sed` on OSX doesn't like `%` delimiters? maybe try `'s/\(^[^\t]\+\t\)\([^/\t]\+\/\)*/\1/'`?

b4n:
(would be easier to test if someone had a real OSX machine instead of trial and error on travis :)

vhda:
I was looking at the `sed -n` part before I saw your updates.
Allow another run and if it doesn't work I'll try your delimiters idea.
(Thanks, btw)

vhda:
Replaced `s%%%` with `s///`.

b4n:
@vhda removing `-n` is *not* trivial, and I can't imagine a sed implementation not having it, because it's just not possible to alter the sed program to not use it.  If you remove `-n`, you need to make sure sed only gets fed the lines that it wants to print, which basically means calling `grep`.

b4n:
OK I see I'm too late :)

vhda:
In this case I don't think it would be a problem because stderr is redirected to `/dev/null` and the output of the script is a single line.

vhda:
I give up, for now.
On the weekend I will try to use my gf's OSX (for the first time) and test this...

fishman:
well the osx build is broken right now. i'll test it as soon as my mac works again. fortunately osx decides to break my build environment with every os update.

vhda:
sed command in OSX doesn't support `[\t]` and using an explicit tab also does not work.
Still debugging, but if anyone has a solution please do share! ;)

b4n:
> sed command in OSX doesn't support `[\t]` and using an explicit tab also does not work.

You mean it doesn't work only in character ranges or anywhere?  (just asking, because you added `[]`s and we use it in both contexts)
It'd be real weird if it did not support tabs *at all*…  does it have options maybe?  and what about in a sed script instead of command line? (ok, that'd be really crazy, but who knows)

b4n:
http://azaleasays.com/2014/03/07/os-x-sed-does-not-recognize-tab-character/ and other links suggest a literal tab should work… did you replace all occurrences of `\t`?

vhda:
Used <kbd>ctrl-v</kbd><kbd>tab</kbd> to insert the literal tab and didn't work. But it did work on grep, though (I normally use grep with `--color` to test regular expressions).
I think we'll need to use `-E` to enable extended regular expressions in the same manner we used `-r` before and one of the following:
1. Replace `\t` with `[[:space:]]`.
2. Use literal tab.
Any preferences?

vhda:
Wait a second, this worked in the command line but is not working in the script.

b4n:
> Used <kbd>ctrl-v</kbd><kbd>tab</kbd> to insert the literal tab and didn't work.

<kbd>^v</kbd> shouldn't be useful if you edit the unit script :)  But ok, very weird…

> I think we'll need to use `-E` to enable extended regular expressions in the same manner we used `-r` before […]

Do you think it's needed for literal tabs?  because if we can avoid adding other non-portable flags we could have the same call on all platforms, and that'd be great.

> 1. Replace `\t` with `[[:space:]]`.
2. Use literal tab.
Any preferences?

`[[:space:]]` is not appropriate here I believe, because we need to match a tab and exactly a tab -- some parser (like CSS) output spaces in the first field, so we need to be careful here.

vhda:
Literal tab does seem to work, but only if not used together with `\+`. BUT WHY!? :(

```
$ cat RAWOUT.tmp | sed 's%\(^[^     ]\).*%\1%'
R
a
a
r

$ cat RAWOUT.tmp | sed 's%\(^[^     ]\+\).*%\1%'
RPC	Units/1795612.js.t/input.js	/^Test.RPC =$/;"	c	class:Test
asyncMethod	Units/1795612.js.t/input.js	/^	asyncMethod: function($/;"	m	class:Test.RPC
asyncRequest	Units/1795612.js.t/input.js	/^	asyncRequest: function($/;"	m	class:Test.RPC
request_id	Units/1795612.js.t/input.js	/^	request_id: 0,$/;"	p	class:Test.RPC

$ cat RAWOUT.tmp | sed 's%\(^[^     ]+\).*%\1%'
RPC	Units/1795612.js.t/input.js	/^Test.RPC =$/;"	c	class:Test
asyncMethod	Units/1795612.js.t/input.js	/^	asyncMethod: function($/;"	m	class:Test.RPC
asyncRequest	Units/1795612.js.t/input.js	/^	asyncRequest: function($/;"	m	class:Test.RPC
request_id	Units/1795612.js.t/input.js	/^	request_id: 0,$/;"	p	class:Test.RPC

$ cat RAWOUT.tmp | sed 's%\(^[^     ]*\).*%\1%'
RPC
asyncMethod
asyncRequest
request_id
```

b4n:
Huh.  Well, this can be worked around if it's really the problem.  Either with `[^	]{1,}` or `[^	][^	]*`.  It's clearly not as good, but if it's all it takes…

b4n:
(of course you need to use `\{1,\}` when not using extended REs)

vhda:
And the `*` also seems to not be "greedy" in OSX sed.

vhda:
```
$ cat RAWOUT.tmp | sed 's%\(^[^     ]\{1,\} \)\([^/ ]\{1,\}/\)*%\1%'
RPC	input.js	/^Test.RPC =$/;"	c	class:Test
asyncMethod	input.js	/^	asyncMethod: function($/;"	m	class:Test.RPC
asyncRequest	input.js	/^	asyncRequest: function($/;"	m	class:Test.RPC
request_id	input.js	/^	request_id: 0,$/;"	p	class:Test.RPC
```

b4n:
Hehe, doesn't that look good? :)

vhda:
Soooo pretty! :D

b4n:
BTW, tested and works unmodified with GNU sed too :) (well, I had to re-put the tabs because GitHub ate them, but that's all)

vhda:
Now look at the expected file of simple.php.t test unit and be amazed! Apparently the special character there breaks GNU sed.

b4n:
Argh, indeed…  I already saw GNU sed having difficulties with some bytes.  And I guess it doesn't break OSX sed, so we have different results there?

vhda:
Removing all `LC_*` variables and setting `LANG=C` appears to help on my system. Could you please test on yours?

b4n:
hey, good thinking!
```shell
$ cat Units/simple.php.t/expected.tags | LANG=C sed 's%\(^[^   ]\{1,\} \)\([^/ ]\{1,\}/\)*%\1%'
```
works here too :)

masatake:
(The expected.tags of *.t test cases are generated by a script. *.h come from old Test directory.
 We have to verify *.t/expected.tags. It an expected.tags under *.t is o.k., it should be renamed to *.d.)

b4n:
@vhda Apparently the env thing doesn't work like this -- at least not on Travis.

BTW, ideally you would split up the changes before the final merge, first to fix the tests on GNU sed, and then on OSX :)

@masatake should we really rename?  I though it behaved the same, and somewhat allows to know the test comes from the old Test, doesn't it?
Anyway the problem here doesn't come from the test itself but from the sed command that failed on lines with "invalid" (as for the locale) bytes.

b4n:
@vhda but OSX is happy, yay :)

masatake:
@b4n, renaming is not must. Verifying is the important. We need something marker which indicates the expected.tags for a *.t is already verified or not. This is a different topic here. I wrote about renaming and verifying here because I want to tell you that you can modify the expected.tags of *.t test case freely for fixing this issue.

b4n:
@masatake oh, okay, makes sense :)

b4n:
@vhda BTW, your previous patch worked for me (Debian unstable), but I actually only need to set `LANG=C`, the rest is not required here (admittedly I don't have LC_* in the env).

vhda:
I was not looking here, sorry...
I'm using Debian Wheezy and it also worked, but it doesn't in Travis.
Let me try the patch using `file` and `iconv`, if it doesn't work we can go back and try playing around with `LANG`. Maybe we could simply detect the encoding with `file` and set `LANG` to that, if it isn't "ascii" or "binary".

vhda:
ffs...

b4n:
@vhda I just created a test repository https://github.com/b4n/travis-sed-test so it should be faster to see the issues :)
Now I need to enabled Travis, but apparently their website is sluggish at the moment…

vhda:
Great idea! Thanks :)

b4n:
Okay, found it!  Travis [has `LC_ALL=en_US.UTF-8`](https://travis-ci.org/b4n/travis-sed-test/builds/52462484#L103), so you need to either unset it (together with all other `LC_*`, but travis only [has `LC_CTYPE`](https://travis-ci.org/b4n/travis-sed-test/builds/52462484#L143)), or simply set it to `C` as it (apparently) overrides the specific ones (and `LANG`).

So I guess the easiest way is to simply `export LC_ALL=C` (like https://github.com/b4n/travis-sed-test/blob/master/check)

vhda:
Sorry for not being to help you further, but I had to travel yesterday.
We need to guarantee that it works on any system, independently of the user's and system's settings.
According to [here](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html) you are correct: `LC_ALL` overrides everything else, so it should be the best option.

vhda:
ups! :D

vhda:
@masatake are these changes in `misc/units` ok? Or would you prefer to implement it some other way?

b4n:
LGTM, and to Travis too! :)

vhda:
Can't believe it's passing! :D

b4n:
@vhda yeah, after all this it kinda sounds like Christmas<sup>1</sup> in February :D

*<sup>1</sup> or whichever is your favorite gift-giving holiday*

vhda:
In Portugal it's Christmas ;)

masatake:
Thank you!

Issue #93, opened 2014-10-21
Build currently failing on os x lion: 'environ' undeclared


lastobelus:
The build is currently failing (on os x lion):

```
./configure
make install
...
gcc -I. -I. -DHAVE_CONFIG_H  -DDATADIR=\"/tmp/share/ctags\" -DPKGCONFDIR=\"/tmp/etc/ctags\" -DPKGLIBEXECDIR=\"/tmp/libexec/ctags\" -g -O2 -Wall -c main.c
main.c: In function ‘sanitizeEnviron’:
main.c:536: error: ‘environ’ undeclared (first use in this function)
main.c:536: error: (Each undeclared identifier is reported only once
main.c:536: error: for each function it appears in.)
make: *** [main.o] Error 1
```

PR #258, opened 2015-02-28
Fix environ retrieval on OSX

b4n:
_NSGetEnviron() returns a pointer to the environ variable, not the
environ variable itself.

---

Issue #93 lead to 1f00f07bbfabfaa275ad8ac1dffc6084415c3724, but [a recent successful Travis OSX build](https://travis-ci.org/fishman/ctags/jobs/52533979#L162) shows a very suspicious compiler warning:

```
main.c:467:4: warning: incompatible pointer types assigning to 'char **' from 'char ***'; dereference with * [-Wincompatible-pointer-types]
        e = _NSGetEnviron();
          ^ ~~~~~~~~~~~~~~~
            *
```

I couldn't find the canonical documentation for `_NSGetEnviron()`, but all other use of it I saw indeed dereference the return value<sup>[1](https://git.gnome.org/browse/glib/tree/glib/genviron.c#n307), [2](http://www.open-mpi.org/community/lists/devel/2007/05/1596.php), [3](https://lists.torproject.org/pipermail/tor-commits/2012-February/039848.html)</sup>.

Though **beware** that I did **not** test this, as I don't have access to an OSX machine.  So it'd be great if someone could try it out first -- especially with something that would actually trigger something visible in that code, like this:
```console
$ SOMETHING='() { ;echo hello;' ./ctags --version >/dev/null
ctags: Warning: reset environment: SOMETHING=() { ;echo hello;
```

### DIFF
--- a/main.c
+++ b/main.c
@@ -464,7 +464,11 @@ static void sanitizeEnviron (void)
 #if HAVE_DECL___ENVIRON
 	e = __environ;
 #elif HAVE_DECL__NSGETENVIRON
-	e = _NSGetEnviron();
+{
+	char ***ep = _NSGetEnviron();
+	if (ep)
+		e = *ep;
+}
 #endif
 
 	if (!e)


masatake:
I'm afraid the case  _NSGetEnviron returns NULL.
How about adding following code for guarding from the case?
{
  char *** eptr = _NSGetEnviron();
  if (eptr)
     e = *eptr;
}

b4n:
> I'm afraid the case _NSGetEnviron returns NULL.

I don't think it's possible as IIUC it's supposed to return a pointer to the `environ` variable, but you're right, let's be on the safe side.
Patch updated.

jonthn:
If you're interested you can find very succinct documentation/information for NSGetEnviron here : 
- https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man7/environ.7.html  (look towards the end of the man page)
- http://opensource.apple.com/source/Libc/Libc-167/crt_externs.c

b4n:
@jonthn I saw the first link, but not the second, thanks.

So indeed it suggests in theory `_NSGetEnviron()` could theoretically return `NULL` for the dynamic code path if the symbol isn't found (though, I guess it's only theoretical).  So anyway, the `NULL` check is good.

b4n:
(and it also shows that `_NSGetEnviron()` really returns a pointer to the `environ` variable, not the variable itself, so the dereference is right)

PR #259, opened 2015-03-04
ada: fulfill formatting conventions

bmateusz:
At first, don't be frightened of the huge changeset!
I made some reformatting and suggest you to check it first with ignore-space-changes flag, there are only a few real changes.
vi:set tabstop=4 shiftwidth=4:

### DIFF
--- a/ada.c
+++ b/ada.c
@@ -310,7 +310,7 @@ static void makeSpec (adaKind *kind);
 /* prototypes of functions for manipulating the Ada tokens */
 static adaTokenInfo *newAdaToken (const char *name, int len,
 								  adaKind kind, boolean isSpec,
- 								  adaTokenInfo *parent);
+								  adaTokenInfo *parent);
 static void freeAdaToken (adaTokenList *list, adaTokenInfo *token);
 static void appendAdaToken (adaTokenInfo *parent, adaTokenInfo *token);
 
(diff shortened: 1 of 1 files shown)


b4n:
Hum, GitHub doesn't shot the diff because it's too large or something, so… here you go:

1. `git show -w` indeed shows there are no actual changes.

2. There are whitespace issues, like:
    ```diff
+static adaTokenInfo *newAdaToken (const char *name, int len,
+                                  adaKind kind, boolean isSpec,
+                                  adaTokenInfo *parent);
```
this last line starts with a space, it should not.  `git show --check` (same option for `git diff`) can show you some (possible) problems.

3. While style consistency is nice, this file was relatively consistent by itself, so I don't know how much sense reformatting it completely makes -- e.g. it makes using `git blame` a whole lot harder.

bmateusz:
@b4n Thank you for the comment, this --check parameter was new for me, but helpful. Fixed the problems.

Regarding your 3rd point, you are right, it doesn't make much sense to make a change like that. But as a newcomer this was the first thing I noticed, this is the only file which has different formatting. This was something that I personally did not like, so this is the reason why I changed that.

vhda:
I tend to agree with @b4n on this. No need to completely modify one file because of its indentation, specially when we don't have a rule for it.
It would be nice to have a vi modeline in each file with the specific settings that file uses, but then again I don't know how many people are using vi in this project :)

ffes:
@vhda I do like to keep the vim modelines. Although I am mainly a Windows developer, I use Linux as well and on a Linux console/xterm I use vim. And I am even writing a plugin for notepad++ (my favorite windows editor) to recognize those modelines ;-)

vhda:
@bmateusz maybe you could add a Vim modeset line to the ada parser file, such that the current indent of the file becomes correct and automatically enabled in Vim? In my opinion this would be a better solution for this particular issue.

PR #264, opened 2015-03-16
A backport of make.c from Geany

### LINKED ISSUE #263: make: add new kind for targets
With a larger Makefile, it would be useful to have tags for the targets, too.

E.g. for `test` in:

    test:
        @echo "Let's ship!"

ffes:
This adds support for targets and fixes #263


ffes:
@b4n Since this is a backport from Geany, could you review it?

ntrel:
I think it would be good to add a comment before this line in `input.mak`:

    $(obj)/raid6int1.c:   F := 6

The `$()` part prevents parsing of a target name.

ntrel:
BTW I haven't actually tested this pull, only Geany.
```make
target:
	cmd -g FOO=BAR
```
The above code creates a macro named `cmd -g FOO`. We can probably accept this for now but at some point it would be nice to fix. Perhaps adding this to `input.mak` too with a FIXME comment would be good.

b4n:
@ffes The import look good.

The Geany parser is not perfect either though (@ntrel pointed one example), but things can always be improved later.

Reading this made me find a bug in the line continuation code (empty continuation lines after a continuation will not be recognized), so that's good :)  I fixed it in Geany, so maybe you want to import that as well (or add it later, your call).

masatake:
Go ahead!
This is quite important feature.

ffes:
I have added the commit @b4n was talking about and added a test case for this.
I also added two test cases to the known bugs @ntrel mentioned.

b4n:
Apart the possibly inadequate result for currently failing tests, LGTM

masatake:
semi-fuzz test is passed:
    make fuzz LANGUAGES=Make VG=1

masatake:
noise test is passed:
make noise LANGUAGES=Make VG=1

ffes:
Since the last comment are about the details of the expected results for bugs yet to fix, I merge this PR.

The details on what should be the expected tags for the known bugs in the make parser can be discussed in the PR for those bugs. We now know these bugs exist and we have reminders not to forget about them through the units test.

PR #265, opened 2015-03-20
Auto-detect missing coffeetags under rbenv

UmkaDK:
Due to the way executables are installed by rbenv, `type coffeetags`
returns true if coffeetags executable is missing in the current ruby
version but is available in one of the other installed rubies.

This commit fixes this issue by explicitly checking rbenv for the
availability of the coffeetags executable within the current ruby
version.

### DIFF
--- a/libexec/drivers/coffeetags
+++ b/libexec/drivers/coffeetags
@@ -27,9 +27,13 @@
 #
 #
 
-if ! type coffeetags > /dev/null 2>&1; then
-    exit 77
-fi
+{
+    if  [[ (! $(type coffeetags)) || \
+        ($(type rbenv) && ! $(rbenv which coffeetags)) ]]
+    then
+        exit 77
+    fi
+} > /dev/null 2>&1
 
 case "$1" in
     --list-kinds*)
--- a/libexec/drivers/coffeetags
+++ b/libexec/drivers/coffeetags
@@ -28,10 +28,16 @@
 #
 
 {
-    if  [[ (! $(type coffeetags)) || \
-        ($(type rbenv) && ! $(rbenv which coffeetags)) ]]
-    then
-        exit 77
+    COMMAND_NOT_FOUND=77
+
+    if type rbenv; then
+        if ! rbenv which coffeetags; then
+            exit ${COMMAND_NOT_FOUND}
+        fi
+    else
+        if ! type coffeetags; then
+            exit ${COMMAND_NOT_FOUND}
+        fi
     fi
 } > /dev/null 2>&1
 


libexec/drivers/coffeetags
```diff
@@ -27,9 +27,13 @@
 #
 #
 
-if ! type coffeetags > /dev/null 2>&1; then
-    exit 77
-fi
+{
+    if  [[ (! $(type coffeetags)) || \
+        ($(type rbenv) && ! $(rbenv which coffeetags)) ]]
```
b4n: `[[ ]]` is a bashism, and this script uses `/bin/sh` so shouldn't use bashisms (or should be a bash script).  Also, I'm not so familiar with `[[ ]]` but it looks a bit weird here… does this check the output of the commands and all the test are for (non) emptiness of the output??

Some testing confirms that a POSIX variant could be:
```shell
[ -z "$(type coffeetags)" -o -n "$(type rbenv)" -a -z "$(rbenv which coffeetags)" ]
```

However, couldn't we just use a normal shell construct, like this?  Only thing it needs is for the various commands to return a meaningful value.
```shell
type rbenv && ! rbenv which coffeetags || ! type coffeetags
```

Anyway, the check looks a bit off to me, because I would imagine someone could be running outside of rbenv yet having rbenv in the PATH, couldn't it?  If so, we'd have to check we are *actually* in a rbenv… anyway, I don't know anything about Ruby, so I might be wrong -- and I don't mind much.

Also, I don't think the test is even exactly the one you intended.  IIUC, what you want is "if `type rbenv` succeeds, then exit 77 if `rbenv which coffeetags` fails.  Else, if `type rbenv` fails, return 77 if `type coffeetags` fails".  This is not exactly what you have here, while it'll probably work in practice if as you say `type coffeescript` always succeed if `rbenv which coffeetags` does. What you have is "if `type coffeetags` fails or if `type rbenv` succeeds and `rbenv which coffeetags` fails, return 77", which means that if both `type rbenv` and `rbenv which coffeetags` succeed but `type coffeetags` does not, return 77.

UmkaDK:
b4n, you've made a number of very good points, so let me address them in order:

> `[[ ]]` is a bashism, and this script uses `/bin/sh`

You are absolutely right. This is an oversight on my part, so I'll update the PR to be POSIX compliant.

> Someone could be running outside of rbenv yet have `rbenv` in their PATH

This is also true. The way rbenv works is that after it is initialised `rbenv` is available as a function. Therefore, testing for the presence of such a function (eg: `[ "$(type -t rbenv)" = "function" ]`) would give us a reliable way to identify active rbenv instances. I'll update the PR to reflect this change.

> What if both `type rbenv` and `rbenv which coffeetags` succeed but `type coffeetags` does not?

When initialises, rbenv creates a surrogate for every executable installed by any version of ruby it manages. These surrogates are updates every time rbenv changes current version of ruby in such a way that they always point to either a real executable or a generic notice stating that this executable is currently unavailable. Thus, I can't think of a single scenario where `rbenv which coffeetags` could succeed but `type coffeetags` would fail.

However, safeguarding ourselves from this unlikely scenario does us no harm, and saves us from a potential headache later, if rbenv ever changes its current behaviour. So, once again, I'll update the PR. :)

libexec/drivers/coffeetags
UmkaDK: b4n, you've made a number of very good points, so let me address them in order:

> `[[ ]]` is a bashism, and this script uses `/bin/sh`

You are absolutely right. This is an oversight on my part, so I'll update the PR to be POSIX compliant.

> Someone could be running outside of rbenv yet have `rbenv` in their PATH

This is also true. The way rbenv works is that after it is initialised `rbenv` is available as a function. Therefore, testing for the presence of such a function (eg: `[ "$(type -t rbenv)" = "function" ]`) would give us a reliable way to identify active rbenv instances. I'll update the PR to reflect this change.

> What if both `type rbenv` and `rbenv which coffeetags` succeed but `type coffeetags` does not?

When initialises, rbenv creates a surrogate for every executable installed by any version of ruby it manages. These surrogates are updates every time rbenv changes current version of ruby in such a way that they always point to either a real executable or a generic notice stating that this executable is currently unavailable. Thus, I can't think of a single scenario where `rbenv which coffeetags` could succeed but `type coffeetags` would fail.

However, safeguarding ourselves from this unlikely scenario does us no harm, and saves us from a potential headache later, if rbenv ever changes its current behaviour. So, once again, I'll update the PR. :)

UmkaDK:
The above commit addresses all of the issues discussed above: tests are now POSIX compliant and are split in such way that `rbenv which coffeetags` is used on systems with `rbenv` and `type coffeetags` is used on systems without it.

After a bit of digging around it turned out that detecting users who are running outside of `rbenv` yet have `rbenv` in their PATH, is unnecessary as `rbenv which` will transparently return the path to the executable even if rb environment is not initialised.

UmkaDK:
The above commit addresses all of the issues discussed above: tests are now POSIX compliant and are split in such way that `rbenv which coffeetags` is used on systems with `rbenv` and `type coffeetags` is used on systems without it.

After a bit of digging around it turned out that detecting users who are running outside of rbenv yet have `rbenv` in their PATH, is unnecessary as `rbenv which` will transparently return the path to the executable even if rb environment is not initialised.

masatake:
Should we handle this in ctags side?

I'm afraid we have to add supports for pyenv, perlenv, phpenv, ...
Preparing the way to override `type' command used in the coffeetags driver via something environment variable is one of approach. However, I'm not sure this is good or not.

masatake:
BTW, thank you for trying coffeetags driver.

UmkaDK:
:) I see your point, but my argument would be that coffeetags is written in Ruby. As such, `rbenv` directly influences the drivers ability to locate the executable. So I do feel that rbenv auto-detection does indeed belong in  the coffeetags driver. 

On a side note, `rbenv` and `rvm` are the primary ways of running multiple version of ruby side by side, which can provide multiple copies of the same executable.  As far as I know, `type` works as expected under `rvm`. However under `rbenv`, if an executable is missing in the current ruby version but is available in one of the other installed rubies, `type` points to a substitute executable which is created by rbenv in order to inform the user in what other versions the executable can be located.

I'm not aware of any circumstances under which pyenv, perlenv, phpenv, etc would affect coffeetags behaviour.

masatake:
o.k.

I got a question. When running coffeetags command, don't we have to use rbenv?

UmkaDK:
Not really… rbenv manages multiple versions of ruby by modifying current working environment in a very transparent way. Thus, we don't need to give `coffeeetags` any special consideration when executing it, since the current environment is inherited by subprocesses.

Here is what rbenv does in a bit more details:

rbenv allows you to run multiple versions of Ruby side-by-side by managing multiple Ruby environments. Once rbenv is installed, all you have to do is go into a directory and do `rbenv install 2.1.1` (for example). This installs your chosen version of ruby in `${RBENV_ROOT}/versions`, and adds shims for all executable to the `${RBENV_ROOT}/shims`. From that point on, every time you cd into the original install directory your shell environment is updated so that the shims always point to the executables in your desired version of Ruby.

The part that confuses `type` command is that shims for executable that do not exist in your desired version of Ruby, but are present in other versions, point to a generic "help script". When executed, this script simply says that the executable is not available in this version of ruby, and lists all other versions of ruby in which that executable is available. After this, the shim exits with `127` (failed). Unfortunately, this means that `type` on a non-existent executable reports its shim and exits with `0` (success).

masatake:
Thank you for the explanation.

PR #267, opened 2015-03-27
Change kind names for Erlang ctags parser

bengtan:
... so the tags will appear in Geany.

I've requested some changes at https://github.com/geany/geany/pull/445.
This is the contribution upstream. Thanks.

### DIFF
--- a/erlang.c
+++ b/erlang.c
@@ -32,8 +32,8 @@ static kindOption ErlangKinds[] = {
 	{TRUE, 'd', "macro",    "macro definitions"},
 	{TRUE, 'f', "function", "functions"},
 	{TRUE, 'm', "module",   "modules"},
-	{TRUE, 'r', "record",   "record definitions"},
-	{TRUE, 't', "type",     "type definitions"},
+	{TRUE, 'r', "struct",   "record definitions"},
+	{TRUE, 't', "typedef",     "type definitions"},
 };
 
 /*


masatake:
This breaks the compatibility.

In my understanding this patch is for introducing the compatibility between ctags at github and geany.
Am I right?
I know its importance. However, this breaks the compatibility between older ctags and newer ctags.
This is more important for me.

masatake:
If I add something like --output-in-geany-compatible-format, does it satisfy your purpose?

bengtan:
Hi,

I don't know anything about the relationship between geany and ctags. I'm just a geany user working on some Erlang code and I wanted to improve the ctags functionality in geany to be more useful.

The geany maintainers said they would like to keep geany ctags as closely in sync with upstream as possible so they suggested that, whatever changes I made in geany, should also be communicated upstream. I'm just following their guidance. 

Perhaps one of the geany maintainers can come and answer your question better than I can.

vhda:
Honestly, I don't understand why Geany does not allow configurable types. It could use defaults, of course, but an (advanced) option could be included for each supported language that would allow the user to choose other values.

There are dozens of of editors out there and a similar amount of plugins for each editor. Does it make sense to add an argument to support each editor/plugin pairs?

b4n:
> The geany maintainers said they would like to keep geany ctags as closely in sync with upstream as possible so they suggested that, whatever changes I made in geany, should also be communicated upstream. I'm just following their guidance. 

I'm the one at fault here.  I should have mentioned that these kind name changes shouldn't be upstreamed, because it's a problem in Geany not in CTags.  The things I though about when I mentioned upstreaming was the logic for extracting the tags and the like, not how they are reported.
Sorry for the confusion.

In the future we'll fix this in Geany so no alteration of kind names will be needed any more.

> Honestly, I don't understand why Geany does not allow configurable types. It could use defaults, of course, but an (advanced) option could be included for each supported language that would allow the user to choose other values.

Yes, the current way of altering parser kinds in Geany is not good.  We plan on moving away from this by having mappings for each parser's kind to the generic types Geany knows semantics about, it's just not done yet.

> There are dozens of of editors out there and a similar amount of plugins for each editor. Does it make sense to add an argument to support each editor/plugin pairs?

Agreed, I don't think adding a new option is sound.  The problem here is in Geany's side, not CTags.

PR #272, opened 2015-04-04
make: Various fixes and improvements

b4n:
A follow up to #264 fixing all bugs I know in Make support, and adding a few features.  Fixes all our current tests, and adds a few.

### DIFF
--- a/Units/make.comment-in-rule.d/expected.tags
+++ b/Units/make.comment-in-rule.d/expected.tags
@@ -1,0 +1,1 @@
+all	input.mak	/^all:$/;"	t
--- a/Units/make.comment-in-rule.d/input.mak
+++ b/Units/make.comment-in-rule.d/input.mak
@@ -1,0 +1,17 @@
+# all this is one single rule
+all:
+	a=1; echo "$$a"
+# foo
+	b=2; echo "$$b"
+  # foo
+	c=3; echo "$$c"
+# foo
+	d=4; echo "$$d"
+# foo
+	e=5; echo "$$e"
+
+
+	f=6; echo "$$f"
+ # foo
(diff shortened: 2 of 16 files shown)


ffes:
FYI: All the tests pass on Windows as well

masatake:
@b4n, I also looked at the code. Though vStringStripTrailing() rejects an empty string, findMakeTags()
accepts the empty string in newMacro:

  vStringTerminate (name);
  vStringStripTrailing (name);
  newMacro (name); <<<<

So I think we need something guard condition like
  
  if (!isEmpty) {
     newMacro(name);
 }

b4n:
> I also looked at the code. Though vStringStripTrailing() rejects an empty string, findMakeTags()
accepts the empty string in newMacro:

Well, I deliberately didn't check for empty strings here because `makeSimpleTag()` does the check on its side so I figured there was no need to add special handling for the unlikely case of an empty defined name.

This said, `isSpecialTarget()` is slightly buggy on empty names too: it accesses the first byte without a check.  This is not really important because `vString`s always have a non-zero-length buffer, but it's not very clean.  I'll fix this part.

With these in mind, do you still think the `newMacro()` call should be guarded?  I don't mind much really.

masatake:
Thank you. I'm wrong. No needs the guard.

b4n:
Added 8a52d51b9174b00d5ecdb67b10913280309f77b8 to fix the small `isSpecialTarget()` issue, even though the bug cannot happen in practice, and even if it did it wouldn't have any impact.

masatake:
@b4n, thank you. Both fuzz and noise targets are passed. It is ready to merge.

PR #274, opened 2015-04-06
Fix D access modifiers

jakeprobst:
So previously ctags would think D was C++ and look for `public:`.
That actually doesn't work in D.
D uses Java/C# style modifiers where each method/var has its own:
`private int a() { ... }`
It also does its own thing:
`private {
    void a() {
        ...
    }
    void b {
        ...
    }
}`

### DIFF
--- a/Units/d-accessmod.d/args.ctags
+++ b/Units/d-accessmod.d/args.ctags
@@ -1,0 +1,1 @@
+--fields=nksSaf
--- a/Units/d-accessmod.d/expected.tags
+++ b/Units/d-accessmod.d/expected.tags
@@ -1,0 +1,35 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Darren Hiebert	/dhiebert@users.sourceforge.net/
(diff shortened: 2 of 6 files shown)


vhda:
It would be nice if you could include a test unit to demonstrate the change you implemented and to avoid future regressions.

masatake:
About the unit test see docs/units.rst.

ntrel:
It's good you're working on this, but D does have `public:` - from http://dlang.org/attribute.html:

```
AttributeSpecifier:
    Attribute :
    Attribute DeclarationBlock

Attribute:
    ...
    ProtectionAttribute

ProtectionAttribute:
    private
    package
    package ( IdentifierList )
    protected
    public
    export
```
So please revert the `simple.d` changes.

jakeprobst:
The documentation was unclear and I couldn't find any examples with the C++ style so I thought it wasn't supported.
It should support all methods of access declarations now.

masatake:
Thank you for contributing.
fuzz target in Makefile says ctags with your patch crashes in following operation::
  
  [yamato@x201]~/var/ctags-github% cat /tmp/broken.d
  a::public
  [yamato@x201]~/var/ctags-github% ./ctags -o - /tmp/broken.d
  zsh: segmentation fault (core dumped)  ./ctags -o - /tmp/broken.d

jakeprobst:
Ok, it it now passes all tests done by `make fuzz`

masatake:
Thank you.

PR #278, opened 2015-04-08
Add removal of autom4te.cache to maintainerclean target

ffes:
At the moment the directory ``autom4te.cache`` is not removed when ``make maintainclean`` is used.
This PR fixes that.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -341,6 +341,7 @@ distclean: clean
 
 maintainerclean: distclean
 	rm -f $(AUTO_GEN)
+	rm -rf autom4te.cache/
 
 #
 # implicit rules


ffes:
For some reason one of the Travis tests fails, but I have no idea why. This PR should not effect the Travis tests. ``make maintainerclean`` is never called by Travis

vhda:
That is strange. Could you please resubmit the patch to force a new travis run? Maybe it was an issue with one of its burial machines.

ffes:
Thanks to http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit found the rebuild button and it turned out to be a Travis problem.

vhda:
Great! And thanks for sharing that link.

b4n:
LGTM

PR #279, opened 2015-04-10
Sh fix a memleak for broken heredoc input and wrong memory access

masatake:
(Both fuzz and noise with VG=1 are passed.)


sh.c
```diff
@@ -179,7 +184,7 @@ static void findShTags (void)
 			{
 				aliasFound = FALSE;
 				functionFound = FALSE;
-				++cp;
+				cp += (*cp == '\0')? 0: 1;
```
b4n: I'd rather write
```c
if (*cp != '\0')
    ++cp;
```
than use a ternary with a no-op

b4n:
Apart from that cosmetic detail, LGTM

b4n:
(I re-ran one of the OSX Travis tests that stalled, all is good now)

masatake:
@b4n, thank you. I reflected your comment to the change.

PR #280, opened 2015-04-12
Sh: deal arguments passed to source command as "source" tag 

masatake:
Request for reviewers:

Tell me better name for the kind.

+       { TRUE, 's', "source", "read files"}


masatake:
Sorry, I created this request with wrong rebased version.

PR #281, opened 2015-04-12
Sh: deal arguments passed to source command as "source" tag

masatake:
Request for reviewers:

Tell me better name for the kind if you can find:

    { TRUE, 's', "source", "read files"}


vhda:
Isn't this equivalent to `#include` and the like? Should we consider adding support for these entries in other language parsers?

Verilog, for example, also has an `` `include ``definition.

masatake:
In my experience this kind of information should be recorded to tags fie.
Actually the parser of python records "import" information.

masatake:
@vhda, I rethink about your comment.
What I wrote is wrong.

Something included or imported should be recorded by default if it makes something namespace.
import of pythin is an example.

If it doesn't make a namespace, it may be recorded. But the recording should be turned off by default.
The user can choose the behavior. #include of C and source of sh are examples.

Recoding import not making a namespace is abuse, however it is quite useful for understanding a set of source files.

vhda:
That makes sense.
But I fail to identify the introducing of an option to enable `K_SOURCE` in the patch you merged. Will you be adding it later?

vhda:
Again, I didn't see the obvious... I apologise for that.

masatake:
Try --sh-kinds=+s .

Issue #282, opened 2015-04-12
Handle files starting with "#compdef" as zsh/sh


blueyed:
With zsh's source (git://git.code.sf.net/p/zsh/code) and installation, there are many completion scripts, which start with a line like the following:

    #compdef git git-cvsserver git-receive-pack git-upload-archive git-upload-pack git-shell gitk tig

Most of them do not have emacs/vim modelines.

It would be nice if ctags would consider these to be sh/zsh, when `--guess-language-eagerly` is used.

`#autoload` should be handled in the same way.

masatake:
Interesting. I prepared a PR.

PR #283, opened 2015-04-13
Use "#compdef" and #autoload" as modeline pattern for guessing zsh input

masatake:
#copmdef and #autoload in the first line of input suggest the input is
zsh. This patch utilize this hint.

Suggested by @blueyed in #282.


blueyed:
Awesome, thanks a lot! :)

PR #284, opened 2015-04-14
Cork, new API for holding and recording scope information

masatake:
See "docs: write about cork API" about API.
New API is just applies to clojure.c. If this patch set is accepted, I will work on the other
parsers.


masatake:
I withdraw this because it doesn't work well when -e option is specified.

PR #287, opened 2015-04-18
Adapt configure process for custom executable name

jonthn:
Add 2 arguments to configure script : --enable-{ctags,etags}-custom-name
At the same time transform the man page to be processed accordingly to those values

Feel free to take this patch as-is or modify it. I'm not sure if I didn't miss an hardcoded value that would benefit or require a modification to work with this change.


masatake:
Thank you for submmiting a PR.
It seems that you wrote from scratch.
I wonder AC_ARG_PROGRAM, a built-in macro of autoconf  is not help for your purpose.

jonthn:
My first try was using AC_ARG_PROGRAM but I must say I'm still learning (and don't really like) autotools but I'll take a second fresh look.

masatake:
It looks like you made more commits stacked on older commits.
Could you merge some of them?
You may have to know two git command:

1. git rebase -i 
    In the interaction mode, you can unify some commits into one with fixup or squash.
    e.g. 9aba681 should be "fixup"'ed to 5e01fa7.
    Reviewers want to know/review the way how to fix ctags itself from you.
    Reviewers don't want to know/review the way how to fix you commit.
    All later ones should be converted to the former ones.

2. git push --force
     With this you can overwrite your branch.

masatake:
In addition could you write about your change to docs/news.rst?
I would like to know an example how to transform the name from configure.

jonthn:
@masatake No problem I was planning to squash those commits but first wanted a review

I will also add the change in news.rst you requested

masatake:
Could you squash before requesting pulling?
Unsquashed change is quite difficult to review. e.g. b97c312 and ccb9cdd.

jonthn:
I didn't request in-between but each time I push a change github automatically update this pull request.
And as I don't know .rst syntax very well I wanted to check on github so you saw those commits, sorry about that

vhda:
There some online editors that you can use for that purpose, a google search will easily give you such a list. Another trick is to copy the text to a comment box in github a use the preview mode.

jonthn:
@vhda I took the easy road and copy/pasted some text above. But thanks for the advice I'll use it next time

masatake:
@jonthn, thank you very much.
This is a quite important feature for building regression test facility.

PR #289, opened 2015-04-22
Support m4 macro language

masatake:


masatake:
Something wrong....

masatake:
Now, it works. It seems that cannot use \< in optlib to share it with people using etags on the other platform.

PR #290, opened 2015-04-23
Move the rest of C code to core subdirectory

masatake:


masatake:
Could you reflect your comment and the result of reviewing directory to sepdri2 branch?

masatake:
Thank you. The next one is related to VPATH as @b4n suggsted.

PR #291, opened 2015-04-24
c: record the "#include" file as a kind 'I'

masatake:
NOTE: chars around included file (", <, and >) are recorded,
too.


ffes:
Great idea! LGTM.

At first I was not sure about the extra characters surrounding the filename, but giving it a second thought I think you are right. When you use those tags for building a tree or something like that it surely is useful information.

vhda:
I only use tags inside vim and it already is able to identify included files by itself without searching for tags. Assuming that there are more people in the same situation as myself, I think that having this as an option would make more sense. Personally, I would have it disabled by default to keep it consistent with older versions, but as long as there is an option I'm OK with it.

vhda:
Should have looked at the patch morecarefully before posting. My bad :)

ffes:
@vhda This feature is an option and is disabled by default.

ffes:
@vhda See your second comment appear when I submitted mine :-)

vhda:
In my defense, I'm still in bed... :P

ffes:
The Linux-clang job of Travis was having troubles to start. I had to restart it a couple of times but now it succeeded.

masatake:
Thank you for comments.
Recording the information between files is new area for ctags.
I will write something interesting application when I cannot find anything more to work on ctags:-P.
Try adding the same kinds to your favourite languages.

masatake:
TODO: Deal #import of objc.

PR #292, opened 2015-04-24
lxcmd: accept -x and -e option

masatake:
xcmd backend cannot provide locations(positions in the source file)
for each tag entries to ctags. The locations don't recorded in a tags
file but are needed for generating a cross reference and TAGS file.

Therefore xcmd has a restriction: it doesn't work with -x and/or -e
options.  In this commit I introduced location
resolver(readSourceLineSlow) to remove the restriction. As the name
shown it is very slow but better than nothing.

Now xcmd based parsers are really first class citizens in ctags world.

Here is the example of generating the corss reference using coffeetags
backend:

%  ./ctags -x --data-dir=./data --libexec-dir=./libexec --options=coffee Units/xcmd-coffeetags.d/input.coffee
Animal           class         3 Units/xcmd-coffeetags.d/input.coffee class Animal
Horse            class        14 Units/xcmd-coffeetags.d/input.coffee class Horse extends Animal
Snake            class         9 Units/xcmd-coffeetags.d/input.coffee class Snake extends Animal
constructor      function      4 Units/xcmd-coffeetags.d/input.coffee constructor: (@name) ->
move             function      6 Units/xcmd-coffeetags.d/input.coffee move: (meters) ->
move             function     10 Units/xcmd-coffeetags.d/input.coffee move: ->
move             function     15 Units/xcmd-coffeetags.d/input.coffee move: ->

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -630,19 +630,8 @@ generator and exuberant-ctags. This is the reason why the name
 *drivers* is used as part of built-in search path.
 
 To write a driver for a tags generator, please read
-"xcmd protocol and writing a driver".
+-"xcmd protocol and writing a driver".
 
-There are some restrictions of utilizing the xcmds:
-
-doesn't work with ``-x``.
-
-  ctags cannot generate cross reference file if
-  ``--<LANG>-xcmd=COMMAND`` is specified.
-
-doesn't work with ``-e``.
-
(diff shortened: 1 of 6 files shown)


masatake:
I withdraw this because it is immature.
Test facility for etags output and xref ouput may be needed.

Issue #294, opened 2015-04-25
Rename 'core' directory ?


jonthn:
Hello,

Today I noticed two issues : 

* When running configure (with the recent changes) you get early on
```
grep: ./ctags.h: No such file or directory
grep: ./ctags.h: No such file or directory
, version
```

because in configure.ac there are 2 variables set by grepping the file $srcdir/ctags.h which doesn't exist anymore as it has been moved into the core directory.

* The second problem is relatively minor but using the name "core" for a directory could be problematic in certain situations and doesn't give a proper ```./configure``` output (as you can see below)

For example in this thread ( http://gnu-autoconf.7623.n7.nabble.com/rm-f-core-cause-some-troubles-td18281.html)

>Naming a directory 'core' is dangerous and non-portable, because the name 'core' is traditionally reserved for core dumps.

```
checking for mkstemp... rm: core: is a directory
yes
checking for opendir... rm: core: is a directory
yes
checking for strerror... rm: core: is a directory
yes
checking for clock... rm: core: is a directory
yes
checking for remove... rm: core: is a directory
yes
checking for truncate... rm: core: is a directory
yes
checking for setenv... rm: core: is a directory
yes
checking for regcomp... rm: core: is a directory
yes
```

masatake:
@jonthn, thank you for reporting. Could you try #295?
@ffes, could you test the branch on windows?

jonthn:
@masatake I've just tried it and it fixes the issues on my side. Thanks for the quick change. :)

masatake:
@jonthn, thank you for checking.

PR #295, opened 2015-04-26
Rename ./core to ./main

masatake:
Suggested by @jonthn in #294.


ffes:
Your search and replace magic worked as expected.

FYI, will be AFK next week

masatake:
@ffes, thank you.

PR #296, opened 2015-04-26
Cork, new API for holding and recording scope information (resubmitting)

masatake:
See "docs: write about cork API" about API.
New API is just applies to clojure.c. If this patch set is accepted, I will work on the other
parsers.


masatake:
etags outut and xref output are tested.

masatake:
TODO: -x option and -e option should be added automatically.

masatake:
> TODO: -x option and -e option should be added automatically.

I found I already implemented the feature. I cannot remember when I did...

PR #301, opened 2015-04-28
landscape: Ignore units

b4n:
I have no clue about it, but apparently someone enabled landscape.io here?  If we do that, we're gonna want to ignore Units.  Patch here (should) do it.

### DIFF
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,0 +1,2 @@
+ignore-paths:
+    - Units


b4n:
Hum.  This doesn't make any sense, [landscape.io is apparently Python-only](https://docs.landscape.io/faq.html)?  So why do we have this?

masatake:
As explained in #310,  landscape is already disabled. We don't need to code to ignore already disabled thing.

PR #303, opened 2015-04-30
 mib: add an optional pattern for IMPORT

masatake:


masatake:
I recognized mib parser needs complete revising.
I withdraw this.

PR #309, opened 2015-05-06
Enable debug.[ch]

masatake:
Features for debugging provided by debug.[ch] is quite useful for development.
We have no plan to release, so there is no reason disabling the features.


b4n:
Maybe we would also add a configure flag to enable/diable debugging?  or at least something like `DEBUG_CFLAGS` or `DEBUG_CPPFLAGS` make variable a caller could override?

like
```make
DEBUG_CFLAGS ?= -DDEBUG -g

.c.$(OBJEXT):
    ... $(DEBUG_CFLAGS) ...
```

masatake:
@b4n, thank you for the suggestion. I reflected your comment to the commits.

PR #312, opened 2015-05-12
c: add <> and "" to include kinds only if --extra=+q is given

masatake:
In c917dc681cf44ebdbe21bcb9d2400f4fa9e0eb34, I introduced
'I' kind to C parser to record the name of included header files.

For containing more information separators around the file names (<,
", >) are recorded. After thinking this is too aggressive as default
behavior.

With this patch the separators are recorded only if
--extra=+q. By default only the names of header files without
separators are recorded. If --extra=+q is given both type of tags are
recorded: only a name of header file and a name of header file with
separators.

### DIFF
--- a/Units/parser-c.r/c-include.d/args.ctags
+++ b/Units/parser-c.r/c-include.d/args.ctags
@@ -1,2 +1,1 @@
---c-kinds=+I
-
+--c-kinds=+Hh
--- a/Units/parser-c.r/c-include.d/expected.tags
+++ b/Units/parser-c.r/c-include.d/expected.tags
@@ -1,13 +1,13 @@
-"../f.h"	input.c	/^#include "..\/f.h"$/;"	I
-"b.h"	input.c	/^#include "b.h"$/;"	I
-"h.h"	input.c	/^#include"h.h"$/;"	I
-"j.h"	input.c	/^#include       "j.h"$/;"	I
-"sys/d.h"	input.c	/^#include "sys\/d.h"$/;"	I
-<../e.h>	input.c	/^#include <..\/e.h>$/;"	I
-<a.h>	input.c	/^#include <a.h>$/;"	I
-<g.h>	input.c	/^#include<g.h>$/;"	I
-<i.h>	input.c	/^#include       <i.h>$/;"	I
-<stdio.h>	input.c	/^   #    	include <stdio.h>$/;"	I
-<sys/c.h>	input.c	/^#include <sys\/c.h>$/;"	I
+../e.h	input.c	/^#include <..\/e.h>/;"	h
+../f.h	input.c	/^#include "..\/f.h"/;"	h
 D	input.c	/^#define D /;"	d	file:
 M	input.c	/^M(void)$/;"	f
+a.h	input.c	/^#include <a.h>/;"	h
+b.h	input.c	/^#include "b.h"/;"	h
+g.h	input.c	/^#include<g.h>/;"	h
+h.h	input.c	/^#include"h.h"/;"	h
+i.h	input.c	/^#include       <i.h>/;"	h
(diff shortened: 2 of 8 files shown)


masatake:
@ffes, how do you think about this change?

b4n:
I find it odd to record the quotes at all, and especially as part of the tag.  So not having it seems good to me, but adding it even with `+q` seems a bit odd to me?

What's the use case of having the quotes?  In C the only meaning is that the current file's directory is searched first when using `"`.  So, OK, a tool *could* make use of this to extract includes and then search for them in the correct order.  But if this is really wanted, IMO it'd be better to have 2 separate tag kind than changing the content.

masatake:
@b4n, thank you for commets. Yes, I'm thinking about a tool. 
Do you have ideas about the kind names? 
How do you think following?
S: System header file
L: Local header file

masatake:
Temporary I used S and L. 


> ./ctags --list-kinds=C | grep '^[A-Z]'
S  included system header file name [off]
L  included local header file name [off]

The names are o.k.? or odd?

b4n:
> Temporary I used S and L. 
> 
> […]
> 
> The names are o.k.? or odd?

I would have used `H` and `h` (for 2 kinds of "headers" -- as `I` is already taken), but I don't know if it's really better.  But having the same letter with different kind seem nice to me, as they represent almost the same thing.

vhda:
Adding noise: it would have an added value if this could be coherent across parsers.

masatake:
@vhda, I don't understand what you wrote.
> Adding noise: it would have an added value if this could be coherent across parsers.

Here, what does "this" point?

masatake:
@ffes, how do you think? You are the initial supporter of the original code:).

After rethinking again, I found the imaginary tool can distinguish between a system header or local header from the regular expression pattern in tags file. So we don't have to different kinds for them.
The tool developer may have to write some extra code.

vhda:
@masatake I didn't have access to the repo at the time, but I just confirmed that no other parser uses `h` or `H`. So if similar support is added then we can use the same letters.

ffes:
Isn't it just a scope? An include tag, with the scope local or system?

masatake:
@vhda, thank you and I understand your intent. However, it is difficult to keep the coherency between parsers. The coherency is ideal but generally it is just restrictions for ctags developers.

 Each language has each own concepts. The concepts may map to ctags kinds.
So assigning kinds should be done by each parser developer.
If we introduce a rule, we may have to confirm a new parser keeps the rule. It means we have to learn the the concept of the language the parser targets. Do you want to do it?
I don't want(We don't have much time, don't we?).

So I think we should not introduce rules for coherency.

We think H or h is good for here. But python has kinds 'i' for import.
's' is used for source in sh.

Who wants the information for such imported or included files? As @b4n wrote, tools may want.
Such tool may wants as many information as possible. The tool may drop non-interesting data internally. For such kind of tools, I already implemented a good option for enabling kinds:

  --*-kinds=*

As far as I know geany utilized unified kinds rule. It is meaningful because it is requirement of the IDE, a real application. The requirement of real application is more important.

masatake:
@ffes, Do you mean something like following:

> foo.h  input.c /^#include <foo.h>/;"  filesystem:system h
   bar.h input.c /^#include "bar.h"/;"    filesystem:local h

Interesting idea but we have to add a new kind filesystem or something here.
Further more system(after filesystem:) and local (after filesystem:) are not appeared on source code.
It is something like a file tag.

masatake:
... I found F kind letter is used many languages.
Though F is also used for file tag... I will open an issue for ths topic. sigh.

masatake:
I found vera uses c preprocessor.

masatake:
As I wrote I just use 'h'.

PR #313, opened 2015-05-12
Trying fix sf-bug:364

masatake:


masatake:
Confirmed by the original bug reporter:
https://sourceforge.net/p/ctags/bugs/364/

PR #316, opened 2015-05-14
c: record goto labels

masatake:
'L' is used  as a letter for the kind.
This kind is disabled by default.


masatake:
make fuzz LANGUAGES=C++,C VG=1
make noise  LANGUAGES=C++,C

are passed.

PR #319, opened 2015-05-15
verilog: Remove redundant comment skip functions

vhda:
This functionality is already implemented in vGetc() function.

I did not submit this at the time because I thought I might still need skipToEOL() and did not want to remove it immediately. But to keep the code clean it's just better to submit this and reintroduce the function later if necessary.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -422,41 +422,6 @@ static int skipPastMatch (const char *const pair)
 	return vGetc ();
 }
 
-static void skipToEOL (void)
-{
-	int c;
-	do
-	{
-		c = vGetc ();
-	} while (c != EOF && c != '\n');
-}
-
-static void skipComments (int c)
-{
-	int p;
-
-	if (c == '/')
-	{
-		c = vGetc ();
-		if (c == '/')
-		{
-			skipToEOL ();
-		}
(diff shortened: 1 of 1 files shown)


vhda:
See issue: https://github.com/fishman/ctags/issues/318#issuecomment-102384551.

masatake:
I understand well you want to remove skipToEOL to kepp the code clean.
However, if you remove it temporary and revert it later, (in my understanding of git) the result of git blame becomes unusable. So it is better that submitting/committing the final fix first then removing skipComments. So the committing history of skipToEOL can be kept.

vhda:
I'll leave this PR open for now. To be completely honest I no longer remember where I intended to use skipToEOL()...

masatake:
I see.

b4n:
LGTM

About `SkipToEOL()`, I don't think it's so complex a function that we should really bother about whether it might get re-introduced someday.

vhda:
I think this is good for merging.
After some analysis I think that at the time I created this patch I was considering using skipToEOL() for pre-processor macros, but I ended using a different approach.
With this said, I don't think I'm going to need this function any longer, so there's no problem in keeping the code clean.

vhda:
P.S. - Rebased code over master

PR #321, opened 2015-05-16
recognize Fortran pointer initialization

QingmingHe:
fortran.c doesn't recognize Fortran pointer initialization like:
```Fortran
type(my_type), pointer :: p => null()
```
This problem has been solved in http://sourceforge.net/p/ctags/bugs/322/. I just copied the code there.


b4n:
FWIW I think there is a lot of change in Geany that weren't (yet) imported:
```
* 257de65 fortran: Make sure not to index kinds out of bounds
* d8954bb Fortran: add support for "len" and "kind" qualifiers
* ed1dc50 Fortran: allow keywords as names
* e47d45e Fortran: fix handling of preprocessor directives on the first line
* ef4c725 Fortran: generate fake tags for anonymous structures, interfaces and enums
* 9520e7f Fortran: parse Fortran 2003 enums
* 329b2cb Fortran: don't generate tags for interface content
* 5a67a4f Fortran: report interfaces
* 29336ff Fortran: display components
* 362a5cc Fortran: properly report scope for programs, subroutines and types
* 7d76063 Fortran: fix handling of pointer association operator
* 1c0fa99 Support F2003 array syntax and F2008 coarrays
* 497bfae Fortran: fix line continuation inside comments
* 2085e8a Symbol browser breaks with END ASSOCIATE
* 194ef60 Tagmanger: recognize Fortran procedure pointers as variables
* 2f31f74 Fix tag generation for Fortran with 'forall' blocks
```

QingmingHe:
Oh, no wonder I've encountered many problems when tagging my project written in Fortran 2003/2008 using Ctags. I'll try to look into Geany and import these changes.

masatake:
@QingmingHe, thank you for taking your time to improve Fortran parser.
I have a question: which commits come from Geany?
( I would like to record the origin for each commit log.)

Surprisingl there are test cases for etags output. I guess you may write them.

QingmingHe:
How can I tell you which commits come from Geany? By amending the commit message? I'm sorry I'm not that familiar with work flow of Git and Github.

Yeah, I'm a heavy Emacs user!

b4n:
> How can I tell you which commits come from Geany?

I guess ideally you'd keep the authoring, date and probably commit message information for the Geany commits.  And perhaps if deemed useful add a note in the commit, like maybe an URL pointing to the relevant commit on *git.geany.org*.

masatake:
Though I added the way to test TAGS(etags) output, could you convert your test case to tags(ctags) output? 
Comparing with TAGS output, tags output is more informative; information like kind is dropped in TAGS output. In other word TAGS provides the subset information of tags. So a test should be done against tags output. The intention of my extending the test facility to cover TAGS output was for testing ctags internal tags to TAGS converter. In many case the converter may work fine.

I know you are an Emacs user. Me, too:-)

masatake:
With git rebase, you can remove "Merge branch 'master' of https://github.com/fishman/ctags" commit from your pull request.

masatake:
As @b4n write, I need the pointer to the original commit for each new commit.
It seems that you wrote test cases. Am I correct? If yes, I would like you to put a test case to a separated commit.

Some commits included modifications for dot files like .gitgireno. It should be in a separated commit.

Your PR is not small. If you split the commits carefully, I can do cherry-pick some of your commits incrementally. As the result, we can reach a hot-spot for reviewing and updating.

QingmingHe:
> Though I added the way to test TAGS(etags) output, could you convert your test case to tags(ctags) output?

I'll convert these tests. Thanks for your explanation.

QingmingHe:
> Why dir-local.el must is here?

It's not indispensably needed. But may provide some

Issue #318, opened 2015-05-14
xcmd recongnizes coffeetags not avaliable  &  verilog.c line 927 error


huanghx1990:
i dl the newest version code just now，and installed in my redhat . then when i type in ctags -R in my uvm_1.2 lib src folder, i got a msg said : 

ctags:Notice:xcmd recongnizes /PATH/usr/libexec/ctags/drivers/coffeetags is not avaliable.
ctags: parsers/verilog.c:927:tagNameList:Assertion 'IsIdentifierCharacter(c)' failed 
Abort.

I wanna use ctags in systemverilog language,which is my proj used lang,but encounter above issues.hope some masters could help fix the error, P,S:i dl the code two days ago i just encounter the first err,but today i encouter the two issues.

vhda:
Could you please guarantee that your UVM source code was extracted correctly?
Until now I only saw the second error when a file ended prematurely.

@masatake could you please comment on the coffeetags error?

masatake:
xcmd related message is just notice. It is not a critical issue as far as not using coffeetags.
For your purpose you can ignore it.

@huanghx1990, could you tell me the version of OS and how did you install ctags. I would like to know the command line especially parameters passed to configure script.

masatake:
About the assertion, we need input file.
I'm afraid it is something to do with multibyte string.
Our ctags may accept only ascii chars.

masatake:
Could you show the output of

              $ /ctags -R   --verbose 2>&1 | tail

? So we can find the input file name which stops ctags.

vhda:
UVM 1.2 is available for download at: http://accellera.org/images/downloads/standards/uvm/uvm-1.2.tar.gz
I'm able to replicate this problem on my side, so allow me to debug the issue.

vhda:
I'll need a bit more time to debug this further, but I think I have made some bad decisions in the SystemVerilog parsing.

masatake:
Shrink says just one byte '/' makes the assertion failure.

> % bash misc/units shrink "./ctags -o /tmp/tags %s" /tmp/uvm-1.2/src/tlm2/uvm_tlm2_time.svh /tmp/tags
[0       9249]...failed(1)
[0          0]...ok(0)
[0       9249]...failed(1)
[0       4625]...failed(1)
[0       2313]...failed(1)
[0       1157]...failed(1)
[0        579]...failed(1)
[0        290]...failed(1)
[0        146]...failed(1)
[0         74]...failed(1)
[...]

PR #323, opened 2015-05-17
verilog: Remove unnecessary Assert() call

vhda:
This assertion was in fact verifying the code, which is not the purpose
of this parser. By removing this check the file will continue to be
parsed and any other contents will be correctly detected.

Rename test cases that were failing because of this assertion call.


vhda:
This should resolve issue #318.

b4n:
LGTM, the code following don't seem to depend on the assertion.

masatake:
It looks o.k.

PR #325, opened 2015-05-18
C family: Fix support for trigraphs and add support for digraphs

b4n:


jonthn:
Thanks for the fix. I just tried it on my file and it works with both value of ```--if0```

masatake:
Other than the definition of the ungetc buffer, LGTM.

PR #326, opened 2015-05-18
C#: Don't ignore the character following an '@'

b4n:
The character following an '@' was dropped if it didn't start a string literal.

This could lead to unexpected problems if '@' was valid in other situations.

---

This change has currently no effect, but could come in handy if at some point anyone is interested with the character following an `@` when the language `hasAtLiteralStrings`.

### DIFF
--- a/main/get.c
+++ b/main/get.c
@@ -753,6 +753,8 @@ process:
 						c = skipToEndOfString (TRUE);
 						break;
 					}
+					else
+						fileUngetc (next);
 				}
 			enter:
 				Cpp.directive.accept = FALSE;


masatake:
LGTM.  You are talking about objective-c, aren't you?

b4n:
No, I'm talking about [whichever language(s) setting `hasAtLiteralStrings`](https://github.com/fishman/ctags/blob/master/main/get.c#L113), which seems to [only be C#](https://github.com/fishman/ctags/blob/master/parsers/c.c#L3145)

b4n:
Ah, apparently it's because Objective-C doesn't use `get.c`.

masatake:
I found @ is used for defining a string literal. Sorry for making noies.

PR #328, opened 2015-05-18
Add annotation as a kind apart from interface (sf-bug:360)

champo:
The use of default values in annotations was added to Java allow for
syntax such as:

    String message() default "{constraints.fieldmatch}";

This syntax caused ctags to go mad since it doesn't expect to find a
keyword after a closing parenthesis. Which makes sense since it's not
allowed anywhere else in the syntax. Thus, the need for an indepedent
annotation kind.

The new kind enables specific behaviour for this use cause without fear
of breaking parsing for other cases.

### DIFF
--- a/Units/review-needed.r/bug1777340.java.t/expected.tags
+++ b/Units/review-needed.r/bug1777340.java.t/expected.tags
@@ -1,3 +1,3 @@
 bug1777340	input.java	/^public @interface bug1777340 {$/;"	i
-d	input.java	/^    String n() default "ninjas; monkeys!";$/;"	f	interface:bug1777340
 m	input.java	/^    String m();$/;"	m	interface:bug1777340
+n	input.java	/^    String n() default "ninjas; monkeys!";$/;"	m	interface:bug1777340
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -35,6 +35,8 @@
 #define isType(token,t)     (boolean) ((token)->type == (t))
 #define insideEnumBody(st)  ((st)->parent == NULL ? FALSE : \
                             (boolean) ((st)->parent->declaration == DECL_ENUM))
+#define insideAnnotationBody(st)  ((st)->parent == NULL ? FALSE : \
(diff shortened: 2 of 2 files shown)


champo:
This fixes #211 completly

masatake:
Thank you for submitting the chage here. I will take a look.

masatake:
(Feel free to comment this PR. I assign myself because I invited @champo from https://sourceforge.net/p/ctags/bugs/360/ .)

masatake:
About the commit log, you wrote "this use cause". I think this should be "this use case". If I'm correct, could you fix your commit log and do "git push --force" ?

About coding style, I don't wan to talk much because I have broken many areas. However I would like  you to follow the style about position of `{` and `}` to other area of ctags source code.

Your code
> if (...) {
} else {
}

the style of ctags
> if (...)
{
}
else
{
}


You introduced a function named processAnotation. It should be processAnnotation, shouldn't it?


Maybe because my English ability I don't understand some part of your commit log.
You use words "annotation kind" and "new kind". What are you talking about? As far as reading diff, you don't introduce any new kind for annotation of Java.


I don't know well about Java. How do you think introducing a new kind 'a', "annotation"?
Could you tell me your idea?

masatake:
http://sourceforge.net/p/ctags/mailman/message/34173798/

champo:
@masatake as far as I understand, annotations are considered interfaces in the formal java spec. See `InterfaceDeclaration` in https://docs.oracle.com/javase/specs/jls/se7/html/jls-18.html. I don't think a different kind should be added.

As far as the PR, I agree with several of your points. I'll get to those as soon as I can!

masatake:
@champo, thank you for the comment. Yes, it is an interface in the spec. However, if ctags records it as 'a'(annotation), I think tools using tags output can do something interests. 
What I can image is it helps a context oriented completion engine.

masatake:
o.k. Let's pend the decision about introducing new kind for annotation.

masatake:
Merged manually. Thank you. About indentation it is not consistent even in c.c. So I did nothing about it.
I just fixed the typo.

Issue #317, opened 2015-05-14
Conflicting in kind letter F


masatake:
Bad news.

Incompatible change may be needed to fix this issue.

F used as file tag in ctags main part, is also used for other kinds in some parsers.
Here is an example of ruby.

> % cat /tmp/foo.rb 
cat /tmp/foo.rb 
class ClassExample
        def self.class_method
            puts "in class_method"
        end
end        
% ./ctags --extra=+f  /tmp/foo.rb ; cat tags | grep -v '!' | grep F
./ctags --extra=+f  /tmp/foo.rb ; cat tags | grep -v '!' | grep F
class_method	/tmp/foo.rb	/^        def self.class_method$/;"	F	class:ClassExample
foo.rb	/tmp/foo.rb	1;"	F

ctags --list-kinds says:

ObjectiveC
    F  Object field
Ruby
    F  singleton methods
Rust
    F  A method
SQL
    F  record fields

masatake:
About regex parser and xcmd parser,  I would like to forbid to use 'F'.
For crafted build-in parsers I will prepare the way to override 'F' kind letter.

I think Not-changing the meaning of (widely used) tags file is the more important than consistency between parsers. Any comments?

vhda:
I think that would be a nice addition, since "main" uses that specifically to create file tags and we do not control what is added in the regex patterns.
Another option would be to abort when both the language parser has an 'F' kind and the include filenames option is enabled.

On the built in parsers this can be modified, but how can we avoid overlooking updates that add kinds using this letter?
It would be nice if "main" would parse the "kinds" variable and guarantee that no parser would do that, but having such a check at run time is just wasting CPU.
Any other ideas? Maybe using pre-processing code?

masatake:
Thank you for comments.

> I think that would be a nice addition, since "main" uses that specifically to create file tags and we do not control what is added in the regex patterns.
Another option would be to abort when both the language parser has an 'F' kind and the include filenames option is enabled.

If a built-in parser uses 'F' for its own purpose and the parser doesn't define an alternative letter for file, ctags stops the execution(assertion).  This is o.k.? ctags can check in per parser initialization stage. Checking may not be so expensive. Checking is done only once when starting the ctags process; not for each input files.

I would like to prepare incrementally. The initial PR will be for regex parser. tmain can be used forr tesintg.

vhda:
That's fine by me.

masatake:
[...]

PR #329, opened 2015-05-19
Make error when F is used in regex parsers

masatake:
See #317. This change is only for regex parsers.


masatake:
I withdraw this PR. See my comment in #317.

Issue #324, opened 2015-05-18
Wrong makeTagEntry usage (sf-bug:356)


masatake:
See http://sourceforge.net/p/ctags/bugs/365/

b4n:
Other languages have the same problem.

One possible way to go would be to require passing a `kindOption` array to `makeTagEntry()` so it could itself check whether to create the tag or not.

This however requires a few non-trivial changes, the most important ones being those under *main/* (parsers are almost trivial to fix, but *c.c* that require a little more work but it totally doable).

The other solution is to simply add the check in each parser missing it (most parsers have it, but not all).

masatake:
@b4n, thank you for fixing quickly. However, I want to think this issue a bit more because this is related to the one of my biggest questions about ctags.

If this issue can be handled main/ side, this kind of bug can be suppressed forever.
As you write, passing a kindOption array is one of solution. I think this one is worth trying.

If accessing the current parserDefinition from makeTagEntry () is permitted, we don't have to pass a kindArray to makeTagEntry. By Introducing getCurrentLanguage or adding parser field to tagsEntryInfo structure, we can pass the parserDefinition object to makeTagEntry(). However, this may break isolation of code. I am not sure.

During reading source code of ctags, I found that accessing to a parserDefinition is  limited in many areas. For me the the limitation is a bit too string. @b4n, how do you think? It seems that this limitation is very basic design decision of ctags. However it is not enough clear for me.

masatake:
I would like to try fixing this in main/ side.
Macros in read.h are the public interface of parserDefinition.

step1. Add isEnabledKindInSourceLanguage(c) to read.h.
step2. Implement makeTagEntryWithKindCheck based on isEnabledKindInSourceLanguage.
step3. use makeTagEntryWithKindCheck where @b4n patched in this pull request.

It will not be so difficult because the places where patching is needed are enumerated by @b4n.

masatake:
@b4n, I implemented step1 and step2. Could you look at them?
The name of functions are a bit too long. If o.k. I will do step 3 bsaed on your PR.

PR #330, opened 2015-05-20
Introduce makeTagEntry variant accepting a given tag only if its kinds is enabled

masatake:
The base for alternative main/ side solution for #324.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -1078,6 +1078,14 @@ extern int makeTagEntry (const tagEntryInfo *const tag)
 	return r;
 }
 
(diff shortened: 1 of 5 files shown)


masatake:
I withdraw this.
To use makeTagEntryMaybe, a parser has to initialize a tagEntryInfo. Howerver, I guess the programmer of the parser may want to check whether a kind associated with the tagEntryInfo before initalizing the tagEntryInfo. So makeTagEntryMaybe will not be used anywhere.

Adding assertion to makeTagEntry may be good idea.

PR #331, opened 2015-05-21
Language own letter for file kind

### LINKED ISSUE #317: Conflicting in kind letter F

masatake:
This fixes #317.

Cod in this PR does:
* permit defining language own letter for file kind,
* let some parsers like Ruby using F for their own kinds use ! as letter for file kind,
* introduce long long flag fileKind for --langdef option; xcmd and regex parsers can use it,
* introduce --list-file-lind option,
* check the letter confliction about file kind when initializing parsers,
* introduce more tmain tests, and
* describe the changes in command line of ctags.


main/parse.h
```diff
@@ -75,6 +93,7 @@ typedef struct {
 	char* name;                    /* name of language */
 	kindOption* kinds;             /* tag kinds handled by parser */
 	unsigned int kindCount;        /* size of `kinds' list */
+	char fileKind;		        /* override letter for file kind */
```
vhda: Could you please align the indent of this comment? OCD alert! ;)

main/parse.c
```diff
@@ -97,6 +99,19 @@ extern const char *getLanguageName (const langType language)
 	return result;
 }
 
+extern const char getLanguageFileKind (const langType language)
+{
+	char kind;
+
+	Assert (0 <= language  &&  language < (int) LanguageCount);
```
vhda: `language >= 0` would be more readable, don't you agree?

main/parse.c
```diff
@@ -1158,6 +1189,28 @@ static void lazyInitialize (langType language)
 /*
 *   Option parsing
 */
+#define COUNT(D) (sizeof(D)/sizeof(D[0]))
+
+static void lang_def_flag_file_kind_long (const char* const optflag, const char* const param, void* data)
```
vhda: Why no camel back?

vhda:
I like this approach because it allows backward compatibility in external tools using ctags, which would not happen if we modified the kind letters currently used in the parsers.

main/parse.h
masatake: (Fixed in my private branch. )

main/parse.c
masatake: ... I have no option but some people may think " 0 <= language < LanguageCount  " is easier to read. See another Assert in main/parse.c. All of the take the same style.

main/parse.c
masatake: This is really one of my bad. (Another one  is the usage of const.)

When I started working on fishman/ctags, I carelessly introduced many name with underscore.
> ~/var/ctags-github/main% grep _ *.c *.h | grep -v define | grep -v undef | grep -v TAG| grep -v ifdef | grep -v off_t | grep -v readtags  | grep -v endif| grep -v size_t | grep -v ifndef | grep -v '[A-Z]_[A-Z]' | wc -l
650

I introduced so many areas. I mixed the naming styles too many areas.
The unified style in all source code is better but I think it is too late to fix.

masatake:
How about error messages, long flags syntax introduced to --langdef, and documents?

main/parse.c
vhda: What is behind us, is behind us.
But can we agree on a naming style and follow it from now on?
At least in this specific case, that code really stands out when compared to the lines above and below!

vhda:
I'll leave that more in depth analysis for @b4n =)

main/parse.c
masatake: Loog at lregex.c, I use `_` for flag related callback functions.
So the name lang_def_flag_file_kind_long is natural for me.

How about following rules?
About file static varaibles and functions we can choose any style.
About extern functions we should use camel style.

> $  grep _  *.h | grep -v define | grep -v undef | grep -v TAG| grep -v ifdef | grep -v off_t | grep -v readtags | grep -v endif| grep -v size_t | grep -v ifndef | grep -v '[A-Z]_[A-Z]' | wc -l
 grep _  *.h | grep -v define | grep -v undef | grep -v TAG| grep -v ifdef | grep -v off_t | grep -v readtags | grep -v endif| grep -v size_t | grep -v ifndef | grep -v '[A-Z]_[A-Z]' | wc -l
44

Most of all 44 hits are introduced by me. They can be fixed.

For field name, I would like to use `_`.

Do the same against *.c in parsers directlry, I got 2368 hits.

masatake: O.K. I will remove _ from main/*.h.

 > $ cd main; grep _ *.h | grep -v #  | grep -v size_t | grep -v ARG_ | grep -v DEBUG_ | grep -v fpos_t | grep -v _QUOTE | grep -v RESCAN_ | grep -v EX_ | grep -v PCOPROC_ | grep -v CONST | grep -v PARSER_LIST | grep -v OPTION | grep -v SO_ | grep -v METHOD_ | grep -v __ | grep -v jmp_buf | grep -v SYMBOL

All underscores are introduced by me.

vhda: Thanks! That makes the code much more consistent.

masatake: Updated rules are:
* function name, type names and struct memeber name in *.h should not use underscore.
* parameter names in function decl in *.h can use underscore.
* in *.c file underscore can be used.

PR #332, opened 2015-05-23
recognize Fortran procedure pointers as variables

QingmingHe:
Recognize Fortran statement such as:
```Fortran
procedure(subprogram_to_point_to), pointer :: my_pointer
```
Import from Geany https://github.com/geany/geany/commit/194ef6010d25e04bf69f705cf94818f68a55f3fa.

### DIFF
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -110,6 +110,7 @@ typedef enum eKeywordId {
 	KEYWORD_pointer,
 	KEYWORD_precision,
 	KEYWORD_private,
+	KEYWORD_procedure,
 	KEYWORD_program,
 	KEYWORD_public,
 	KEYWORD_pure,
@@ -273,6 +274,7 @@ static const keywordDesc FortranKeywordTable [] = {
 	{ "pointer",        KEYWORD_pointer      },
 	{ "precision",      KEYWORD_precision    },
 	{ "private",        KEYWORD_private      },
+	{ "procedure",      KEYWORD_procedure    },
 	{ "program",        KEYWORD_program      },
 	{ "public",         KEYWORD_public       },
 	{ "pure",           KEYWORD_pure         },
@@ -1109,6 +1111,7 @@ static boolean isTypeSpec (tokenInfo *const token)
 		case KEYWORD_logical:
 		case KEYWORD_record:
 		case KEYWORD_type:
+		case KEYWORD_procedure:
 			result = TRUE;
 			break;
 		default:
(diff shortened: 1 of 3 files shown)


parsers/fortran.c
```diff
@@ -110,6 +110,7 @@ typedef enum eKeywordId {
 	KEYWORD_pointer,
 	KEYWORD_precision,
 	KEYWORD_private,
+        KEYWORD_procedure,
```
b4n: style: spaces instead of a tab.  (this doesn't seem to be in the original patch)

b4n:
FTR, Geany has geany/geany@f02b70949adf2659ebc833a2c8075595a7625ff3 as test case

QingmingHe:
> FTR, Geany has geany/geany@f02b709 as test case

This test can not pass for fishman/ctags. Reasons are when ctags parse:
```Fortran
subroutine sub(x)
  real, intent(inout) :: x(:)

  integer :: i

  do i=1,size(x)
    x(i) = 0.5**i
  end do
end subroutine sub
```
```
sub	input.f90	/^subr/;"	s
```
will be written to tags file. This is an truncate line issue of Fortran parser.

b4n:
While it looks like something that should be fixed, how does it prevent tests having that from passing?  It seems to be reliable (always producing the same result), and it's only the tag excmd, which should not be so very important here, is it?

QingmingHe:
I've submit an issue https://github.com/fishman/ctags/issues/333 and we may discuss the truncation problem there.

masatake:
@QingmingHe, you can define expected.tags as you want four your purpose of testing.

> sub input.f90   /^subr/;"   s

Temporary, how about accepting this output?
 So you can test the fortran parser code you imported from geany, and ctags may  pass the test case coming from geany.  The truncating things can be fixed later.

Your original test case is a good because it is simple. The reason I want to use the test case of geany instead is that we want to close the code base of ctags and that of tagmanager of geany as much as possible.

masatake:
Looks good.
Next one is the final request from me.

Some of your commits including the modifications reflecting comments here.

> git log --oneline master...    
e204fe0 Use the Fortran procedure test from Geany
8d7af7b remove lines start with ! in fortran-procedure test
6e9975f replace spaces with tab
25f085d A Fortran procedure test
109a709 recognize Fortran procedure pointers as variables

I would like to final version. 109a709, 6e9975f can be merged into one.
25f085d and 8d7af7b can be dropped because it is overwritten by e204fe0.
"git rebase -i" will help you.

After rewriting your history, could you do "git push --force"? 
So I can merge the perfect commits.

Thank you.

masatake:
Excellent!

masatake:
Thank you.

QingmingHe:
My pleasure.

PR #335, opened 2015-05-27
Fortran: fix handling of pointer association operator

QingmingHe:
Handling Fortran pointer association as:
```Fortran
procedure(ProcOne), pointer :: MyProc => null()
```
Both of the code and test case are imported from Geany.


masatake:
Thanks.

PR #336, opened 2015-05-28
go: fix missing '-' from '<-' in function prototype

techee:
The whole <- token representation has to be inserted into the prototype,
not just the single character <.

@b4n This is the fix to the problem you observed in Geany.


b4n:
LGTM

masatake:
Thank you.

PR #337, opened 2015-05-28
Fortran: recognize extends qualifier

QingmingHe:
Recognize Fortran 2003 type extension feature which introduces OOP for Fortran language.

### DIFF
--- a/Units/fortran-extends-qualifier.d/expected.tags
+++ b/Units/fortran-extends-qualifier.d/expected.tags
@@ -1,0 +1,9 @@
+color	input.f90	/^    integer :: color$/;"	k	type:shape
+filled	input.f90	/^    logical :: filled$/;"	k	type:shape
+length	input.f90	/^    integer :: length$/;"	k	type:rectangle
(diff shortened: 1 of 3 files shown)


masatake:
Thank you. Looks good to me.

(I don' t know fortran.) It seems that rectangle is derived from shape.
ctags has ability to record the type derivation with --field=+i. How about recording a parent type, rectangle in fortran? It is meaningful?


If yes, here is hints for implementation:

In your patch ctags just does skipOverParen after finding "extends" keyword.
If ctags analyses the token(s?) between the pair of the paren after "extend", parent type can be recored.

tag->extensionFields.inheritance of addOtherFields in c.c is an example of implementing --field=+i.

QingmingHe:
Your seggestion is quite meaningful! I've not released that **--field=+i** option can be used in Fortran parser. I will borrow some ideas from c.c. Maybe I cound fix it this weekend.

masatake:
o.k. I will merge this. Give a patch supporting --field=+i in new PR.
Thanks.

PR #339, opened 2015-05-29
Fortran interface content

QingmingHe:
+ Report Fortran interface by default.
+ Don't report interface content as it is not the definition of Fortran procedures. We don't need these data in tags file. I think that when you press **control-]** in Vim, you want to jump to the definition of procedure rather than its corresponding interfaces.

These two patches are from Geany.


vhda:
Please include the fixes to the existing tests in this PR.

QingmingHe:
ok

QingmingHe:
> Do we have to keep /* ... */ here?

If we want to be consistent with Geany, we have to.

I've rebase the change of code and fix of existing test into one commit(https://github.com/QingmingHe/ctags/commit/ec207733660c3f40aa29e43b5ada26bbcfcb5ced). Is it suitable?

masatake:
@QingmingHe, o.k. Don't remove /* */ in the commit. Could you stack one more commit which removes /* */?

QingmingHe:
oh. the pointer may be wrong. I will fix in seconds.

masatake:
All, how do you think turning on the kind for interface of fortran by default?
ctags @ sf.net turns off it by default. It means there are some impact on tools.

I have 3 ideas.

1. Don't turn on it for keeping the compatibility. Our user must enable it when tags for interface is needed. For testing, args.ctags file is needed: put --fortran-kind=+i to the file.

2. Just turn on. We are going to our own way. 

3. Introduce flavor. If running ctags with --flavor=sf.net, ctags turns off the kind by default.
    If running ctags with --flavor=github, ctags turns on the kind by default.

I am interested in 3. With 3. you can add --flavor=geany. Kind definition is one of the gap between our code and geany code. However, some work is needed in main part.


@QingmingHe, are you using fortran and ctags daily?

QingmingHe:
> are you using fortran and ctags daily?

Yes, I'm working on a project with many Fortran 2003/2008 features. My work flow heavily relies on ctags. That's the reason why I want to improve the ctags Fortran parser and make it support new language standards.

The reasons I turn on Fortran interface kind is model Fortran projects uses interfaces to wrap multiple procedures. With tags for Fortran interface generated, we can experience fluent completion and navigation when coding. Without interface tags, we have to use **grep** to find it. In this sense, user tends to turn it on as he uses ctags longer.


As the compatibility you've mentioned. I think it's really very hard to fix. There are gaps between sf.net ctags and Geany ctags. The question is whether we want to be consistent with sf.net or Geany.

> Introduce flavor. If running ctags with --flavor=sf.net, ctags turns off the kind by default.
If running ctags with --flavor=github, ctags turns on the kind by default.

I think history burden is too heavy. Do you really want to add this option? progammers are prone to accept new tool if is better than the old ones.

As with this PR, you can ask other Fortran programmers and ctags users whether Fortran interface kind should be turned on by default or not.

vhda:
If I was able to understand Fortran Interfaces correctly, they behave as prototypes. That is, in a source code you have the interface declaration and the respective procedure declaration with the same name. Both should be able to provide the same information in the tags file, but when you want to jump to a procedure declaration you don't want to go to the interface instead. I think this is the reason why prototypes are disabled by default.

That said, I would keep the interface disabled by default for coherency reasons. But first I'd like to understand why it appears that the procedure tag and the interface tag do not provide the same information... it is possible I'm overlooking something.

QingmingHe:
The syntax of Fortran is a little bit stange. There are serveral kinds of usage of **interface**:

+ **Wrap** procedure prototypes. In this case interface can be **anonymous** or not. If not anonymous, the name of interface is not limited to the procedure name. In the example below, **suba** and **subb** are prototypes where I don't want to jump. That's the reason I **skip the content of interface inside it**. May be we could add a **prototype** kind in Fortran and turn it off by default as the content of interface inside it behaves much like prototype in C.
```Fortran
! the name of interface don't make difference
interface sub_interface
  subroutine suba(a, b)
    integer, intent(in) :: a
    integer, intent(inout) :: b
  end subroutine suba
  ! I don't want to jump here as it is prototype of subb
  subroutine subb(a, b)
    integer, intent(in) :: a
    integer, intent(inout) :: b
  end subroutine subb
.
.
.
! you can wrap more prototypes
end interface sub_interface
```

+ Define generic subprogram. In this case, the tag for the interface **wrap** is significant for that interface itself defines a **new generic subprogram**. So we should turn on the Fortran interface kind by default in this sense. Again, in this example, we don't want to jump to content of interface inside it.

**The definition** of generic subprogram:
```Fortran
MODULE SWAPPER
       ! you can treat me as a new procedure
       INTERFACE SWAP
              MODULE PROCEDURE SWAP_R, SWAP_I, SWAP_C
       END INTERFACE
CONTAINS

       SUBROUTINE SWAP_R(A, B)
       IMPLICIT NONE
       REAL, INTENT (INOUT)      :: A, B
       REAL                      :: TEMP
               TEMP = A ; A = B ; B = TEMP
       END SUBROUTINE SWAP_R

       SUBROUTINE SWAP_I(A, B)
       IMPLICIT NONE
       INTEGER, INTENT (INOUT)   :: A, B
       INTEGER                   :: TEMP
               TEMP = A ; A = B ; B = TEMP
       END SUBROUTINE SWAP_I

       SUBROUTINE SWAP_C(A, B)
       IMPLICIT NONE
       CHARACTER, INTENT (INOUT) :: A, B
       CHARACTER                 :: TEMP
                  TEMP = A ; A = B ; B = TEMP
       END SUBROUTINE SWAP_C
END MODULE SWAPPER
```
The usage of generic subprogram:
```Fortran
program main
  integer :: a, b
  ! The cursor is here, when I press C-] in Vim,
  ! I want to jump to the generic subprogram defined by interface
  call swap(a, b)
end program main
```

+ Operator overloading. This is something like generic subprogram only that the name of subprogram is the operator or assignment.
```Fortran
        interface assignment(=)
          module procedure matrix_from_real, matrix_from_matrix, &
                           vector_from_real, vector_from_vector
        end interface
                
        interface operator(+)
          module procedure matrix_add, vector_add
        end interface

        interface operator(-)
          module procedure matrix_sub, vector_sub
        end interface

        interface operator(*)
          module procedure matrix_mul, vector_mul, matrix_vector_mul
        end interface

        interface operator(/)
          module procedure matrix_div, matrix_vector_div
        end interface
```

In conclusion, maybe a better approach is turn on interface kind by default and add other Fortran prototype kind to tag the interface content inside it. What do you think @masatake @vhda

masatake:
@QingmingHe, thank you for detailed explanation.

> In conclusion, maybe a better approach is turn on interface kind by default and add other Fortran prototype kind to tag the interface content inside it. What do you think @masatake @vhda

Agree with you. interface should be turned on by default. This means we should cherry-pick ec20773.
Something  prototype is needed. It means we should not merge  30cb282.

@QingmingHe, I would like to cherry-pick ec20773 and  bd5bb76 (with modifying expected.tags.); and close this PR. If you are o.k., could you try to implement "Something  prototype"? "P" can be used for the purpoues:-)

Am I correct?

vhda:
I think that @masatake got everything right, as usual :)

Apologise for not saying anything else but time has been short. And this week ended with a 22h trip from Portugal to San Francisco...

QingmingHe:
> I would like to cherry-pick ec20773 and bd5bb76 (with modifying expected.tags.); and close this PR.

Just close this PR and I'll provide cleaner one. There are more expected.tags to be modified according to ec207733.

> could you try to implement "Something prototype"? "P" can be used for the purpoues:-)

ok.

masatake:
@QingmingHe, I see. Thank you

PR #341, opened 2015-05-29
Fortran: support --fields=+i option

QingmingHe:
This is an improvement of commit 918b281, see also https://github.com/fishman/ctags/pull/337.

For example:
```Fortran
type, extends(shape) :: rectangle
```
The parser just record the word **shape** between paranthesis as the parent derived type and set e.extensionFields.inheritance.


masatake:
I would like to avoid using file static variable unless it is really needed.
Based on your code, I wrote alternative version which uses just local variable.

https://github.com/masatake/ctags/tree/fortran-extends-qualifier-local-variable-based-implementation

Could you look at this?

I wonder using token->secondary for recoding parent type is acceptable or not...

QingmingHe:
Thanks, your implementation is better. I'm considering extend your code to support more fields such as **access**, **implementation** ... as we can get more information from qualifiers. I'm working on it and just wait for a few days. This PR will be limited to the inheritance field and I'll add other fields in following PRs.

> I wonder using token->secondary for recoding parent type is acceptable or not...

Maybe not. The original author just want to use token->secondary to record **end (type|interface|module|...)** token. We'd better not break it.

masatake:
o.k. Add parent_type field and use it for the purpose.
Should I merge my d8a3511? and then will you work on the commit?
In the case I would like you to introduce parent_type field in your side.

QingmingHe:
> Should I merge my d8a3511? and then will you work on the commit?

Need not. I've already pull your commit on which I'm working on. I'll push the modified commit in the weekend.

masatake:
I'm running "make noise LANGUAGE=Fortran VG=1" now.

masatake:
noise target found a memory leak:

Run
> valgrind --leak-check=full -v ./ctags -o - --language-force=Fortran FILENAME

With following intput https://gist.github.com/masatake/6af481ba4f9f678f2c9b

valgrind says
> ==6067== HEAP SUMMARY:
==6067==     in use at exit: 216 bytes in 6 blocks
==6067==   total heap usage: 7,765 allocs, 7,759 frees, 644,128 bytes allocated
==6067== 
==6067== Searching for pointers to 6 not-freed blocks
==6067== Checked 217,456 bytes
==6067== 
==6067== 112 (48 direct, 64 indirect) bytes in 2 blocks are definitely lost in loss record 4 of 4
==6067==    at 0x4A06BCF: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==6067==    by 0x4118F5: eMalloc (routines.c:216)
==6067==    by 0x4135DA: vStringNew (vstring.c:76)
==6067==    by 0x4136ED: vStringNewCopy (vstring.c:117)
==6067==    by 0x4208AC: makeParentType (fortran.c:1260)
==6067==    by 0x421A76: skipOverParensFull (fortran.c:1110)
==6067==    by 0x421B77: parseExtendsQualifier (fortran.c:1266)
==6067==    by 0x421B77: parseQualifierSpecList (fortran.c:1322)
==6067==    by 0x422097: parseDerivedTypeDef (fortran.c:1693)
==6067==    by 0x422097: parseTypeSpec (fortran.c:1234)
==6067==    by 0x4220F4: parseTypeDeclarationStmt (fortran.c:1419)
==6067==    by 0x4226C5: parseDeclarationConstruct (fortran.c:1887)
==6067==    by 0x4226C5: parseSpecificationPart (fortran.c:1947)
==6067==    by 0x422CA2: parseModule (fortran.c:2035)
==6067==    by 0x422CA2: parseProgramUnit (fortran.c:2185)
==6067==    by 0x422CA2: findFortranTags (fortran.c:2231)
==6067==    by 0x40ED03: createTagsForFile (parse.c:1666)
==6067==

masatake:
``` diff
diff --git a/parsers/fortran.c b/parsers/fortran.c
index f849d80..a476eb2 100644
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -1257,6 +1257,8 @@ static boolean skipStatementIfKeyword (tokenInfo *const token, keywordId keyword
 
 >static void makeParentType (tokenInfo *const token, void *userData)
 {
+	if (((tokenInfo *)userData)->parentType)
+		vStringDelete (((tokenInfo *)userData)->parentType);
 	((tokenInfo *)userData)->parentType = vStringNewCopy (token->string);
 }
```

QingmingHe:
I've found some errors in your input file https://gist.github.com/masatake/6af481ba4f9f678f2c9b
```Fortran
  type, extends(shape :: rectangle
```
should be
```Fortran
  type, extends(shape) :: rectangle
```
In the first case, **::** and **rectangle** will be parsed after parsing **shape** in parseExtendsQualifier and **parentType** will point to another block of memory which causes memory leakage.

Or, if I could catch your purpose, you just want to test the first case.......

masatake:
(If possible) ctags should have any memory leakage even if input is broken.
However, we don't care about the quality of tags file for the broken input. We can just generate garbage output from garbage input.

With the patch in my comment, the memory leakage doesn't hapen because the parentType pointing another block of memory is freed when qualifierToken allocated at parseQualifierSpecList().

You may wonder when qualifierToken is freed. It is frreed via trashbox. A trashbox keeps the reference of qualifierToken. A trashbox frees qualifierToken when trashBoxDelete function is called at craeteTagsForFile() in main/parse.c.

Do I make sense?

QingmingHe:
Thanks for your explanation.

masatake:
o.k. I will merge you changes. Then I merge my small memory related change.

QingmingHe:
ok

masatake:
Thank you.

PR #345, opened 2015-06-02
Use verbose() to report unavaiable xcmd backends

### LINKED ISSUE #343: xcmd: coffeetags is not available
ctags complains that coffeetags is not available. In my opinion no warning should be displayed in ctags for a feature that the user is not running.
Would it be possible to update the code such that this warning is only shown if the availability of coffeetags would improve the tags processing of the files being parsed?

masatake:
Close #343


main/lxcmd.c
```diff
@@ -307,7 +307,7 @@ static boolean loadPathKinds  (xcmdPath *const path, const langType language)
 			if (status > 0 
 			    && WIFEXITED (status) 
 			    && (WEXITSTATUS (status) == XCMD_NOT_AVAILABLE_STATUS))
-				notice ("xcmd recognizes %s is not available", argv[0]);
+				verbose ("xcmd recognizes %s is not available\n", argv[0]);
```
vhda: I would use `"xcmd cannot find %s"`, or similar.
@ffes : as our English language reviewer what do you think? :)

masatake: Thank you. For a while I think this sentence is something wrong.
How about:

                  %s reports the backend driver is not avaiable.

?

blueyed: I suggest:

    xcmd: the %s backend is not available.

masatake:
@blueyed, thank you. I use yours.

PR #347, opened 2015-06-03
Couple more exuberant renames

ffes:
There are still quite some places left where the Exuberant should be renamed, but some of those files need to go into the website, others we might delete anyway because they are outdated. But for now these two are here to stay, so they need the rename.

And the PROGAM_URL is likely to change again in the near future


masatake:
Please, go ahead. So we can receive bug reports.

Issue #340, opened 2015-05-29
Homebrew Formula & website


cweagans:
Hello! I just wanted to let you know that I created a Homebrew formula for Mac users that allows them to install Universal Ctags. You can find it here: https://github.com/cweagans/homebrew-universal-ctags

I'd love to move that repo into the universal-ctags organization. If you want to add me to the org, I'll transfer it!

I'd also be interested in helping you guys out with a website, if you're interested. We can use Github Pages for simplicity, and I already own ctags.io, so we could use that for the web address. I helped Neovim with similar tasks and things seem to be running pretty well.

cweagans:
@ffes I think you're the person that I'd need to talk to about this, right?

ffes:
I really like this!
But for now I think it is too early to move your repo to the organization. Not even the main source repo has been moved yet.

I am not a Mac user myself, but I understand that homebrew is a repo for Mac users to get additional Linux(like) tools. And by adding your own tap you can add packages not provided by homebrew (similar to PPA for Ubuntu).

I see there is a formula for exuberant ctags in the homebrew repo.
Is there a change to get them to make the switch to universal? If so, is there any need for us to maintain our own homebrew repo? What would it take to get them to make the switch do you think?

ffes:
And I personally would love to have ctags.io as our URL for our project, but it is not my decision. We have to discuss this with the other regular contributors.

cweagans:
Homebrew is a package manager, yeah. Short term, this tap is probably the right way to go. When Universal Ctags has a 1.0.0 release, then we can submit a PR to the official homebrew repo to change over to Universal Ctags instead. For now, though, this provides an easy way for somebody to get started with Universal Ctags. I don't foresee a lot of maintenance for this tap - it shouldn't change that much as long as the build process stays the same.

Regarding the website, I'd suggest Github Pages for the actual site on ctags.io, and then Read The Docs for any documentation that we want to publish (perhaps on docs.ctags.io)

cweagans:
@ffes The repo has been moved. Would you like to move the Homebrew formula too?

cweagans:
Also, is there a more real-time communication method that contributors are using? IRC or Gitter or Slack or something?

blueyed:
> Also, is there a more real-time communication method that contributors are using? IRC or Gitter or Slack or something?

See https://github.com/universal-ctags/ctags/issues/344#issuecomment-107753545.

I think IRC is fine.  `#ctags` is unused currently.
[...]

PR #350, opened 2015-06-04
Assert a kind passed through a tag entry info should be enabled

masatake:
See #340.

Checking whether a kind is enabled or disabled is a  responsibility of parser side.
This patch set is for adding an assert() to makeTagEntry to reject a taginfo which is associated with disabled kind. This commits may have some performance penalty.


vhda:
Careful with the underscores in the function names again! ;)
Just for the sake of discussion: how hard do you think would be to completely move the ignore code from parsers into core? That is, the parser would blindly add all tags and core would filter out the disabled ones. In my opinion this would simplify parser creation/maintenance.

masatake:
>Careful with the underscores in the function names again! ;)
The underscores are used in file static functions. 

About the responsibility between parsers and main, I think your approach is better. (Here I don't care about performance.) However, as far as reading the source code of ctags for a year, it requires changes of the internal design of ctags.  
Functions in entry.c cannot access to the parser own data directly, and a kind table where enable/disable is managed is defined in parser source file as a file static variable. It seems that this is strong rule in design of ctags.

Redesigning the parser class and allowing from any place in ctags source code in object oriented way are attractive idea. However, it takes long time and I'm not sure it is really right approach or not.

For trying such fundamental changes, we have to improve test facility. 

See also #330.

Issue #349, opened 2015-06-04
Running ctags with --options=.ctags under windows (sf-bug:366)


masatake:
http://sourceforge.net/p/ctags/bugs/366/

Is this reproduce with universal-ctags, too?

@ffes, could you look at this?

ffes:
I guess the reporter of the problem on SF is @vadz

I can confirm that it is a problem on Windows in our repo as well.
Haven't tested it yet under Linux, but I don't think it is platform specific.

.ctags and ctags.cnf both give the error. But ctags.conf doesn't give the error.
Adding --verbose gives a clue.

```
Considering option file /ctags.cnf: not found
Considering option file /etc/ctags.conf: not found
Considering option file /usr/local/etc/ctags.conf: not found
Considering option file C:\cygwin\home\Frank\.ctags: not found
Considering option file C:\cygwin\home\Frank\ctags.cnf: not found
Considering option file .\.ctags: reading...
Considering option file .\ctags.cnf: reading...
Reading initial options from command line
  Option: --options=ctags.cnf
Considering option file ctags.cnf: already considered
ctags.exe: cannot open option file "ctags.cnf" : No error
```

Both filenames that fail are the default filenames already already considered.

In options.c in parseFileOptions() the return value remains false when the file is already considered. 
https://github.com/universal-ctags/ctags/blob/master/main/options.c#L2309
I have will prepare a fix

masatake:
Oh, sorry, it is not platform specific. Feel free to assign to me.

masatake:
Approach:

1. Introduce parseFileOptions(const char* const fileName, bool whetherCheckSameFile);
    This function does the same as current parseFileOptions but calling checkSameFile only if whetherCheckSameFIle is true.

2. Rewrite parseFileOptions
   New parseFileOptions jsut calls paresFileOptions(fileName, TRUE);

3. Modify processOptionFile
    if (!checkSameFile()) parseFileOptions(fileName, FALSE));

Discussion:
   Should we notify to the user that ".ctags is already loaded".

   Loading .ctags of current directory may be unsafe it xcmd is specfined in the .ctags.

   checkSameFile function uses a file static variable named CheckFile. We should not do so.
   Extend stringListHasTest function: boolean stringListHasTest (const stringList *const current, boolean (*test)(const char* s, void *), void *user_data);

vhda:
Why not simply make `parseFileOptions` return TRUE when file is already considered?
[...]

PR #356, opened 2015-06-08
Set return value when option file already considered

ffes:
This fixes issue #349

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -2306,7 +2306,10 @@ static boolean parseFileOptions (const char* const fileName)
 	const char* const format = "Considering option file %s: %s\n";
 	CheckFile = fileName;
 	if (stringListHasTest (OptionFiles, checkSameFile))
+	{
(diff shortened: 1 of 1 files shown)


masatake:
Could you try to make a test under Tmain?

masatake:
I found extra hack was needed in ctags/tmain side. So I wrote a test case for this one by myself. Thank you.

vhda:
Seeing this PR being closed without merging left me confused. Had to check the commit log to understand what happened.

masatake:
@vhda, I'm sorry. I added extra commits to the merged branch.
Should I merge before adding extra commits?

vhda:
Don't worry too much, it's just that I mostly track the updates through github's emails. When this type of things happen or when commits are directly pushed I miss those updates.
But you could have merged, rebased your commits over the latest commit and PR those updates. Although this requires a bit more work, it would give more visibility to the changes, specially if people work like I do :)

PR #359, opened 2015-06-09
json: Fix handling of tags containing a dot

### LINKED ISSUE #357: Assertion failed: function makeJsonTag, file parsers/json.c
When I run `ctags -R .` on my project I get the following output after maybe just 2 seconds:

```
Assertion failed: (token->scopeKind > TAG_NONE && token->scopeKind < TAG_COUNT), function makeJsonTag, file parsers/json.c, line 128.
Abort trap: 6
```

b4n:
Closes #357.

### DIFF
--- a/Units/dotted-names.d/expected.tags
+++ b/Units/dotted-names.d/expected.tags
@@ -1,0 +1,19 @@
+0	input.json	/^    {$/;"	o	array:phone.numbers
+1	input.json	/^    {$/;"	o	array:phone.numbers
+address	input.json	/^  "address": {$/;"	o
+age	input.json	/^  "age": 25,$/;"	n
+children	input.json	/^  "children": [],$/;"	a
+city	input.json	/^    "city": "New York",$/;"	s	object:address
+first.name	input.json	/^  "first.name": "John",$/;"	s
+height_cm	input.json	/^  "height_cm": 167.6,$/;"	n
+is.alive	input.json	/^  "is.alive": true,$/;"	b
+last.name	input.json	/^  "last.name": "Smith",$/;"	s
+number	input.json	/^      "number": "212 555-1234"$/;"	s	object:phone.numbers.0
+number	input.json	/^      "number": "646 555-4567"$/;"	s	object:phone.numbers.1
+phone.numbers	input.json	/^  "phone.numbers": [$/;"	a
+postal.code	input.json	/^    "postal.code": "10021-3100"$/;"	s	object:address
+spouse	input.json	/^  "spouse": null$/;"	z
+state	input.json	/^    "state": "NY",$/;"	s	object:address
+street.address	input.json	/^    "street.address": "21 2nd Street",$/;"	s	object:address
+type	input.json	/^      "type": "home",$/;"	s	object:phone.numbers.0
+type	input.json	/^      "type": "office",$/;"	s	object:phone.numbers.1
--- a/Units/dotted-names.d/input.json
+++ b/Units/dotted-names.d/input.json
@@ -1,0 +1,25 @@
+{
+  "first.name": "John",
+  "last.name": "Smith",
+  "is.alive": true,
+  "age": 25,
+  "height_cm": 167.6,
+  "address": {
+    "street.address": "21 2nd Street",
+    "city": "New York",
+    "state": "NY",
+    "postal.code": "10021-3100"
+  },
+  "phone.numbers": [
+    {
+      "type": "home",
+      "number": "212 555-1234"
+    },
+    {
+      "type": "office",
+      "number": "646 555-4567"
+    }
+  ],
+  "children": [],
+  "spouse": null
+}
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -54,11 +54,18 @@ extern boolean vStringAutoResize (vString *const string)
 	return ok;
 }
 
+extern void vStringTruncate (vString *const string, const size_t length)
+{
+	Assert (length <= string->length);
+	string->length = length;
+	vStringTerminate (string);
+	DebugStatement ( memset (string->buffer + string->length, 0,
+	                         string->size - string->length); )
+}
+
 extern void vStringClear (vString *const string)
 {
-	string->length = 0;
-	string->buffer [0] = '\0';
-	DebugStatement ( memset (string->buffer, 0, string->size); )
+	vStringTruncate (string, 0);
 }
 
 extern void vStringDelete (vString *const string)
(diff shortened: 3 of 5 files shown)


b4n:
However, this raises the question of how to report qualified scopes as virtually any byte can happen in a JSON tag, so using `.` can lead to ambiguities. Maybe using something non-printable like ETX, EOT or FF would make more sense?  Another solution could be to report non-qualified scopes, but that's likely to introduce ambiguities in many files, which isn't better.

masatake:
@b4n, for discussion, could you put a short but unwanted example json input here?

b4n:
> How about introducing following function?
>> void vStringTruncate (vString *string, size_t length);
> […]

I first searched for such a function indeed.  Though, if it's only useful for this parser maybe adding a new vString API isn't needed, but if it can be useful some place else (and such a function could probably), yeah we should add it.

masatake:
Can I ask you adding such function?

b4n:
> @b4n, for discussion, could you put a short but unwanted example json input here?

The test cases shows it already, but imagine this JSON snippet:
```JSON
{
  "some.name.with": {
    "dot": 1
  },
  "some": {
    "name": {
      "with": {
        "dot": 2
      }
    }
  }
}
```
Now if we run ctags on it (aligned for clarity):
```console
$ ./ctags -o- -n input.json
dot             input.json	3;"	n	object:some.name.with
dot             input.json	8;"	n	object:some.name.with
name            input.json	6;"	o	object:some
some            input.json	5;"	o
some.name.with  input.json	2;"	o
with            input.json	7;"	o	object:some.name
```
As you see, both occurrences of `dot` have the same scope.  However, in practice the first one is only one element (`some.name.with`), while the second is 3 elements (`some`, `name` and `with`).  The problem here being that the scope separator character is present in a tag.

Were we using e.g. ETX (ASCII 3), we would have this:
```console
 $ ./ctags -o- -n input.json
dot             input.json	3;"	n	object:some.name.with
dot             input.json	8;"	n	object:somenamewith
name            input.json	6;"	o	object:some
some            input.json	5;"	o
some.name.with  input.json	2;"	o
with            input.json	7;"	o	object:somename
```

For most parsers it isn't really an issue because it's easy to pick a character (or a string) that just cannot be part of a tag name.  E.g. in C using `.`, or PHP using `::` -- those are not valid inside identifiers, and there isn't really any tag to generate but from identifiers.  This issue only arises when a parser allows virtually anything in the tags, like e.g. reST/Markdown/AsciiDoc/… sections -- or JSON, as entries are strings.

b4n:
> Can I ask you adding such function?

Sure.  Do I make another PR or do I add this function in this one and start using it right away?

masatake:
Could you do in this one?

b4n:
Done.  As usual GitHub UI is confused when commits are not in chronological order, but well.

masatake:
I'm thinking following simpler notation:

```
dot             input.json  3;" n   object:some.name.with
dot             input.json  8;" n   object:with
name            input.json  6;" o   object:some
some            input.json  5;" o
some.name.with  input.json  2;" o
with            input.json  7;" o   object:name
```

Showing full path(?) in scope information is handy understandable. But developers of applications which use tags may want coherency. With above tags output, an application rebuild the full path of scope from
object: scope information and line number.

As you write this is not only about json. Here in discussion let's forget the compatibility.
We would like to decide the basic policy of the notation of scope information.

Here is my idea.
* if we can expect we can keep consistency, we can use full path.
* if we cannot, we just use relative path.

This issue is related to --extra=+q option. We can decide the notation of name slot(the first column)
as we like.

b4n:
> With above tags output [using unqualified scopes], an application rebuild the full path of scope from object: scope information and line number.

This is unfortunately not completely true, because although here it can be guessed with line numbers, it wouldn't be possible if all tags (or some unfortunate subset) were on the same line: in this case unqualified scopes become ambiguous. Also, I don't really like the idea of requiring applications to implement such heuristics when we can avoid it.  And simply imagine there was e.g. a second `with` like this:
```JSON
{
  "some": {
    "name": {
      "with": {
        "with": 1,
        "dot": 2
      }
    }
  }
}
```
It would now be impossible to know whether `dot`'s scope is `some.name.with` or `some.name.with.with`.

Qualified scopes are unambiguous if we assume there can only be one tag of the same name at the same scope (plus the last one sharing the same type).  While this assumption is not completely true either, it generally is good enough at the very least.

If we want something truly unambiguous, we could e.g. give IDs to tags and use this as the scope.  This would be totally incompatible and a lot less practical to read for a human, but I guess an application could deal with it reasonably easily.  However, this *requires* scopes to be real tags inside a same file, and that might not be wanted: one example that comes to mind is the class itself in a C++ implementation side (where the class is generally not defined or declared directly):
```C++
Foo::Foo() {}
Foo::~Foo() {}
```
We'd need a tag and and ID for the class `Foo`, not just the current scope of `class:Foo`.

Another more realistic idea might be to output several `scope` extra fields, one for each level (in order), like `object:some	object:name	object:with`.  Maybe however some parsers won't be able to give a real type for all levels, so maybe we would need a generic one like `scope` -- but that'd be a detail.

All this said, I think that it would most probably be good enough to just use a highly unlikely byte, like a low non-printable ASCII character.
And while at it, we might want to lower the need for applications to really know each parser by having a way to query the separator (many parsers already use something specific).

> Here is my idea.
* if we can expect we can keep consistency, we can use full path.
* if we cannot, we just use relative path.

Well, this would mean the scope ambiguity is left unresolved for some parsers.

And while it could work to some extend, then how could an application know whether the scope was qualified and then shouldn't be resolved further or not?  Well, actually this point is somewhat moot as applications already need to know many things about each parser already.  And BTW, some parsers already report unqualified scopes AFAIK.

masatake:
Thank you for great explanation.
How about adding adding the line number of parent like:

```
some            input.json  2;" o
name            input.json  3;" o   object:some scope-hint:line:2
with            input.json  4;" o   object:name scope-hint:line:3
with            input.json  5;" o   object:with scope-hint:line:4
dot             input.json  6;" o   object:with scope-hint:line:4
```

The next challenge your parepre may be:

```
{
  "some": {
    "name": {
      "with": { "with": 1,
        "dot": 2
      }
    }
  }
}
``

How about adding adding the depth of scope like:

some            input.json  2;" o   
name            input.json  3;" o   object:some scope-hint:line:2,depth:0
with            input.json  4;" o   object:name scope-hint:line:3,depth:1
with            input.json  5;" o   object:with scope-hint:line:4,depth:2
dot             input.json  6;" o   object:with scope-hint:line:4,depth:2

masatake:
With cork API used in clojure parser implementing scope-hint may not be difficult.

masatake:
(BTW, how can I this thread be made open status?)

PR #362, opened 2015-06-10
python: Fix resetting scope on anonymous blocks (CTags bugs:356)

b4n:
The previous fix, coming from [CTags bug #1988026], was incorrect if
the parent was not a root-level element, as it checked the level name
(unqualified) against the parent name (qualified).

However, there is no need to check the level name, all what counts is
the indentation level itself: if it's smaller than an existing level,
it ends it.

This fixes [CTags bug #356].

[CTags bug #1988026]: https://sourceforge.net/p/ctags/bugs/227/
[CTags bug #356]: https://sourceforge.net/p/ctags/bugs/356/


vhda:
LGTM

masatake:
LGTM. Could you merge the commit?

Issue #62, opened 2014-08-07
avoid infinite loop


masatake:
--timeout option or something to avoid a parser jumps into infiinite loop.

masatake:
https://lwn.net/Articles/656294/
Interesting article. I'm doing similar thing with https://github.com/masatake/srpmix7.

In these use case Ctrl-C should be pressed by software.

masatake:
Now I understand what you wrote 4 years ago are correct.
timeout should be implemented in the client side.
If ctags has a bug, we should get the report and fix it.

So how can we make a user report such bug easier.
In the situation that ctags run with -R option enters an infinite loop in a user site, what we would like to know is the input that causes the infinite loop. If the ctags process can report the current input file, we can ask the user to send the input file by specifying the file name.

`--verbose` option can report such information, however, it is too noisy.
I'm thinking about introducing `--report-progress` or something new option.

PR #366, opened 2015-06-12
csharp: add missing keywords

kusma:
These don't do all that much with the new keywords yet, but at least they prevent ctags from incorrectly treating keywords as identifiers.


ffes:
You combined 4 commits in one PR, but only one is related to the subject: C#. The others are more general changes to c.c 
Curious what others think of that.

I like adding those C# keywords.

masatake:
@ksuma, thank you for sending a PR.

Do your changes have effects the output of ctags(tags file)?
If yes, we need a test case. See docs/units.rst.

kusma:
@ffes: The first two are simply preparatory patches that makes the following patches a bit easier to review. The last one fixes some keywords that were incorrectly marked as not-supported in C and C++.

@masatake: There's some changes, but the only cases I've been able to trigger are with invalid code, where I use keywords as variable names. Does that warrant test-cases? If so, I'll make them...

masatake:
... The first two screw the history recorded in git. Unified single style may be beautiful. However, we  don't seek the beautify. 

https://github.com/universal-ctags/ctags/commit/ad245287b7eda9f4db2dfa103da694ea3ca926b2
is very special. This is really dangerous.

I wonder what I should do with f5b0f23. I have some clear criterias.

1. Even if input is broken, ctags should not crash. 
   Some members don't agree with this (https://github.com/universal-ctags/ctags/issues/62). However, this is important for me.

2. For broken input, ctags can generate garbages as far as keep the file format of tags.

3. For right input, ctags should try to generate good output. Sometimes this is quite difficult.

> @masatake: There's some changes, but the only cases I've been able to trigger are with invalid code, where I use keywords as variable names. Does that warrant test-cases? If so, I'll make them...

For invalid code, ctags can generate garbages. This is the base line of discussion.

Here following invalid code:
> int continue;

Currently ctags generates:

> continue	/tmp/foo.c	/^int continue;$/;"	v

Yes, "continue" is a keyword of c. It looks acceptable for me.

Something wrong for ctags users including you?

For 
> int if;

ctags generates nothing. Do you say these are inconsistent?

I cannot find good criteria in this situation.

Of course, a keyword table should be expected held keywords of the language.
With such expectation I have a plan to write a new parser guessing engine based on keyword tables. So I want a correct keyword tables. But this is my story.

kusma:
1eb6739 and 224cfd6 more or less undo the damage from d97bf7e. I can easily drop the latter one, but there's litte benefit of dropping the first one, as the re-wrapping in f5b0f23 pretty much touches the same lines.

Alternatively, I could stay away from keeping this list alphabetized, but since it *was* alphabetized (and tab-indented) before d97bf7e, it seems this would just continue down a path that looks like a simple mistake in the first place. And the history-damage was already done in d97bf7e, as far as I can tell.

b4n:
Hum, I'm curious: why add those keywords in the first place?  If they don't change how the parser handles *valid* input, I see very little gain in adding them.  Parsers only need to recognize what they care about, everything else is useless.

I'm not saying those keywords shouldn't be added, but I don't think they should be added for the sake of adding all the keywords.

> And the history-damage was already done in d97bf7e, as far as I can tell.

That one may have broke indentation, but as far as I can see it only touched lines that actually needed changing (the whole keyword table gained a new column, for D).

All this said, I don't have a strong opinion on indentation fixes.  On one hand, I like consistent indentation.  OTOH, I don't like messing with the "browsability" of the history with no good reason.  Though Git is quite handy when it comes to following changes to file's parts (like `git log -L`), so maybe it's not that much of a problem to mess up the output of `git blame`, not sure.

masatake:
As @b4n wrote, I also know the original motivation of adding keywords.

If you don't have any strong motivation, I will not introduce your commits now but I will pick them when I implement keyword based guessing engine.

kusma:
Yeah, I don't have a strong-motivation, so I'm fine with waiting with these patches until they become useful.

masatake:
o.k. someone will reopen this when s/he try to solve #378.

Issue #361, opened 2015-06-10
Test cases under review-needed.r (C)


masatake:
Test cases under review-needed.r comes from e-ctags.
expected.tags files under the directory is generated by u-ctags.
We have to understand the meaning of test cases and add args.ctags if necessary.
If args.ctags is added, expected.tags must be updated, too.

This may be hard work. We are looking for volunteers.

If you reviewed test cases of langauge X, do

0. mkdir Units/parser-X.r,
1. rename test case from *.t to *.d, and
2. move the *.d to Units/parser-X.r/.

I did some part of C language. See parser-c.r.
I will continue work on C language.

If you are interested in working on this task, open an issue and declare the language you will work on.
The volunteer needs the knowledges of Units test facility. See docs/*.rst, especailly docs/units.rst.

Let's make review-needed.r empty!

masatake:
objective-c is done.

cweagans:
Sorry for the noise. Just switching the "Release blocker" tag for the "Initial release" milestone so that everything is tracked in one place.

masatake:
```
% for x in */input*; do ../../ctags --print-language $x; done | sed -e 's/.*: //' | sort | uniq -c
< in */input*; do ../../ctags --print-language $x; done | sed -e 's/.*: //' | sort | uniq -c
      1 Ada
      6 Asm
      1 Asp
      3 Basic
      2 C
     38 C#
     30 C++
      1 Cobol
      1 D
      1 DosBatch
      8 Eiffel
      2 Erlang
      1 Falcon
      1 Flex
      1 HTML
     11 Java
     17 JavaScript
      2 Make
      2 MatLab
      2 NONE
      5 OCaml
      1 Pascal
      6 Perl
      7 PHP
     12 Python
      2 Ruby
      2 SML
     20 SQL
      1 Tcl
      4 Tex
      2 Vera
      9 Verilog
      2 VHDL
      6 Vim
```

masatake:
The most of all python test cases are reviewed.

PR #368, opened 2015-06-12
Add Qingming He to docs/developers.rst

QingmingHe:
I'm a Fortran programmer and would like to be one of the developers of universal-ctags.

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -40,6 +40,14 @@ Masatake YAMATO <yamato@redhat.com>
 	introduced optlib and xcmd. I'm working on Fedora. I don't
 	have access to the other platforms.
 
+Qingming He <906459647@qq.com>
+
+	I'm mainly a Fortran developer and I use ctags combined with Emacs to
+	handle my projects. My goal is to improve the Fortran parser to make
+	it support Fortran standards from 77 to 2008 and maybe 2015 to be
+	released in the near future. I'm also interested in improving the lisp
+	parsers (elisp and scheme).
+
 Vitor Antunes <vitor.hda@gmail.com>
 
 	I've been working with Verilog for most of the last 10


masatake:
Welcome!

I have some requests.

0. Please, join the reviewing patches, especially about Fortran parser.
1. If your change is small,  you can commit without making a PR. If not small, please, make a PR.
   @b4n is the authority of parsers/.
2. check geany's code base before trying to fix a bug. We want to make the code gap between ctags and geany small.
3. See #361. When you have time, please, review Units/review-needed.r/*.f.t, fortran related test cases.

BTW, M-x c-guess helps you teach emacs indentation style of the current buffer. It is quite handy and is one of my main work in GNU emacs.

Happy tagging!

masatake:
I sent an invitation. But I'm new to the operation of github organization. So I'm not sure it works as I expected. If you can, please merge this change by yourself. If you get a trouble, write here. @ffes may help us.

ffes:
If needed I gladly help

QingmingHe:
> Please, join the reviewing patches, especially about Fortran parser.

This is what I should do here.

> If your change is small, you can commit without making a PR. If not small, please, make a PR. @b4n is the authority of parsers/.

Just ok, I can determine what is small and what is not.

> check geany's code base before trying to fix a bug. We want to make the code gap between ctags and geany small.

Some of the patches can be imported directly but some are not as there is already gap between these two repos. If there is difference I'll open a PR.

> See #361. When you have time, please, review Units/review-needed.r/*.f.t, fortran related test cases.

Glad to do so.

masatake:
BTW, do you use scheme daily task? If yes, which is your favourite implementation?
I use gauche. I'm looking for readtags.scm.

QingmingHe:
Not daily task, just use it for fun. I use guile.

I really think the elisp parser should be split from lisp.c and re-write an elisp parser. Current implementation is too simple.

QingmingHe:
What do you mean by **readtags.scm**?

masatake:
"Too simple" is also applicable to clojure.c.

As far as I know Python and Perl have a library to load tags file and make data structure.
readtags.c in ctags can do the same thing.

It is handy for me if gauche can load tags file.

Issue #373, opened 2015-06-14
Possible bug writing pattern to tags file


ffes:
While testing the rest.c parser from Geany, I run into a something strange. That parser has a bug and adds the underline as pattern. There seems to be an easy fix by adding one line to the makeRestTag() `e.pattern = e.name;`

But with this fix when the tags line is written to the file the `/^` and `$/` are not added. This is because at https://github.com/universal-ctags/ctags/blob/master/main/entry.c#L905  the pattern is just written to the file. This seems to be a bug to me. I think this test should be in writePatternEntry() just above that.

So before I open a PR for that I would like your opinion on this. Maybe I just overlooked something.

This issue needs to be solved before I can add the reStructuredText parser and with that add nestlevel.[ch] aka #315

b4n:
> While testing the rest.c parser from Geany, I run into a something strange. That parser has a bug and adds the underline as pattern. There seems to be an easy fix by adding one line to the makeRestTag() `e.pattern = e.name;`

I didn't look into it, but I guess the real fix would be to have the proper line/position information on the tag, e.g. the line with the title and not the one with the underline.

> But with this fix when the tags line is written to the file the `/^` and `$/` are not added. […] This seems to be a bug to me.

While it might be sensible to change this as you said, AFAIK *no* parsers currently set `entry->pattern`, so it might just be considered an internal detail, and parsers *should not* be setting it.  Also if they really wanted to do it, it might make sense for them to chose not to use completely anchored patterns (imagine e.g. minified CSS or JS).

ffes:
> I didn't look into it, but I guess the real fix would be to have the proper line/position
> information on the tag, e.g. the line with the title and not the one with the underline.

Yes, storing the file position was not that difficult either. I solved the problem with that. I will prepare a PR for reStructuredText.

> While it might be sensible to change this as you said, AFAIK no parsers currently
> set `entry->pattern`, so it might just be considered an internal detail, and parsers
> should not be setting it.

I leave it for now as it is. I will investigate the use of the pattern member of the struct some time later.

> Also if they really wanted to do it, it might make sense for
> them to chose not to use completely anchored patterns (imagine e.g. minified CSS or JS).

Minified JS and CSS result in quite useless tags anyway IMHO. In my tags plugin for Notepad++ I ignore tags with patterns longer then 1024 chars in JS and CSS.

masatake:
I've tried rst parser. It prints the line number and the pattern for a tag well.
[...]

PR #375, opened 2015-06-15
Add reStructuredText parser

### LINKED ISSUE #315: Import nestlevel.[hc] from geany
It is nice. I want it.

Not only just importing, I want to add a variant API specialized to cork.

> extern void nestingLevelsPush(NestingLevels *nls,
	const vString *name, int type);

This is the original version.

> extern void nestingLevelsPushIndex(NestingLevels *nls, int index);

Here INDEX is a value returned from makeTagEntry.

ffes:
The parser is copied and improved from Geany. Since the parser relies on nestlevel.[ch] that one is copied as well.

The original Geany parser has a bug. It adds the underline as search pattern. This bug has been fixed. See #373.

By adding nestlevel.[ch] this largely fixes #315


main/parsers.h
```diff
@@ -51,6 +51,7 @@
 	PythonParser, \
 	RexxParser, \
 	RubyParser, \
+	RestParser, \
```
cweagans: Rest has a different meaning to a lot of people. Thought about calling it RstParser?

vhda:
I wonder if I could make use of this new nestlevel infrastructure... and I wonder how many parsers could benefit from it...
Any takers?

cweagans:
A markdown parser would be another good place to use it :)

main/parsers.h
ffes: Good suggestion

ffes:
> A markdown parser would be another good place to use it :)

Geany has a markdown parser. I want to copy it as well. But first this one ;)

ffes:
There is a compile error on Travis that didn't happen on Windows. I will solve that

ffes:
> I wonder if I could make use of this new nestlevel infrastructure... 
> and I wonder how many parsers could benefit from it...

The nestlevel comes from the python parser. See #315 for more details.

main/parsers.h
blueyed: That would apply to classes and method names then, too?!

Maybe `ReST` could be used instead?

ffes: > That would apply to classes and method names then, too?!

If you mean things like RestKinds[] and findRestTags()... Yes, it would.

> Thought about calling it RstParser
> Maybe ReST could be used instead?

I have no strong option on "Rst" or "ReST", just a slight preference for Rst (but mainly because it looks better in [CamelCasing](https://en.wikipedia.org/wiki/CamelCase))

* "Rst" is the most common used file extension. And note the URL of the homepage: http://docutils.sourceforge.net/rst.html
* "ReST" matches the official abbrev spelling best.

blueyed: > a slight preference for Rst (but mainly because it looks better in CamelCasing)

Makes sense.

masatake:
(When writing a new parser, please consider using cork API.)

steveno:
It took me forever to find it, but there is some documentation on the cork API here:

https://github.com/universal-ctags/ctags/blob/master/docs/internal.rst

ffes:
FYI: Pushed all changes. I think it is ready to merge

masatake:
(About cork API, @steveno, feel free to fix internal.rst or open a PR or an issue if it is not clear.)

vhda:
The filename of the parser appears inconsistent with everything else. Don't you think it would be better to rename it to rst.c?

ffes:
@vhda good catch!

ffes:
> It seems that all patterns can be converted to extensions:

I see the difference. It can be found in EXTENTING.html as well, but that is not very clear indeed. I'll fix that as well.

PR #382, opened 2015-06-18
Fortran type bound procedure

QingmingHe:
+ Report type bound procedures as TAG_METHOD.
+ Parse type bound procedures, including procedure, finalizor and generic procedure.
+ Support **--fields=+m** options for deferred type bound procedure.


masatake:
LGTM.

QingmingHe:
Do you think **non_overridable** is a implementation type? non_overridable procedure means that this procedure should not be overriden by sibling class/type.

masatake:
I guess you mean whether non_overridable should be a member of enum eImplementation or not.

We can know only the parts of languages ctags supports. So answering this kind of question is not easy. However, there is a fundamental approach: think whether adding such information helps tools utilizing tags file or not.

I'm working for ctags for helping my daily work: code reading. So let's assume the tool is for code reading.

I think this information is helpful. Consider IMPL_ABSTRACT. Consider the situation a user of a tool utilizing tags file look for a name X. If the user find X is marked as virtual, the user may find s/he must visit concrete types(classes?) derived(or inherited?) from X to know X more.

In this case tags file doesn't only tell the position but also gives a hint for the next step of code reading: you may have to track the derived types, and don't take time to look for the implementation of X.

BTW, IMP_FINAL is not defined in Java parser. The same discussion can be applied?
How about IMPL_NON_OVERRIDABLE? Consider the situation a user of a tool utilizing tags file look for a name Y. If the user find X is marked as non_overridable, the user may find s/he can stop the tracking; Y is the end point.
Here tags file gives a hint for the next step of code reading.

QingmingHe:
Thanks. I will add IMP_NON_OVERRIDABLE later.

> BTW, IMP_FINAL is not defined in Java parser. The same discussion can be applied?

**final** in java is much like **non_overridable** in Fortran. So the discussion can be applied.

BTW, how can you utilize these information in Emacs as tags in etags format only tell us little data. Or you know a Emacs package can use tags in ctags format?

masatake:
Good question. I am not a ctags user:-P.

http://savannah.gnu.org/projects/vtags
vtags can read tags file. However, as far as I know there is no tool which utilizes
kinds and other extra fields of tags file. 

If really you are interested in emacs interface, you can use our repository of course.
If you want just alternative to etags.el, I recommend you to join emacs-devel list.

I have interests. But a command line tool like readtags is more attractive.
I would like to write an elisp running as front end of readtags when I cannot find any more to do in this project:-P

masatake:
I wonder whether final or non_overridable is better.

However, ctags.1 says
> implementation
                   When  present,  this  indicates a limited implementation (abstract vs. concrete) of a routine or class, where
                   value is specific to the language ("virtual" or "pure virtual" for C++; "abstract" for Java).

So non_overridable is better. But keep this open for awhile. 
I'm recruiting one knows fortran now.

QingmingHe:
> But a command line tool like readtags is more attractive.

I've tried readtags under Cygwin (I've no access to other platform), it consumes ~0.2 secnds to find tags, which is really very slow. How does it perform under your development environment?

Another, do you know how to update tags file incrementally. I know GNU global http://www.gnu.org/software/global/ has this feature. Update tags file via **global** command is very easy:
```Shell
global --single-update=<file>
```
I've write a elisp package to update tags file in etags format incrementally. It's easy for that tags for a file are grouped together in etags format. What I do is deleting tags for a file and then using command **ctags -e -o TAGS -a \<file\>** to append tags in the end of **TAGS** file. But it's a tricky for ctags format as tags are sorted. Do you plan to add this feature to ctags?

masatake:
> I'm recruiting one knows fortran now. 
I have not got answer. So let's merge your commit.

masatake:
Could you tell me the size of tags file you used?
(readtags does binary search, therefore I guess the impact of size of tags file is not so big.)

For me it is enough fast:
>[yamato@x201]~/var/ctags-github% wc -l ~/var/linux/tags 
1919853 /home/yamato/var/linux/tags
[yamato@x201]~/var/ctags-github% time ./readtags -t ~/var/linux/tags sys_open
<performing binary search>
sys_open	fs/open.c	/^SYSCALL_DEFINE3(open, const char __user *, filename, int, flags, umode_t, mode)$/
sys_open	include/linux/syscalls.h	/^asmlinkage long sys_open(const char __user *filename,$/
./readtags -t ~/var/linux/tags sys_open  0.00s user 0.00s system 72% cpu 0.002 total

I'm thinking about incremental updating. There are two approaches.
1) Adding ability to load tags file to ctags.
    See lxcmd.c. I wrote code to read data that has tags syntax.
    (NOTE: It is different from what readtags does.)
    Currently it is used to read the output of external command and write parsed data to tags file.
    However, we can use another purpose; read tags contents(loading) into memory of ctags process.
    During loading we can filter the tag entry related to a file which should be updated.

2) Introducing another tool for managing tags files like you wrote
     tagstool merge|update|....
     
I'm not sure which approach is better.
The both approaches need code to load tags file into data structures.
 lxcmd.c can be used as starting point.

QingmingHe:
> Could you tell me the size of tags file you used?

~1M. It may be a problem of Cygwin. Cygwin is famous for its slowness.

> lxcmd.c can be used as starting point.

Agree. Emacs and Vim already have its own tool to update tags file (https://github.com/xolox/vim-easytags and https://github.com/mattkeller/etags-update). We need not do the redundant work. I'm reading lxcmd.c ...

You can merge this PR.

masatake:
(generally I expect a member of the organization merges his/her change by himself/herself.)

QingmingHe:
OK. Next time I will.

vhda:
If I ever find the time, I would like to make ctags be able to also identify objects. To do so, it would be nice if it could read an already existing tags file containing the tags for all my existing classes.

masatake:
... I will lose my time for sleeping if I enter the discussion about this topic:-P Though tags file is well designed only a few people utilizes its possibility. 

Currently I focus on stablizing the features we have implemented and finding more members who can maintain a parser like @QingmingHe. So I will not touch this attractive area.

See  parseXcmdPath of lxcmd.c and 
writeCtagsEntry, recordTagEntryInQueue of entry.c. I think they can help you writing a tool.

Of course you can write your own tool with something scripting language like python.
Feel free to make a new repository in the orgnazatoin for prototying.
Instead of writing a tool, writing a blog entry about your ideal tool. I would like to read what you are thinking.

QingmingHe:
> ... I will lose my time for sleeping if I enter the discussion about this topic

It's also very late in China. I'll go sleep. A meeting waits for me tomorrow.

> Instead of writing a tool, writing a blog entry about your ideal tool. I would like to read what you are thinking.

Maybe in the weekend, I will also share my immature elisp package aiming at handling tags file.

PR #385, opened 2015-06-20
Documents dir cleanup

masatake:


README.md
```diff
@@ -1,26 +1,19 @@
...
-```bash
-$ autoreconf -vfi
-$ ./configure --prefix=/where/you/want # defaults to /usr/local
-$ make
-$ make install # may require extra privileges depending on where to install
-```
-
-After installing the `ctags` executable will be in `$prefix/bin/`.
-
+To build with configure see `docs/autotools.rst` for more information.
```
ffes: Maybe it would be useful to add `GNU/Linux` somewhere in this line?

masatake:
In my understanding configure is for improving the portablity, not just for GNU system.
How about
> To build with Autools...
instead?

PR #391, opened 2015-06-21
Fortran: support --fields=+S

QingmingHe:
Parse signature of Fortran subprogram and prototype. The signature of Fortran language has three forms:

1. No function arguments at all
```Fortran
   subroutine suba
   end subroutine
```
2. No function arguments but with parens
```Fortran
   subroutine suba()
   end subroutine
```
3. With function arguments
```Fortran
   subroutine suba(a, b)
     interger :: a, b
   end subroutine
```

In the first and second case, signature will set to NULL and will not write to tags file. In the last case, signature will be written with **--fields=+S** options.

### DIFF
--- a/Units/parser-fortran.r/fortran-signature.d/args.ctags
+++ b/Units/parser-fortran.r/fortran-signature.d/args.ctags
@@ -1,0 +1,2 @@
+--fields=+S 
+--Fortran-kinds=+LP
--- a/Units/parser-fortran.r/fortran-signature.d/expected.tags
+++ b/Units/parser-fortran.r/fortran-signature.d/expected.tags
@@ -1,0 +1,9 @@
+a	input.f90	/^    integer :: a$/;"	L	subroutine:fnb	file:
+a	input.f90	/^    integer a,/;"	L	subroutine:suba	file:
+arg	input.f90	/^      type(atype) :: arg$/;"	L	prototype:subb	file:
+b	input.f90	/^    integer a, b$/;"	L	subroutine:suba	file:
+fna	input.f90	/^  real function fna(/;"	f	module:test_signature
+fnb	input.f90	/^  subroutine fnb$/;"	s	module:test_signature
+suba	input.f90	/^  subroutine suba /;"	s	module:test_signature	signature:(a, b)
+subb	input.f90	/^    subroutine subb /;"	P	module:test_signature	signature:(arg)
+test_signature	input.f90	/^module test_signature$/;"	m
--- a/Units/parser-fortran.r/fortran-signature.d/input.f90
+++ b/Units/parser-fortran.r/fortran-signature.d/input.f90
@@ -1,0 +1,25 @@
+module test_signature
+
+  interface
+    subroutine subb (arg)
+      type(atype) :: arg
+    end subroutine
+  end interface
+  
+contains
+  subroutine suba &
+      (a, &
+      ! comment
+    b)
+    integer a, b
+  end subroutine
+
+  real function fna()
+    use, intrinsic :: iso_c_binding
+  end function
+
+  subroutine fnb
+    integer :: a
+  end subroutine
+
+end module test_signature
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -204,6 +204,7 @@ typedef struct sTokenInfo {
 	tagType tag;
 	vString* string;
 	vString* parentType;
+	vString* signature;
 	impType implementation;
 	boolean isMethod;
 	struct sTokenInfo *secondary;
@@ -384,6 +385,7 @@ static void ancestorPop (void)
 	Ancestors.list [Ancestors.count].lineNumber = 0L;
 	Ancestors.list [Ancestors.count].implementation = IMP_DEFAULT;
 	Ancestors.list [Ancestors.count].isMethod   = FALSE;
+	Ancestors.list [Ancestors.count].signature  = NULL;
 }
 
 static const tokenInfo* ancestorScope (void)
@@ -457,6 +459,7 @@ static tokenInfo *newToken (void)
 	token->string       = vStringNew ();
 	token->secondary    = NULL;
 	token->parentType   = NULL;
+	token->signature    = NULL;
 	token->implementation = IMP_DEFAULT;
 	token->isMethod     = FALSE;
 	token->lineNumber   = getSourceLineNumber ();
@@ -472,6 +475,7 @@ static tokenInfo *newTokenFrom (tokenInfo *const token)
 	result->string = vStringNewCopy (token->string);
 	token->secondary = NULL;
 	token->parentType = NULL;
+	token->signature = NULL;
 	return P (result, deleteToken);
 }
 
@@ -481,6 +485,7 @@ static void deleteToken (tokenInfo *const token)
 	{
 		vStringDelete (token->string);
 		vStringDelete (token->parentType);
+		vStringDelete (token->signature);
 		deleteToken (token->secondary);
 		token->secondary = NULL;
 		eFree (token);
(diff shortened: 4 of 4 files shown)


masatake:
> subroutine suba(a, b)
     interger :: a, b
   end subroutine

It seems that "(a, b)" is recorded as signature.

I got a question. Following code is valid in Fortran?
> subroutine suba(a, b)
     someThingOtherType :: a, b
   end subroutine

If, yes, I think users may want the type informations are also recorded like
> signature:(a, b)someThingOtherType :: a, b

QingmingHe:
Code like this is normal and beautiful:
```Fortran
subroutine suba(a, b)
  someThingOtherType :: a, b
end subroutine
```
Its' easy to get the type information here. But Fortran compiler also recognize ugly code like this:
```Fortran
subroutine suba(a, b)
  real c
  real d
  integer :: a, e
  real(my_kind), allocatable, intent(in) :: b(:,:)
end subroutine
```
Type of **a** and **b** is declared after other local variables. It' a tricky for ctags to get the type information for that ctags report **suba**, **c**, **d**, **a**, **e**, **b** in a sequence (The local variables are also reported in Fortran parser). Ctags cannot get the type information of **a** before it parses **integer :: a, e**. But if ctags parse this statement just for obtaining signature, it will lose the information of **c** and **d**. Ctags can not parse backward as far as I know.

masatake:
Did you know cork API?
With extending cork API, you can implement forward referencing.

Look at docs/internal.rst.

When you call makeTagEntry with "suba", you will get an index. Let's call the index I_suba.
After parsing integer :: a, e and real(my_kind), allocatable, intent(in) :: b(:,:), you can ready to
build perfect signature string. With the string you can update the signature slot of I_suba.

Currently the API is minimized. However, if you need you can extend it.

cork API will help you to delay the decision about what should be written to tags file.

If you don't think about such perfect solution now, let's think about forward compatibility.
If there is perfect code for generating signature string, how the output format is?
Try to make the current imperfect output subsets of the perfect output.

Do I make sense?

@b4n, could you give us input about this topic. I guess signature strings may be utilized in geany. 
From the tool side, what is a preferred solution?

QingmingHe:
Let's summon @b4n again.
Say the source code is
```Fortran
subroutine suba(a, b)
  real c
  real d
  integer :: a, e
  real(my_kind), allocatable, intent(in) :: b(:,:)
end subroutine
```
I've two ideas to show the signature.
The first one is include type data along with argument variables:
```
signature:(interger :: a, real(my_kind), allocatable, intent(in) :: b(:,:))
```
Or show variables in the first parens and type information in the second:
```
signature:(a, b)(integer :: a, real(my_kind), allocatable, intent(in) :: b(:,:))
```
I vote for the second for that I tend to give argument variable a long_and_detailed name such as **material**, **geometry** rather than **a**, **b**. The data inside first parens is adequate. If I want to get more information, I'd like to jump to the definition directly as jump via etags is convenient in Emacs. I can hardly find the name of variable in the first case. For user who want to get variable names along with type information, he can refer to the second parens.
Besides, it's easy to make it an option to determine whether show data inside first or second parens via eldoc-mode of Emacs (I dont know other editors).

masatake:
The second one is o.k.

> signature:(a, b)(integer :: a, real(my_kind), allocatable, intent(in) :: b(:,:))

(a, b) part is must and you already implemented it. Could you merge this change?
The rest is optional and you will implement it, won't you:-P

Don't forget escaping tabs in the rest part.

b4n:
Sorry for the delay.

Well, signatures are typically used for 2 things by tools:
1. Calltips (hinting to the user, the most common usage)
2. Getting the list of parameter names, and possibly types.

For 1, IMO we need something that is natural for the language.  What most (all?) parers do is they just report whatever was used as the signature in the code -- typically everything inside the parentheses for non-K&R C-style things.

2 however is a hack, and requires extra understanding of the particular languages.  We could have a unified signature format, but that wouldn't be a really great solution here.  If we really want to support that, we probably should rather introduce a new "parameter" kind that works like local variables.

So IMO only the calltip usage is really important, and there it is directly targeted as the programmer, so it should IMO be as natural as possible for the language.  If Fortran uses `subroutine suba(a, b) integer :: a, b`, IMO the signature for subroutine `suba` should be `(a, b) integer :: a, b` -- or maybe `(a, b) integer:: a integer:: b`, but IMO it's better and simpler to just respect what was in the actual code.
But basically, use whatever feels the more natural for a Fortran programmer I guess.

masatake:
@b4n, thank you very much.

My understanding is that the parenthesis  are not needed in the 2nd proposed format.

> signature:(a, b) integer :: a, real(my_kind), allocatable, intent(in) :: b(:,:)

@QingmingHe, could do as you think better based on the criteria given by @b4n.

QingmingHe:
> My understanding is that the parenthesis are not needed in the 2nd proposed format.

It seems more natural to have parenthesis in the second. We can understand that the first is a succinct signature and the second is a detailed one.

b4n:
Well, I don't know Fortran but I would tent to agree with @masatake given your code examples.  But again, it's not very important as it's basically just something human-readable.

QingmingHe:
o.k. it's just a personal taste to have parenthesis. I will accept the second format. What I really cares is how to show function API via Emacs eldoc.

QingmingHe:
Is the discussion applicable to other languages? Will other parser write signature like this:
```
(name1, name2) type of name1, type of name2
```

b4n:
> Is the discussion applicable to other languages? Should other parser other than C parser write signature like this: […]

IMO no, they should write whatever is natural for the language, e.g. probably reproduce the language's own syntax there.

QingmingHe:
But it will make it hard for outer tools to parse signature and do some useful things without unified format.

QingmingHe:
Or at least, there should be some rules. Maybe classify into three kinds of language:
1. c-style.
```
(type name1, type name2)
```
2. type-free
```
(name1, name2)
```
3. some style I can't give it a name
```
(name1, name2) type of name1, type of name2
```

b4n:
As I said earlier, while currently parsing the arguments might be needed for some features, it's IMO not the way to go.  If we want to support this, we should probably simply add an extra tag type (or field, but probably rather type) for arguments, that would work like variables: name and type.  This would be a lot more robust and easy than parsing an argument list, and wouldn't require "foreign" calltips.

Having unified calltips would probably be okay-ish for languages like C, C++, Java, JavaScript, PHP (but what about `$` in the identifiers?), maybe even Fortran, but probably not all languages.
Also, it becomes quite complex to parse with some types, e.g. in C `(void(*)(void(*)(int a, void *d) f, void* d) f, void *d, int a)` is perfectly valid, but very subtle to parse.

b4n:
> Maybe classify into three kinds of language: […]

Types 1 and 3 seem to be identical in what they provide, so having both would go against uniformity.
If we *really* wanted a uniform format, I'd rather go for `([[type] arg1[, [type] arg2[, ...]]])` (e.g. C-style, with the type being optional), but once again IMO it's fragile and not the way to go if we want to let callers understand parameters.

b4n:
(BTW type 3 is mostly K&R C)

QingmingHe:
Type 3 is not k&r c. It's languages declaring types of arguments outside parenthesis like Fortran.

b4n:
> Type 3 is not k&r c. It's languages declaring types of arguments outside parenthesis like Fortran.

Well, K&R style is this:
```C
int mul(a, b)
  int a;
  int b;
{
  return a*b;
}
```
Seems pretty similar to me :)

QingmingHe:
BTW, no signature generated for K&R style functions (I just test the example you gives). Is it a bug?

b4n:
> BTW, no signature generated for K&R style functions (I just test the example you gives). Is it a bug?

Indeed, and yeah it must be a bug.  The version of the C parser in Geany does handled those correctly (well, it only gives the part within the parentheses, but still)

QingmingHe:
> The rest is optional and you will implement it, won't you:-P

o.k. In the first step I will replace Ancestors by cork API. They seem do the same thing.

PR #394, opened 2015-06-25
docs: add a document about criterial of reviewing

masatake:
Feel free updating this.

### DIFF
--- a/docs/criteria.rst
+++ b/docs/criteria.rst
@@ -1,0 +1,124 @@
+Criteria in patch accepting
+=============================================================================
+
+These are what we would like potential contributors to know.
+
+Origin of changes
+---------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


ffes:
I am working on `contribution.md` when GitHub finds this file in the repo it will show it at the top of every new issue and PR.

But we should discuss some standards we want to maintain. Like a units test, tab settings, eol style and some basic code style like how to place curly braces and C or C++ style comments.

masatake:
@ffes, thank you. 

Do you think the strict style guide is needed?
Of course tailing white spaces should be removed before sending a PR.
I don't care the place of curly braces and the others. 

If this is a newly started project, I will care. However, the style was not unified in
the original code. If we define a style guide, we should fix whole the old code.

PR #396, opened 2015-06-25
add nim language parser

jangko:


masatake:
It seems that I make you confused.
I merged your commit mistakenly but I reverted.
So currently your commits are not in universal-ctags repository.

Could you rebase your commits on the latest "master" branch of the repository?


The assertion I show is just for telling the issue in your code.
With the assertion, run ctags with your test inputs.
You will see the failure of assertion.

jangko:
it already near midnight here, I think I will do it tomorrow
Thanks a lot

PR #404, opened 2015-06-28
Support Fortran file extensions .f03, .f08 and .f15

QingmingHe:
See http://gcc.gnu.org/onlinedocs/gfortran/GNU-Fortran-and-GCC.html
See also http://fortranwiki.org/fortran/show/File+extensions


b4n:
apart that the first link doesn't mention `f15`, and the second suggests `f95`, `f03` and `f15` are a bad idea, LGTM

masatake:
LGTM.

PR #405, opened 2015-06-29
Add C++11 override support(sf-bugs:367)

vmiklos:
Revert the code change and see how the Derived::foo() declaration is missing from the tags file.

Bugreport: https://sourceforge.net/p/ctags/bugs/367/

### DIFF
--- a/Units/cpp-override.d/input.cpp
+++ b/Units/cpp-override.d/input.cpp
@@ -1,0 +1,18 @@
+class Base
+{
+public:
+	virtual void foo() = 0;
+};
+
+class Derived : public Base
+{
+	virtual void foo() override;
+};
+
+void Base::foo()
+{
+}
+
+void Derived::foo()
+{
+}
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -2289,7 +2289,12 @@ static boolean skipPostArgumentStuff (
 					break;
 
 				default:
-					if (isType (token, TOKEN_NONE))
+					/* "override" and "final" are only keywords in the declaration of a virtual
+					 * member function, so need to be handled specially, not as keywords */
+					if (isLanguage (Lang_cpp) && isType (token, TOKEN_NAME) &&
+						strcmp ("override", vStringValue (token->name)) == 0)
+						;
+					else if (isType (token, TOKEN_NONE))
 						;
 					else if (info->isKnrParamList  &&  info->parameterCount > 0)
 						++elementCount;
--- a/Units/cpp-override.d/args.ctags
+++ b/Units/cpp-override.d/args.ctags
@@ -1,0 +1,2 @@
+--c++-kinds=+p
+--fields=+S
--- a/Units/cpp-override.d/expected.tags
+++ b/Units/cpp-override.d/expected.tags
@@ -1,0 +1,6 @@
+Base	input.cpp	/^class Base$/;"	c	file:
+Derived	input.cpp	/^class Derived : public Base$/;"	c	file:
+foo	input.cpp	/^	virtual void foo() = 0;$/;"	p	class:Base	file:	signature:()
+foo	input.cpp	/^	virtual void foo() override;$/;"	p	class:Derived	file:	signature:()
+foo	input.cpp	/^void Base::foo()$/;"	f	class:Base	signature:()
+foo	input.cpp	/^void Derived::foo()$/;"	f	class:Derived	signature:()


parsers/c.c
```diff
@@ -2038,6 +2038,7 @@ static void processToken (tokenInfo *const token, statementInfo *const st)
 		case KEYWORD_CATCH:     skipParens (); skipBraces ();           break;
 		case KEYWORD_CHAR:      st->declaration = DECL_BASE;            break;
 		case KEYWORD_CLASS:     checkIsClassEnum (st, DECL_CLASS);      break;
+		case KEYWORD_OVERRIDE:  st->declaration = DECL_BASE;            break;
```
b4n: This seems unnecessary at best to me

b4n:
Otherwise LGTM, but maybe you should include the signatures in the test output as this affects them (use `--fields=+S`).

vmiklos:
Done for both. :-)

masatake:
@vmiklos, thank you for submitting your patch here.

> signature:() override

Should we deal with "override" as a part of signature?

vmiklos:
If "const" is part of the signature (and it is), then I would say "override" should be, too. But no strong opinion on that, my primary problem was that if "override" is used, then the member function declaration was not found at all.

mawww:
Well, C++ does consider const part of the function signature (and overloads depending on it), but override is just a validation check. Now I do not know if ctags signatures follow the same semantics as C++ signature.

masatake:
See #391. Signature is a hot topic.

@mawww, thank you for the comment.

Could you review the following sentences. (I would like to write this to docs/criteria.rst.)
The format of signature field is depends on language and its evaluation/interpretation should be done by a tool which has knowledges of the target language.

Units/cpp-override.d/input.cpp
```diff
@@ -0,0 +1,18 @@
+class Base
+{
+public:
+	virtual foo() = 0;
```
b4n: BTW, it would be better to have valid C++ code here, e.g. give a type to all these declarations.

vmiklos:
I've updated the example to be valid C++ code.

vmiklos:
I've uploaded one more version, this no longer puts "override" to the signature based on the above discussion.

b4n:
Someone just pointed out to me in https://github.com/geany/geany/pull/544 that `override` isn't really a keyword, so this change would break valid code using the identifier `override`:
```C++
class Foo : Base {
  virtual void foo() override;
  virtual void override();
};
```

So `override` should be handled in some very specific way.  I came up with a strcmp-based patch, which while not very pretty avoids breaking other code.

vmiklos:
Aha -- then no idea how to proceed, maybe take this patch as-is, and and deal with the keyword problem as a follow-up patch?

b4n:
> Aha -- then no idea how to proceed, maybe take this patch as-is, and and deal with the keyword problem as a follow-up patch?

Well, the problem is that this patch actually breaks all use of `override` as an identifier, so it should probably be fixed before merging it.

You can see how I managed to do it in https://github.com/geany/geany/pull/544 -- it's not very pretty, but it works.  (note that this also handles `final`, but that the code there isn't yet in sync with CTags, so there will be some adaptation needed)

vmiklos:
Fair enough, I kept the test code, but for the rest I just ported your Geany changes. :-)

b4n:
LGTM, although it might be interesting to also ass tests checking whether `override` is still accepted as an identifier (I know it is, but not to break it in the future).

Also, I was also importing the commits from my Geany PR here, and PR'd that as #413.  e1d6ab072fcfba0814d73f65f8c53cb1bf4108c4 contains the same (well, plus `final` support) change.
I don't mind updating my PR after if we merge this one, but we could also merge #413 instead.  Opinions?

vhda:
@b4n I know this was a typo, but I can't stop from laughing:
ass tests!? :D

vhda:
Given that @vmiklos based his work on yours, I don't think that it should be a problem to merge your bigger PR. Obviously, as long as @vmiklos doesn't mind :)

vmiklos:
No, I don't mind. :-)

vhda:
As previously discussed, this was included in #413

masatake:
Thank you for contributing and sorry for my late reply.

vhda:
I should have thanked @vmiklos 
My apologies for not doing so when closing this PR.

PR #406, opened 2015-06-30
Parse Fortran 2003 enum

QingmingHe:
Fortran 2003 enums is:
```Fortran
enum, bind(c)
  enumerator [::] name1 [= val], ...
end enum
```
See https://books.google.com.hk/books?id=40UkLNQuAeoC&pg=PA114&lpg=PA114&dq=fortran+enumerator+type&source=bl&ots=WTou0LNC5z&sig=kV-2jyz4T2qUlCQ3T4jx3WeRJkc&hl=zh-CN&sa=X&ei=JZeSVbSSH5DYoAShrI6gDA&ved=0CEcQ6AEwBTgK#v=onepage&q=fortran%20enumerator%20type&f=false page 114
and http://www-01.ibm.com/support/knowledgecenter/SSGH4D_10.1.0/com.ibm.xlf101a.doc/xlflr/enum.htm%23enum

We shoud note that
+ bind(c) is mandatory
+ enum shoud have no name
+ enumerator should have no qualifier

The implementation of Geany is different https://github.com/geany/geany/commit/9520e7f7d7b7a9570db2feb165653ef1d68e547a which is based on
 http://docs.cray.com/books/S-3692-51/html-S-3692-51/z970507905n9123.html. The syntax is different:
+ kind selector can be used instead of bind(c)
+ enum can be given a name

The docs of ISO http://www.j3-fortran.org/doc/year/04/04-007.pdf page 66 don't allow for syntax of cray.

### DIFF
--- a/Units/parser-fortran.r/fortran-enum.d/expected.tags
+++ b/Units/parser-fortran.r/fortran-enum.d/expected.tags
@@ -1,0 +1,45 @@
+Constants	input.f90	/^module Constants$/;"	m
+E_e	input.f90	/^  real, parameter :: E_e /;"	v	module:Constants
+Named1	input.f90	/^  enum :: Named1$/;"	E	module:Constants
+Named2	input.f90	/^  enum Named2$/;"	E	module:Constants
+Named3	input.f90	/^  enum(8) Named3$/;"	E	module:Constants
+Named4	input.f90	/^  enum*8 Named4$/;"	E	module:Constants
+Named5	input.f90	/^  enum(8) :: Named5$/;"	E	module:Constants
+Named6	input.f90	/^  enum*8 :: Named6$/;"	E	module:Constants
+Named7	input.f90	/^  enum, bind(c) :: Named7$/;"	E	module:Constants
+a	input.f90	/^    enumerat/;"	N	module:Constants
+b	input.f90	/^    enumerator :: a, b,/;"	N	module:Constants
+black	input.f90	/^    enumerator :: red =1, blue, black /;"	N	module:Constants
+blue	input.f90	/^    enumerator :: red =1, blue,/;"	N	module:Constants
+bronze	input.f90	/^    enumerator gold, silver, bronze$/;"	N	module:Constants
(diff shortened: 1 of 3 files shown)


QingmingHe:
I've found the syntax is different in http://docs.cray.com/books/S-3692-51/html-S-3692-51/z970507905n9123.html (GNU Fortran compiler doesn't recgnize the examples)

QingmingHe:
The question is whether we should support compiler specific features or not.

QingmingHe:
As the original implementation is just a subset of that of Geany, I fully import that of Geany.

I'm not quite sure whether kind selector and enum name are standard Fortran 2003 syntax. But at least they are allowed by Cray compiler and discussed here http://www.j3-fortran.org/doc/year/00/00-121.pdf.

masatake:
Do you mean there are two fortran syntax(fortran2003 and cray) for enum definition?
Is it impossible to accept both syntax in ctags? Is there serious confliction?

I'm not sure I can give you a constructive comment I would like to see a typical problematic fortran causing the confliction. In addtion I would like to see tags output generated by ctags following fortran 2003 syntax and by ctags following cray syntax.


Surprisingly,  createTagsForFile allows a parser doing multiple path parsing. If parising by a syntax is failed, paring by another syntax can be run.  If we cannot find any good solution, you can use this feature.

QingmingHe:
Sorry for replying late. I moved both dormitory and researching room in last week, having no access to Internet, feeling being dropped by the world...

> Do you mean there are two fortran syntax(fortran2003 and cray) for enum definition?

Not two syntax, Cray compiler has its own extension to standard Fortran 2003. Besides, the extension syntax is discussed to be incorporated in to standard syntax.

> Is it impossible to accept both syntax in ctags? Is there serious confliction?

As I mentioned, standard Fortran 2003 is a subset of Cray. So if ctags support Cray, it will also support standard F2003. There should be no confliction.

As the the trashbox will be removed in #425, I will modify this patch after merging #425.

masatake:
The commits of #425 is merged.

Issue #384, opened 2015-06-19
Utilize coverall


masatake:
I don't want to open more issues:-P But this is worth to be opened.

Does anyone have interest this area?
I wrote a small initial version and made a branch enable-gcov.
Feel free hacking it.

masatake:
I feel difficulties to review patches for unknown languages. Only test cases protect us.
So I will concentrate on this issue.

k-takata:
I use lcov and [lcoveralls](https://github.com/pcolby/lcoveralls) to utilize coveralls in [my project](https://github.com/k-takata/Onigmo).
My .ravis.yml: https://github.com/k-takata/Onigmo/blob/master/.travis.yml
Result: https://coveralls.io/github/k-takata/Onigmo?branch=master

masatake:
I'm working on this issue now. I refer https://github.com/eddyxu/cpp-coveralls.
Soon, I will make a new experimental branch for testing.

masatake:
... nothing happens.
@k-takata, I'm using `coverall` branch not to break `master` branch. Is this o.k.?

k-takata:
When I ran lcov on my local machine, I had to move all *.gcda and *.gcno into the project top directory before running lcov. Without that, lcov couldn't find the source codes. I don't know why.

> Is this o.k.?

I think it's OK.

masatake:
I read the log of travis. lcov is run but something wrong is reported.
I guess it is something to do with source/building directory structure. I have to inspect it more.

https://travis-ci.org/universal-ctags/ctags/jobs/72836837

Thank you.

masatake:
```
3.60s$ if [ $CC = 'gcc' ] && [ $TRAVIS_OS_NAME = 'linux' ]; then mv main/*.gcda main/*.gcno parsers/*.gcda parsers/*.gcno .; lcov -c -d . -o coverage.info && lcoveralls --retry-count 5 coverage.info; fi

Capturing coverage data from .

Found gcov version: 4.6.3

Scanning . for .gcda files ...

Found 68 data files in .

Processing yacc.gcda

Processing windres.gcda

Processing vim.gcda

Processing vhdl.gcda

Processing verilog.gcda

Processing tex.gcda

Processing tcl.gcda

Processing sql.gcda

Processing sml.gcda

Processing slang.gcda

Processing sh.gcda

Processing scheme.gcda

Processing rust.gcda

Processing ruby.gcda

Processing rst.gcda

Processing rexx.gcda

[...]

PR #412, opened 2015-07-01
Enable gcov

masatake:
Usage:
After doing make clean, you can build coverage measuring ready
ctags by make COVERAGE=1. At this time *.gcno files are generated
by the compiler. *.gcno files can be removed with make clean.

After building ctags, you can run run-gcov target.  When running
*.gcda files.  The target runs ctags withh all input files under
Units/**/input.*; and call gcov. Human readable result is printed. The
detail can be shown in *.gcov. files. *.gcda files and *.gcov files
can be removed with make clean-gcov.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 test.*.diff
 *.gcno
 *.gcda
+*.gcov
 *.o
 *.od
 *.obj
--- a/Makefile.in
+++ b/Makefile.in
@@ -38,7 +38,7 @@ endif
 DEFS	= @DEFS@
 CPPFLAGS = @CPPFLAGS@
 CFLAGS	= @CFLAGS@
-LDFLAGS	= @LDFLAGS@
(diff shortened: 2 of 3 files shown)


steveno:
If you're going to start enabling coverage report, you might as well go for the gold:

https://coveralls.io/

masatake:
@steveno, I'm sorry but what is "gold" here?
Do you mean gold liker(https://en.wikipedia.org/wiki/Gold_%28linker%29)?
Or something well know expression?
Sorry for stupid question.

I'm thinking about using coverall (#384).

steveno:
Sorry :( 

It's an English expression that's actually a reference to the Olympics. You "go for gold" in that if you're going to do something, you go all the way, and do the best you possibly can.

I was suggesting that if you're going to enable gcov support, you should also enable the coveralls service to go with it. Gives you a nice button in the readme file displaying how much the project is covered by a test case.

masatake:
Oh, I see. Thank you.

PR #413, opened 2015-07-01
C++11 stuff (override, final and noexcept)

b4n:
Add support for some C++11 things:
* `final` contextual keywords on classes and structs
* `final` and `override` contextual keywords on methods (see also #405)
* `noexcept`

See also https://github.com/geany/geany/pull/544.  Note that some Geany commits referenced there are not yet merged into master, as I PR this here first to get comments and avoid further divergence.

### DIFF
--- a/Units/cxx11-final.d/args.ctags
+++ b/Units/cxx11-final.d/args.ctags
@@ -1,0 +1,2 @@
+--c++-kinds=+p
+--fields=+S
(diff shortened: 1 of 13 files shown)


vmiklos:
Can a maintainer comment on this PR, please? #405 is kind of rejected, as this one is preferred, but as far as I see nothing happened to either of them for a week.

Thanks!

potuz:
Overrides in QT are typically done by adding the macro Q_DECL_OVERRIDE, a small variation of this patch here could handle those as well.

PR #414, opened 2015-07-01
Revive strict args order

masatake:
Background:

      $ ctags -o - main.c --list-kinds

Both the result of parsing main.c and the list of kinds go to stdout.
A tool consuming output may be confused.

As far as reading option.c ctags doesn't allow such dangerous order of
option specification.

A boolean, NonOptionEncountered becomes true when the command line parser handles
an input file name initially.

checkOptionOrder checks the order for short options.

initOnly field of parametricOption and booleanOption structures provides
the way to declare an order sensitive long options.

These facilities were just ignored.
I am afraid this behavior was introduced by me when working on optlib.
However, exuberant ctags shipped as part of Fedora also ignores the order.

These patches for fxing the facilities.

This breaks the compatibility between exuberant ctags in semantics of command line
interpretation.

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -2081,10 +2081,10 @@ static booleanOption BooleanOptions [] = {
  *  Generic option parsing
  */
 
-static void checkOptionOrder (const char* const option)
+static void checkOptionOrder (const char* const option, boolean longOption)
 {
 	if (NonOptionEncountered)
-		error (FATAL, "-%s option may not follow a file name", option);
+		error (FATAL, "-%s%s option may not follow a file name", longOption? "-": "", option);
 }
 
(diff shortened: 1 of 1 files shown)


main/options.c
```diff
@@ -2081,10 +2081,10 @@ static booleanOption BooleanOptions [] = {
  *  Generic option parsing
  */
 
-static void checkOptionOrder (const char* const option)
+static void checkOptionOrder (const char* const option, boolean more_hyphen)
```
vhda: Could I ask you to rename the argument to `longOption`?
Makes it more clear what TRUE means in this case.

masatake:
@vhda,  thank you.

PR #415, opened 2015-07-01
Fortran fix bugs reported by valgrind

masatake:


QingmingHe:
o.k. Seems I've made such mistakes for several times. I'll move to Linux system. Cygwin doesn't support valgrind.

QingmingHe:
Thanks, I've moved to ubuntu and tested with valgrind.

PR #416, opened 2015-07-02
Fix  fuzz-hitting-assertions/eiffel-fha-0

masatake:
An assertion is failed when unexpected input given.
I cannot find a good way to recover for such broken input in Eiffel. So in this fix, I make ctags just stop parsing.


vhda:
We had decided as a group against trash box.
Again, removing the `Assertion()` call fixes this error and does not introduce a regression in the remaining test cases (which results in completely removing the `else` part). You could also replace it with a `goto getNextChar` without introducing any memory leak, as far as I can see.

Again, what are we trying to achieve? Do we want to **verify** or **parse** the input?

b4n:
I fully agree with @vhda.  The worse that can happen here is reporting and unknown token, which should never be worse than an unexpected one.

IMO, adding such complexity (and additionally hiding it), plus adding more `longjmp()`, should be avoided.  We should rather drop existing `longjmp()` usage than introduce more.  But we had this discussion already didn't we :)

masatake:
What I am trying to achieve are following two things:

1. avoiding entering infinite loop for any input.
2. avoiding memory leak for any input

As you wrote just stopping parsing for broken input not ideal for the 
purpose of ctags.

However, to achieve above two goals quickly, just stopping is easy.


Just removing introduces a  memory leak.
```
$ git diff                                                      
diff --git a/parsers/eiffel.c b/parsers/eiffel.c
index ad2e4df..0106793 100644
--- a/parsers/eiffel.c
+++ b/parsers/eiffel.c
@@ -770,7 +770,7 @@ getNextChar:
                        else
                        {
                                token->type = TOKEN_UNDEFINED;
-                               Assert (! isType (token, TOKEN_UNDEFINED));
+                               /* Assert (! isType (token, TOKEN_UNDEFINED)); */
                        }
                        break;
        }
$ make fuzz VG=1 LANGUAGES=Eiffel 
Semi-fuzzing (Eiffel)                                       
....
[valgrind-error Eiffel]                 Units/review-needed.r/attributes.cs.t/input.cs
```

b4n:
> Just removing introduces a memory leak.

That's likely because the parser uses `longjmp()` for handling EOF already.  I'll fix this.

b4n:
> Just removing introduces a memory leak.

Hum, does it?  I can reproduce this with or without the assertion.

with eiffel.c from master:
```console
$ valgrind --leak-check=full ./ctags --language-force=Eiffel -o- Units/review-needed.r/attributes.cs.t/input.cs
==30062== Memcheck, a memory error detector
==30062== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==30062== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==30062== Command: ./ctags --language-force=Eiffel -o- Units/review-needed.r/attributes.cs.t/input.cs
==30062== 
MyDllimportClass	Units/review-needed.r/attributes.cs.t/input.cs	/^[DllImport] public class MyDllimportClass {}$/;"	c
==30062== 
==30062== HEAP SUMMARY:
==30062==     in use at exit: 208 bytes in 7 blocks
==30062==   total heap usage: 6,246 allocs, 6,239 frees, 345,101 bytes allocated
==30062== 
==30062== 208 (40 direct, 168 indirect) bytes in 1 blocks are definitely lost in loss record 7 of 7
==30062==    at 0x4C28C20: malloc (vg_replace_malloc.c:296)
==30062==    by 0x411C15: eMalloc (routines.c:216)
==30062==    by 0x41D15A: newToken (eiffel.c:617)
==30062==    by 0x41DD2D: parseType (eiffel.c:855)
==30062==    by 0x41DCBF: parseGeneric.isra.3 (eiffel.c:849)
==30062==    by 0x41DF35: parseClass (eiffel.c:1217)
==30062==    by 0x41DF35: findEiffelTags (eiffel.c:1245)
==30062==    by 0x40EE7F: createTagsForFile (parse.c:1687)
==30062==    by 0x4103FE: createTagsWithFallback (parse.c:1731)
==30062==    by 0x4103FE: parseFile (parse.c:1809)
==30062==    by 0x40942D: createTagsForEntry (main.c:241)
==30062==    by 0x402C37: createTagsForArgs (main.c:286)
==30062==    by 0x402C37: makeTags (main.c:432)
==30062==    by 0x402C37: main (main.c:517)
==30062== 
==30062== LEAK SUMMARY:
==30062==    definitely lost: 40 bytes in 1 blocks
==30062==    indirectly lost: 168 bytes in 6 blocks
==30062==      possibly lost: 0 bytes in 0 blocks
==30062==    still reachable: 0 bytes in 0 blocks
==30062==         suppressed: 0 bytes in 0 blocks
==30062== 
==30062== For counts of detected and suppressed errors, rerun with: -v
==30062== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

masatake:
Thank you. I am runinng noise target with VG=1.

vhda:
Just for the sake of documentation, using the `goto` solution:

```console
> valgrind --leak-check=full ./ctags --language-force=Eiffel -o- Units/fuzz-hitting-assertions.r/eiffel-fha-0.b/input.e 
==30792== Memcheck, a memory error detector
==30792== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==30792== Using Valgrind-3.10.0 and LibVEX; rerun with -h for copyright info
==30792== Command: ./ctags --language-force=Eiffel -o- Units/fuzz-hitting-assertions.r/eiffel-fha-0.b/input.e
==30792== 
==30792== 
==30792== HEAP SUMMARY:
==30792==     in use at exit: 0 bytes in 0 blocks
==30792==   total heap usage: 5,368 allocs, 5,368 frees, 302,326 bytes allocated
==30792== 
==30792== All heap blocks were freed -- no leaks are possible
==30792== 
==30792== For counts of detected and suppressed errors, rerun with: -v
==30792== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

and

```console
> valgrind --leak-check=full ./ctags --language-force=Eiffel -o- Units/review-needed.r/attributes.e.t/input.e 
==30793== Memcheck, a memory error detector
==30793== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==30793== Using Valgrind-3.10.0 and LibVEX; rerun with -h for copyright info
==30793== Command: ./ctags --language-force=Eiffel -o- Units/review-needed.r/attributes.e.t/input.e
==30793== 
ATTRIBUTES	Units/review-needed.r/attributes.e.t/input.e	/^expanded class ATTRIBUTES$/;"	c
anchored	Units/review-needed.r/attributes.e.t/input.e	/^	anchored: like one$/;"	f	class:ATTRIBUTES
eight	Units/review-needed.r/attributes.e.t/input.e	/^	eight: DOUBLE is 2.0$/;"	f	class:ATTRIBUTES
five	Units/review-needed.r/attributes.e.t/input.e	/^	five, six: INTEGER is unique$/;"	f	class:ATTRIBUTES
four	Units/review-needed.r/attributes.e.t/input.e	/^	four: DOUBLE$/;"	f	class:ATTRIBUTES
nine	Units/review-needed.r/attributes.e.t/input.e	/^	nine: STRING is "abc"$/;"	f	class:ATTRIBUTES
one	Units/review-needed.r/attributes.e.t/input.e	/^	one: INTEGER$/;"	f	class:ATTRIBUTES
seven	Units/review-needed.r/attributes.e.t/input.e	/^	seven: INTEGER is 1$/;"	f	class:ATTRIBUTES
six	Units/review-needed.r/attributes.e.t/input.e	/^	five, six: INTEGER is unique$/;"	f	class:ATTRIBUTES
three	Units/review-needed.r/attributes.e.t/input.e	/^	three: HASH_TABLE [ARRAY [DOUBLE], STRING]$/;"	f	class:ATTRIBUTES
two	Units/review-needed.r/attributes.e.t/input.e	/^	two: ARRAY [STRING]$/;"	f	class:ATTRIBUTES
==30793== 
==30793== HEAP SUMMARY:
==30793==     in use at exit: 0 bytes in 0 blocks
==30793==   total heap usage: 5,558 allocs, 5,558 frees, 309,225 bytes allocated
==30793== 
==30793== All heap blocks were freed -- no leaks are possible
==30793== 
==30793== For counts of detected and suppressed errors, rerun with: -v
==30793== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

masatake:
@vhda, I wonder why goto works fine.
As @b4n wrote, as far as using longjmp, we have risk of memory leaks.
In other work, as faras using longjmp/setjmp trashbox may be needed to avoid memory leaks.

vhda:
@masatake didn't we decide somewhere that we would stop using longjmp/setjmp?
But even if we keep using it, theoretically we could always free any allocated memory prior to the longjmp call, correct?

b4n:
> @vhda, I wonder why goto works fine.

It doesn't actually, @vhda called it on the wrong file, it should be `valgrind --leak-check=full ./ctags --language-force=Eiffel -o- Units/review-needed.r/attributes.cs.t/input.cs`

vhda:
Ups...
I confirm that I see it now.

b4n:
> But even if we keep using it, theoretically we could always free any allocated memory prior to the longjmp call, correct?

Sure, but this means that we need to know what to free at the longjmp() call site, probably making everything even more complex.

vhda:
@b4n I was just trying to justify that the code can still be made correct even without introducing trashBox.
I still prefer the current approach that we've been implementing in the various parsers.

masatake:
I withdraw this.

As @b4n showed, removing longjmp is better; (1)the assertion can be remove safely, (2)memory leak is removed, and (3)ctags can continue to find something to be tagged in input.

My patch can do only (1) and (2). 

As @vhda, we decided not using longjmp in the past discussion. However, In this patch I didn't try to remove longjmp. I estimated removing longjmp was not easy. So I just tried removing assertion in safe way; just aborting the processing.

@b4n removed the longjmp incredible quickly. I cannot image I do it. After reading the patch, I learnt some points.
(1) When passing a data from a variable to another valriable, do it by deep copying if there is a gap between extents(lifetime) of two variables. Passing by reference makes a trouble of memory management.
(2) After calling a function for reading a token, token->type == TOKEN_EOF must be check. It it is true, caller must return the control to upper layer after releasing resources in the function.

Above method may be applicable to flex.c and tex.c.
About ada, I need more inspection whether longjmp can be removed in low cost.

fortran.c and c.c are exceptions.  As far as doing grep on the code, the usage of longjmp is more complex than ther others. We will continue to use. (Actually I sent a PR for fortran.c using trashBox .)

PR #417, opened 2015-07-02
sql: Fix assertion on invalid or incomplete input

### LINKED ISSUE #399: SQL: fix the failure of fuzz-hitting-assertions/sql-fha-0

b4n:
parseRecord() is called without any checks, so need to be forgiving
with unexpected input.

Closes #399.


parsers/sql.c
```diff
@@ -1004,8 +1004,9 @@ static void parseRecord (tokenInfo *const token)
 	 */
 	if (!isType (token, TOKEN_OPEN_PAREN))
 		readToken (token);
+	if (!isType (token, TOKEN_OPEN_PAREN))
+		return;
```
vhda: Isn't this the same check as two lines above?

b4n: yes, but the first advances token.  it's "if I'm called on a non-paren, then I advance, but if next isn't a paren either I give up"

vhda: My bad.

masatake:
(I'm running noise .)

masatake:
It passed all tests. I'll merge this. Thank you.

masatake:
It passed all tests. I'll merge this. Thank you.

masatake:
@b4n, you made Units/sql.r directory. Can I rename it to parser-sql.r? (or should I remove "parser-" prefixes from Units/parser-<LANG>.r directories?)

b4n:
@masatake You can if you want, but I find the "parser" prefix a bit redundant as Units/ only contains parser tests.  But do as you prefer.

masatake:
@b4n, thank you replying. Units directory is still chaotic. So I will add the prefix.
In the future we can remove them.

vhda:
I think the current pattern is to have "<parser name>-<unit name>.d".
So the only thing missing would be the "-<unit name>" part.

masatake:
@vhda, sorry but I don't understand well. Could you tell me more?

vhda:
Hmmm... Maybe it was me who did not understand correctly.
Up until now we have not used the `parser-` prefix. Do you intend to move the `Tmain` units into `Units`? Is that the reason why you have added the `parser-` prefix?

masatake:
Oh, I see.

There are test cases that are not categorized into parser-<LANG>.r directories.

e.g. xcmd-etags.d and  extension-with-template-suffix.d.
Ideally I move such cases to Tmain. However, it takes cost.

So I decide to use "parser-" prefix.

vhda:
Well, personally I think that even without the `parser` prefix it should be easy to identify test groups. My `verilog-*` units are a good example of that. Adding that prefix only increases directory name size.
Nevertheless, if you feel that it is really necessary then I won't oppose to it.

b4n:
@vhda @masatake is only talking about `.r` (recursive) directories, e.g. directories grouping test cases, not individual test case directories.

vhda:
The comment still applies: having `verilog.r` should be enough to identify that this is parser related, no?

But it's possible that there's a corner case that I'm not remembering and for this reason I won't oppose to keeping the `parser-` prefix.

b4n:
> The comment still applies: having `verilog.r` should be enough to identify that this is parser related, no?

Yeah I agree and that's why I named SQL's one `sql.r`.  But just like you I don't mind either way, so IMO @masatake can do as he prefers.

PR #419, opened 2015-07-03
Encoding option

mattn:
When using `-R` and source codes is encoded with several encodings, tags file contains the encodings. So we can't jump to the location matched in tags file.
This patch add `--encoding` option to convert encoding from the source code to utf-8. For example, input.java is encoded as cp932, input.js is encoded as euc-jp, others are utf-8. Below should work well.

```
$ ctags --encoding=utf-8 --encoding-javascript=euc-jp --encoding-java=cp932 -R .
```
If you build ctags on windows mingw32, you must specify `WITH_ICONV=yes` like below.

```
C:\dev\ctags>mingw32-make -f mk_mingw.mak WITH_ICONV=yes
```


masatake:
I found Tmain test facility doesn't have enough ability to compare tags file.
I'm fixing the facility now.

mattn:
> I'm sorry but could you rename this to list-encodings to align other --list-*s options.

done

masatake:
Could you rebase on master branch and push again?
I introduced two new features to Tmain for testing your patch.
1. testing tags output, not stdout.
2. the way to skip a test for the case multibytes feature is disabled.

I update your test case locally. I will send a PR to your tree after you rebase your encoding-option branch on our master brach.

masatake:
One more requests about commit log.
Could you add "encoding: " prefix for your commits (including my commit about EncodingMap)?

If possible I would like you to squash many commits into one or a few.

mattn:
squashed

masatake:
Thanks.

1. It seems that --encoding option is not appeared in help messages(LongOptionDescription). Am I missing something?

2. "econding" is searched from --help output.  In stead could you use --list-features option?
    "multibyte" should be searched in this case.

3. Could you specify both input.java and init.js in run.sh instead of -R . ?
    So the test case becomes more specific.

4. Could you add "encoding: " prefix for each title of commit messages?

mattn:
done

mattn:
I'll fix this indentation.
https://github.com/mattn/ctags/commit/eecc38cb1749ab5fdaad9443847f5f588d68a0e7#diff-de50371a57c9c1e9bed8e1eb45920306R104

k-takata:
@mattn I don't mind if you squash my two commits into your commit.
I think it's better to remove `(Issue #14)` from my commit message, because the issue number is not for this repository. It's confusing.

mattn:
@k-takata ok, will do.

mattn:
@masatake done. please take a look.

mattn:
I am a principle to remain in the commit log everyone's contribution, but I'll do squash if you want a short commit log.

main/mbcs.c
```diff
@@ -0,0 +1,115 @@
...
+	if (!utf8)
+		return FALSE;
+	mbcs = vStringValue (string);
+retry:
+	if (iconv (iconv_fd, &mbcs, &mbcs_len, &utf8ptr, &utf8_len) == (size_t) -1)
+	{
+		eFree (utf8);
+		return FALSE;
+	}
+	if (errno == EILSEQ)
```
b4n: this will never happen because if `iconv()` encounters an invalid sequence it returns -1, for which you aborted already:

> […] The conversion can stop for four reasons:
>
>   1. An  invalid multibyte sequence is encountered in the input.  In this case, it sets `errno`
>          to `EILSEQ` and returns `(size_t) -1`.  `*inbuf` is left pointing  to  the  beginning  of  the
>          invalid multibyte sequence.

b4n: hum, worse, it might happen if iconv() failed *in the previous run*, as `errno` is not reset.

main/mbcs.c
```diff
@@ -0,0 +1,115 @@
...
+	utf8_size = utf8ptr - utf8;
+
+	vStringClear (string);
+	while (vStringSize (string) <= utf8_size + 1)
+		vStringAutoResize (string);
+	memcpy (vStringValue (string), utf8, utf8_size + 1);
+	vStringLength (string) = utf8_size;
+	eFree (utf8);
+
+	iconv (iconv_fd, (void*)0, (void*)0, &utf8ptr, &utf8_len);
```
b4n: hum, what? 

> A  different case is when inbuf is `NULL` or `*inbuf` is `NULL`, but `outbuf` is not `NULL` and `*outbuf` is not `NULL.`  In this case, the `iconv()` function attempts to set `cd`'s conversion  state to  the  initial  state and store a corresponding shift sequence at `*outbuf`.  At most `*outbytesleft` bytes, starting at `*outbuf`, will be written.  If the output buffer  has  no  more   room  for  this reset sequence, it sets `errno` to `E2BIG` and returns `(size_t) -1`.  Otherwise, it increments `*outbuf` and decrements `*outbytesleft` by the number of bytes written.

Don't you rather want this one?: (e.g. all `NULL`s)

> A third case is when `inbuf` is `NULL` or `*inbuf` is `NULL`, and `outbuf`  is  `NULL`  or  `*outbuf`  is  `NULL.`  In this case, the `iconv()` function sets `cd`'s conversion state to the initial state.

b4n: BTW, in C we have `NULL` as a better and more correct version of `(void*)0` ;)

main/options.c
```diff
@@ -230,6 +236,12 @@ static optionDescription LongOptionDescription [] = {
  {1,"      Include reference to 'file' in Emacs-style tag file (requires -e)."},
  {1,"  --exclude=pattern"},
  {1,"      Exclude files and directories matching 'pattern'."},
+#if defined(HAVE_ICONV) && defined(iconvlist)
+ {1,"  --encoding=encoding"},
+ {1,"      Specify encoding of all of source files."},
+ {1,"  --encoding-<LANG>=encoding"},
+ {1,"      Specify encoding of the LANG source file."},
```
b4n: file*s* I think

main/parse.c
```diff
@@ -1781,6 +1784,62 @@ static void printGuessedParser (const char* const fileName, langType language)
...
+		}
+		EncodingMapMax = language;
+	}
+	if (EncodingMap [language])
+		eFree (EncodingMap [language]);
+	EncodingMap [language] = eStrdup(encoding);
+}
+
+extern boolean processLanguageEncodingOption (const char *const option,
+									const char *const parameter __unused__)
```
b4n: it's used, actually

main/parse.c
```diff
@@ -1803,6 +1862,12 @@ extern boolean parseFile (const char *const fileName)
 	{
 		if (Option.filter)
 			openTagFile ();
+
+#ifdef HAVE_ICONV
+		openConverter (EncodingMap && language <= EncodingMapMax ?
+				EncodingMap [language] : Option.encoding);
```
b4n: shouldn't this also check whether `EncodingMap [language]` is `NULL`?
`(EncodingMap && language <= EncodingMapMax && EncodingMap [language]) ? EncodingMap [language] : Option.encoding`

main/parse.h
```diff
@@ -178,6 +178,12 @@ extern void printLanguageAliases (const langType language);
 extern void printLanguageList (void);
 extern boolean parseFile (const char *const fileName);
 
+#ifdef HAVE_ICONV
+extern boolean processLanguageEncodingOption (const char *const option,
+								   const char *const parameter __unused__);
```
b4n: no need to tag unused in the prototype (especially as it's used :)

main/routines.c
```diff
@@ -545,10 +555,24 @@ extern const char *baseFilename (const char *const filePath)
...
+
+		for (p = filePath  ;  *p != '\0'  ;  ++p)
+		{
+			ml = mblen(p, MB_LEN_MAX);
+			if (ml > 1)
+				p += ml - 1;
+			else if (*p == PathDelimiters [i] && p > tail)
+				tail = p;
+		}
+#else
```
b4n: why?  if you mean to support non-ASCII-compatible encodings, this seems incorrect to me.  If you don't, what's the point?
And anyway, if we really want to support on-ASCII-compatible encodings, we probably need a lot more work (including converting the input files content, as all parsers assume ASCII).

But note that of course, UTF-8 is ASCII-compatible.

main/mbcs.c
```diff
@@ -0,0 +1,115 @@
...
+	utf8_size = utf8ptr - utf8;
+
+	vStringClear (string);
+	while (vStringSize (string) <= utf8_size + 1)
+		vStringAutoResize (string);
+	memcpy (vStringValue (string), utf8, utf8_size + 1);
+	vStringLength (string) = utf8_size;
+	eFree (utf8);
+
+	iconv (iconv_fd, NULL, NULL, &utf8ptr, &utf8_len);
```
b4n: hum, what? 

> A  different case is when inbuf is `NULL` or `*inbuf` is `NULL`, but `outbuf` is not `NULL` and `*outbuf` is not `NULL.`  In this case, the `iconv()` function attempts to set `cd`'s conversion  state to  the  initial  state and store a corresponding shift sequence at `*outbuf`.  At most `*outbytesleft` bytes, starting at `*outbuf`, will be written.  If the output buffer  has  no  more   room  for  this reset sequence, it sets `errno` to `E2BIG` and returns `(size_t) -1`.  Otherwise, it increments `*outbuf` and decrements `*outbytesleft` by the number of bytes written.

Don't you rather want this one?: (e.g. all `NULL`s)

> A third case is when `inbuf` is `NULL` or `*inbuf` is `NULL`, and `outbuf`  is  `NULL`  or  `*outbuf`  is  `NULL.`  In this case, the `iconv()` function sets `cd`'s conversion state to the initial state.


(repost as GitHub ditched it when you updated the `NULL`s)

b4n:
Don't we want a way to generate something else than UTF-8?  I love UTF-8, but if people have files in weird encodings why wouldn't they rather target that one and only fix a few files?  Supporting this would mean adding a `--output-encoding` option or similar.

main/parse.c
mattn: you are right. will fix

main/routines.c
mattn: cp932 have some letters that contains \x5c in trailing byte of DBCS(double byte character set).
So it should be skipped.

main/parse.c
mattn: so no need to do?

main/mbcs.c
mattn: so no need to do?

mattn:
in japanese, there are three encodings that representing japanese. cp932, euc-jp, iso-2022-jp.
and in the world, many encodings are there. so tags file should be the encoding that can be included like utf-8.

b4n:
While converting the output is nice, if we really want to have complete proper support for all kind of encodings we'd need to convert the *input* passed to parsers, not what parsers output.  All parsers expect ASCII-compatible input, that is any of the ASCII-based codepages, and UTF-8.  If one feeds them e.g. UTF-16, they won't be able to do anything sensible.

So to fix this we'd need to convert the whole (parsed) input as whatever fits (using UTF-8 would probably be a good choice as it can represent mostly everything) if the input is not ASCII-compatible.  It'd be a little tricky if we wanna support non-UTF-8 output as I suggested above, as we could imagine the whole file content not to be convertible to e.g. ISO-8859-1, but the output part to be (if incompatible bytes are not present in the tags).  And even a bit trickier if we won't wanna convert if the input is ASCII-compatible.  But nothing impossible.

b4n:
> in japanese, there are three encodings that representing japanese. cp932, euc-jp, iso-2022-jp.
and in the world, many encodings are there. so tags file should be the encoding that can be included like utf-8.

I don't see the point.  Yes UTF-8 is better.  But what if I want only ISO-8859-1 but I have one file in UTF-8?  (the UTF-8 file containing a character that ISO-8859-1 can represent, e.g. `é`)

mattn:
> I don't see the point. Yes UTF-8 is better. But what if I want only ISO-8859-1 but I have one file in UTF-8? (the UTF-8 file containing a character that ISO-8859-1 can represent, e.g. é)

Then you don't need to specify `--encoding`. ctags will do nothing.

main/routines.c
b4n: Hum, so it's a special case for one single encoding that happens to be "somewhat" ASCII compatible?  But then, what about e.g. UTF-16?  UTF-32?  IMO this would require real converting, knowing the input, and then working on a known result.

b4n:
> Then you don't need to specify `--encoding`. ctags will do nothing.

I do, because my UTF-8 file contains a `é` (`\xa9\xc3`), and I don't wanna see `Ã©`.

main/parse.c
b4n: yeah so just drop the `__unused__`

main/mbcs.c
b4n: it depends, if you want to accept invalid input and replace the offending bytes, you need to handle it *when `iconv()` returned `-1`*.  Otherwise, just drop it.
BTW, the common UTF-8 "invalid byte" character is `�`, not `?`.

main/parse.c
mattn: Ah, good catch. Thanks.

main/routines.c
mattn: Users who want `--encoding` option are user who use text editor with utf-8. Users who don't want it are not user of utf-8 encoding. I guess.

main/mbcs.c
mattn: I want to handle invalid bytes to replace it to `?`. because vim also can't handle it.

mattn:
filled NULL into iconv().

mattn:
> I do, because my UTF-8 file contains a é (\xa9\xc3), and I don't wanna see Ã©.

Well, why you need `--encoding`? You shouldn't need conversion since you have utf-8 file and tags file be utf-8.

b4n:
> Well, why you need `--encoding`? You shouldn't need conversion since you have utf-8 file and tags file be utf-8.

No no, my point is, imagine I have this:
* *file1.c* (ISO-8859-1)
* *file2.c* (ISO-8859-1)
* *file3.c* (UTF-8)
* *file4.c* (ISO-8859-1)

And I want a ISO-8859-1 output, because it's my preferred encoding (the UTF-8 file is a anomaly).

main/mbcs.c
b4n: So handle the case where `errno` is `EILSEQ` in the code path checking if `iconv()` returned `-1`:
```C
if (iconv (...) == (size_t) -1)
{
    if (errno == EILSEQ)
    {
        /* handle the invalid sequence */
    }
    else
    {
        eFree (utf8);
        return FALSE;
    }
}
```

mattn:
So we've better to rename `--encoding` to `--input-encoding` and add `--output-encoding` ?

main/routines.c
b4n: First, this code path has nothing to do with `--encoding` AFAICT.

Then, well, as said if we wanted to do it right, we should find the path encoding (basing this on the locale is *somewhat* correct) and convert it to e.g. UTF-8, which we can trust.  What is here is IMO a hack at best.

If we really want to keep this code, I'd say we need at least a comment explaining why it's OK to do this wrong for the designed purpose (so someone like me won't wonder what the heck is this :)

b4n:
If we want to be flexible, yes.  But in the end apart that not supporting this would seem like an arbitrary limitation, I don't care much myself, UTF-8 is my preferred encoding anyway :)

main/routines.c
mattn: > First, this code path has nothing to do with --encoding AFAICT.

Sorry I was confused. This code is needed for cp932. Will add comment.

mattn:
added `--output-encoding`.

main/mbcs.h
```diff
@@ -0,0 +1,25 @@
...
+*
+*   This module contains functions for checking multibyte character set.
+*/
+
+#include "general.h"  /* must always come first */
+#include "vstring.h"
+
+#ifdef HAVE_ICONV
+
+extern boolean isConverting ();
```
b4n: BTW, proper C needs `(void)` for "no argument", `()` is indefinite.

main/entry.c
```diff
@@ -182,6 +182,10 @@ static void addPseudoTags (void)
 		writePseudoTag ("TAG_PROGRAM_NAME",    PROGRAM_NAME, "", NULL);
 		writePseudoTag ("TAG_PROGRAM_URL",     PROGRAM_URL,  "official site", NULL);
 		writePseudoTag ("TAG_PROGRAM_VERSION", PROGRAM_VERSION, "", NULL);
+#ifdef HAVE_ICONV
+		if (Option.outputEncoding)
+			writePseudoTag ("TAG_FILE_ENCODING", "utf-8", "", NULL);
```
b4n: `"utf-8"` here should probably be `Option.outputEncoding`, right?

mattn: u are right.

masatake:
@ffes, @cweagans, I have a platform related question.
On your platfrom(Windows and OSX) where ctags run, is iconv command available?
If you I would like know its option. Does your iconv command have -l or --list option?

mattn:
On Windows
```
ANSI_X3.4-1968 ANSI_X3.4-1986 ASCII CP367 IBM367 ISO-IR-6 ISO646-US ISO_646.IRV:1991 US US-ASCII CSASCII
UTF-8
ISO-10646-UCS-2 UCS-2 CSUNICODE
UCS-2BE UNICODE-1-1 UNICODEBIG CSUNICODE11
UCS-2LE UNICODELITTLE
ISO-10646-UCS-4 UCS-4 CSUCS4
UCS-4BE
UCS-4LE
UTF-16
UTF-16BE
UTF-16LE
UTF-32
UTF-32BE
UTF-32LE
UNICODE-1-1-UTF-7 UTF-7 CSUNICODE11UTF7
UCS-2-INTERNAL
UCS-2-SWAPPED
UCS-4-INTERNAL
UCS-4-SWAPPED
C99
JAVA
CP819 IBM819 ISO-8859-1 ISO-IR-100 ISO8859-1 ISO_8859-1 ISO_8859-1:1987 L1 LATIN1 CSISOLATIN1
ISO-8859-2 ISO-IR-101 ISO8859-2 ISO_8859-2 ISO_8859-2:1987 L2 LATIN2 CSISOLATIN2
ISO-8859-3 ISO-IR-109 ISO8859-3 ISO_8859-3 ISO_8859-3:1988 L3 LATIN3 CSISOLATIN3
ISO-8859-4 ISO-IR-110 ISO8859-4 ISO_8859-4 ISO_8859-4:1988 L4 LATIN4 CSISOLATIN4
CYRILLIC ISO-8859-5 ISO-IR-144 ISO8859-5 ISO_8859-5 ISO_8859-5:1988 CSISOLATINCYRILLIC
ARABIC ASMO-708 ECMA-114 ISO-8859-6 ISO-IR-127 ISO8859-6 ISO_8859-6 ISO_8859-6:1987 CSISOLATINARABIC
ECMA-118 ELOT_928 GREEK GREEK8 ISO-8859-7 ISO-IR-126 ISO8859-7 ISO_8859-7 ISO_8859-7:1987 CSISOLATINGREEK
HEBREW ISO-8859-8 ISO-IR-138 ISO8859-8 ISO_8859-8 ISO_8859-8:1988 CSISOLATINHEBREW
ISO-8859-9 ISO-IR-148 ISO8859-9 ISO_8859-9 ISO_8859-9:1989 L5 LATIN5 CSISOLATIN5
ISO-8859-10 ISO-IR-157 ISO8859-10 ISO_8859-10 ISO_8859-10:1992 L6 LATIN6 CSISOLATIN6
ISO-8859-13 ISO-IR-179 ISO8859-13 ISO_8859-13 L7 LATIN7
ISO-8859-14 ISO-CELTIC ISO-IR-199 ISO8859-14 ISO_8859-14 ISO_8859-14:1998 L8 LATIN8
ISO-8859-15 ISO-IR-203 ISO8859-15 ISO_8859-15 ISO_8859-15:1998 LATIN-9
ISO-8859-16 ISO-IR-226 ISO8859-16 ISO_8859-16 ISO_8859-16:2001 L10 LATIN10
KOI8-R CSKOI8R
KOI8-U
KOI8-RU
CP1250 MS-EE WINDOWS-1250
CP1251 MS-CYRL WINDOWS-1251
CP1252 MS-ANSI WINDOWS-1252
CP1253 MS-GREEK WINDOWS-1253
CP1254 MS-TURK WINDOWS-1254
CP1255 MS-HEBR WINDOWS-1255
CP1256 MS-ARAB WINDOWS-1256
CP1257 WINBALTRIM WINDOWS-1257
CP1258 WINDOWS-1258
850 CP850 IBM850 CSPC850MULTILINGUAL
862 CP862 IBM862 CSPC862LATINHEBREW
866 CP866 IBM866 CSIBM866
MAC MACINTOSH MACROMAN CSMACINTOSH
MACCENTRALEUROPE
MACICELAND
MACCROATIAN
MACROMANIA
MACCYRILLIC
MACUKRAINE
MACGREEK
MACTURKISH
MACHEBREW
MACARABIC
MACTHAI
HP-ROMAN8 R8 ROMAN8 CSHPROMAN8
NEXTSTEP
ARMSCII-8
GEORGIAN-ACADEMY
GEORGIAN-PS
KOI8-T
MULELAO-1
CP1133 IBM-CP1133
ISO-IR-166 TIS-620 TIS620 TIS620-0 TIS620.2529-1 TIS620.2533-0 TIS620.2533-1
CP874 WINDOWS-874
VISCII VISCII1.1-1 CSVISCII
TCVN TCVN-5712 TCVN5712-1 TCVN5712-1:1993
ISO-IR-14 ISO646-JP JIS_C6220-1969-RO JP CSISO14JISC6220RO
JISX0201-1976 JIS_X0201 X0201 CSHALFWIDTHKATAKANA
ISO-IR-87 JIS0208 JIS_C6226-1983 JIS_X0208 JIS_X0208-1983 JIS_X0208-1990 X0208 CSISO87JISX0208
ISO-IR-159 JIS_X0212 JIS_X0212-1990 JIS_X0212.1990-0 X0212 CSISO159JISX02121990
CN GB_1988-80 ISO-IR-57 ISO646-CN CSISO57GB1988
CHINESE GB_2312-80 ISO-IR-58 CSISO58GB231280
CN-GB-ISOIR165 ISO-IR-165
ISO-IR-149 KOREAN KSC_5601 KS_C_5601-1987 KS_C_5601-1989 CSKSC56011987
EUC-JP EUCJP EXTENDED_UNIX_CODE_PACKED_FORMAT_FOR_JAPANESE CSEUCPKDFMTJAPANESE
MS_KANJI SHIFT-JIS SHIFT_JIS SJIS CSSHIFTJIS
CP932
ISO-2022-JP CSISO2022JP
ISO-2022-JP-1
ISO-2022-JP-2 CSISO2022JP2
CN-GB EUC-CN EUCCN GB2312 CSGB2312
CP936 GBK MS936 WINDOWS-936
GB18030
ISO-2022-CN CSISO2022CN
ISO-2022-CN-EXT
HZ HZ-GB-2312
EUC-TW EUCTW CSEUCTW
BIG-5 BIG-FIVE BIG5 BIGFIVE CN-BIG5 CSBIG5
CP950
BIG5-HKSCS BIG5HKSCS
EUC-KR EUCKR CSEUCKR
CP949 UHC
CP1361 JOHAB
ISO-2022-KR CSISO2022KR
CP856
CP922
CP943
CP1046
CP1124
CP1129
CP1161 IBM-1161 IBM1161 CSIBM1161
CP1162 IBM-1162 IBM1162 CSIBM1162
CP1163 IBM-1163 IBM1163 CSIBM1163
DEC-KANJI
DEC-HANYU
437 CP437 IBM437 CSPC8CODEPAGE437
CP737
CP775 IBM775 CSPC775BALTIC
852 CP852 IBM852 CSPCP852
CP853
855 CP855 IBM855 CSIBM855
857 CP857 IBM857 CSIBM857
CP858
860 CP860 IBM860 CSIBM860
861 CP-IS CP861 IBM861 CSIBM861
863 CP863 IBM863 CSIBM863
CP864 IBM864 CSIBM864
865 CP865 IBM865 CSIBM865
869 CP-GR CP869 IBM869 CSIBM869
CP1125
EUC-JISX0213
SHIFT_JISX0213
ISO-2022-JP-3
ISO-IR-230 TDS565
RISCOS-LATIN1
```

masatake:
@mattn, thanks. Did you get the list with 'iconv -l'?

mattn:
Yes, I did `iconv -l`.

masatake:
My colleague gave me. MacOSX has iconv -l, too. 
@mattn, I'm very sorry, but could you remove --list-encodings related code from your patch set?
(I asked him to implement --list-encodings option at https://github.com/vim-jp/ctags/issues/25 )

mattn:
NP! Will do :)

mattn:
done

masatake:
> $ make tmain

reports:

> Failed tests
encoding-option/stderr-compare
encoding-option/exit-compare
encoding-option/tags-compare

masatake:
It seems that --input-encoding-<LANG>= doesn't work:

> [yamato@x201]~/var/ctags-github% ./ctags --input-encoding-java=utf-8
ctags: Unknown option: --input-encoding-java
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding=utf-8     
ctags: No files specified. Try "ctags --help".
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding-Java=utf-8
ctags: Unknown option: --input-encoding-Java
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding-c=utf-8   
ctags: Unknown option: --input-encoding-c
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding-C=utf-8
ctags: Unknown option: --input-encoding-C

mattn:
did you do `./configure --enable-iconv` ?

masatake:
Yes. I added --enable-iconv when running configure.
Actually --input-encoding=utf-8 works fine.
I wonder I'm testing older version.
```
[yamato@x201]~/var/ctags-github% git log --oneline | head -3
9ee0f52 Merge branch 'encoding-option' of https://github.com/mattn/ctags into mattn-encoding-option
0f1980e encoding: add unit test for --input-encoding, --output-encoding option
0ef2f8e encoding: add --input-encoding, --output-encoding option to convert encodings
[yamato@x201]~/var/ctags-github% ./ctags --help | grep input-
  --input-encoding=encoding
  --input-encoding-<LANG>=encoding
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding-c=utf-8 main/options.c
ctags: Unknown option: --input-encoding-c
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding-C=utf-8 main/options.c
ctags: Unknown option: --input-encoding-C
[yamato@x201]~/var/ctags-github% ./ctags --input-encoding=utf-8 main/options.c
[yamato@x201]~/var/ctags-github% head -2 tags
!_TAG_FILE_ENCODING	UTF-8	//
```

mattn:
```
--input-encoding-=
```

Ah, I miss the last dash.

mattn:
Sorry, fixed.

masatake:
make tmain says:

```
Testing encoding-option
skipped
```

No testing is conducted.

mattn:
test passed

masatake:
Updating the patch has been so frequently, I could not track the reviewers comments.
Does any issue remain?

mattn:
I'm thinking, I want to add few tests into encoding-option.d. But I don't have a time to do just now.

masatake:
I see. Please, do it when you have time.

mattn:
added two tests.

* input-encoding-option.d: test of --input-encoding is working
* output-encoding-option.d: test of --input-encoding is working
* invalid-encoding-option.d: test of failing with specify invalid encoding name.

masatake:
@mattn, thank you. 

I have no more thing. @b4n, how about you?

main/mbcs.c
```diff
@@ -0,0 +1,100 @@
...
+}
+
+extern boolean isConverting ()
+{
+	return iconv_fd != (iconv_t) -1;
+}
+
+extern boolean convertString (vString *const string)
+{
+	size_t utf8_len = 0, mbcs_len, utf8_size;
```
b4n: all references to "UTF8" here are misleading, as it might not be UTF8 (anymore).
I might instead use `src` and `dest`, `input` and `output`, or `raw` and `conv` or something.

main/mbcs.c
```diff
@@ -0,0 +1,100 @@
...
+	char *utf8, *mbcs, *utf8ptr;
+	if (iconv_fd == (iconv_t) -1)
+		return FALSE;
+	mbcs_len = vStringLength (string);
+	utf8_len = mbcs_len * 4;
```
b4n: This is probably still OK as I don't know an encoding (actually) using more than 4 bytes per code point, but this might warrant a comment

main/options.c
```diff
@@ -302,6 +312,10 @@ static optionDescription LongOptionDescription [] = {
  {1,"       Output list of language mappings."},
  {1,"  --map-<LANG>=[+]map"},
  {1,"       Alternative version of --langmap option."},
+#ifdef HAVE_ICONV
+ {1,"  --output-encoding=encoding"},
+ {1,"      Specify encoding to write tags file."},
```
b4n: might be interesting to mention it defaults to UTF-8

b4n: If only `--output-encoding` is provided, nothing happens at all.  It's relatively confusing, so I'd say that we either need to warn if ``output-encoding| is used without `--input-encoding`, or make `--input-encoding` default to something sensible when `--output-encoding`  is given.

b4n:
The commit message for e0d2287 contains some things that shouldn't be here IMO.  Comments on design decision or details are great, but "fixed leak" or "fixed AH_TEMPLATE usage" shouldn't be here IMO as it just acknowledge the first version of the patch had issues.

b4n:
Apart that, looks good

mattn:
@b4n Done. Thanks for your review. Since I'm not native english speaker, some wording in commit-comment may not  be correct.

main/mbcs.c
```diff
@@ -0,0 +1,117 @@
...
+#include "mbcs.h"
+#include "routines.h"
+
+static iconv_t iconv_fd = (iconv_t) -1;
+
+extern boolean openConverter (char* inputEncoding, char* outputEncoding)
+{
+	if (!inputEncoding || !outputEncoding)
+	{
+		static boolean warn;
```
b4n: technically this is valid, but I'd rather see an explicit initialization `= FALSE`

main/mbcs.c
```diff
@@ -0,0 +1,117 @@
...
+		if (!warn)
+		{
+			if (outputEncoding)
```
b4n: I'd rather use `! inputEncoding`

main/mbcs.c
```diff
@@ -0,0 +1,117 @@
...
+			{
+				error (WARNING, "--input-encoding is not specified");
+				warn = TRUE;
+			}
+			if (inputEncoding)
```
b4n: same here, I'd rather use `! outputEncoding`

main/options.c
```diff
@@ -302,6 +312,11 @@ static optionDescription LongOptionDescription [] = {
  {1,"       Output list of language mappings."},
  {1,"  --map-<LANG>=[+]map"},
  {1,"       Alternative version of --langmap option."},
+#ifdef HAVE_ICONV
+ {1,"  --output-encoding=encoding"},
+ {1,"      Specify encoding to write tags file. Default value is UTF-8,"},
+ {1,"      --output-encoding is provided. If not, nothing happen."},
```
b4n: I guess the  `--output-encoding` mention in the description shuold rather be `--input-encoding`?

I guess I'd write something like this:  "The encoding to write the tag file in.  Defaults to UTF-8 if `--input-encoding` is specified, otherwise no conversion is performed."
or something like it.

main/mbcs.c
mattn: This mean:
--output-encoding is specified but not --input-encoding provided.

So second part is needless. Then I removed in latest.

main/options.c
mattn: will fix

main/mbcs.c
b4n: I get what it's here for, but the message says "input encoding not specified" yet the test is "output encoding".  This only works because you know that at this point either input our output encoding is NULL, but not both -- so IMO, it'd be more clear and simple to just perform the actual test the warning is about.

masatake:
Reviews here may be very familiar with non-native English. I train them every day:-)

@mattn, I've tried to write about your patch and summarize the comments here. Could you look at this? (I have to write this anyway to update ctags(1) man page.) Feel free to use the parts of mine in your commit log. I've referred your comment given to me at https://github.com/vim-jp/ctags/issues/25

```
Japanese programmers sometimes use Japanese language in comments in
source code.

ctags didn't care it. ctags just reads input as just bytes sequence and
use them as is when writing tags entries.

In other hand vim cares it. When loading a file, vim converts the file
content internal format with one of encodings specified in
fileencodings variable.

As the result of this difference, vim cannot move the cursor to the
definition of a tag as users expect with pattern matching.

Good news is that there is a way to notify vim the encoding used in a
tags file with _TAG_FILE_ENCODING pseudo tag in the tag file.

This patch solves this issue utilizing _TAG_FILE_ENCODING pseudo tag.

This patch introduces two type of options to 

--input-encoding=IN
--output-encoding=OUT

As specified encoding with these options ctags converts input from IN
encoding to OUT encoding. Converted input is passed to parsers. As the
result tags output is encoded in OUT encoding.  In addition OUT is
specified in the top tags file as value for _TAG_FILE_ENCODING pseudo tag.
As OUT utf-8 is as default.

With --input-encoding-<LANG>=IN, you can specify <LANG> own
input encoding. It overrides the global default value given with
--input-encoding.

Acceptable IN and OUT can be listed with `iconv -l' or `iconv --list'.
It is up to platform where ctags runs.

NOTE: The parsers still deal with input as bytes sequence.
```

mattn:
If you don't mind, I want to put whole of your comment.

main/mbcs.c
mattn: If either --input-encoding or --output-encoding is NULL, this function will crash. The first check is for this reason. For the half of the conditions, the second check are satisfied in the first check.

masatake:
Of course it is o.k. But please, check the contents. Not about English, just about technical aspect.

Here is Japanese translation.

```
日本のプログラマはプログラム中のコメントに日本語を使うことがあります。

ctagsは日本語を考慮しません。なんであれ、バイト列として扱います。
そしてそのまま結果をtagsファイルへ書き込みます。

一方vimは考慮します。ファイルを読み込むとき、fileencodings変数に
指定されたエンコーディングの一つに従い内部表現へ変更します。

その結果、vimはユーザの期待する通りに、パターンマッチによってcursorを
タグの定義へ移動させることができません。

幸いなことに、tagsファイル中で_TAG_FILE_ENCODING疑似タグを使う
ことで、vimへtagsファイルのencodingが何かを伝えることはできます。

このパッチは、_TAG_FILE_ENCODING疑似タグここで指摘した問題を解決します。

2種類のオプションを追加します。

--input-encoding=IN
--output-encoding=OUT

これらのオプションの指定に従ってctagsは 入力を INからOUTへコード変換
します。変換た結果が各言語のパーザへ渡されます。結果としてtagsファイルは
OUTで指定したencodingで出力されます。さらにOUTで指定したencodingが
_TAG_FILE_ENCODING疑似タグの値としてtagsファイルの先頭に記載されます。
OUTのデフォルトはutf-8です。

--input-encoding-<LANG>=IN を使うと、<LANG>固有のencodingを指定できます。
--input-encoding-<LANG>=INによる指定は--input-encodingに優先します。

INあるいはOUTで指定できるencodingのリストは iconv -lあるいは iconv --list
で一覧できます。プラットフォームによって異なります。

ただし書き: パーザは相変らず入力をバイト列として扱います。

```

mattn:
@masatake perfect :100:

b4n:
> Japanese programmers sometimes use Japanese language in comments in
source code.

it's not limited to Japanese programmers ;)  Many (young) programmers not speaking English very well will often use their native language in comments and strings, and they will often contain non-ASCII characters :)

> Converted input is passed to parsers

That's not true, that's not what the current patch does.  The conversion only happens in `readLine()`, which is used to read the line to fetch the pattern right before writing the output -- but *not* what reads input source file data for parses.

If we wanted to convert everything (which as already mentioned would be a lot larger but would be needed to really support non-ASCII-like input files), we'd need to convert in e.g. `iFileGetLine()` (I didn't investigate much).  But actually we need to convert even before that, because `iFileGetLine()` has to recognize line ends -- and then we might need to convert each character, which can be tricky and/or expansive.

masatake:
@mattn, thank you for checking. But please go your bed.

I will reflect the comments from @b4n, tomorrow.

Late night in Japan. Very tired.

masatake:
Updated.
I reflect the comment from @b4n.

Please, feel free to copy and paste from  any parts of my comment to your commit log.

I hope this one is the last:-)

```
Japanese programmers sometimes use Japanese language in comments in
source code. Of course it is not limited to Japanese. People may use
their own native language in some case. In such case conding becomes
an issue.

ctags didn't care it. ctags just reads input as just bytes sequence and
use them as is when writing tags entries.

In other hand vim cares it. When loading a file, vim converts the file
content internal format with one of encodings specified in
fileencodings variable.

As the result of this difference, vim cannot move the cursor to the
definition of a tag as users expect with pattern matching. ctags writes
patterns in tags file.

Good news is that there is a way to notify vim the encoding used in a
tags file with _TAG_FILE_ENCODING pseudo tag in the tag file.

This patch solves this issue utilizing _TAG_FILE_ENCODING pseudo tag.

This patch introduces two type of options to 

--input-encoding=IN
--output-encoding=OUT

As specified encoding with these options ctags converts input from IN
encoding to OUT encoding. ctags uses the conveted strings when writing
pattern parts of tags lines. As the result tags output is encoded in
OUT encoding.  In addition OUT is specified in the top tags file as
value for _TAG_FILE_ENCODING pseudo tag.  As OUT utf-8 is as default.

NOTE: Converted input is NOT passed to language parsers.
      The parsers still deal with input as bytes sequence.

With --input-encoding-<LANG>=IN, you can specify <LANG> own
input encoding. It overrides the global default value given with
--input-encoding.

The example usage can be found in Tmain/{input,output}-encoding-option.d.

Acceptable IN and OUT can be listed with `iconv -l' or `iconv --list'.
It is up to platform where ctags runs.

To enable the option, libiconv is needed in your platform. In addition --enable-iconv
must be specified to configure before making ctags. On windows mingw32, you must specify WITH_ICONV=yes like below.

       C:\dev\ctags>mingw32-make -f mk_mingw.mak WITH_ICONV=yes
```

k-takata:
@masatake
I found some typos.

> In such case conding becomes

'conding' -> 'coding' or 'encoding'

> vim converts the file content internal format

Maybe 'into' is missing?
'vim converts the file content into internal format'

> ctags uses the conveted strings when writing

'conveted' -> 'converted'

masatake:
@k-takata, thank you.
I reedit my comment directly.

masatake:
Thank you for contribution.

masatake:
(TODO: updating docs/news.rst.)

k-takata:
I think it's better to update ctags.1.in too.

PR #421, opened 2015-07-05
tex: don't make a tag for a zero length name

masatake:
ctags warns when following input gives:

     \section{}

because a zero length name is passed to makeTagEntry.

This patch makes the tex parser not pass such
zero length names.


b4n:
LGTM

PR #427, opened 2015-07-07
debug: Print currently parsed file and line number on asserting

### LINKED ISSUE #420: Reporting the current input file name when assertion failure
I'm looking for a neat way to print the current input file name when an assertion is made failure.
If you have a good idea or code, please, let me know.

b4n:
If an assertion occurs during parsing of a file, print the location in the parsed input where the assertion occurred.

Closes #420.

### DIFF
--- a/main/debug.c
+++ b/main/debug.c
@@ -13,6 +13,8 @@
 #include "general.h"  /* must always come first */
 
 #include <ctype.h>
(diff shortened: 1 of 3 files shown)


b4n:
This mimics glibc's `assert()`, but adds an extra line with currently parsed file information if available.

PR #430, opened 2015-07-10
c++: Handle C++11 '= delete'

vmiklos:
In C++03 '= 0' was only valid if the member function was virtual, but in
C++11 '=' may be used for non-virtual member functions, too.

Without the change, the 'Del::operator =' declaration is not found in
the testcase.

### DIFF
--- a/Units/cxx11-delete.d/args.ctags
+++ b/Units/cxx11-delete.d/args.ctags
@@ -1,0 +1,2 @@
+--c++-kinds=+p
+--fields=+S
--- a/Units/cxx11-delete.d/expected.tags
+++ b/Units/cxx11-delete.d/expected.tags
@@ -1,0 +1,3 @@
+Del	input.cpp	/^    Del(const Del &) = delete;$/;"	p	class:Del	file:	signature:(const Del &)
+Del	input.cpp	/^class Del$/;"	c	file:
+operator =	input.cpp	/^    void operator=(const Del& rDel) = delete;$/;"	p	class:Del	file:	signature:(const Del& rDel)
--- a/Units/cxx11-delete.d/input.cpp
+++ b/Units/cxx11-delete.d/input.cpp
@@ -1,0 +1,6 @@
+class Del
+{
+private:
+    Del(const Del &) = delete;
+    void operator=(const Del& rDel) = delete;
+};
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -2655,7 +2655,7 @@ static void analyzeParens (statementInfo *const st)
 				 ! st->gotParenName  &&
 				 (! info.isParamList || ! st->haveQualifyingName  ||
 				  c == '('  ||
-				  (c == '='  &&  st->implementation != IMP_VIRTUAL) ||
+				  (c == '='  &&  st->implementation != IMP_VIRTUAL && !isLanguage (Lang_cpp)) ||
 				  (st->declaration == DECL_NONE  &&  isOneOf (c, ",;"))))
 		{
 			token->type = TOKEN_NAME;


masatake:
Thank you for sending a PR.
I don't know c++ well, so I would like to ask some questions?

1. In your test case, there two '= delete;'.  Both are important in your patch? (I guess only '= delete' after operator = is important.)

2. The string 'delete' is important here? In other word it is something like keyword? (I guess any name is o.k.)

```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.h                        
class Del
{
    void operator=(const Del& rDel) = something_member;
};
[yamato@x201]~/var/ctags-github% ./ctags -o - --c++-kinds=+p /tmp/foo.h
Del	/tmp/foo.h	/^class Del$/;"	c
rDel	/tmp/foo.h	/^    void operator=(const Del& rDel) = something_member;$/;"	m	class:Del
[yamato@x201]~/var/ctags-github% git checkout vmiklos-cxx11-delete 
Switched to branch 'vmiklos-cxx11-delete'
[yamato@x201]~/var/ctags-github% make                                  
dir="$(dirname parsers/c.o)"; [ -d "$dir" ] || /usr/bin/mkdir -p "$dir"
gcc -I. -I. -I./main -DHAVE_CONFIG_H  -DDEBUG -DDATADIR=\"/home/yamato/var/ctags-github/tinst-root/share/ctags\" -DPKGCONFDIR=\"/home/yamato/var/ctags-github/tinst-root/etc/ctags\" -DPKGLIBEXECDIR=\"/home/yamato/var/ctags-github/tinst-root/libexec/ctags\" -g -O2 -Wall  -o parsers/c.o -c parsers/c.c
gcc  -o ctags main/args.o main/entry.o main/flags.o main/get.o main/htable.o main/keyword.o main/lregex.o main/lxcmd.o main/main.o main/mbcs.o main/nestlevel.o main/options.o main/parse.o main/pcoproc.o main/read.o main/routines.o main/sort.o main/strlist.o main/tg.o main/trashbox.o main/vstring.o parsers/ada.o parsers/ant.o parsers/asm.o parsers/asp.o parsers/awk.o parsers/basic.o parsers/beta.o parsers/c.o parsers/clojure.o parsers/css.o parsers/cobol.o parsers/dosbatch.o parsers/eiffel.o parsers/erlang.o parsers/falcon.o parsers/flex.o parsers/fortran.o parsers/go.o parsers/html.o parsers/jscript.o parsers/json.o parsers/lisp.o parsers/lua.o parsers/make.o parsers/matlab.o parsers/objc.o parsers/ocaml.o parsers/pascal.o parsers/perl.o parsers/php.o parsers/python.o parsers/rexx.o parsers/rst.o parsers/ruby.o parsers/rust.o parsers/scheme.o parsers/sh.o parsers/slang.o parsers/sml.o parsers/sql.o parsers/tcl.o parsers/tex.o parsers/verilog.o parsers/vhdl.o parsers/vim.o parsers/windres.o parsers/yacc.o main/debug.o  
[yamato@x201]~/var/ctags-github% ./ctags -o - --c++-kinds=+p /tmp/foo.h
Del	/tmp/foo.h	/^class Del$/;"	c
operator =	/tmp/foo.h	/^    void operator=(const Del& rDel) = something_member;$/;"	p	class:Del
```
You patch is meaningfull even if 'delete' is not used. It means the title(subject?) of this PR can be changed to something more better one. Am I wrong?

vmiklos:
1. It's a common idiom to mark "non-copyable objects" as such (both the copy constructor and the assignment operator having = delete), but it's not a strict requirement, it's still valid to use = delete only for one of them.

2) Yes, "delete" is important, it expresses that the ctor/operator is intentionally only declared, not defined. Without this, you would get an error message only from the linker, but with this, the compiler already can fail the build -> more user-friendly.

The other special string that's allowed in this context is "default" (to force the generation of compiler-default copy ctor/assignment operator, even in cases where they would not be generated by default).

More on this e.g. here: <http://en.cppreference.com/w/cpp/language/default_constructor>.

For your something_member example, a C++ compiler won't accept it. :-)

masatake:
LGTM.

@b4n, could you look at this?
If you are o.k., please, merge this.

b4n:
I'd need to investigate it a bit further, because for some reason Geany's C++ parser is not affected by the issue, yet has virtually the same `analyzeParens()` code than *without* this PR.
I'll try to look at this more seriously during the week.

BTW, a C++ guy actually suggested those maybe should be *removed* from the output (or perhaps tagged differently), see https://github.com/geany/geany/issues/545 (4th item).
But this is a slightly different topic as currently u-ctags breaks on such declarations (I tested and it generated only 2 tags, `Del` (correct) and `rDel` (utter crap)), so probably parsing it correctly is the first step anyway.

vmiklos:
Hi,

I agree with @b4n , that `rDel` is really unexpected. :-) Is there anything I should do or could please one of the maintainers merge this PR?

Thanks,

Miklos

vmiklos:
Ping, could you please let me know if any action is needed from my side?

masatake:
@vmiklos, my understanding is that you don't have to take any action now. Please, wait with the response from @b4n.

b4n:
I'm sorry I won't be able to work on ctags until the end of next week.

So while I don't like diverging further from Geany's c.c, I won't hold this back further if you want to merge it, as the fix seems to mostly make sense indeed.  Otherwise I'll try to dig into this the week after.

masatake:
@b4n, thank you for replying. As I wrote in #459, synchronizing the parser code with geany is quite important.
We will wait for you.

masatake:
@b4n, any progress?
If you don't have time, I will merge this in this Friday.

b4n:
I should be able to look at this before Friday -- otherwise anyway better merge it than wait ages for me ;)

masatake:
o.k. I will merge on this Saturday...after this, please look at #459 . We need your lead.

masatake:
Saturday at Japan.

masatake:
@vmiklos, thank you for contributing.

b4n:
Friday here :)
But I just given up like 15s ago and was about to tell you to merge it anyway: the diff between CTags and Geany's *c.c* is too large and subtle and I can't seem to see which part would affect this behavior.  So for the moment, patch here LGTM.

masatake:
@b4n, than you for replying.
Could you look at #459?
Now I understand there are big gap between C parser.

So I worked on diff and lua. They are synchronized well.
Can we share the parser code between two projects?
I hope sharing the code makes your load lower.

masatake:
...about c parser,  merging test cases is very important.

PR #432, opened 2015-07-11
fix objc property error

SolaWing:
there have two bugs: 
first, in `ObjcKinds`, `Protocol` and `property` type both used token `p`, this may cause some conflict. 
I changed the protocol to upper case `P`.

second, in `parseProperty` function, it never set next operation, so if entering parsing property, it won't stop parsing property,  and after other type will parse fail.
here I fix it and set next to the `parseMethods` func, where parseProperty be used from.

### DIFF
--- a/Units/parser-objectivec.r/objectivec_property.h.d/expected.tags
+++ b/Units/parser-objectivec.r/objectivec_property.h.d/expected.tags
@@ -1,5 +1,5 @@
 Person	input.h	/^@interface Person : NSObject {$/;"	i	interface:
-age	input.h	/^-(id)initWithAge:(int)age;$/;"	p	interface:Person
+initWithAge:	input.h	/^-(id)initWithAge:(int)age;$/;"	m	interface:Person
 m_age	input.h	/^        int m_age;$/;"	F	interface:Person
 m_name	input.h	/^        NSString *m_name;$/;"	F	interface:Person
 personAge	input.h	/^@property(readonly) int personAge;$/;"	p	interface:Person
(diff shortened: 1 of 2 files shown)


masatake:
Could you split your PR into two?

I would like to merge p/P patch first but I would like to the result of travis.

SolaWing:
I add the unit test for the use of property, Protocol. and fix the indent typo

masatake:
Could you split your PR into two?

I would like to merge p/P patch first.

vhda:
Does it make sense to keep the "fix" part of the comment?

b4n:
> Does it make sense to keep the "fix" part of the comment?

Nope, agreed "fix" should be remove from the comment.

Also the existing tests need to be fixed:

* *objectivec_protocol.h.d* needs to use the new `P` as the expected kind for the protocol tag, and
* *objectivec_property.h.d* should be updated to reflect the fix -- AFAIU with my non-existing Objective-C knowledge, the property parse fix does fix this test case, as `initWithAge` (method) seem more like the correct tag than `age` (property).
So actually there already is a test case for the change :)

When fixed and test pass again, it'll probably be good as the code changes LGTM.

SolaWing:
thanks for reply. `objectivec_property.h.d` itself is enough as a test case. I rebase to newest master and fix corresponding unit test.

by the way, the split PR is here #435

masatake:
O.K. Now you can drop 43330bf after rebasing to master.

masatake:
Thank you.

PR #435, opened 2015-07-12
fix property Protocol type error

SolaWing:
here is the first property and protocol conflict bug.
later I may write some unittest for the second, parse property bug.


masatake:
Thank you.
However, this doesn't pass the test under travis.

Could you look at failed test cases under Units directory?
You may have to fix expected.tags files.

b4n:
As @masatake suggested, the expected result need to be fixed to use the new kind letter for protocols (e.g. replace `p` with `P`)

masatake:
Fine. Thank you.

PR #443, opened 2015-07-15
Run tmain under travis (2days)

masatake:
The output of tmain target is not so polished but the target itself is helpful.


masatake:
Surprisingly, tmain is passed on travis without any changes!

masatake:
(2 days) on the title means I will keep open this 2 days to get comments.
If you are interested in this PR but you don't have time now, extend the days with your name like
(masatake: 7days).

The organization member can set this.

PR #445, opened 2015-07-16
Introduce {optional} long flag for --langdef(5 days)

masatake:
You can define a parser but disabled by default with this ``optional``
long flag. Following is an example to disable ``foo`` parser
in definition::

	$ ctags --langdef=foo'{optional}'

This is helpful to add a parser for rather minor language.  This
option is useful in the case you want to add a parser which large parts
of users may will not use. You can enable it with putting
``--languages=+foo`` or ``--languages=all``; and can put it
to your .ctags.


b4n:
Apparently it reuses something that already exists for the option format, but otherwise I would have suggested rather providing a different option, like `--langdef-optional=foo` or something, that I'd think be a more clean command-line API.
But I don't mind much, it's just a though from the top of my head.

masatake:
Thank you for the comment. This is the first time to get comment about the long flag{} syntax.
After getting your comment and I have thought  about this syntax again. I will withdraw this.

I introduced the syntax for --regex-<LANG> option. e.g.

```
--regex-m4=/^[[:space:]]*(m4_)?s?include\([[`]?([^]')]+)[]']?\)/\2/I,inclusion/{exclusive}{optional}
```

How do you think about this?

PR #451, opened 2015-07-22
Add an updated ocaml parser

toroidal-code:
This was a complete refactor of the ocaml parser that I was inspired to do by the horrible results the original gave for usage by emacs' ECB code browser. There are still a couple edge cases, but in general this is much more reliable and properly handles the majority of important tags.

This was originally on my own personal ctags repo, but after finding this, I thought it would be best to submit it for review and merging into the actual project.


toroidal-code:
...It looks like the tests may need to be updated as well.

toroidal-code:
Tests have been updated.

vhda:
Could I ask you to revert the changes that are only style related?
Examples:

* Use tabs in left side indent.
* Open brackets in a single line.
* Keep a space between function name and parentheses.
* Undo multi-line to one-line function declarations.
* Remove any existing extra white spaces at the end of the line.

That will not only make the code more consistent across the project, but will also make it easier for us to identify and analyze the real code changes.

toroidal-code:
This is really the best I can do with an auto-formatter, as I never actually kept the old formatting around, not having planned on ever releasing the code. I included the `.clang-formatter` simply because you might be able to shed some light on what could be improved.

.clang-format
```diff
@@ -0,0 +1,49 @@
+---
```
masatake: This is nothing to do with ocaml parser.
This one can be useful for all C files in ctags.
However, we have .indent.pro.

Units/ocaml_empty.ml_etags.t/expected.tags-e
```diff
@@ -1,2 +0,0 @@
-
-input.ml,0
```
masatake: Why this should be dropped?

Units/ocaml_empty.ml_xref.t/expected.tags-x
```diff
@@ -1 +0,0 @@
-Input            module        1 input.ml
```
masatake: Why this should be dropped?

parsers/ocaml.c
```diff
@@ -25,39 +25,43 @@
...
-	{TRUE, 'e', "Exception", "An exception"}
+	{ TRUE, 'c', "class", "classes" },
+	{ TRUE, 'm', "method", "Object's method" },
+	{ TRUE, 'M', "module", "Module or functor" },
+	{ TRUE, 'v', "variable", "Global variable" },
+	{ TRUE, 'p', "val", "Signature item" },
+	{ TRUE, 't', "type", "Type name" },
+	{ TRUE, 'f', "function", "A function" },
+	{ TRUE, 'C', "constructor", "A constructor" },
+	{ TRUE, 'r', "record_field", "A 'structure' field" },
```
masatake: Interesting. We have to think about this.

parsers/ocaml.c
```diff
@@ -91,102 +96,108 @@ typedef enum {
...
+	{ "->", Tok_To },
+	{ ":", Tok_Of },
+	{ "true", Tok_Val },
+	{ "false", Tok_Val }
 };
 
 static langType Lang_Ocaml;
 
-static boolean exportLocalInfo = FALSE;
+boolean exportLocalInfo = FALSE;
```
masatake: Why this should be a global variable?

parsers/ocaml.c
```diff
@@ -594,13 +629,15 @@ static const char *contextDescription (contextType t)
 	case ContextValue:
 		return "value";
 	case ContextModule:
-		return "Module";
+		return "module";
```
masatake: Why this is downcase'ed?

parsers/ocaml.c
```diff
@@ -1899,17 +2180,13 @@ static void ocamlInitialize (const langType language)
 	initKeywordHash ();
 }
 
-extern parserDefinition *OcamlParser (void)
+extern parserDefinition * OcamlParser (void)
 {
-	static const char *const extensions[] = { "ml", "mli", "aug", NULL };
```
masatake: aus should not be rmeoved.

parsers/ocaml.c
```diff
@@ -1899,17 +2180,13 @@ static void ocamlInitialize (const langType language)
...
-	static const char *const aliases[] = { "tuareg", /* mode name of emacs */
```
masatake: Don't remove aliases.

masatake:
First of all, thank you for contacting us. I agree this is the place for submitting your private patch.

However, we cannot merge your patches into our tree.
The patch changes too many things at once.
It is difficult to review. I want you to split the patch into smaller ones.
In some case we want to split the PR itself.
We also need test cases.

We also take care about compatibility. You changed "Record field" to "record_field".
This breaks the compatibility of output. Is this needed?
ECB code browse needs this?

Units/ocaml_empty.ml_etags.t/expected.tags-e
toroidal-code: This is done to match behavior to other parsers, such as the C parser, which gives an empty file when fed an empty file as input.

parsers/ocaml.c
toroidal-code: Ecb has a number of issues dealing with uppercase names and especially with tags that have spaces in them, that was the main reason for altering those tags. I also wanted to match closer to the c/c++/java tags, as it reduced the need to implement custom tag interpretation in ecb (which is significantly more messy to deal with).

Units/ocaml_empty.ml_xref.t/expected.tags-x
toroidal-code: I removed top-level module annotations because if ecb sees the module, it immediately discards all other tags as they aren't considered "top-level". I don't think a top-level module should be apparent in the tags anyways, as it's not explicitly declared in the file, but instead an inherent aspect of how the OCaml compiler interprets filenames.

.clang-format
toroidal-code: Ah, excellent. Didn't see that.

parsers/ocaml.c
toroidal-code: Good question. I have no good answer, and changing it back doesn't break anything, so I've reverted this change.

toroidal-code:
I have no problem splitting into chunks, however I don't really know _how_, as I don't know what functional aspects could be split off other than the introduction of the lookahead token.

What kind of testcases would you like? I updated the ones that are already present, so are there any more than you would like introduced?

vhda:
For what it takes when I diff'ed the changes ignoring white spaces it didn't seem that many changes. Tonight I'll try to refactor the changes that I will share it with the PR creator.

Units/ocaml_empty.ml_xref.t/expected.tags-x
masatake: Just a question; ECB refers tags file, TAGS file or -x output?
(I don't want to change tags output.)

masatake:
@vhda, thank you. How do you think about merging whitespace related changes first?
As you wrote before we need a guide line about changes about whitespaces for potential contributors.
This PR can be a good testbed for the guide line.

There is a good news. ocaml parser is not in geany. So we can change ocaml parser as we want.
(About the other parsers which are in geany, we should not introduce whitespace related changes before sharing the parser codes with geany. If introducing, the tasks for sharing become complicated.)

masatake:
It seems that we can split this commit into 3 categories.

1. indent/whitespaces/style related changes
    They don't change the output.
2. changes related to ecb
3. the else

About 2, we have to deal with the changes carefully.
ecb may be good and important software. However, u-ctags is not only for ecb.
Not a few tool depends on tags output. So I don't want to change. We need to know what ecb expects/requires to u-ctags and its output. The expectations and requirements will be evaluated.
I guess they will not be ocaml specific. If the expectations and requirements are acceptable we 
want to fix all parers, not only ocaml. So I want @toroidal-code to open an separated issue.
@vhda may give more deatiled comment. 

Our interest is 3. This will be interesting. You just say "Add an updated ocaml parser".
But I wonder what you really have done.

toroidal-code:
What kind of style changes are we discussing? I'm confused about this topic, as I was under the impression that the current style of the commit (which matches the original file) was the one desired.

Details for the commit are in the expansion of the description, it's actually pretty long, as I tried to be as verbose as possible. You can view it by clicking the ellipses next to the commit name in this thread.

ECB isn't the actual system that deals with ctags, that responsibility is delegated to [semantic](https://github.com/toroidal-code/cedet/tree/master/lisp/cedet/semantic) to create the necessary emacs tag-tree. Semantic's ctags parser already has support for the majority of ctags' output tags as can be seen [here](https://github.com/toroidal-code/cedet/blob/master/lisp/cedet/semantic/ectags/parse.el#L131), so I don't think any changes are really needed on this side, in fact OCaml is the only language I've noticed problems with. 

Semantic, among other things, includes a full-blown bison-compatible parser generator, as well as a relatively complete C and Java parser. Ctags is only used as a last resort/fallback tag-generator if emacs doesn't have a built-in parser for the language being analyzed, such as bash, ASM, BASIC, FORTRAN, Pascal, Perl, Python, TCL, and (obviously) OCaml.

toroidal-code:
@vhda could you please use the inline notes as masatake did? I can't respond on a per-comment basis the way you are currently commenting.

vhda:
Sorry, I did not see your updated commit until now. It is indeed much more focused on the introduced changes.
I've added some comments to the commit, but the input from @masatake and @b4n is much more valuable than mine ;)

Thanks a lot for helping us in improving the OCaml parser and for the extra effort you had in preparing the PR for review.

vhda:
But that was what I did!
I'll try to understand what happened tomorrow morning, ok?

masatake:
(About sematic parser's ctags parser)

Thank you for the explanation about the ECB. I remember both ECB and semantic.

> so I don't think any changes are really needed on this side, in fact OCaml is the only language I've noticed problems with. 

Could you enumerate the problems in newly opened issue?
Could you remove the fixes for the problem from this PR?

masatake:
> What kind of style changes are we discussing? I'm confused about this topic, as I was under the impression that the current style of the commit (which matches the original file) was the one desired.

Coding style about white spaces, newline, redundant {} and so on. They don't change the output.

masatake:
If you refers a ticket in sf-project, could you use prefix like sf-bugs:92382, sf-patchs:39204, or something like I used in docs/tracking.rst. If you use ``#``, github web engine refers issue of this project(universal ctags)itself. It makes confusion.

toroidal-code:
@vhda Huh. Maybe you commented directly on the commit instead of on the PR? There's a tab at the top of this page labeled "Files Changed". Adding a line note on that diff instead of the actual commit shouldn't add a new post to the thread. ... I think.

@masatake It's fine, I'll work on the update, squash it down, and then push it to this thread.

Units/ocaml_empty.ml_xref.t/expected.tags-x
toroidal-code: Semantic uses the direct "tags" format from stdout. It builds commands and executes them in a subshell. The commands look something like `/usr/bin/ctags --language-force=ocaml --sort=no --excmd=number --fields=aKStsim --ocaml-kinds=cmMvptfCre -f - ~/code/ocaml/test.ml`

vhda:
@toroidal-code my bad, sorry!
But were you able to at least see them all?

parsers/ocaml.c
```diff
@@ -176,7 +182,7 @@ typedef struct _lexingState {
 } lexingState;
 
 /* array of the size of all possible value for a char */
-static boolean isOperator[1 << (8 * sizeof (char))] = { FALSE };
+boolean isOperator[1 << (8 * sizeof (char))] = { FALSE };
```
b4n: should stay static

parsers/ocaml.c
```diff
@@ -732,7 +757,6 @@ static void popSoftContext ( void )
 static contextType popStrongContext ( void )
 {
 	int i;
-
```
b4n: style, should probably be kept (but I don't mind myself)

parsers/ocaml.c
```diff
@@ -749,10 +773,29 @@ static contextType popStrongContext ( void )
...
+static void jumpToMatchContext ( void )
+{
+	int i;
+	for (i = stackIndex; i >= 0; --i)
+	{
+		if (stack[i].type == ContextMatch)
+		{
+			stackIndex = i + 1;
+			toDoNext = stack[i].callback;	// this should always be
+			// matchPattern
```
b4n: comment would better be aligned (or on one single line before)

parsers/ocaml.c
```diff
@@ -873,23 +920,33 @@ static void killCurrentState ( void )
...
+	return popped;
 }
 
-/* used to prepare tag for OCaml, just in case their is a need to
+/* Keep track of our _true_ line number and file pos,
+ * as the lookahead token gives us false values. */
+static unsigned long ocaLineNumber;
+static fpos_t ocaFilePosition;
+
+/* used to prepare an OCaml tag , just in case their is a need to
```
b4n: while at it changing this comment, let's fix it: s/their/there/, remove space before the comma

parsers/ocaml.c
```diff
@@ -873,23 +920,33 @@ static void killCurrentState ( void )
...
  * add additional information to the tag. */
 static void prepareTag (tagEntryInfo * tag, vString const *name, ocamlKind kind)
 {
 	int parentIndex;
 
-	initTagEntry (tag, vStringValue (name));
+	/* Ripped out of read.h initTagEntry, because of line number
+	 * shenanigans.
```
b4n: I'd rather keep the `initTagEntry()` call and only override the required fields,  or use the new `initTagEntryFull()`.
This can help if we need to add new fields in the future.

parsers/ocaml.c
```diff
@@ -912,7 +969,7 @@ static void addTag (vString * const ident, int kind)
 	}
 }
 
-static boolean needStrongPoping = FALSE;
+boolean needStrongPoping = FALSE;
```
b4n: should remain static

parsers/ocaml.c
```diff
@@ -970,25 +1033,27 @@ static void typeRecord (vString * const ident, ocaToken what)
...
+    {
+        globalScope (ident, what, whatNext);
     }
 	toDoNext = &globalScope;
 }
 
-static tagEntryInfo tempTag;
-static vString *tempIdent;
+tagEntryInfo tempTag;
+vString *tempIdent;
```
b4n: ditto

parsers/ocaml.c
```diff
@@ -1060,15 +1126,54 @@ static void typeDecl (vString * const ident, ocaToken what)
...
+		 * function definition */
+		addTag (ident, K_VAL);
+		toDoNext = &globalScope;
+		break;
+
+	case Tok_Val:	/* Can be a weiiird binding, or an '_' */
+	case OcaIDENTIFIER:
+		addTag (ident, K_VAL);
+		toDoNext = &globalScope;	// sig parser ?
+		break;
```
b4n: this case is the very same as above, maybe it would make sense to merge them?  Or maybe not, if it's only a mere luck the code is the same, just a suggestion.

parsers/ocaml.c
```diff
@@ -1093,7 +1199,9 @@ static void typeSpecification (vString * const ident, ocaToken what)
 			 * AnotherModule can mistakenly be took
 			 * for a constructor. */
 			if (! OcamlKinds[K_CONSTRUCTOR].enabled)
+			{
 				vStringClear (tempIdent);
+			}
```
b4n: doesn't matter to me, but adding curly braces is not required

parsers/ocaml.c
```diff
@@ -1484,8 +1608,7 @@ static void methodDecl (vString * const ident, ocaToken what)
 
 /* name of the last module, used for
  * context stacking. */
-static vString *lastModule;
-
+vString *lastModule;
```
b4n: should remain static

parsers/ocaml.c
```diff
@@ -1509,38 +1632,57 @@ static void moduleSpecif (vString * const ident, ocaToken what)
...
 		toDoNext = &globalScope;
+		break;
 	}
 }
 
 /* parse :
  * module name = ...
  * then pass the token stream to moduleSpecif */
-static void moduleDecl (vString * const ident, ocaToken what)
+
```
b4n: I don't think adding an empty line between the comment and function is a good idea

parsers/ocaml.c
```diff
@@ -1509,38 +1632,57 @@ static void moduleSpecif (vString * const ident, ocaToken what)
...
+static void moduleDecl (vString * const ident, ocaToken what, ocaToken whatNext)
 {
 	switch (what)
 	{
+	case OcaKEYWORD_rec:
+		/* recursive modules are _weird_, but they happen */
 	case OcaKEYWORD_type:
-		/* just ignore it, name come after */
+		/* this is technically a special type, but whatever */
 		break;
```
b4n: why is there special (empty) handling for those while the default case is also to do nothing?

parsers/ocaml.c
toroidal-code: The reason I didn't merge it is because I was thinking about the possibility of extending the parser to be able to handle the special signature syntax that interfaces use. I'd prefer to keep it separate so that extension is easier to consider, despite being currently repetitive.

parsers/ocaml.c
```diff
@@ -1587,7 +1730,16 @@ static void globalLet (vString * const ident, ocaToken what)
 		/* We ignore this token to be able to parse such
 		 * declarations :
 		 * let (ident : type) = ...
-		 */
+		 * but () is the toplevel function name, so fake ourselves
+		 * as an ident and make a new function */
+		if (whatNext == Tok_PARR)
+		{
+			vString *fakeIdent = vStringNewInit ("()");
```
b4n: wat?  ocaml is weird :)

parsers/ocaml.c
```diff
@@ -1655,10 +1829,25 @@ static void globalScope (vString * const UNUSED (ident), ocaToken what)
...
-	case OcaKEYWORD_end:
+	case OcaKEYWORD_end:;
+		contextType popped = killCurrentState ();
+
+		/** so here, end can legally be followed by = or and in the
+		 * situation of
+		  * module M : sig ... end = struct ... end  and
+		  * module M struct ... end and N = struct ... end
+		  * and we need to make sure we know we're still inside of a
+		 * struct */
```
b4n: weird mixed comment indent

parsers/ocaml.c
```diff
@@ -1718,8 +1913,12 @@ static void localScope (vString * const ident, ocaToken what)
 		toDoNext = &mayRedeclare;
 		break;
 
+	/** An in keyword signals the end of the previous context and the
+      * start of a new one. */
```
b4n: weird indent

parsers/ocaml.c
toroidal-code: It really is, haha

parsers/ocaml.c
```diff
@@ -1730,12 +1929,22 @@ static void localScope (vString * const ident, ocaToken what)
...
 		waitedToken = Tok_CurlR;
 		comeAfter = &localScope;
-		contextualTillToken (ident, what);
+		contextualTillToken (ident, what, whatNext);
 		break;
 
 		/* Yeah imperative feature of OCaml,
 		 * a ';' like in C */
 	case Tok_semi:
+		/* ';;' case should end all scopes */
```
b4n: if `;;` is a special token (and not just `;` followed by a second `;` -- e.g. that `;;` and `; ;` aren't the same) it should probably rather be recognized as such?  (e.g. emit something like `Tok_souble_semi` or something)

toroidal-code: Probably? I don't actually remember why I didn't do that. The OCaml lexer treats `;;` as a special token, so it would make sense to do the same here.

parsers/ocaml.c
```diff
@@ -1809,9 +2024,20 @@ static void localScope (vString * const ident, ocaToken what)
...
 	const char *filename = getSourceFileName ();
+
+	FILE *file = fopen (filename, "r");
+	fseek (file, 0, SEEK_END);
+	long fileSize = ftell (file);
+	fclose (file);	// CLOSE!
+	if (fileSize == 0)
+	{
+		return;
+	}
```
b4n: please, do not play with IO like this.  Not only it adds some (probably unnecessary) overhead, but it also complicates possible future move to a non-file-backed IO layer.  So please use another way of determining if the file is empty.

BTW, it can even be better, as this only finds empty files if they are truly empty, but not if they contain e.g. a newline, or some whitespace, or a comment, etc.
What about rather checking if some tags were emitted, or if it's not good enough, if a token was generated?  And emit the module tag at the end of the processing (which should probably be alright, as I don't think CTags has any promise on the tag ordering -- does it?

parsers/ocaml.c
toroidal-code: I think I tried to do that, but I had some serious compiler errors from what I remember. I think it was specifically related to the linking stage. Don't really recall. But I do know I tried using `initTagEntry()` first. I'll have to take a look at `initTagEntryFull()`.

parsers/ocaml.c
```diff
@@ -1858,31 +2084,70 @@ static void clearStack ( void )
...
 
+#pragma clang diagnostic ignored "-Wattributes"
+/* clang's -licm (Loop Invariant Code Motion) optimization pass actually
+ * alters the main while loop. This has the (inadvertant and incorrect)
+ * side-effect of moving the line/filePos update statements _outside_ the loop,
+ * because it never sees them get updated inside of it ('File' is in another
+ * object file). On GCC and versions of clang bundled with LLVM 3.5 and up,
+ * the 'optnone' attribute is respected, and everything works predictably.
+ * ... I'm still going to email the clang dev team about this. */
+__attribute__ ((optnone))
```
b4n: This should be removed, has been fixed in 585d60d5a4ea4c6f2e96c9cd65dae87fe52d25c2 (and I already tried to fix it in exuberant-ctags a long while back, see https://sourceforge.net/p/ctags/mailman/message/30813660/ for patch and details).

But anyway, before this change the compiler was actually is totally right optimizing this out.  As the `File` global was `const`, it was totally valid to assume its value never changed, leading to the behavior you experienced.  This is NOT the optimizer's fault, but our (previous) code's which used a plain hack to avoid altering the content of the `File` global.

parsers/ocaml.c
toroidal-code: ctags can emit an ordered tag list, which is what emacs/semantic uses, so unfortunately I don't think that's viable. Token-presence checking looks like a good possibility, and I'm a little annoyed that I didn't consider that as C-style File I/O is really not my forte.

parsers/ocaml.c
toroidal-code: Oh! Excellent! I'll get rid of that then.

parsers/ocaml.c
```diff
@@ -1902,14 +2167,10 @@ static void ocamlInitialize (const langType language)
 extern parserDefinition *OcamlParser (void)
 {
 	static const char *const extensions[] = { "ml", "mli", "aug", NULL };
-	static const char *const aliases[] = { "tuareg", /* mode name of emacs */
-					       "caml",	 /* mode name of emacs */
-					       NULL };
```
b4n: should probably not be removed

parsers/ocaml.c
```diff
@@ -1902,14 +2167,10 @@ static void ocamlInitialize (const langType language)
...
 	parserDefinition *def = parserNew ("OCaml");
 	def->kinds = OcamlKinds;
 	def->kindCount = KIND_COUNT (OcamlKinds);
 	def->extensions = extensions;
-	def->aliases = aliases;
```
b4n: neither should this

parsers/ocaml.c
vhda: What about creating the module tag together with the first tag found?

parsers/ocaml.c
```diff
@@ -119,48 +123,50 @@ typedef struct sOcaKeywordDesc {
...
+	{ "mod "      , Tok_Op               },
+	{ "land "     , Tok_Op               },
+	{ "lor "      , Tok_Op               },
+	{ "lxor "     , Tok_Op               },
+	{ "lsl "      , Tok_Op               },
+	{ "lsr "      , Tok_Op               },
+	{ "asr"       , Tok_Op               },
+	{ "->"        , Tok_To               },
+	{ ":"         , Tok_Of               },
+	{ "true"      , Tok_Val              },
```
vhda: I hate these end-of-line spaces, but I still would prefer to have a valid `git blame` available even if that means keeping the extra spaces. But that's me... @b4n ?

parsers/ocaml.c
toroidal-code: because a `module type` gives the type signature for the module, and could possibly used to build a more complete type annotation for the functions contained in the module itself, as types aren't explicitly annotated and usually inferred in OCaml, meaning ctags has _no_ idea what the types of anything are.

parsers/ocaml.c
toroidal-code: It looks like that's the current behavior. ...weird. I was certain I had changed that. But I guess not. I'll update my other comment to reflect this.

parsers/ocaml.c
b4n: okay, if it's better for future handling, it's fine by me

parsers/ocaml.c
b4n: as `initTagEntry()` was used before I doubt it would be a problem…

parsers/ocaml.c
b4n: > What about creating the module tag together with the first tag found?

Sounds fair, though it might then mean the source position would not be (by default) on the first line/pos.  But that'd be a matter of recording the source position at the start of the file and use it later.

b4n: Though, maybe a translation unit might not be considered empty even if it doesn't generate tags?  Hence why I suggested when finding the first non-empty token

parsers/ocaml.c
```diff
@@ -1864,25 +2067,55 @@ static void findOcamlTags (void)
...
+	nextSt.name = vStringNew ();
+	nextSt.cp = fileReadLine ();
 	toDoNext = &globalScope;
-	tok = lex (&st);
+	nextTok = lex (&nextSt);
+
+	/* prime the lookahead token */
+	st = nextSt;	// preserve the old state for our first token
+	st.name = vStringNewCopy (st.name);
+	st.cp = (const unsigned char *) temp_cp->buffer;
```
b4n: shouldn't the buffer be copied from `nextSt.cp` first like [below](#diff-9ee068438f29821d9f8a0cea175b87acR2110)?

parsers/ocaml.c
b4n: I totally agree, especially as it also did hide a bug: `OcaKEYWORD_function` is *never* mapped here, so no token using it are ever created, and then all code paths using it area effectively dead.  Bot "fun" and "function" are mapped to `OcaKEYWORD_fun`.

Also, if ignoring whitespace changes, we see what actually was added: `"if", OcaKEYWORD_if` and `":", Tok_Of`

b4n: Ideally the `diff -w` and `diff` should touch the same lines (well, you get the point).

Also as much as possible I guess not adding unnecessary changes (like adding or removing optional `{}`s) would be optimal for diff beauty/review ease, but it matters a lot less.

b4n: I'd suggest applying this to undo unnecessary changes (I kept e.g. indentation fixes when other code was touched right around and it would have looked funny):
```diff
diff --git a/parsers/ocaml.c b/parsers/ocaml.c
index ee4d8fc..2d9f191 100644
--- a/parsers/ocaml.c
+++ b/parsers/ocaml.c
@@ -123,50 +123,50 @@ typedef struct sOcaKeywordDesc {
 typedef ocamlKeyword ocaToken;
 
 static const ocaKeywordDesc OcamlKeywordTable[] = {
-	{ "and"       , OcaKEYWORD_and       },
-	{ "begin"     , OcaKEYWORD_begin     },
-	{ "class"     , OcaKEYWORD_class     },
-	{ "do"        , OcaKEYWORD_do        },
-	{ "done"      , OcaKEYWORD_done      },
-	{ "else"      , OcaKEYWORD_else      },
-	{ "end"       , OcaKEYWORD_end       },
-	{ "exception" , OcaKEYWORD_exception },
-	{ "for"       , OcaKEYWORD_for       },
-	{ "fun"       , OcaKEYWORD_fun       },
-	{ "function"  , OcaKEYWORD_fun       },
-	{ "functor"   , OcaKEYWORD_functor   },
+	{ "and"       , OcaKEYWORD_and       }, 
+	{ "begin"     , OcaKEYWORD_begin     }, 
+	{ "class"     , OcaKEYWORD_class     }, 
+	{ "do"        , OcaKEYWORD_do        }, 
+	{ "done"      , OcaKEYWORD_done      }, 
+	{ "else"      , OcaKEYWORD_else      }, 
+	{ "end"       , OcaKEYWORD_end       }, 
+	{ "exception" , OcaKEYWORD_exception }, 
+	{ "for"       , OcaKEYWORD_for       }, 
+	{ "fun"       , OcaKEYWORD_fun       }, 
+	{ "function"  , OcaKEYWORD_fun       }, 
+	{ "functor"   , OcaKEYWORD_functor   }, 
 	{ "if"        , OcaKEYWORD_if        },
-	{ "in"        , OcaKEYWORD_in        },
-	{ "let"       , OcaKEYWORD_let       },
-	{ "match"     , OcaKEYWORD_match     },
-	{ "method"    , OcaKEYWORD_method    },
-	{ "module"    , OcaKEYWORD_module    },
-	{ "mutable"   , OcaKEYWORD_mutable   },
-	{ "object"    , OcaKEYWORD_object    },
-	{ "of"        , OcaKEYWORD_of        },
-	{ "rec"       , OcaKEYWORD_rec       },
-	{ "sig"       , OcaKEYWORD_sig       },
-	{ "struct"    , OcaKEYWORD_struct    },
-	{ "then"      , OcaKEYWORD_then      },
-	{ "try"       , OcaKEYWORD_try       },
-	{ "type"      , OcaKEYWORD_type      },
-	{ "val"       , OcaKEYWORD_val       },
+	{ "in"        , OcaKEYWORD_in        }, 
+	{ "let"       , OcaKEYWORD_let       }, 
+	{ "match"     , OcaKEYWORD_match     }, 
+	{ "method"    , OcaKEYWORD_method    }, 
+	{ "module"    , OcaKEYWORD_module    }, 
+	{ "mutable"   , OcaKEYWORD_mutable   }, 
+	{ "object"    , OcaKEYWORD_object    }, 
+	{ "of"        , OcaKEYWORD_of        }, 
+	{ "rec"       , OcaKEYWORD_rec       }, 
+	{ "sig"       , OcaKEYWORD_sig       }, 
+	{ "struct"    , OcaKEYWORD_struct    }, 
+	{ "then"      , OcaKEYWORD_then      }, 
+	{ "try"       , OcaKEYWORD_try       }, 
+	{ "type"      , OcaKEYWORD_type      }, 
+	{ "val"       , OcaKEYWORD_val       }, 
 	{ "value"     , OcaKEYWORD_value     }, /* just to handle revised syntax */
-	{ "virtual"   , OcaKEYWORD_virtual   },
-	{ "while"     , OcaKEYWORD_while     },
-	{ "with"      , OcaKEYWORD_with      },
-
-	{ "or"        , Tok_Op               },
-	{ "mod "      , Tok_Op               },
-	{ "land "     , Tok_Op               },
-	{ "lor "      , Tok_Op               },
-	{ "lxor "     , Tok_Op               },
-	{ "lsl "      , Tok_Op               },
-	{ "lsr "      , Tok_Op               },
-	{ "asr"       , Tok_Op               },
-	{ "->"        , Tok_To               },
+	{ "virtual"   , OcaKEYWORD_virtual   }, 
+	{ "while"     , OcaKEYWORD_while     }, 
+	{ "with"      , OcaKEYWORD_with      }, 
+
+	{ "or"        , Tok_Op               }, 
+	{ "mod "      , Tok_Op               }, 
+	{ "land "     , Tok_Op               }, 
+	{ "lor "      , Tok_Op               }, 
+	{ "lxor "     , Tok_Op               }, 
+	{ "lsl "      , Tok_Op               }, 
+	{ "lsr "      , Tok_Op               }, 
+	{ "asr"       , Tok_Op               }, 
+	{ "->"        , Tok_To               }, 
 	{ ":"         , Tok_Of               },
-	{ "true"      , Tok_Val              },
+	{ "true"      , Tok_Val              }, 
 	{ "false"     , Tok_Val              }
 };
 
@@ -435,9 +435,13 @@ static ocamlKeyword lex (lexingState * st)
 		retType = lookupKeyword (vStringValue (st->name), Lang_Ocaml);
 
 		if (retType == -1)	/* If it's not a keyword */
+		{
 			return OcaIDENTIFIER;
+		}
 		else
+		{
 			return retType;
+		}
 	}
 	else if (isNum (*st->cp))
 		return eatNumber (st);
@@ -591,7 +595,9 @@ static int getLastNamedIndex ( void )
 	for (i = stackIndex - 1; i >= 0; --i)
 	{
 		if (vStringLength (stack[i].contextName) > 0)
+		{
 			return i;
+		}
 	}
 
 	return -1;
@@ -728,7 +734,9 @@ static void popLastNamed ( void )
 static void popSoftContext ( void )
 {
 	if (stackIndex <= 0)
+	{
 		toDoNext = &globalScope;
+	}
 	else
 	{
 		stackIndex--;
@@ -1025,9 +1033,13 @@ static void exceptionDecl (vString * const ident, ocaToken what,
 	ocaToken whatNext)
 {
 	if (what == OcaIDENTIFIER)
+	{
 		addTag (ident, K_EXCEPTION);
+	}
 	else /* probably ill-formed, give back to global scope */
+	{
 		globalScope (ident, what, whatNext);
+	}
 
 	toDoNext = &globalScope;
 }
@@ -1502,7 +1514,7 @@ static void letParam (vString * const ident, ocaToken what,
 			break;
 
 			/* parse something like
-			 * ~varname:type
+			 * ~varname:type 
 			 * or
 			 * ~varname
 			 * or
@@ -1513,7 +1525,7 @@ static void letParam (vString * const ident, ocaToken what,
 			break;
 
 			/* Optional argument with syntax like this :
-			 * ?(bla = value)
+			 * ?(bla = value) 
 			 * or
 			 * ?bla */
 		case '?':
@@ -1737,9 +1749,7 @@ static void globalLet (vString * const ident, ocaToken what, ocaToken whatNext)
 
 	case Tok_Val:
 		if (vStringValue (ident)[0] == '_')
-		{
 			addTag (ident, K_FUNCTION);
-		}
 		pushStrongContext (ident, ContextFunction);
 		requestStrongPoping ();
 		toDoNext = &letParam;
@@ -1889,18 +1899,18 @@ static void localScope (vString * const ident, ocaToken what, ocaToken whatNext)
 		toDoNext = &mayRedeclare;
 		break;
 
-	/* An in keyword signals the end of the previous context and the
-	 * start of a new one. */
+		/* An in keyword signals the end of the previous context and the
+		 * start of a new one. */
 	case OcaKEYWORD_in:
 		popLastNamed ();
 		pushEmptyContext (&localScope);
 		toDoNext = &mayRedeclare;
 		break;
 
-	/* Ok, we got a '{', which is much likely to create
-	 * a record. We cannot treat it like other [ && (,
-	 * because it may contain the 'with' keyword and screw
-	 * everything else. */
+		/* Ok, we got a '{', which is much likely to create
+		 * a record. We cannot treat it like other [ && (,
+		 * because it may contain the 'with' keyword and screw
+		 * everything else. */
 	case Tok_CurlL:
 		toDoNext = &contextualTillToken;
 		waitedToken = Tok_CurlR;
@@ -1908,8 +1918,8 @@ static void localScope (vString * const ident, ocaToken what, ocaToken whatNext)
 		contextualTillToken (ident, what, whatNext);
 		break;
 
-	/* Yeah imperative feature of OCaml,
-	 * a ';' like in C */
+		/* Yeah imperative feature of OCaml,
+		 * a ';' like in C */
 	case Tok_semi:
 		/* ';;' case should end all scopes */
 		if (whatNext == Tok_semi)
@@ -1919,8 +1929,8 @@ static void localScope (vString * const ident, ocaToken what, ocaToken whatNext)
 			break;
 		}	/* else fallthrough */
 
-	/* Every standard operator has very high precendence
-	 * e.g. expr * expr needs no parentheses */
+		/* Every standard operator has very high precendence
+		 * e.g. expr * expr needs no parentheses */
 	case Tok_Op:
 		toDoNext = &mayRedeclare;
 		break;
@@ -2011,9 +2021,7 @@ static void computeModuleName ( void )
 	long fileSize = ftell (file);
 	fclose (file);	// CLOSE!
 	if (fileSize == 0)
-	{
 		return;
-	}
 
 	int beginIndex = 0;
 	int endIndex = strlen (filename) - 1;
```

parsers/ocaml.c
b4n: What about this?
```diff
diff --git a/parsers/ocaml.c b/parsers/ocaml.c
index 2d9f191..f321fd1 100644
--- a/parsers/ocaml.c
+++ b/parsers/ocaml.c
@@ -2016,13 +2016,6 @@ static void computeModuleName ( void )
      */
     const char *filename = getSourceFileName ();
 
-    FILE *file = fopen (filename, "r");
-    fseek (file, 0, SEEK_END);
-    long fileSize = ftell (file);
-    fclose (file);  // CLOSE!
-    if (fileSize == 0)
-        return;
-
     int beginIndex = 0;
     int endIndex = strlen (filename) - 1;
     vString *moduleName = vStringNew ();
@@ -2082,7 +2075,6 @@ static void findOcamlTags (void)
 
     initStack ();
 
-    computeModuleName ();
     tempIdent = vStringNew ();
     lastModule = vStringNew ();
     lastClass = vStringNew ();
@@ -2095,6 +2087,9 @@ static void findOcamlTags (void)
     toDoNext = &globalScope;
     nextTok = lex (&nextSt);
 
+    if (nextTok != Tok_EOF)
+        computeModuleName ();
+
     /* prime the lookahead token */
     st = nextSt;    // preserve the old state for our first token
     st.name = vStringNewCopy (st.name);
```

b4n:
@toroidal-code could you please add fixup commits during the review process, and only squash them together in the end?  It would help seeing the actual changes of each update -- yet the overall diff would be updated appropriately.  And in the end, just squash them together so it's clean.  Thanks!

toroidal-code:
@bfn Sure! I'm having trouble applying the diffs because GitHub converted the tabs to spaces. Would you mind uploading the raw diff output using gist?

b4n:
@toroidal-code even better, I added commits on a branch ;) https://github.com/b4n/fishman-ctags/commits/toroidal-code/master%2Breview

And don't worry about authorship when squashing, I don't mind.

b4n:
(it might not be against your latest version though, might require cherry-picking or rebasing)

parsers/ocaml.c
b4n: fair enough

parsers/ocaml.c
toroidal-code: No, if we did that, there'd be no lookahead.  `lex(&nextSt)` advances that cp, and we want `st.cp` to be a step behind it. In the loop below `lex(&nextSt)` follows the assignment, advancing the lookahead's cp without modifying `st`'s.

At least, that's what I remember.

b4n: well, here you copy `nextSt` (which contains the *first* token at this point) to `st`, so I would think you should also copy the `cp` `lex(&nextSt)` above maybe have moved.

Assuming you just want to copy the `nextSt` already read (so you read it again (below) to get the actual next (second) token), I can't imagine you don't want to keep the whole original (first token) `nextSt`, including `cp`.

BTW, IMO it'd be less confusing if you'd first read in `st` and *then* in `nextSt`, instead or reading twice in `nextSt` and copying in `st`:
```diff
diff --git a/parsers/ocaml.c b/parsers/ocaml.c
index f321fd1..a9baed8 100644
--- a/parsers/ocaml.c
+++ b/parsers/ocaml.c
@@ -2082,19 +2082,21 @@ static void findOcamlTags (void)
     vStringCopyS (voidName, "_");
     vString *temp_cp = vStringNew ();
 
-    nextSt.name = vStringNew ();
-    nextSt.cp = fileReadLine ();
+    st.name = vStringNew ();
+    st.cp = fileReadLine ();
     toDoNext = &globalScope;
-    nextTok = lex (&nextSt);
+    tok = lex (&st);
 
-    if (nextTok != Tok_EOF)
+    if (tok != Tok_EOF)
         computeModuleName ();
 
     /* prime the lookahead token */
-    st = nextSt;    // preserve the old state for our first token
-    st.name = vStringNewCopy (st.name);
+    nextSt.name = vStringNewCopy (st.name);
+    /* backup previous cp and swap it, to have the real one (fileReadLine()
+     * result) in the token we actually read */
+    vStringCopyS (temp_cp, (const char *) nextSt.cp);
+    nextSt.cp = st.cp;
     st.cp = (const unsigned char *) temp_cp->buffer;
-    tok = nextTok;
     ocaLineNumber = File.source.lineNumber;
     ocaFilePosition = File.filePosition;
     nextTok = lex (&nextSt);
```
or am I missing something?

parsers/ocaml.c
masatake: Now I understand the need of this change.
This line is needed to make the kind name used in scope information and the kind name in OcamlKinds struct, which is used in --list-kinds, the same.

You updated expected.tags. It is correct.

masatake:
I'm very sorry to be late.  I will merge the changes incrementally.

masatake:
See  #1482.

masatake:
Merged. @toroidal-code, @b4n, thank you and sorry to take too long time till merging.

Issue #455, opened 2015-07-22
Ruby: `class << self` not recognized as block that defines class-level methods


mislav:
This method definition:

```rb
class C
  class << self
    def foo() end
  end
end
```

is not recognized as class-level method definition (`F` kind). Instead, it gets parsed as instance method (`f` kind).

```
$ ctags --version
Universal Ctags Development, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Jul 21 2015, 12:41:18
  Addresses: <dhiebert@users.sourceforge.net>, https://github.com/universal-ctags/ctags
  Optional compiled features: +wildcards, +regex, +debug, +option-directory, +coproc
```

masatake:
(I will convert this to a unit test case. If someone already started to do, let me know here.)

masatake:
@mislav, could you look at #456?

b4n:
@masatake I'm on it, I have a fix for this

masatake:
@b4n, thank you. Please, use the test case I converted from this issue.

PR #456, opened 2015-07-23
units, ruby: new bug case reported in #455

masatake:
This test case represents a bug report of #455.
The input.rb is completely based on a comment by @mislav in #455.


masatake:
Withdraw this. #458 includes this.
Thank you, @mislav and @b4n.

PR #457, opened 2015-07-24
Adding 'ph' suffix to the list of Perl extensions

ymnick:


b4n:
LGTM, but maybe adding a test case for recognition would be best?

masatake:
This is so obvious so I think we don't need a test case.

b4n:
fair enough

PR #461, opened 2015-07-25
Sort --help output

k-takata:
The output of `ctags --help` was not sorted in alphabetical order after
the support of encoding options.


masatake:
Thanks.

PR #462, opened 2015-07-26
Fix build errors and warnings with VC2015 (7 days)

k-takata:
* VC2015 has snprintf.
* VC doesn't have unistd.h.
* __USE_GNU is already defined in makefiles.
* Definition of isLanguageKindEnabled was different from its
  declaration.
* Couldn't link with libiconv. Imported @koron's patch:
  https://github.com/koron/ctags/commit/f09817755ac218425e7e5c265f70b808a835f46b

### DIFF
--- a/main/e_msoft.h
+++ b/main/e_msoft.h
@@ -41,7 +41,9 @@
 
 # define HAVE__FINDFIRST 1
 # define HAVE_DIRECT_H 1
-# define snprintf _snprintf
+# if _MSC_VER < 1900
+#  define snprintf _snprintf
+# endif
 # define findfirst_t intptr_t
 
 # ifndef _CRT_SECURE_NO_DEPRECATE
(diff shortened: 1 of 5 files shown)


masatake:
LGTM.

I would like @ffes, our platform maintainer, to look at the changes.
However, he may be in busy. So I set "7 days" period for reviewing.

ffes:
Just tested with VS2013 and everything still works as expected.

Will test VS2015 ASAP. Have to use installation of colleague for now.

ffes:
Just checked VS2015 as well. Compiles just fine and some basic tests works as expected.

One thing to consider:
The .vcxproj file gets a minor update. All references to v12 of the compiler are updated to v14.
So I think the *_vs2013.* files should get a *_vs2015.* equivalent some day.
But that should not keep us from merging this one.

masatake:
@ffes, thank you. Could you merge this?

Issue #466, opened 2015-07-27
ctags fails to compile in gcc 4.5


vhda:
Compile fails with the following message:

```
gcc -I. -I. -I./main -DHAVE_CONFIG_H  -DDEBUG -DDATADIR=\"/remote/de02h25/vhugo/.local/ctags/share/ctags\" -DPKGCONFDIR=\"/remote/de02h25/vhugo/.local/ctags/etc/ctags\" -DPKGLIBEXECDIR=\"/remote/de02h25/vhugo/.local/ctags/libexec/ctags\" -g -O2 -Wall  -o main/entry.o -c main/entry.c
In file included from main/entry.c:43:0:
main/options.h:136:3: error: redefinition of typedef ‘OptionLoadingStage’
main/parse.h:21:34: note: previous declaration of ‘OptionLoadingStage’ was here
make: *** [main/entry.o] Error 1
```

**Update**
The following warnings also appear during the compile in this version of gcc:

```
gcc -I. -I. -I./main -DHAVE_CONFIG_H  -DDEBUG -DDATADIR=\"/remote/de02h25/vhugo/.local/ctags/share/ctags\" -DPKGCONFDIR=\"/remote/de02h25/vhugo/.local/ctags/etc/ctags\" -DPKGLIBEXECDIR=\"/remote/de02h25/vhugo/.local/ctags/libexec/ctags\" -g -O2 -Wall  -o main/parse.o -c main/parse.c
main/parse.c: In function ‘nominateLanguageCandidates’:
main/parse.c:213:14: warning: ‘spec’ may be used uninitialized in this function

gcc -I. -I. -I./main -DHAVE_CONFIG_H  -DDEBUG -DDATADIR=\"/remote/de02h25/vhugo/.local/ctags/share/ctags\" -DPKGCONFDIR=\"/remote/de02h25/vhugo/.local/ctags/etc/ctags\" -DPKGLIBEXECDIR=\"/remote/de02h25/vhugo/.local/ctags/libexec/ctags\" -g -O2 -Wall  -o main/tg.o -c main/tg.c
main/tg.c: In function ‘tg_nonalnum_get_char’:
main/tg.c:130:3: warning: implicit declaration of function ‘isblank’
```

masatake:
I'm very sorry. I stop my rush.

masatake:
Could you try the head?

vhda:
[...]

PR #467, opened 2015-07-27
main: Fix double inclusion of OptionLoadingStage typedef

vhda:
Fix for #466

### DIFF
--- a/main/options.h
+++ b/main/options.h
@@ -184,6 +184,8 @@ extern vString* expandOnDriversPathList (const char* leaf);
 extern langType getLanguageComponentInOption (const char *const option,
 					      const char *const prefix);
 
+extern boolean processXcmdOption (const char *const option, const char *const parameter, OptionLoadingStage stage);
+
 #endif  /* _OPTIONS_H */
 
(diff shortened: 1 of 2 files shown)


masatake:
As you know already this comes from circular dependencies in header inclusion. 

I could find the best way when I introduced OptionLoadingStage.

How about moving all *Option to options.h?
I cannot say it is the best but I feel moving just processXcmdOption is a bit inconsistent with other process*Option functions.

vhda:
What you say makes sense and it apparently works.
Other than the `/process.*Option/` functions, do you think we should move any other definitions to options.h?

masatake:
I think  /print.*Option/ also should be moved to options.h.

vhda:
From a coherency point of view, should we move anything else?

masatake:
No. I think we should not move anymore.

vhda:
Could you please check now?

masatake:
Fine, thank you.

vhda:
Simple changes, so I won't wait longer for merging :)

PR #468, opened 2015-07-27
beta: convert the encoding to utf-8(1 day)

masatake:
It seems that coverall (or lcoveralls command) expects source code
files are encoded in UTF-8.


k-takata:
That is the restriction of lcoveralls command (not coveralls).

masatake:
I updated the commit log. @k-takata, thank you.

PR #469, opened 2015-07-28
Scheme: handle newline in define/set! line (5 days)

masatake:
The parser assumes a name is within the same line where define/set! appears.
This causes a crash.


masatake:
For S-expression based language like lisp and scheme, we need a way to define specialized parser.

```
--langdef=scheme
--sexp-scheme=define-macro/cadr/m,macro
--sexp-scheme=define-syntax/cadr/s,syntax
````
Kind letter [a-z] should be reserverd in ctags side.
Kind letter [A-Z] is for users.

```
--langdef=gauche{parent=scheme}
--sexp-gauche=define-class/cadr/c,class
--sexp-gauche=define-method/cadr/m,method
```

So ctags can be adapted to your define-xxx.

**Comment by vhda** [comment]
I would avoid the a `--sex*` option ;)

**Comment by masatake** [comment]
http://www.gnu.org/software/emacs/manual/html_node/elisp/List-Motion.html

**Comment by QingmingHe** [comment]
What about **defadvice** in elisp:
```lisp
(defadvice kill-region (before some-name activate)
  ...)
```
you can't use something like cadr to teach ctags to catch **some-name**. But --regex-* is useful in this case.

masatake:
Cadaddr

masatake:
I have no idea to hadle define-values.
```
(define-values (x y) ...
```

masatake:
c*adr

QingmingHe:
you use * to match anything of a list?

masatake:
Yes!
Still do you say regex is better?

QingmingHe:
I think ctags can treat
```lisp
(define-values (x y) ...)
```
automatically.
User use cadr to catch (x y) and ctags find it to be a list, so ctags can tag each item of the list to be a variable or something else.

QingmingHe:
en, seems better

masatake:
We need destructuring here; ctags shoud make two tag entries: one for x, another for y.

QingmingHe:
another case:
```lisp
(unless (symbolp some-symbol)
   (defcustome some-symbol ...
```
Ctags should parse lisp sexp by sexp rather line by line.

masatake:
The same discussion can be done for xml based parser with xpointer.

--langdef=svg
--xpointer-svg=...

masatake:
With sexp based aproach we can deal with lisp families. A difficulty is how to handle the differences in comment syntax. Here we need code prism library. I wrote about it somewhere issue about keyword based parser guessing.

Issue #473, opened 2015-07-30
Implement Perl 6 parser


dtikhonov:

dtikhonov:
Done -- closing.

PR #479, opened 2015-08-02
Implemented a basic parser for Perl 6 - issue #473

dtikhonov:
An interesting change is the introduction of selectLanguage function
pointer to parserDefinition.  This is done to select Perl 5 or Perl
6 parser for files with extension .pm -- Perl 6 uses .pm and .pm6 for
its modules.

### DIFF
--- a/Units/perl6-bunch1.d/expected.tags
+++ b/Units/perl6-bunch1.d/expected.tags
@@ -1,0 +1,51 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Darren Hiebert	/dhiebert@users.sourceforge.net/
+!_TAG_PROGRAM_NAME	Universal Ctags	//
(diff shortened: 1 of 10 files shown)


masatake:
I like "select language" feature. It will work better than my tg.

Could you separate the "select language" code and perl6 parser code to two commits (or two PRs.)?

Could you write "select language"  to docs/guessing.rst?

I would like to use it for `*.m` (ObjectiveC v.s. matlab) and `*.h` (C++ v.s. ObjectiveC).
In this case ObjectiveC needs two selectors; one for fighting matlab and the other for figihting C++.
So a language should can have more than one selectors.

dtikhonov:
@masatake,

> Could you separate the "select language" code and perl6 parser code to two commits (or two PRs.)?

Will do.

> Could you write "select language" to docs/guessing.rst?

Will do.

> I would like to use it for *.m (ObjectiveC v.s. matlab) and *.h (C++ v.s. ObjectiveC).
> In this case ObjectiveC needs two selectors; one for fighting matlab and the other for figihting C++.
> So a language should can have more than one selectors.

I am not sure how I'd implement multiple selectors per language.  Let's do it separately -- I'll just submit what I have for now.

dtikhonov:
Closing the request: I will submit two merge requests: one for the language selector feature and one for the Perl 6 parser.

PR #480, opened 2015-08-02
Selection logic and Perl 6 support.

dtikhonov:
This request is instead of request #479.  As discussed in the other pull request, the following changes were made:
1. selection logic and Perl 6 parser have been separated into two commits.
2. New function type has been renamed to selectLanguage.
3. If selectLanguage does not return a value (it really should, though), we fall back to arbitration.
4. docs/guessing.rst has been updated.
5. Whitespace has been cleaned up.

Once merged, issue #473 should be closed.


masatake:
Could you squash cae5d8e ?

PR #481, opened 2015-08-02
Selection logic and Perl 6 support.

dtikhonov:
This request is instead of request #479. As discussed in the other pull request, the following changes were made:
1. selection logic and Perl 6 parser have been separated into two commits.
2. New function type has been renamed to selectLanguage.
3. If selectLanguage does not return a value (it really should, though), we fall back to arbitration.
4. docs/guessing.rst has been updated.
5. Whitespace has been cleaned up.

Once merged, issue #473 should be closed.

### DIFF
--- a/Units/perl6-bunch1.d/expected.tags
+++ b/Units/perl6-bunch1.d/expected.tags
@@ -1,0 +1,51 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Darren Hiebert	/dhiebert@users.sourceforge.net/
+!_TAG_PROGRAM_NAME	Universal Ctags	//
+!_TAG_PROGRAM_URL	https://github.com/universal-ctags/ctags	/official site/
+!_TAG_PROGRAM_VERSION	Development	//
+JSONPrettyActions	input.pm	/^my class JSONPrettyActions {$/;"	c
+JSONPrettyGrammar	input.pm	/^my grammar JSONPrettyGrammar {$/;"	g
+TOP	input.pm	/^    method TOP($\/) {$/;"	m
+TOP	input.pm	/^    token TOP       { ^ \\s* [ <object> | <array> ] \\s* $ }$/;"	t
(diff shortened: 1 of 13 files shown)


masatake:
I have requested many things because I would like to understand the design of  selectLanguage.

masatake:
Please, merge the commits after replacing LANG_AUTO with LANG_IGNORE.

masatake:
I would like to know the file name of perl6 interpreters  anyway.
That should be listed in parser->aliases. (We can do it after merging.)

dtikhonov:
Please have another look, @masatake.

masatake:
When I run perl6 parser on 32bit machine, I got a crash.
I inspected it in this morning and I found general.h was not included first.
Please, merge the commits after fixing it.

masatake:
LGTM. Please, merge the commits.

dtikhonov:
> LGTM. Please, merge the commits.

Excellent.  Done!

masatake:
selectLanguage is really useful.

Thank you.

BTW, could you tell me the name of perl6 interpreter you are using?
As far as I know  "perl6" is a name of language specification.
I would like to put the name to parser->aliases.

dtikhonov:
I used Rakudo to test my Perl 6 code; the name of the interpreter program is `perl6`.

The funny part is that I don't know Perl 6 -- just enough reading the docs to make a ctags parser for it! :)

masatake:
Oh, I see.

vhda:
Out of curiosity: why add a new parser instead of expanding the available one, like is done in C, C++, etc?

dtikhonov:
> Out of curiosity: why add a new parser instead of expanding the available one, like is done in C, C++, etc?

There are several reasons:

1. While related, Perl (5) and Perl 6 are different languages.
1. The code in perl.c has grown too complex for my liking.  Instead of a) making it more complex or b) refactoring the whole thing, potentially introducing bugs, it is better to get a clean slate.
1. A single Perl 5/6 parser (and I played with this idea and had more or less working implementation) would necessarily have to pick which mode it is in: Perl 5 or Perl 6.  I was afraid that I would not be able to figure out the language correctly each time, thus leaving the user without an option for a manual override.
1. `c` stands for `constant` in Perl 5; in Perl 6, it makes sense to have it stand for `class`.

vhda:
Thanks @dtikhonov

PR #484, opened 2015-08-05
Embed git rev in  -- version and tags files(5 days)

masatake:


masatake:
This chagens supoort only platforms where gmake and sh are available.

masatake:
An entry for .ctags is removed.

PR #485, opened 2015-08-06
main, lregex: remove {optional} long flag (3 days)

masatake:
{optional} long is redundant. It can be replaed with --kind-<LANG>=-
option.

Note about compatibilities:  I introduced it and exuberant ctags doesn't have it.


b4n:
LGTM

PR #490, opened 2015-08-07
Fix a benign-for-now off-by-one bug in Perl parser

dtikhonov:
This bug in findPerlTags() might become less benign should the
implementation of fileReadLine() change.


dtikhonov:
> LGTM. Did you try "make noise"? It is useful to hunt this kind of bug.

No.  I will check it out.

PR #493, opened 2015-08-07
Add Perl6 parser to VS project files as well (1day)

ffes:
The new Perl6 parser is not yet added to the Visual Studio project files.

### DIFF
--- a/win32/ctags_vs2013.vcxproj
+++ b/win32/ctags_vs2013.vcxproj
@@ -94,6 +94,7 @@
   <ItemGroup>
     <ClCompile Include="..\main\htable.c" />
     <ClCompile Include="..\main\nestlevel.c" />
+    <ClCompile Include="..\main\selectors.c" />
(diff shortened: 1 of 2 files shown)


vhda:
Wasn't trashbox.h removed from the repository?

masatake:
Feel free to add a script to check VS project files. We can integrate it to git-hook in the future.

ffes:
First I need to check the comment of @vdha hopefully tomorrow.

Adding a script to check this seems a lot of work. If AppVeyor is used it will do the check for us.

ffes:
The WiFi in our second stay is not that good, so I don't think I will be
able to do or respond too much until the weekend when I'm back home.

vhda:
Enjoy your vacations instead. At least from my point of view, since we still don't have neither release date nor a release feature plan.

masatake:
@ffes, enjoy your staying.

masatake:
I merge this.
I will remove trashbox.h later.

PR #495, opened 2015-08-08
Solve confliction between objc and matlab with selector(5 days)

masatake:
This commits do two things.

1. revise --langmap and --map-<LANG>=options, and
2. replace 2gram based arbitrator with a selector to solve confliction between ObjectiveC and MatLab about an extension ".m".

`2.` doesn't require 1. directly. However, test cases for 2. need 1.

About the option I have to update documents. I will do it after stabilizing the behavior of the options.


main/selectors.c
```diff
@@ -81,11 +101,48 @@ tastePerlLine (const char *line)
...
-            return lang;
+    /* Default to Perl 5 */
+    return selectByLines (input, tastePerlLine, TR_PERL5);
+}
+
+static const char *
+tasteObjectiveCOrMatLabLines (const char *line)
+{
+#define TR_OBJC "ObjectiveC"
+#define TR_MATLAB "MatLab"
```
vhda: Is using a `#define` better than `static const`?
Since you don't have a corresponding `#undef` at the end of the file, isn't there a (very low) risk that this define affects other files?

main/selectors.c
```diff
@@ -81,11 +101,48 @@ tastePerlLine (const char *line)
...
+	     || startsWith (line, "typedef "))
+	return TR_OBJC;
+    else {
+	if (startsWith (line, "function ")) {
+	    const char *p = line + strlen ("function ");
+	    while (isspace(*p))
+		p++;
+	    if (*p != '\0' && *p != '(')
+		return TR_MATLAB;
+	}
```
vhda: I like this approach! :)
Out of curiosity, did you compare the run time of both cases?

masatake:
I converted "#define ... " to static const char* at the top-level.

main/selectors.c
b4n: > Since you don't have a corresponding `#undef` at the end of the file, isn't there a (very low) risk that this define affects other files?

There is no risk unless someone includes the `.c` itself somewhere (which no one should do).  And in this case, getting all declarations is expected, and probably what one would expect.  It being macros, or file-local variables or functions.

vhda:
I have to be honest: I really hate the macros.
But I come from a Verilog world, where all project files are compiled in a single run and so pre-processor accidents are much more prone to happen, as they are kept defined during the compilation of multiple files.
I also prefer to have the memory allocation clearly defined. Even though a `#define` is used, the `char []` will still need to be allocated by the compiler, correct?

But again, I have limited C experience so I will accept your decision.

b4n:
> I also prefer to have the memory allocation clearly defined. Even though a `#define` is used, the `char []` will still need to be allocated by the compiler, correct?

Yes, a literal will be stored in the executable, like any literal -- after all the C preprocessor just pre-processes, the resulting code is normal code.
One slight advantage of using literals instead of explicit variables might be that you will only have one single definition for a given literal (e.g. you can use as many occurrences of the `"foo"` literal, you'll still only have one single data entry for it, no matter how you use it).

As for "clearly defining" the memory allocation, maybe if you're not used to C it might look less magical, but OTOH you probably either don't really care or know what happens -- and in any case it's probably less "magical" than what other languages do, is it :)

Anyway, I would myself have kept the macros, but I don't mind either way, using either `const char []` or plain `strlen()` is fine too -- BTW, the optimization of `STRLEN()` is probably not useful, modern compilers are likely to optimize `strlen()` calls if they can anyway (at least GCC does it -- even at `-O0` for plain literals like `strlen("hello")` apparently).

vhda:
I apologize for extending this discussion, but I'm taking this opportunity to improve my C knowledge.

>One slight advantage of using literals instead of explicit variables might be that you will only have one single definition for a given literal (e.g. you can use as many occurrences of the "foo" literal, you'll still only have one single data entry for it, no matter how you use it).

I would expect the exact opposite behavior of what you describe. That is, I thought that when using a literal multiple times in the code through a `#define` that is interpreted during the pre-processing stage, then it will not guarantee that the compiler only has a single literal. In fact, I would only expect this behavior when using a `const char []` because we are explicitly declaring a **single** literal. This was one of the main reasons why I preferred the latter!
But I guess that compilers nowadays optimize all these decisions, so if they find the same literal multiple times in the code they only allocate memory once. Possibly they will only replicate these literals if having them available in different memory zones might improve execution time, correct?

Issue #494, opened 2015-08-08
Some tests fail on MSYS2


k-takata:
I have uploaded the results of `make check` on MSYS2.
https://gist.github.com/k-takata/d3ea1ddb3c5c5e186348

Some tests fail because the executable name is shown in the fullpath:

```diff
--- Tmain/alias-for-unknown-language.d/stderr-actual.txt	2015-08-08 16:01:59.918759200 +0900
+++ Tmain/alias-for-unknown-language.d/stderr-expected.txt	2015-07-09 22:37:01.408106900 +0900
@@ -1 +1 @@
-C:\WORK\universal-ctags\ctags.exe: Unknown language "nosuchlang" in "alias-nosuchlang" option
+ctags: Unknown language "nosuchlang" in "alias-nosuchlang" option
```

Some other tests seems to fail with another reason.
E.g. https://gist.github.com/k-takata/d3ea1ddb3c5c5e186348#file-tmain_lregex-list-kinds-d_stderr-diff-txt

Related: #478

masatake:
About the quoted lines, baseFilename in routines.c dones't work as we expected.
@k-takata, can I ask you to look this function more?




Tmain is rather newer test harness than Units. Both mainly I wrote. In Units features (ctags --list-feathres) are tested before running a test case. The features available in the target ctags is compared with features a test case requires. e.g. Units/xcmd-etags.d/features.

In some test cases in Units, this features tesing are conducted. 
In most of all test cases in Units, this features testing is not conducted.

Features causing trouble here will be +wildcards, +regex, +cproc. However, +wildcard, +regex should be available. These two require libregex and libfnmatch. Ctags build system utilizes the bundled library source code when they are not available. Therefore features +wildcard, +regex should be always available.

Anyway I would like to know --list-features output on the environment/platform.

masatake:
BTW, @k-takata, will you have more issues and ideas to improve ctags?
(I guess the answer is yes.)

We want you to work as a member of universal ctags team.
Do you have interest being a member?
Smaller fixing like s/yamoto/yamato/ can be done without sending a PR:-P.

If you have, could you add your name to docs/developers.rst and sending a PR?

k-takata:
> Anyway I would like to know --list-features output on the environment/platform.

You can see the result of `./ctags --version` at here:
https://ci.appveyor.com/project/k-takata/ctags/build/1.0.69/job/8gf0t6sgurclt837#L775

```
[...]

PR #497, opened 2015-08-09
Fix msys2 ExecutableName

k-takata:
When compiled with MSYS2, `ExecutableName` was not set properly.
There were two problems:

1. `MSDOS_STYLE_PATH` was not defined when using autotools on MSYS2.
2. File extension was not removed from `ExecutableName`.

Related: #494


masatake:
I want the command line passed to a compiler short.
Instead of passing -D to the compiler utilizing config.h is better.

I think AC_DEFINE or AC_DEFINE_UNQUOTE will help us.

Following patch is not tested.

```
index f6ebad5..ccd66cf 100644
--- a/configure.ac
+++ b/configure.ac
@@ -117,6 +117,8 @@ AH_TEMPLATE([NEED_PROTO_FTRUNCATE],
 	a prototype for, or a type mismatch using, the following function, define
 	this label and remake.])
 
+AH_TEMPLATE([EXEEXT],
+	[File extension for exectables.])
 
 
 # Report system info
@@ -396,6 +398,8 @@ AC_C_CONST
 AC_OBJEXT
 AC_EXEEXT
 
+AC_DEFINE(EXEEXT)
+
 AC_MSG_CHECKING(if struct stat contains st_ino)
 AC_TRY_COMPILE([#include <sys/stat.h>
 				#include <stdlib.h>], [
```

k-takata:
> I think AC_DEFINE or AC_DEFINE_UNQUOTE will help us.

AC_DEFINE didn't work but AC_DEFINE_UNQUOTE worked. Thanks.

masatake:
LGTM.

vhda:
This is making my _spider sense_ tingle...
Shouldn't we update the Units scripts instead of changing the output of the ctags command?
I'm particularly concerned about the removal of the extension. I understand that it is very, **very** unlikely that we ever have both a `ctags.exe` and a `ctags.com` (or `.bat`) installed in the same system, but in Windows the executable name includes the extension and is essential to uniquely identify the program.
I will not oppose to this change, but I'd like to state my position nevertheless :)

masatake:
@k-takata, does following patch helps you?

```
diff --git a/misc/units b/misc/units
index 4361820..775736a 100755
--- a/misc/units
+++ b/misc/units
@@ -1667,6 +1667,10 @@ tmain_run ()
 	r=$?
 	echo $r > ${subdir}/exit-actual.txt
 
+	sed -i -e 's|^ctags.exe|ctags|' ${subdir}/stderr-actual.txt
+	sed -i -e 's|^ctags.cmd|ctags|' ${subdir}/stderr-actual.txt
+	sed -i -e 's|^ctags.bat|ctags|' ${subdir}/stderr-actual.txt
+
 	if [ $r = $CODE_FOR_IGNORING_THIS_TMAIN_TEST ]; then
 	    run_result skip '/dev/null' "$(cat ${subdir}/stdout-actual.txt)"
 	    for a in ${subdir}/*-actual.txt; do

```

I don't want to touch "*-actual.txt" to avoid being confused.
We should introduce "*-filtered.txt" like units test harness does. But we can do it later.

k-takata:
Updated.
I don't think we need to care about ".cmd", ".bat" or some other extensions, only ".exe" is enough.

masatake:
Fine!

Let's go ahead.

PR #501, opened 2015-08-10
Fix that make check couldn't run in a different builddir

k-takata:
If `configure` was executed in a different directory form the srcdir, `make check` didn't work.

E.g.

```sh
$ cd $srcdir   # cd to the top of the u-ctags srcdir
$ mkdir build
$ cd build
$ ../configure
   ...
$ make check
   ...
/bin/sh: misc/units: No such file or directory
Makefile:439: recipe for target 'tmain' failed
make: *** [tmain] Error 127
```

Note: This PR enables to run `make check` in a different builddir, but the results of `make check` are still stored under `$(srcdir)/Unit` or `$(srcdir)/Tmain`. It would be prefer to store the results in the builddir.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -394,24 +394,28 @@ check: tmain units
 fuzz: TIMEOUT := $(shell timeout --version > /dev/null 2>&1 && echo 1 || echo 0)
 fuzz: $(CTAGS_TEST)
 	@ \
-	c="misc/units fuzz \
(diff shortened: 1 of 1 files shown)


b4n:
Code changes LGTM as a first step.

> It would be prefer to store the results in the builddir.

Agreed, `srcdir` shouldn't be touched in normal operation

masatake:
LGTM.

PR #506, opened 2015-08-10
Dont allow tab in tag name

### LINKED ISSUE #499: whitespaces in tags file (--dont-use-whitespaces-in-name should be added)
I recognized this issue during writing for a test case for rpsec input( #453 )

Loot at our bible, docs/format.rst:

```
   {tagname}
    Any identifier, not containing white space..
```

Some parsers violate this rule.
About lua, I fixed in #491. This is just a bug of parser. Not so interesting.
There is case that a target language, not a parser allows whitespaces in
an object which a parser for the target language will record as a tag.

In all of the cases the target languages are for documentation tools. 
Here is an example.

```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.rst 
Joy of hacking universal ctags
========================================
[yamato@x201]~/var/ctags-github% ./ctags -o - /tmp/foo.rst 
Joy of hacking universal ctags  /tmp/foo.rst    /^Joy of hacking universal ctags$/;"    c
```

tex is another example:

```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.tex        
\section{The joy of hacking ctags}

[yamato@x201]~/var/ctags-github% ctags -o -  /tmp/foo.tex

The joy of hacking ctags    /tmp/foo.tex    /^\\section{The joy of hacking ctags}$/;"   s
```

This one is reproducible in exuberant ctags.

What we should do?

I think the most of tools reading tags just look at tab. So spaces are accetable.
Actually we don't receive any bug report about it.
However, we have to write "universal ctags violates this rule. universal ctags allows spaces in tagname."
Alternative approach is replacing a space with `_`.

The real issue is a tab.  If a tag name includes a tab, what we should do?
My idea is just throwing away the tag in a parser level, not in entry.c till we receive a bug report.

Any comment?

masatake:
Close #499.


main/entry.c
```diff
@@ -1021,7 +1021,14 @@ static void writeTagEntry (const tagEntryInfo *const tag)
 	else if (Option.etags)
 		length = writeEtagsEntry (tag);
 	else
+	{
+		/* This assertion should be at initTagEntry.
+		   However, ada parser does delayed initialization;
+		   it passes NULL pointer as a name to initTagEntry; and
+		   builds the name later in parsing. */
+		Assert (strchr (tag->name, '\t') == NULL);
```
b4n: Maybe duplicate this in `initTagEntry()`, like `Assert (tag->name == NULL || strchr (tag->name, '\t') == NULL)`?  It would be a lot handier for everyone (but the Ada parser), especially as now `Assert()` reports input file position.

masatake: I see. I put the similar check to both initTagEntryFull and makeTagEntry.

PR #508, opened 2015-08-11
fortran: Cleanup parseNumeric() function

vhda:
A temporary vString was being created unnecessarily.

### DIFF
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -877,15 +877,12 @@ static vString *parseInteger (int c)
 
 static vString *parseNumeric (int c)
 {
-	vString *string = vStringNew ();
-	vString *integer = parseInteger (c);
-	vStringCopy (string, integer);
-	vStringDelete (integer);
+	vString *string = parseInteger (c);
 
 	c = getChar ();
 	if (c == '.')
 	{
-		integer = parseInteger ('\0');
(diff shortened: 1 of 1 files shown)


masatake:
LGTM.

b4n:
LGTM, though I'd also reduce the scope of `integer` (in each `if`s)

vhda:
Yeah, on a second thought that makes sense. If both ifs are false there's no reason to allocate that variable.
Will update it as soon as possible.

b4n:
Reducing scope is also rumored to help compiler optimizing (but I never actually checked whether it was actually true).

vhda:
Updated.

parsers/fortran.c
```diff
@@ -877,14 +877,12 @@ static vString *parseInteger (int c)
...
-	vString *integer = parseInteger (c);
-	vStringCopy (string, integer);
-	vStringDelete (integer);
+	vString *string = parseInteger (c);
 
 	c = getChar ();
 	if (c == '.')
 	{
+		vString *integer;
 		integer = parseInteger ('\0');
```
b4n: could be on one single line, `vString *integer = parseInteger ('\0');`

vhda: Done.

b4n:
nice, looks great to me

PR #509, opened 2015-08-11
travis: Compile mingw32 separately

vhda:
Add new environment variable TARGET to separate normal Unix compilation
from mingw32 and exclude the combination clang + mingw32 from the test
matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,16 +38,27 @@ addons:
       - binutils-mingw-w64-i686
       - gcc-mingw-w64-i686
 
+env:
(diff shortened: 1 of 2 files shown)


vhda:
Coverage is enabled both in gcc and clang. Well fix that tomorrow, sorry.

vhda:
Coverage now is only run in GCC and Linux environments, when not compiled with mingw32 target.
Mingw32 target now is disabled in OSX systems.
Also updated script to fail if any of the intermediate commands failed.

Issue #511, opened 2015-08-11
make check stores the results always in the srcdir


k-takata:
Even if `configure` was executed in a different directory form the srcdir, `make check` stores the results in the srcdir not in the builddir.
It would be prefer to store the results in the builddir.

Related: #501

masatake:
I would like to know the requirement.

$(builddir) is not found in Makefile.in. Should I define it by myself?

masatake:
Now I may understand the requirement.
Could you try #513? It is for tmain target.

I will work on units target. It takes rather longer  time. units is much more complicated than tmain. sigh.

masatake:
Is this issue solved now?

k-takata:
Yes.

PR #513, opened 2015-08-11
Store the results of tmain and units to separeted builddir

masatake:


masatake:
How I tested the change:
```
$ mkdir /tmp/b
$ cd /tmp/b
$ ~/ctags/configure --srcdir=/home/masatake/ctags
$ make tmain
```

k-takata:
Does this PR stores only tmain results to separated directory?
How about units results?

masatake:
About Units, I have not implemented yet. But I would like to get feed-back about this change.

k-takata:
I confirmed that this PR works fine for tmain.

masatake:
Is there portable way to get the first character of a string in a shell script?
```${x:0:1}```` is not portable available in dash.
```$(expr substr $x 1 1)``` doesn't work on macosx.

Do I have to use sed here?

masatake:
I should use grep.

masatake:
It works. I'm a bit tired.

vhda:
https://wiki.ubuntu.com/DashAsBinSh

And have done rest :)

masatake:
I trued the changes with dash. I found "source" is not available dash.

masatake:
@k-takata, these changes for units target solves #511?

k-takata:
Yes. I confirmed that the changes works fine.

masatake:
Thank you for confirmation.

PR #514, opened 2015-08-12
options.c: Avoid redefining asprintf

oscarfv:
MinGW64 defines asprintf when __USE_MINGW_ANSI_STDIO is defined.

MSYS2 makepkg-mingw system automatically defines this macro. Without
this change, makepkg-mingw fails because asprintf is redefined in
options.c.

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -479,7 +479,7 @@ static vString* getHome (void)
 	}
 }
 
-#if defined(_WIN32)
+#if defined(_WIN32) && !defined(__USE_MINGW_ANSI_STDIO)
 
 /* Some versions of MinGW are missing _vscprintf's declaration, although they
  * still provide the symbol in the import library.


masatake:
@k-takata, could you look at this when you have time?

k-takata:
I'm wondering why gcc doesn't warn this redefinition on my environment.

> makepkg-mingw fails because asprintf is redefined in options.c.

@oscarfv Could you explain how to reproduce this?

oscarfv:
@k-takata :
When you use `makepkg-mingw` it sets `CPPFLAGS` like this:

    CPPFLAGS="-D_FORTIFY_SOURCE=2 -D__USE_MINGW_ANSI_STDIO=1"

(See `/etc/makepkg_mingw32.conf` in your MSYS2 install).

For reproducing the problem, either add `-D__USE_MINGW_ANSI_STDIO=1` to your build script, or set `CPPFLAGS` before invoking `configure`, or use `makepkg-mingw` with the `PKGBUILD` file in https://github.com/Alexpux/MINGW-packages/pull/739 after removing `CFLAGS=-U__USE_MINGW_ANSI_STDIO` on that file.

The gist is that `asprintf` and others are optionally defined in the MinGW-w64 `stdio.h`, with `__USE_MINGW_ANSI_STDIO` being the switch, and `makepkg-mingw` automatically defines that macro.

k-takata:
Thank you for the explanation. Now I understand.
LGTM.

masatake:
@k-takata, thank you for reviewing.
In our team, you and @ffes are the windows experts. @ffes is in vacation, so please, merge the commits when you say "LGTM".

k-takata:
Done.

masatake:
Thank you.

PR #515, opened 2015-08-12
Fix that writing tags to stdout didn't work with MSYS2(masatake: till Aug 23)

k-takata:
There were two problems with tempfile handling on MSYS2.

1. MinGW-w64 has mkstemp(), but it deletes a tempfile automatically when the file descriptor is closed. However u-ctags expects that the file remains.
2. tempdir was set to '/tmp', but it's not prefer for Windows. Use `TMP` environment for the temporary directory.

Related: #494

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -357,7 +357,7 @@ fi
 # Checks for header files
 # -----------------------
 
-AC_CHECK_HEADERS_ONCE([dirent.h fcntl.h stat.h stdlib.h string.h])
+AC_CHECK_HEADERS_ONCE([dirent.h fcntl.h io.h stat.h stdlib.h string.h])
 AC_CHECK_HEADERS_ONCE([time.h types.h unistd.h])
 AC_CHECK_HEADERS_ONCE([sys/dir.h sys/stat.h sys/times.h sys/types.h sys/wait.h])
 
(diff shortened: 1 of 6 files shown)


masatake:
How about mkstemp? mkstemp tells a caller the name of created file.
As you wrote in the initial message, it removes the file when closing it. For this behavior we can make the closing delayed.

Does I miss anything?

k-takata:
I confirmed that other processes cannot open the temporary file created by `mkstemp` (and not closed yet) on MSYS2.

So, [your proposal](https://github.com/k-takata/ctags/commit/6b38524db47e1972bc7b3fb2eaef4864d590b934#commitcomment-12693349) can be used for internal sort mode but it cannot be used for external sort mode.

Additionally, MSVC doesn't have `mkstemp`.

masatake:
I see.
How about popen instead of system in externalSortTags? 
(I strongly believe sort on the platform accept standard in.:-)

k-takata:
Updated. I used `dup2` and `system` to redirect the stdin.

masatake:
I will review. these changes.

k-takata:
Updated. I will squash the commits later.

k-takata:
Squashed. (Some parts are moved to another commit.)

PR #516, opened 2015-08-12
units: Avoid unexpected pathname conversion on MSYS2(masatake: till Aug 23)

k-takata:
MSYS2 wrongly handles regex options as Unix style path, and convert it to MS-DOS style path.
Setting `MSYS2_ARG_CONV_EXCL` avoids it.

See: http://sourceforge.net/p/msys2/wiki/Porting/#filesystem-namespaces
Related: #494

### DIFF
--- a/misc/units
+++ b/misc/units
@@ -41,6 +41,7 @@ SHOW_DIFF_OUTPUT=
 #
 _CMDLINE=
 _CMDLINE_FOR_SHRINKING=
+_PREPERE_ENV=
 readonly _DEFAULT_CATEGORY=ROOT
 readonly _TIMEOUT_EXIT=124
 readonly _VG_TIMEOUT_FACTOR=10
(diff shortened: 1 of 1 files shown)


b4n:
LGTM, but I don't think the guard is really needed, defining `MSYS2_ARG_CONV_EXCL` on non-MSYS should be harmless.

masatake:
If possible I would like you to add `MSYS2_ARG_CONV_EXCL=--regex-` to CMDLINE.tmp.
The file helps people rerun failed test cases separately.

Approach:

1. Set MSYS2_ARG_CONV_EXCL=--regex- in Makefile.in only if `$(uname -o` = "Msys")` is true.
2. In run_record_cmdline function of misc/units, check whether MSYS2_ARG_CONV_EXCL is set or not.
    If setting MSYS2_ARG_CONV_EXCL is found the function, add arguments printf called
    in run_record_cmdline to record `MSYS2_ARG_CONV_EXCL=--regex-`.

(This is applicable to LC_ALL in 
prepare_environment, too)

k-takata:
Updated.

masatake:
I will review these changes.

masatake:
Thank you. LGTM.

k-takata:
Thanks for reviewing. Merged.

PR #521, opened 2015-08-18
Fix that isSameFile() in routines.c didn't work on MSYS2

k-takata:
MinGW has stat.st_ino but it doesn't contain a valid value. Disable it.
Some tests fail on MSYS2 without this fix.


masatake:
LGTM.

k-takata:
Thanks for reviewing.

PR #522, opened 2015-08-19
Rust parser and test updates

SiegeLord:
- Update Rust examples to compile with modern Rust
- Handle some syntax changes (`where` bounds) that were made close to the 1.0 release.


masatake:
Thank you.

PR #525, opened 2015-08-22
Javascript fix null tag emission stage 2

### LINKED ISSUE #471: jscript: null tag for broken input
```
$ cat y.js
function){
$ ./ctags -o - y.js > /dev/null
ctags: Warning: ignoring null tag in y.js
```

masatake:
Close #471.

The first commit about --_exit-when-warning was needed to make minimal unwanted input
which caused "null tag" warnings.


parsers/jscript.c
```diff
@@ -1402,7 +1402,11 @@ static boolean parseStatement (tokenInfo *const token, tokenInfo *const parent,
 		}
 	}
 
-	copyToken(name, token, TRUE);
+	if ( isType (token, TOKEN_IDENTIFIER) && isKeyword (token, KEYWORD_NONE) )
+		copyToken(name, token, TRUE);
+	else
+		/* Something broken input. Throw away the current token. */
```
vhda: "Something broken input" does not sound right.
Maybe "Broken input" is sufficient?

masatake:
@vhda, @b4n, thank you for comments. I updated my changes.

masatake:
@b4n, I updated the commits based on your suggestion.

PR #534, opened 2015-08-23
css: Don't include any character below ASCII space in the tag name

### LINKED ISSUE #530: Assertion failed on parsing malformed CSS file (main/entry.c:1108)
How to reproduce
```
$ echo -e '0["\t' > min.css
$ ctags min.css
ctags: main/entry.c:1108: void initTagEntryFull(tagEntryInfo *const, const char *const, unsigned long, const char *, fpos_t, const char *): Assertion `name == NULL || strchr (name, '\t') == NULL' failed.
ctags: main/entry.c:1108: parsing min.css:1 as CSS
Aborted
```

this is found by afl-fuzz

b4n:
Characters below space are control characters that are likely to break
tag output format (especially HT, LF, CR, etc.).

While this change is not totally correct (some of those would actually
be valid CSS I think), they are not very important and some are
currently breaking the CTags format rules.

Closes #530.

----
It's not visible in GitHub UI, but the test case contains all control characters between NUL and SP, one per line.


masatake:
Do you think replacing all of them with a space is bettern than following rules for tagfield 
defined in docs/format.rst?

```
  - When a value contains a "\\t", this stands for a <Tab>.
  - When a value contains a "\\r", this stands for a <CR>.
  - When a value contains a "\\n", this stands for a <NL>.
  - When a value contains a "\\\\", this stands for a single '\\' character.
```

masatake:
Can I take over this?


I have thought about this topic long time.
I would like to generalize your patch; ctags main part repairs the name only if the  kind of a tag entry passed to makeTagEntry requests ctags does so. If there is no such request in a kind, main part does nothing other than calling Assert.

```
enum SpecialCharacterHandlingStrategy {
  DO_NOTHING,
  STRIP,
  ESCAPE,
  ...
typedef struct sKindOption {
	boolean enabled;          /* are tags for kind enabled? */
	int letter;               /* kind letter */
	const char* name;         /* kind name */
	const char* description;  /* displayed in --help output */
        SpecialCharacterHandlingStrategy strategy;
} kindOption;
```

vhda:
Don't you think it would be better to not call assert? That way, if there is an error in the beginning of the file the remaining tags are still identified.

masatake:
@vhda, ... the assertion calls bring us many bug reports. I would like to fix them if possible.
When none of us cannot find time to fix bugs, let's remove the assertions.

b4n:
> Do you think replacing all of them with a space is bettern than following rules for tagfield
defined in docs/format.rst?  […]

Hum, no, probably not indeed.

> Can I take over this?

Sure.

> I have thought about this topic long time.
> I would like to generalize your patch; ctags main part repairs the name only if the kind of a tag entry passed to makeTagEntry requests ctags does so. If there is no such request in a kind, main part does nothing other than calling Assert.

Addition of an option seem a bit overkill to me.  It indeed makes some kind of sense if we want to debug invalid characters in names for languages that don't actually allow them, but this would probably be better served by a check in the parser side as it could know what characters should actually be allowed instead of just rejecting what the CTags format doesn't like (which is mostly irrelevant to the parser itself).

Anyway, as you know I like the idea of fixing this in the main part, I don't think all parsers should know how to handle some specific chars in the output format.

vhda:
I like the idea of having this sanitizer in the main part, but since we are proposing to escape special characters, wouldn't it be more consistent to also escape the white space? Would this solution be more compatible with the previous standard of not using white spaces at all? (Guess not, but asking doesn't hurt, right?;)

I still need to confirm the details of this in the Verilog standard, but if I remember correctly Verilog does not allow for square brackets in keywords unless the keyword is escaped (starts with `\`), in which case these brackets are allowed. If I implement the support for this in Verilog parser, what would be the output of ctags with the new sanitizer?

---

In my opinion it's not ctags responsibility to verify that the source code being parsed is correct. Ctags should make a best effort to find as many tags as possible. If I mistakenly type `ã` instead of `a`, ctags should still output this tag (and any other tags it finds later in the source file or files) even if  this is not a valid keyword in the source language. Code validation is the responsibility of compiler, linter, interpreter, simulator, etc.

With this said, a bug related with an `assert()` call should be considered show stopper because it might make the user unable to parse its source code. A bug related with an invalid/incorrect output tag is not so important, as the user can still obtain the remaining tags. So I really do not think it is better to have an `assert()` just to get more bugs reported. This might even work the other way around, because the user might consider going to another tags generator because this doesn't work!

masatake:
I wrote the infrastructure code for improving this area.
I would like to clear the situation. a TagEntryInfo is passed to makeTagEntry function. The name field of the TagEntryInfo includes 'unwanted' character. What ctags should do in this case?

Our choince may be:
A. Abort with assertion (current implementation),
B. Just skip the tag, and/or
C. Escape the character.
D. Write out to tags file though it violates FORMAT (for debugging prupose?)

Am I correct?


Let's call them policies. With the infrastructure code we can impement any of them.

@vhda, your opinion is ctags should not do A. How about B and C?
If these for policies enough, please, add E.
My idea is developer can choose one of the policies for each kinds of a parser.
The issue in my idea is which one should be the default policy.

masatake:
Any way I will make a PR. Temporarily I will implement only A, B, and D.

b4n:
IMO:

> A. Abort with assertion (current implementation),

As said I can also understand the developers finding it somewhat handy (though as said there could easily be more useful ways to check this).

> B. Just skip the tag, and/or

I don't see the point.

> C. Escape the character.

This is the way to go IMO.

> D. Write out to tags file though it violates FORMAT (for debugging prupose?)

This is not acceptable as it might render the generated tag file unusable.

So IMO: *C* > *A* > *B* > *D*

masatake:
(
> > B. Just skip the tag, and/or
> I don't see the point.

I should say this as "Just ignore the tag and don't print it to tags file.".
)

masatake:
> So IMO: C > A > B > D
o.k. I see.

@vhda, how about you?

b4n:
> I should say this as "Just ignore the tag and don't print it to tags file.".

Yeah I get what the behavior would be, but I don't see a use case for silently skipping such a tag.  I see this behavior as a poor man's implementation of *C* (e.g. just ignore the tag not because we want to but because we don't properly deal with it).

vhda:
I prefer solution C.
But maybe we could keep your approach of having the parser configure the desired behavior such that a warning message is displayed (to stderr) if a tag is escaped and the parser was not configured to support such mechanism. This way the parsing does not abort, the tag is written and the user/developer is informed of an unexpected tag.

masatake:
@vhda, I see. Than you. "Warning" with enough information is good idea. Even warning we can get a bug report:-P

I'm not sure how to combine all these ideas here. Soon I will submit an initial implementation.
I would like to get comment for it.

masatake:
(I must reread https://github.com/universal-ctags/ctags/pull/162, too)

masatake:
Now kindOption can be passed to makeTagEntry.
I will prepare next set of patches.

masatake:
As usual, both of you are correct. "per-kind" is overkill. I recognized this after implementing.
I throw away the implementation. and I'm rewriting simplified one.

If an unexpected char is found in writing stage, ctags main part does:
```
if (!Option.quiet) => warning();
if (Option.fatalWarnings) => error(abort);
repair;
```

I apply this policy only to tag name and scope name.
Other fields also should not include such characters but I cannot solve all at once.

masatake:
...it seems that what we need is per-field policy, not per-kind policy....

masatake:
An issue dealt by this PR is solved by another way in #582.

PR #535, opened 2015-08-24
main: fix a memory leak in language map operation

masatake:
The leak is found in option-corpora-conflicts.d test
case with VG=1.


vhda:
It appears that `keepItem` is never true. Are you expecting a scenario where such is necessary?
And if the situation presents itself, wouldn't it be better to force the use of `vStringNewCopy()` to have the memory management be done where needed?

masatake:
Updated.

PR #546, opened 2015-08-31
Add -Wall to compilation flags 

steveno:
After adding Wall I had to make two changes to the code:
* Remove an unused variable
* Remove an unused parameter


steveno:
I changed the makefile back and squashed the commits. Hopefully this should be good to go now.

masatake:
Thank you.

PR #548, opened 2015-09-03
Added Device Tree Source(DTS) parser (3 days)

ARivottiC:


masatake:
Test cases are neeed.
See https://github.com/universal-ctags/ctags/blob/master/docs/units.rst
and files under Units directory.

ARivottiC:
Tests added as requested.

masatake:
Thanks.
LGTM. I will keep the open for 3 days to get more comments from other reviewers.

masatake:
( It seems that DTS utilizes C preprocessor. So we can use get.c to capture included files.
I will work on this after merging commits submitted here.)

source.mak
```diff
@@ -113,7 +114,7 @@ SOURCES = $(MAIN_SOURCES) $(PARSER_SOURCES) $(DEBUG_SOURCES)
 
 ENVIRONMENT_HEADERS = e_msoft.h
 
-ENVIRONMENT_SOURCES = 
+ENVIRONMENT_SOURCES =
```
vhda: Although I agree with the change, since only white spaces are modified I would prefer not to have this included in the patch.

source.mak
```diff
@@ -46,6 +46,7 @@ PARSER_SOURCES =				\
 	$(PARSER_DIR)/cobol.c			\
 	$(PARSER_DIR)/diff.c			\
 	$(PARSER_DIR)/dosbatch.c		\
+	$(PARSER_DIR)/dts.c			\
```
vhda: A similar change is required in `win32/ctags_vs2013.vcxproj` and `win32/ctags_vs2013.vcxproj.filters`, such that we at least try not to break win32 build.

ARivottiC: Made a new commit with this. Wasn't able to test it dough.

source.mak
ARivottiC: Yes, that was my OCD, nevermind :).

win32/ctags_vs2013.vcxproj
```diff
@@ -197,4 +198,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
\ No newline at end of file
+</Project>
```
vhda: Use `git reset -p HEAD^` to revert this EOF specifically.

win32/ctags_vs2013.vcxproj.filters
```diff
@@ -312,4 +315,4 @@
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
vhda: Use `git reset -p HEAD^` to revert this EOF specifically.

source.mak
vhda: The change is still in the commits, right?

vhda:
Please squash "Fixed line endings to DOS" onto "Added support for win32 builds.".
And the result of this should be squashed into your first commit where you add the new files.

ARivottiC:
@vhda not used to squash commits. I hope it is as expected.

vhda:
Uhhh... I think you lost some commits along the way.
The test case is missing now :(

Sorry for all the trouble, but if we need to automatically find a regression using `git bisect` it is important that all commits compile cleanly, even in Windows.

masatake:
Can I merge these?

vhda:
The test is there. I'm not sure if I didn't see it before or if it was added in the meanwhile.
But the extra new lines are still in the patch, which would be cleaner without them.

masatake:
@vhda, thank you. I will remove the extra new line.

masatake:
@ARivottiC, thank you.

masatake:
...though I remove the extra new line, the future contributor will add it accidentally.
So I change my mind; I will do nothing till building on Windows gets failed.

k-takata:
When someone edit *.vcxproj with VisualStudio, VisualStudio might automatically remove the extra new line, I think. (Building on Windows will not fail even with the new line.)

masatake:
> When someone edit *.vcxproj with VisualStudio, VisualStudio might automatically remove the extra new line, I think.

So we don't have to do it now. The future contributor will fix it. :-P.

PR #549, opened 2015-09-07
Fix a bug found in AppVeyor

masatake:
This is a trial to solve an issue reported in #494.
Let's see what happens on AppVeyor.


b4n:
`isalpha()` depends on the locale, so it makes sense it has different behavior depending on the system.  BTW, most `is*()` functions from `ctype.h` are locale-dependent, and that might be something we wanna check, as we likely don't want that in parsers.

masatake:
@b4n, thank you for the suggestion.
How do you think set locale to C in ctags side?
So we can ignore some parts of is* related things.
Bad idea?

PR #550, opened 2015-09-07
AppVeyor: Install dos2unix on Cygwin

k-takata:
Some tests were skipped on Cygwin on AppVeyor because Cygwin version of
dos2unix was not installed. (Another version of dos2unix which comes
with Git is already installed but it doesn't work well when we run our
tests.)


k-takata:
Related: #308

masatake:
Thank you!

PR #551, opened 2015-09-08
Tag macros and header files in dts

masatake:
See the each commit log and the change in the test case.

### DIFF
--- a/Units/dts-simple.d/args.ctags
+++ b/Units/dts-simple.d/args.ctags
@@ -1,0 +1,1 @@
+--kinds-DTS=+h
--- a/Units/dts-simple.d/expected.tags
+++ b/Units/dts-simple.d/expected.tags
@@ -1,2 +1,4 @@
 0x00	input.dts	/^    phandle = <0x00>    $/;"	p
+bar	input.dts	/^#define bar /;"	d	file:
+foo.dtsi	input.dts	/^#include "foo.dtsi"/;"	h
 label	input.dts	/^label: test {$/;"	l
--- a/Units/dts-simple.d/input.dts
+++ b/Units/dts-simple.d/input.dts
@@ -1,5 +1,8 @@
 /dts-v1/;
 
+#include "foo.dtsi"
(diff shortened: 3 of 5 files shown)


ARivottiC:
Nice, I was wondering about this but didn't had the time. Thank you @masatake.

masatake:
@ARivottiC, I have a question about dts.
As far as reading dts and dtsi files in the latest linux kernel code, I cannot find any line where phandle is used. What does it mean?  (labels are used widely.)

ARivottiC:
@masatake, a phandle is a reference for another node:

node1 { ... }
node2 {
    a-property = <&node1>
}

When converting Device Tree Blobs (dtb) to dts, you'll get something like this:

node1 { 
    phandle = <0x01>
}
node2 {
    a-property = <0x01>
}

masatake:
@ARivottiC, do you mean we can see lines using phandle only when converting Device Tree Blobs (dtb) to dts?

ARivottiC:
@masatake that's the use case I'm used to. In fact, my contribution is due to the fact I was comparing two dtb files and it was a pain going back and fourth with the phandle references.

masatake:
```
node1 { ... }
node2 {
a-property = <&node1>
}
```
How about node1?
When reading  `a-property = <&node1>` line of the original dts file, I guess the reader may want to know the line where node1 is defined.

ARivottiC:
My bad! "node1" is a label, should be something like:

node1: node@0 {...}

masatake:
I see. Thank you.

masatake:
I would like to ask you one more question.

As far as reading kernel code `};` is used to close an elemenet(?). But elements in your examples are closed
with just `}'. I cannot find `;`. Do you have any idea about this difference?

(I'm working on implementing scope in dts. To track the scope I have to the closer of elements.)

ARivottiC:
Again you are right. I forgot the ;.

masatake:
Thank you. I will add ';' to the test case.

PR #552, opened 2015-09-09
Add total lines to file tag

masatake:
See the expected.tags.


main/parse.c
```diff
@@ -1772,6 +1775,26 @@ extern void makeFileTag (const char *const fileName)
 	}
 }
 
+static void makeEOFTag (const char *const fileName)
```
vhda: Why not merge `makeEOFTag` into `makeFileTag` and only call the later where the first is now called?

masatake: I see. I will introduce makeFileComon static function to merge the both code.

vhda: Please check my commit at https://github.com/vhda/ctags/tree/masatake-add-total-lines-to-file-tag

masatake: Oh, I see. Your one is better. Please, merge yours.

vhda: You already have the PR opened! :)
Just drop your "refactor" commit and cherry-pick my extra commit. Everything should be ok after that, or so I hope.

PR #554, opened 2015-09-11
SML: avoid accessing `enabled' field of wrong type

### LINKED ISSUE #528: Assertion failed on parsing malformed SML file (main/parse.c:1372)
How to reproduce:
```
$ echo and > and.sml
$ ctags and.sml
ctags: main/parse.c:1372: boolean isLanguageKindEnabled(const langType, char): Assertion `kindOpt' failed.
ctags: main/parse.c:1372: parsing and.sml:1 as SML
Aborted
```

this issue is found by afl-fuzz

masatake:
This is reported and suggested by @kcwu.

Quoted from the report from @kcwu (#528):

    I found I can reproduce it with clang but not with gcc. The problem is

     72| static void makeSmlTag (smlKind type, vString *name)
     73| {
     74|         tagEntryInfo tag;
     75|
     76|         if (! SmlKinds [type].enabled)
     77|                 return;

    where type = K_NONE = -1.
    It's out of bound access for SmlKinds[-1] and thus depends on memory layout.

Close #528.


parsers/sml.c
```diff
@@ -31,7 +32,8 @@ typedef enum {
 	K_SIGNATURE,
 	K_STRUCTURE,
 	K_TYPE,
-	K_VAL
+	K_VAL,
+	K_LAST_ELT = K_VAL
```
b4n: I'd use something like `K_COUNT /* must be last */` and simply keep it last, to avoid having to update it when adding a kind.  But that's mostly style, so it's your call.

b4n:
LGTM

masatake:
@b4n, thank you for reviewing. I updated as you suggested.

b4n:
last commit LGTM, but apparently 3 more commits invited themselves here, yet they don't seem related ^^

masatake:
Oh, I'm sorry. Pushed again.

b4n:
LGTM now :)

masatake:
@b4n, thank you for taking time.

PR #555, opened 2015-09-16
configure.ac: Remove redundant $TMPDIR environment variable test.

p-montanus:
Redundant TMPDIR environment variable tests are exists in configure.ac.


masatake:
Thank you.

PR #558, opened 2015-09-16
Pass kind option structure to makeTagEntry

masatake:
Currently a kind letter and name are passed to makeTagEntry when making a name as a tag.
 This patch set allows a paser passing a kind descriptor(kindOption) instead of them.

We can extend the kind descriptor later so that we can reduce code and implement interesting features based on this patch set.

e.g.
(1) kind enable/disable checking can be done in main part side,
(2) policy how ctags delas broken name in tag when printing, 
(3) flavor defining alternative kind letter and name, and
(4) automatic qualifie tag generation.


main/kind.h
```diff
@@ -0,0 +1,37 @@
+/*
+*   Copyright (c) 1998-2003, Darren Hiebert
```
vhda: Darren? ;)

main/parse.h
```diff
@@ -90,7 +66,7 @@ typedef struct {
 	char* name;                    /* name of language */
 	kindOption* kinds;             /* tag kinds handled by parser */
 	unsigned int kindCount;        /* size of `kinds' list */
-	char fileKind;		           /* override letter for file kind */
+        kindOption* fileKind;          /* kind for overriding the default fileKind */
```
vhda: Bad indent

main/kind.h
masatake: The definition of kindOption is just moved from parse.h. So I keep the Copyright notice.

PR #561, opened 2015-09-17
Fix configure use obsolete detecting macros

p-montanus:
AC_TRY_COMPILE() and AC_TRY_RUN() are obsolete. Succeeding AC_COMPILE_IFELSE() and AC_RUN_IFELSE() should be used.


b4n:
Not tested but LGTM

masatake:
Tested on RHEL5.11 (with some proposed patches).

masatake:
Thank you.

Issue #524, opened 2015-08-22
Ruby parser add support for --extra=q option


lowski:
Hi!
Thanks all your hard work you put into this project.

I was wondering it would be possible to add support for `--extra=q` option for Ruby parser. When that option is specified tags should also include class with its namespace, i.e. `A::B`) not only B.

masatake:
I don't know Ruby. Could you tell me the rule about scope separators? Which should I use `::` or `.`?
One more when you want to enable --extra=q? I guess ctags developers don't have much interest --extra=q because, making `q` entries can be done in post processors which reading tags file generated by ctags.

I myself don't have much chance to use ctags, so I would like to know how people use it, especially about `--extra=q`.

vhda:
This is mostly useful in vim, where you don't have access to the extra attributes of the tags. So if you want to search for the contents of, let's say, a class you need to search for `class_name\..*` or similar.
It is possible that this limitation exists in other tools, so having this support is always positive.

masatake:
@vhda, thank you for explanation.

I have got an inde to implement it in mainly main side.
 But many parts must be modified. It will take long time.

Only the issue is about separators.
In my understanding a separator can be choose by relations between the kinds of parent and the kinds of child.

e.g.
If the kind of parent is class and the kind of child is method, the separator will be `.`.
If the kind of parent is namespace and the kind of child is class, the separator can be `::`.
A parse should have a table(separators table?).
main part looks at the table when generating "q" related entry.

lowski:
@masatake basically when class / module is nested (`A::B`), `--extra=q` would generate 2 records in ctags - one for `class B` and the other one for `class A::B`. It would be useful in vim so that I could immediately jump to the correct class if I highlight `A::B`. Currently I need to select `B` but if there are more `B` classes in other modules I would see all of them on the selection list.

The idea came from here - https://github.com/tmm1/ripper-tags/issues/48. Currently I uses that gem to correctly generate ctags file for ruby.

Unfortunately I'm not very familiar with algorithm for generating tags file.

masatake:
How about method?
A.method?

dgutov:
@masatake `A#method` for an instance (normal) method, and `A.method` for a class method (one defined with e.g. `def self.method` inside `class A`).

masatake:
The basic infrastructure is implemented as part of solving #787.
About python, fullQufalified tag is implemented on the infrastructure, and works fine.
Next one is ruby.

[...]

PR #562, opened 2015-09-17
Regex with scope

masatake:
Introduce scope long flag to allow a regex based parser attaching scope information to tags.
e.g.

```
    $ cat /tmp/input.pp
    class foo {
	include bar
    }

    $ cat /tmp/pp.ctags
    --langdef=pp
	    --map-pp=+.pp
	    --regex-pp=/^class[[:blank:]]*([[:alnum:]]+)[[[:blank:]]]*\{/\1/c,class,classes/{scope=push}
	    --regex-pp=/^[[:blank:]]*include[[:blank:]]*([[:alnum:]]+).*/\1/i,include,includes/{scope=ref}
	    --regex-pp=/^[[:blank:]]*\}.*//{scope=pop}{exclusive}

    $ ~/var/ctags/ctags --options=/tmp/pp.ctags -o - /tmp/input.pp
    bar	/tmp/input.pp	/^    include bar$/;"	i	class:foo
    foo	/tmp/input.pp	/^class foo {$/;"	c
```

### DIFF
--- a/Units/regex-with-scope.d/expected.tags
+++ b/Units/regex-with-scope.d/expected.tags
@@ -1,21 +1,21 @@
 NS1	input.foo	/^ns NS1$/;"	n
 x	input.foo	/^define x {$/;"	d	namespace:NS1
-y	input.foo	/^	define y {$/;"	d	definition:x
-v_y_0	input.foo	/^	       var v_y_0$/;"	v	definition:y
-v_y_1	input.foo	/^	       var v_y_1$/;"	v	definition:y
-z	input.foo	/^	define z {$/;"	d	definition:x
-v_z_0	input.foo	/^	       var v_z_0$/;"	v	definition:z
-v_z_1	input.foo	/^	       var v_z_1$/;"	v	definition:z
-a	input.foo	/^	       define a {$/;"	d	definition:z
+y	input.foo	/^	define y {$/;"	d	definition:NS1.x
+v_y_0	input.foo	/^	       var v_y_0$/;"	v	definition:NS1.x.y
+v_y_1	input.foo	/^	       var v_y_1$/;"	v	definition:NS1.x.y
+z	input.foo	/^	define z {$/;"	d	definition:NS1.x
+v_z_0	input.foo	/^	       var v_z_0$/;"	v	definition:NS1.x.z
+v_z_1	input.foo	/^	       var v_z_1$/;"	v	definition:NS1.x.z
+a	input.foo	/^	       define a {$/;"	d	definition:NS1.x.z
 NS2	input.foo	/^ns NS2$/;"	n
 p	input.foo	/^define p {$/;"	d	namespace:NS2
-q	input.foo	/^   define q {$/;"	d	definition:p
+q	input.foo	/^   define q {$/;"	d	definition:NS2.p
 v_g1	input.foo	/^var v_g1$/;"	v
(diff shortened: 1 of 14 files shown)


Units/regex-with-scope.d/expected.tags
```diff
@@ -0,0 +1,20 @@
+NS1	input.foo	/^ns NS1$/;"	n
+x	input.foo	/^define x {$/;"	d	namespace:NS1
+y	input.foo	/^	define y {$/;"	d	definition:x
+v_y_0	input.foo	/^	       var v_y_0$/;"	v	definition:y
```
vhda: Discussion: would it make sense to have "definition:NS1.x.y" instead?

masatake: Tank you. You are right. Very interesting.

```
[jet@localhost]~/var/ctags% cat /tmp/foo.py 
cat /tmp/foo.py 
def x():
    def y():
        def z():
            pass
[jet@localhost]~/var/ctags% ./ctags -o - /tmp/foo.py 
./ctags -o - /tmp/foo.py 
x	/tmp/foo.py	/^def x():$/;"	f
y	/tmp/foo.py	/^    def y():$/;"	f	function:x	file:
z	/tmp/foo.py	/^        def z():$/;"	f	function:x.y	file:
[jet@localhost]~/var/ctags% cat /tmp/foo.c
cat /tmp/foo.c
struct x {
  struct y {
    int a;
  } z;
};
[jet@localhost]~/var/ctags% ./ctags -o - /tmp/foo.c
./ctags -o - /tmp/foo.c
a	/tmp/foo.c	/^    int a;$/;"	m	struct:x::y	file:
x	/tmp/foo.c	/^struct x {$/;"	s	file:
y	/tmp/foo.c	/^  struct y {$/;"	s	struct:x	file:
z	/tmp/foo.c	/^  } z;$/;"	m	struct:x	typeref:struct:x::y	file:
```

It seems that full qualified tag is needed as scope name.
Scope context is tracked with a stack, it may be possible to make full qualified tag.
The basic idea is already show in #524.I have some code.

One of ths issue is the command line interface to specify scope separator(operator or combinator?).
You wrote `definitions:NS1.x.y`. However, if `.` is for string concatenation in this language,
the user may want ... `definitions:NS::x->y` or something natural in the language. The worst
case is that a language allows `.` being part of a name.

In addition, a language neutral tool may want to know which characters (or strings) are the separator for the tag to separate the each components of a scope name.

How about following one:
```
--separator-<LANG>=+<KIND1><KIND2><SEPARATOR>
```
Applying this to foo:
```
--separator-foo=+nd::
--separator-foo=+dd->
```
We need a wild card. Kind letter `*` must be ctags side.

About the output, ... we may have to embed it in pseudo tags.
```
!_TAG_LANGUAGE_SEPARATOR\tnd\t/::/;"language:foo
!_TAG_LANGUAGE_SEPARATOR\tdd\t/->/;"language:foo
```

I have never thought about pseudo tags well so more considerations are needed.
As the initial step I will introduce `.` as default value for scope separator.

masatake:
@vhda, I added code to implement what you suggested. 
I use a term full-scope-context. How do you think about the term?
full-qufalified-scope name is better? or do you have other idea?

Based on this commit I will try to solve #524.

masatake:
I'll clean up the code introduced in this PR when visiting #524 again.

PR #563, opened 2015-09-18
Use mkdir-p without checking its availablity

masatake:

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -27,7 +27,6 @@ pkgconfdir=@sysconfdir@/@PACKAGE_NAME@
 libexecdir=@libexecdir@
 pkglibexecdir=@libexecdir@/@PACKAGE_NAME@
 SLINK	= @LN_S@
-MKDIR_P = @MKDIR_P@
 STRIP	= @STRIP@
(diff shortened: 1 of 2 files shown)


vhda:
Do you think there's no problem in doing this?

p-montanus:
In unit: target
```
units: $(CTAGS_TEST)
        @ \
        builddir=$$(pwd); \
        mkdir -p $${builddir}/Units && \
```
and tmain: target
```
tmain: $(CTAGS_TEST)
        @ \
        \
        builddir=$$(pwd); \
        mkdir -p $${builddir}/Tmain && \
```
use 'mkdir -p', It seems no problem.

masatake:
o.k. Let's assume mkdir -p is always available.

> Do you think there's no problem in doing this?
I don't know but there is a race conditoin. I will do as @p-montanus wrote.
I cannot ignore your request about RHEL5:-)

masatake:
I will prepare a new patch.

masatake:
If  continuous-integration/appveyor  and  continuous-integration/travis-ci/pr say O.k., I will merge this commit.

PR #564, opened 2015-09-19
Fix configure convenience macro

p-montanus:
When checking header file contains string "yes" such as "_Bool prompt_yes_or_no(void);", CHECK_HEADER_DEFINE() returned wrong result.


masatake:
LGTM. Thank you.

Issue #570, opened 2015-09-22
Add support for Matlab variables


petobens:
After seeing the work done by @vhda in #544, I was wondering if it is possible to add support for Matlab variables. Consider for example the following `foo.m` file:

```matlab
A = magic(4);
R = randn(3,4,5);
```

It would be great if tags were generated for these variable.
Thanks in advance!

vhda:
Please try #571

masatake:
The original issue is solved by the commit.

PR #571, opened 2015-09-22
matlab: Parse variables in Matlab cod

vhda:


vhda:
Fixes #570, but since there isn't proper variable declaration in Matlab every assignment is identified as a declaration of a variable. So this solution is less than ideal.

petobens:
@vhda thank you! Seems to work for me but I understand if you prefer not to merge it. No idea how to workaround the fact that Matlab doesn't have a proper variable declartion.

vhda:
Let's wait for comments from the rest of the team (who have been oddly quiet lately).

petobens:
@vhda ok. Thanks once again.

masatake:
@vhda, I'm sorry I'm bit busy now.
Could you show me unwanted input?
My guessing is:
```
A=1;
A=2;
```
In my example A is tagged twice, this is what you don't want?
If yes, do you want ctags to tag only the first A?
How about the seoncd A?
If #569 is merged, the second A should be marked as "reference"?

I would like to know idea case for future development.

vhda:
Your guessing is correct.

Generally speaking, a tag is created when a variable is declared. But this is a problem in languages like Matlab, where the syntax does not include variable declaration. This is also true in Python, which does not use a regex parser but it also generates two tags if we assign the same variable twice. So, at least we would be consistent even if we merge this PR.

Going forward, it might make sense to create a mechanism where we only create a tag on the first occurrence in the current context. The context requirement would allow the use the same variable name in different functions, for example. This could probably not be too hard to implement in the cork mechanism, if I understood its functionality correctly. But parsers that do not use it will not benefit from the change.

Yes, if #569 is merged, the second A should be marked as "reference".
(P.S. - While reviewing #569, could you also take a peek at #80 again?)

masatake:
> Generally speaking, a tag is created when a variable is declared. But this is a problem in languages like Matlab, where the syntax does not include variable declaration. This is also true in Python, which does not use a regex parser but it also generates two tags if we assign the same variable twice. So, at least we would be consistent even if we merge this PR.

Yes. Let's merge this.


----
Interesting topic.

About cork, as you wrote, it is possible to implement what we need.
I will introduce cork more to parsers; the first step is ruby.

After reading your comment, "reference" is not so simple. A name is referred variety reasons.
For this example "A" is referred as lvalue...It seems that there is a concept "reference kind".
Some ideas:
```
reference:D - Not reference, a definition.
reference:R - Generic. 
reference:A - Assignment(for variable)
reference:r - The value is used.
...
```
I wonder wheter ctags should deal with such detail or not.
I should implement -X option first. About "reference" we have to discuss more.

petobens:
> Generally speaking, a tag is created when a variable is declared. But this is a problem in languages like Matlab, where the syntax does not include variable declaration. This is also true in Python, which does not use a regex parser but it also generates two tags if we assign the same variable twice. So, at least we would be consistent even if we merge this PR.

Not really sure if I completely understand the discussion but I just want to mention that with `R` files if one writes:

```R
A <- 1
A <- 2
```

then two tags are generated for the variable `A`.

masatake:
@petobens, do you want ctags to make a tag entry for each "A" ?
I guessed you may want only the first one.

petobens:
@masatake , I'm not really sure. I believe that a tag entry for the first one is definitely necessary. However tags for each entry might be useful . For example in Vim I use a tag explorer buffer: if there is an entry for each variable then I can see how that variables changes with each new assignment (this might be useful or not, I guess in the end it's a matter of personal preference).  
But I understand that maybe the standard use is to use tags to jump to the definition (i.e to the first variable declaration).  So really I cannot decide for one or the other.

screenshot-24-09-2015_11-08-55

masatake:
@petobens, thank you for explanation. I understand the second one is useful. However, the first one may be more important than the other. Anyway we will merge the change.

petobens:
@masatake great. Thank you!

vhda:
@masatake just noticed the typo in the subject of the commit. Allow me to fix it tonight and I'll merge the PR afterwards.

masatake:
@vhda, sure!.

petobens:
Thank you!!

PR #575, opened 2015-09-25
Replace complex m4 builtins in configure.ac with suitable AS_TR_CPP

p-montanus:
In configure.ac, some m4 builtins are used directory.
autoconf provides 'm4_'-prefixed alias macros for m4 builtin macros, and original non 'm4_'-prefixed macros are not recommended. Equivalent AS_TR_CPP provided by autoconf is more suitable rather than complex these m4 builtins.

AS_TR_CPP first appeared in autoconf-2.50a (2001-06-26).
autoreconf-2.59 and autoreconf-2.69 with '-Wall' report no warnings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -568,7 +568,7 @@ AC_DEFUN([CHECK_PROTO], [
 	[],
 	[
 		AC_MSG_NOTICE([adding prototype for $1])
-		AC_DEFINE(patsubst([NEED_PROTO_NAME], [NAME], translit([$1], [[a-z]], [[A-Z]])))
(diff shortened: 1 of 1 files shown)


b4n:
Mostly LGTM, see small remarks above. And indeed using `AS_TR_CPP` is a lot less crazy :)

p-montanus:
Thank you for reviewing.
Revised.

b4n:
LGTM, thanks :)

One remark, maybe the middle commit is not useful as all its changes are actually removed in the next one.  But that's not important, it's a valid step anyway.

p-montanus:
Yes, middle commit is just informational.
It aims to put pressure on a person to do not use non 'm4_'-prefixed macros.

b4n:
fair enough :]

PR #578, opened 2015-09-25
Fixing of license & copyright headers

viccuad:
This pull request mostly adds "version 2." to the license already stated in the headers, and adds some missing headers and copyright info.

I have left the copyright holders out for files that weren't mostly done by 1 or 2 persons (one can use git blame to obtains those always).

Also, I have added some license info with the GPL-2 to select files that didn't have a header or license info at all (given that the previous work was GPL-2, exuberant-ctags).

### DIFF
--- a/misc/budge
+++ b/misc/budge
@@ -4,7 +4,7 @@
 #
 #   This program is free software; you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
-#   the Free Software Foundation; either version 3 of the License, or
+#   the Free Software Foundation; either version 2 of the License, or
 #   (at your option) any later version.
 #
 #   This program is distributed in the hope that it will be useful,
--- a/misc/budge
+++ b/misc/budge
@@ -1,4 +1,20 @@
 #!/bin/sh
+#
+#   Copyright (C) 2014 Masatake YAMATO
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
 # ----------------------------------------------------------------------------------------
 # Show the ratio of the number of files handled by ctags to the number of all files of ctags source directory.
 # -----------------------------------------------------------------------------------------
--- a/misc/src-check
+++ b/misc/src-check
@@ -1,4 +1,19 @@
 #!/bin/sh
+#
+#   Copyright (C) 2014 Masatake YAMATO
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 check_include_general_h_first()
 {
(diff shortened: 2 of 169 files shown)


viccuad:
Added @k-takata suggestion :)

misc/budge
```diff
@@ -1,4 +1,20 @@
 #!/bin/sh
+#
+#   Copyright (C) 2014 Masatake YAMATO
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
```
b4n: that's v3

b4n:
I guess nobody would complain as it probably (see below) is the spirit (as COPYING is v2), and IANAL, but I'm not sure whether this is totally legal (or even right): didn't the file referencing GPL without a version referred to v1?  Or the latest?  Or "v2 or newer" in the spirit?

masatake:
I got the same question posted by @b4n.
ctags itself tells us the original intent is v2+:
```
[jet@localhost]~/var/ctags% ctags --version
ctags --version

  Compiled: Sep 30 2014, 12:58:02
  Addresses: <dhiebert@users.sourceforge.net>, http://ctags.sourceforge.net
  Optional compiled features: +wildcards, +regex
[jet@localhost]~/var/ctags% ctags --license
ctags --license

  Compiled: Sep 30 2014, 12:58:02
  Addresses: <dhiebert@users.sourceforge.net>, http://ctags.sourceforge.net
  Optional compiled features: +wildcards, +regex

This program is free software; you can redistribute it and/or
modify it under the terms of the GNU General Public License
as published by the Free Software Foundation; either version 2
of the License, or (at your option) any later version.
```

viccuad:
@masatake @b4n: 
 
Good news; the GPL license itself says: (extract from the GPL license)

> Each version is given a distinguishing version number.  If the Program
specifies a version number of this License which applies to it and "any
later version", you have the option of following the terms and conditions
either of that version or of any later version published by the Free
Software Foundation.  If the Program does not specify a version number of
this License, **you may choose any version ever published by the Free Software
Foundation**.

The original code printout says it's under the GPL2+ (and I'm not sure that is totally legally vinculant, but shows the intent), and the code itself has headerss with  "GPL", so that's good. IANAL, but seems one could put GPL2+.

Seems like I made a mistake appending "version 2", and it should be "version 2 or more".

viccuad:
@masatake 
> Could you remove my name (Masatake YAMATO), here. 
> This is mattn's work. 
( Tmain/input-encoding-option.d/run.sh )
( Tmain/output-encoding-option.d/run.sh )

> Could you remove my name (Masatake YAMATO), here. 
> This is Dmitri's work.
( main/selectors.h )
( main/selectors.c )

Fixed.

In my defense, 'git blame' shows that there is a mixture of you and Dmitri and mattn in those files (but maybe you only did refactoring) :)

b4n:
@viccuad Ah, nice!  Ok, so GPL v2+ seem fine to update indeed.

viccuad:
@masatake said:
> Is "Universal Ctags team" acceptable here instead of my name? (misc/travis-check.sh)

IANAL, but this is what i think:

'git blame' shows that the code is from Vitor Antunes. I would say that without an explicit note from Antunes that the copyright is transeferred to "whoever is currently working on Universal Ctags" (since there isn't a  comprehensive list of contributors to check for), then, the Universal Ctags team is no the copyright holder. 

The most sensible thing would be to just drop out the copyright note and be done with it. The license is there, so people can use i.

viccuad:
b4n said:
> that's v3 (misc/budge)

changed to 2.

misc/src-check
```diff
@@ -1,4 +1,19 @@
 #!/bin/sh
+#
+#   Copyright (C) 2014 Masatake YAMATO
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
```
b4n: v3

misc/tinst
```diff
@@ -1,4 +1,20 @@
 #!/bin/sh
+#
+#   Copyright (C) 2014 Masatake YAMATO
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
```
b4n: v3

misc/travis-check.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/sh -e
+#
+#   This program is free software; you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation; either version 3 of the License, or
```
b4n: v3

masatake:
LGTM. I will merge in a day if there is no more comments.

masatake:
Merged.
@viccuad, thank you very much.
The reviewers, thank you.

PR #582, opened 2015-09-29
Escape unexpected characters in tagname

masatake:
This is generic version of a commit proposed in #534 by @b4n.
Instead of handling 'unexpected character' in each parser, this commit tries to handle it in main part.
1ddfa66 is the heart of this commits.

Escaping the characters breaks compatibility. At least php and vim get the impact.
docs/format is updated temporary. Based on this commit I will try to fix afl-fuzz releated bugs reported by @kcwu. Through the fixing I will find more things I have to take care.

Building scope string is not optimized. I will do it after solving #524 based on this commit.
I don't touch code releated to TAGS file.


masatake:
Though I open this PR, not I think just replacing a control character with a whitespace is better than backslash + something.

masatake:
In #162 @b4n provides more complete function for escaping with no string copy.
This is the way to go?

In other hand I may need %N and %-N for extending -x output.

masatake:
I reread #162. Mine is mixed approaches explained in #162.

```
1. Filter the offending bytes when creating the tag entry, in
   `makeTagEntry()`.  I see two implications worth noting with this
    approach:
    A. It requires string duplication that may be avoided with other
       solutions.
    B. It is an arbitrary limitation of the in-memory format, which may
       or may not be wanted (e.g. when making CTags a library).
2. Filter the offending bytes when writing the tag entry.  This has the
   advantage of only affecting the output format (the very goal of the
   change) and not the internal representation, so it has very low
   implication on how internals handle these.
``` 

My code duplicates the strings when printing, not when `makeTagEntry()`.
I don't understand 1.B well. but I think my approach can overcome the limitation.
With combining newly introduced fieldDesc structure in #588, we can customize escape function for each field.

About the escape function itself what @b4n wrote does much more than mine. It seems that I 
should follow @b4n's implementation.

masatake:
This is not updated as I wrote. Just rebased to the latest master.

Good news is that all afl-fuzz test cases are passed.
```
Summary (see CMDLINE.tmp to reproduce without test harness)
  #passed:                                521
  #FIXED:                                 13
	afl-fuzz/github-issue-529-bas
	afl-fuzz/github-issue-530-css
	afl-fuzz/github-issue-531-make
	afl-fuzz/github-issue-532-vim
	afl-fuzz/github-issue-536-flex
	afl-fuzz/github-issue-537-beta
	afl-fuzz/github-issue-538-sql
	afl-fuzz/github-issue-539-js
	afl-fuzz/github-issue-540-fortran
	afl-fuzz/github-issue-541-json
	afl-fuzz/github-issue-542-vhdl
	afl-fuzz/github-issue-543-eiffel
```

steveno:
Great work @masatake !

masatake:
(The documentation is updated.)

masatake:
Poeple who are reading this, could you look at "EXCEPTION"s in https://github.com/masatake/ctags/commit/3bf575f60eeae9cfb353206ecc5840e673e29ce5  ?

This commit explains the incompatible format changing of tags file generated by universal-ctags. 
I myself doesn't use ctags much(I'm not a vi user), so I need feed-back from real user, especially who writes a tool utilizing tags files.

masatake:
I merge this. I will close associated issues and PRs.

Feel free to open a new issue or give a comment here.

PR #587, opened 2015-09-30
build: Add '--with-ctags-{data,sysconf,libexec}dir' configure options

### LINKED ISSUE #580: Missing configure flag for dir naming (/etc/ctags, /usr/share/ctags, /usr/lib/x86_64-linux-gnu/ctags)
I have looked at `configure --help` but I can't seem to find a flag for this. 

Right now, if I install universal-ctags I get (among other things):

/etc/**ctags**/* (optlib, preload and corpora)
/usr/lib/x86_64-linux-gnu/**ctags**/*  (optlib, preload and corpora)
/usr/share/**ctags**/* (optlib, preload and corpora)

It would be useful to have a `configure` flag for changing those ctags dirs into other string (and maybe defaulting to universal-ctags?). 

That would enable universal-ctags to be installed in parallel with other ctags projects, without the possibility of collisions in those directories.

p-montanus:
Add '--with-ctags-datadir', '--with-ctags-sysconfdir', and '--with-ctags-libexecdir' configure options to control universal-ctags specific installation directories.

fix #580.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -26,11 +26,11 @@ libdir	= @libdir@
 incdir	= @includedir@
 mandir	= @mandir@
 datadir = @datadir@
-pkgdatadir = @datadir@/@PACKAGE_NAME@
+pkgdatadir = @pkgdatadir@
 sysconfdir = @sysconfdir@
-pkgsysconfdir = @sysconfdir@/@PACKAGE_NAME@
+pkgsysconfdir = @pkgsysconfdir@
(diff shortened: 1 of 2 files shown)


masatake:
I didn't test this but I merged.

I need this for reworking optlib.  We may have to allow tinst specifying '--with-ctags-datadir', '--with-ctags-sysconfdir', and '--with-ctags-libexecdir'.

PR #590, opened 2015-10-01
main, json: allow null tag

### LINKED ISSUE #492: json: "Warning: ignoring null tag"
```
[yamato@x201 ctags-github]$ cat null.json
[{"":1}]
[yamato@x201 ctags-github]$ ./ctags -o - null.json  
ctags: Warning: ignoring null tag in null.json
0	null.json	/^[{"":1}]$/;"	o
```

masatake:
Close #492.

A null tag was warned and was not outputted.

    $ cat null.json
    [{"":1}]
    $ ./ctags -o - null.json
    ctags: Warning: ignoring null tag in null.json
    0   null.json   /^[{"":1}]$/;"  o

From the view of language spec, it is the right thing that json parser
outputs an empty string though ctags warns it. However, in generally
outputting an empty string as a tag is abnormal.

In this commit, allowNullTag field is added to parserDefinition. If
the field is set to TRUE, a null tag is never warned and is outputted.

Currently only JSON parser set the field to TRUE.


masatake:
I need this to work on another bugs.
Feel free commenting the changes.

b4n:
I'm a bit late, but while indeed technically it's correct for JSON (or JS), it's not actually really useful either.  Also, readtags wouldn't handle such a tag IIRC.

masatake:
@b4n, thank you for the comment. I agree with you.
See #602.

Everytime I understand what you mean after implementing a wrong version. sigh.

Issue #543, opened 2015-08-24
Assertion failed on parsing malformed Eiffel file (main/entry.c:1108)


kcwu:
How to reproduce
```
$ echo -e 'class feature d alias"\t' > min.e
$ ctags min.e
ctags: main/entry.c:1108: void initTagEntryFull(tagEntryInfo *const, const char *const, unsigned long, const char *, fpos_t, const char *): Assertion `name == NULL || strchr (name, '\t') == NULL' failed.
ctags: main/entry.c:1108: parsing min.e:1 as Eiffel
Aborted
```

this is found by afl-fuzz

masatake:
Thank you for all the reports. I will look them later.

masatake:
Closed by #582.

Issue #542, opened 2015-08-24
Assertion failed on parsing malformed VHDL file (main/entry.c:1108)


kcwu:
How to reproduce
```
$ echo -e 'constant"\t' > min.vhdl
$ ctags min.vhdl
ctags: main/entry.c:1108: void initTagEntryFull(tagEntryInfo *const, const char *const, unsigned long, const char *, fpos_t, const char *): Assertion `name == NULL || strchr (name, '\t') == NULL' failed.
ctags: main/entry.c:1108: parsing min.vhdl:1 as VHDL
Aborted
```

this is found by afl-fuzz

masatake:
Closed by #582.

Issue #541, opened 2015-08-24
Assertion failed on parsing malformed Json file (main/entry.c:1108)


kcwu:
How to reproduce
```
$ echo -e '{"\\\t' > min.json
$ ctags min.json
ctags: main/entry.c:1108: void initTagEntryFull(tagEntryInfo *const, const char *const, unsigned long, const char *, fpos_t, const char *): Assertion `name == NULL || strchr (name, '\t') == NULL' failed.
ctags: main/entry.c:1108: parsing min.json:1 as JSON
Aborted
```

this is found by afl-fuzz

masatake:
Closed by #582.

Issue #540, opened 2015-08-24
Assertion failed on parsing malformed Fortran file (main/entry.c:1108)


kcwu:
How to reproduce
```
$ echo -e 'type;generic"\t' > min.f
$ ctags min.f
ctags: main/entry.c:1108: void initTagEntryFull(tagEntryInfo *const, const char *const, unsigned long, const char *, fpos_t, const char *): Assertion `name == NULL || strchr (name, '\t') == NULL' failed.
ctags: main/entry.c:1108: parsing min.f:1 as Fortran
Aborted
```

this is found by afl.fuzz

masatake:
Solved by #582.

PR #593, opened 2015-10-01
units: add cases derived from bugs found by afl-fuzz

masatake:
@kcwu reported many bugs. They are found by afl-fuzz.
This commit adds many cases converted from the bug reports.

The related issues are:

#543, #542, #541, #540, #539, #538, #537, #536, #532, #531, #530, #529, and #528


masatake:
Let's debug them.

PR #594, opened 2015-10-01
main: remove unused Option.kindLong field and --kind-long option(incompatible change)

masatake:


b4n:
LGTM, but that could be an incompatible change from e-ctags as now u-ctags would fail parsing the option, instead of ignoring it.

masatake:
Yes, but we introduced many (not large) incompatible changes already.
Releasing initial public version of u-ctags is good and rare chance to clean up and to introduce incompatible changes.

b4n:
Makes sense.

PR #595, opened 2015-10-01
Don't allow space in kind name (incompatible changes)

masatake:
This introduces incompatible change between e-ctags.

As written in format.rst, only alphabetical char should be used in name part of 
tagfield:

```
{tagfield}
		See below.

A tagfield has a name, a colon, and a value: "name:value".

* The name consist only out of alphabetical characters.  Upper and lower case
  are allowed.  Lower case is recommended.  Case matters ("kind:" and "Kind:
  are different tagfields).
```

TODO:

* make an error if other than alphabetical char is used in regex parser definition, and
* sanitize kind names  returned from xcmd.


parsers/c.c
```diff
@@ -384,7 +384,7 @@ typedef enum {
 
 static kindOption JavaKinds [] = {
 	{ TRUE,  'c', "class",         "classes"},
-	{ TRUE,  'e', "enum constant", "enum constants"},
+	{ TRUE,  'e', "enumConstant", "enum constants"},
```
b4n: alignment seem broken (no biggie)

parsers/fortran.c
```diff
@@ -227,7 +227,7 @@ static boolean ParsingString;
 
 /* indexed by tagType */
 static kindOption FortranKinds [] = {
-	{ TRUE,  'b', "block data", "block data"},
+	{ TRUE,  'b', "blockData", "block data"},
```
b4n: ditto

b4n:
LGTM

masatake:
@dtikhonov, could you look at my change for perl perser?

dtikhonov:
> @dtikhonov, could you look at my change for perl perser?

@masatake, looks OK.

masatake:
@dtikhonov.
BTW, have you watched the video?
https://www.youtube.com/watch?v=1k4qDMokR-A

dtikhonov:
@masatake, the video looks very cool, thanks!

PR #600, opened 2015-10-02
Allow empty name for tags marked as placehodler

masatake:
I have introduced many broken code about cork+scope-stack.
dts parser uses these new technologies. So u-ctags cannot parse linux kernel source code well.
This commits fix one of them.

See the test case about the bug.


masatake:
(priviate memo: after merging this fix-corrupted-cork-queue comes next.)

PR #602, opened 2015-10-02
main: don't make a tag for an empty string even if AllowNullTag is TRUE

masatake:
Revisit #492, #590.

As suggested by @b4n in #590, a tag for an empty string is not useful.
With this commit ctags doesn't record such tags even if AllowNullTag is TRUE.

In other word, AllowNullTag field just controls  the warning about
the tags for empty strings; ctags doesn't warn if it is TRUE.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -1152,9 +1152,11 @@ extern int makeTagEntry (const tagEntryInfo *const tag)
 	Assert (tag->name != NULL && strchr (tag->name, '\t') == NULL);
 	Assert (getSourceLanguageFileKind() == tag->kind || isSourceLanguageKindEnabled (tag->kind->letter));
 
-	if (tag->name [0] == '\0' && (!getSourceLanguageAllowNullTag()) && (!tag->placeholder))
+	if (tag->name [0] == '\0' && (!tag->placeholder))
 	{
-		error (WARNING, "ignoring null tag in %s(line: %lu)", vStringValue (File.name), tag->lineNumber);
+		if (!getSourceLanguageAllowNullTag())
+			error (WARNING, "ignoring null tag in %s(line: %lu)",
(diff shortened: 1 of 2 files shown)


main/entry.c
```diff
@@ -1152,10 +1152,18 @@ extern int makeTagEntry (const tagEntryInfo *const tag)
 	Assert (tag->name != NULL && strchr (tag->name, '\t') == NULL);
 	Assert (getSourceLanguageFileKind() == tag->kind || isSourceLanguageKindEnabled (tag->kind->letter));
 
-	if (tag->name [0] == '\0' && (!getSourceLanguageAllowNullTag()) && (!tag->placeholder))
+	if (tag->name [0] == '\0')
 	{
-		error (WARNING, "ignoring null tag in %s(line: %lu)", vStringValue (File.name), tag->lineNumber);
-		goto out;
+		if (tag->placeholder)
```
b4n: this could be kept as part of the above `if`: `if (tag->name [0] == '\0' && (!tag->placeholder))`

b4n:
apart from the construction, LGTM.

Another way to go could be fixing readtags to allow an empty tag, but I'm not sure whether we'd consider a 0-length tag name compatible with the historical CTags format.  I guess the question would be whether users handle it correctly; but if readtags itself doesn't I'm afraid many other might not as well.

masatake:
@b4n, what is the "the construction" ? My guessing is "condition". Am I correct?

b4n:
@masatake yeah sorry, I just meant how the code was written as per my remark on the diff, so yeah, the condition.

masatake:
Oh, I see. I will revise the condition.

masatake:
Revised.

b4n:
LGTM, thank you

masatake:
Thank you.

PR #603, opened 2015-10-02
Fix corrupted cork queue

masatake:
There was a critical bug in cork+context(scope) stack code.
Though the stack must be initialized each time the input file is changed, it was not.
For a parser written in --regex-<LANG>= patterns, it is initialized well. However, for a crafted parser with regex, it it not. dts is such a parser.

Related bugs are also fixed in this commits. All these needed to make a tags file for linux kernel source code; dts is heavily used in the source tree.


b4n:
LGTM

PR #605, opened 2015-10-03
Initialize parser lazily

masatake:
* Make "units" a bit faster.
* Regex patterns and keywords can be used declarative way.

### DIFF
--- a/main/parse.h
+++ b/main/parse.h
@@ -58,6 +58,11 @@ typedef struct {
 } tagRegexTable;
 
 typedef struct {
+	const char *name;
+	int id;
+} keywordTable;
+
+typedef struct {
 	/* defined by parser */
 	char* name;                    /* name of language */
 	kindOption* kinds;             /* tag kinds handled by parser */
@@ -76,6 +81,8 @@ typedef struct {
 	boolean allowNullTag;
 	tagRegexTable *tagRegexTable;
 	unsigned int tagRegexCount;
+	keywordTable *keywordTable;
+	unsigned int keywordCount;
 
 	/* used internally */
 	unsigned int id;               /* id assigned to language */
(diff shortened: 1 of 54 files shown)


main/parse.h
```diff
@@ -67,6 +79,10 @@ typedef struct {
 	unsigned int method;           /* See PARSE__... definitions above */
 	boolean useCork;
 	boolean allowNullTag;
+	tagRegexTable *tagRegexTable;
```
b4n: elements should probably be const

main/parse.h
```diff
@@ -67,6 +79,10 @@ typedef struct {
...
+	unsigned int tagRegexCount;
+	keywordTable *keywordTable;
```
b4n: ditto

parsers/ant.c
```diff
@@ -15,22 +15,21 @@
 #include <string.h>
 #include "parse.h"
 
+static tagRegexTable antTagRegexTable [] = {
```
b4n: this and other tables should probably be all static

b4n: sorry, s/static/const/

b4n:
I didn't really check the initialization logic changes, but using tables looks like a nice idea.

masatake:
@b4n, thank you. I added a new commit where I reflected your comment.

vhda:
Is keeping `KIND_COUNT` name so incorrect? Because if we kept it the number of modified files would decrease drastically.

masatake:
@vhda, CURRENTLY I don't want to provide COUNT macros for each data type: KIND_COUNT, REGEX_COUNT, FIELD_COUNT, KEYWORD_COUNT, and FLAG_COUNT. 

Defining macros for each type itself is not bad. However, I don't want to do it now because the header files of ctags are not well organized. When we can split the headers file for each data type, I would like to introduce COUNT macros for each data type.

vhda:
Understood and it makes sense.
Thanks! :)

masatake:
O.k. I will merge this. This makes units faster. 

I really need these commits because I'm using i386(http://ascii.jp/elem/000/000/214/214977/). It takes really long time to run units.

PR #606, opened 2015-10-04
Code cleanup

steveno:
* Remove unused parameter
* Add some missing parens

### DIFF
--- a/main/lxcmd.c
+++ b/main/lxcmd.c
@@ -525,12 +525,15 @@ extern boolean hasXcmdKind (const langType language, const int kind)
 
 #ifdef HAVE_COPROC
 static void printXcmdKind (xcmdPath *path, unsigned int i,
-			   const char* const langName, boolean allKindFields, boolean indent)
+                           const char* const langName, boolean allKindFields, boolean indent)
 {
 	unsigned int k;
 
 	if (!path[i].available)
 		return;
+
+        if (allKindFields && indent)
+            printf ("%s", langName);
 
 	for (k = 0; k < path[i].n_kinds; k++)
 		printKind (path[i].kinds + k, allKindFields, indent);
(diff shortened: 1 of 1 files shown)


steveno:
@masatake I misunderstood what the TODO was referencing. I thought the "warning" was the missing parens.

steveno:
It looks like the build is failing. I'll look into fixing the issues and get back to this. I'll also look into trying to write a test case. Thank you for help/tips!

masatake:
@steveno, please, se my inline comments for your patch. The test failed because you removed a parameter from a function for long regex flag.

main/lxcmd.c
```diff
@@ -826,8 +828,10 @@ static boolean parseExtensionFields (tagEntry *const entry, char *const string,
 
 						}
 					}
-					else
-						; /* TODO: warning */
+					else {
+                                            /*TODO Handle warning*/
+                                            ;
+                                        }
```
b4n: Style doesn't appear consistent here.  Should probably be either left as is (how is it harmful?) or follow common style, something like:
```C
else
{
	/* TODO: Handle warning */
}
```

b4n:
All newly inserted lines should follow the indentation style (e.g. tabs, width=4)

steveno:
@masatake I've made the changes to my patch. 

@b4n I've made the commit message more explicit to answer your question.

steveno:
@masatake I've lost your comment on making the test. Would you be kind enough repost it?

masatake:
@b4n wrote:
> All newly inserted lines should follow the indentation style (e.g. tabs, width=4)

I am afraid that I myself may violate the style.
Could you make an ideal editor setting with
https://www.wireshark.org/tools/modelines.html
?
So people can follow the rule a bit easier.

masatake:
@steveno, see #607.

PR #614, opened 2015-10-07
 Optlib translator

masatake:
Please, don't merge this.
This one is for testing the translator releated code on travis/appveyor.

I'm looking for the way to link data/optlib/*.ctags into ctags itself.
One of benefits is I don't have to think about where these files should be installed.


masatake:
For msvc, we should provide source code which includes C code generated by ctagst.
... it is not easy for me.

masatake:
(Instead of dropping optlib, I would like to merge this before releasing.)

vhda:
@masatake I think this deserves a bit more discussion.
If I understand correctly we'd maintain the original files in `--regex` format and then this script would automatically generate an equivalent C file.
While I see benefits in having such script for initial conversion and creation of that C file, I don't see it's advantage after that initial conversion. If we have the C file, why not update it directly? Aren't we making this more complicated than necessary?
Again, with the script we could stop delivering optlib configuration files in our package (whenever it is ready) and have all parsers in C format, even if using regular expressions. I see this as an advantage.
But maintaining both files, even if one is automatically generated, I think is too much.

masatake:
@vhda, thank you for comments.
We will maintain generated C files.
Because some platforms may not have bash, we have to bundle translated C files to .tar.gz. However, it is like configure script; we will maintain configure.ac instead of configure.

masatake:
Even is a parser is loaded as an optlib via `--options` or is generated by the translator, a test case for the parser must be able to be run without modifying args.ctags... How can I do it? The units test harness must be hacked.

masatake:
I forgot writing benefits.  
> There are two type of source files using regex engine. 1. foo.ctags files under data/optlib and 2. foo.c files under parsers. Here I will focus on 1.

This is rather about 2. Sometimes I update the interface between main part and built-in regex parser like html.c. The latest my work about this area is "initializing  per language regex table lazily". When I introduce or update the interface, I have to make similar modifications on all built-in regex based parsers. It is not small cost for me. The translator will suppress the cost of mine.

cweagans:
@masatake Can you clarify why this needs to be part of the 1.0.0 release?

masatake:
@cweagans, optlib is one of my early work on universal-ctags. 
Files under data/optlib/*.ctags are installed to the system and are loaded to ctags with this mechanism when starting.

However, I got comments about optlib; the directory layout of optlib is too complex.
I had been thinking what was the best layout. However, I could not find a good answer.
In other hand I wrote some *.ctags files. Some of them will be useful for general purpose.
I would like to distribute it with universal ctags.

If I introduce the translator, we can remove some part of optlib; and embed the *.ctags based parsers
into ctags binary. In addition "make install" becomes simple because we don't have to install *.ctags files into system directory.

We can revive optlib after releasing.

masatake:
I will work on this after merging automake branch.

masatake:
This is the next step for introducing automake.
With translator files which were installed to /usr/share/ctags can be embedded into ctags itself.
This simplifies the installation process.

TODO: docs, options.c and units script must be updated.

masatake:
It works!

masatake:
> TODO: docs, options.c and units script must be updated.

I have reviewed my changes on options.c. It is o.k. No more change is needed.
Now /usr/share/ctags/optlib and /usr/share/ctags/preload are gone.
Only /etc/ctags/optlib and /etc/ctags/prelad are used.


I'm thinking about rename /etc/ctags to /etc/ctags.d to align ~/ctags.d.

masatake:
@vhda, I got a question ago.

I would like to answer again. Here are points of introducing the translator.

(1) We maintain .ctags file only. We don't maintain *.c file generated from .ctags file.
(2) For environment where bash is not available, .c files are generated in autogen.sh.
     The generated files are bundled into .tar.gz. So ctags can be built without bash from tar.gz.
(3) For environment where bash is not available, generated .c files are commited to our git repository.
    The generated files are available when doing git clone. So ctags can be built without bash from git repository.

A critical drawback is that a .c file cannot be update when .ctags is modified on the platform where bash is not available.
An advantage is that all ctags executable is needed to use parsers defined in .ctags files.(An exception is a parser using xcmd.)

https://github.com/masatake/ctags/commit/a4ce4c9f07f32d64303ad0588353ab7de5c87e8a

masatake:
The last task is done.
Ready to merge.

masatake:
The next step for 1.0.0 is introducing automake.

PR #617, opened 2015-10-08
Don't require bash in tmain sorted help message

masatake:
See also #609.


vhda:
I like the solution you implemented.
Looks pretty nice!

masatake:
Thank you for reviewing.
Can we close #609, too?

vhda:
Let me update it and I'll merge it afterwards.

PR #630, opened 2015-10-13
fix #629 infinite loop with fortran parser on C file

### LINKED ISSUE #629: infinite loop with fortran parser on C file
How to reproduce
```
./ctags.20151012 --language-force=fortran ctags/Units/parser-c.r/enum.c.d/input.c
```

It's found by forcing language as fortran against universal-ctags unit test files.

QingmingHe:
When parsing C enum, Fortran parse try to find an "end" token in
parseEnumBlock. There is no such key word at end of C enum statement and
the parser falls into infinite loop. To solve it, add an end-of-file
test when trying to find "end" token. If reach end of file, jump out of
the loop.

### DIFF
--- a/Units/parser-fortran.r/bug629.d/args.ctags
+++ b/Units/parser-fortran.r/bug629.d/args.ctags
@@ -1,0 +1,1 @@
+--language-force=Fortran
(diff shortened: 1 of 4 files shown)


b4n:
LGTM

b4n:
Just maybe we want an explicit test case for this?  not sure it's needed as we can already reproduce it relatively easily with fuzz

masatake:
It is better to have a test case.
@QingmingHe, could you add it Units/parser-fortran.r ?

masatake:
LGTM.

Issue #627, opened 2015-10-13
assertion failed on parsing malformed Perl file (perl.c:196)


kcwu:
How to reproduce
```
$ echo 'use constant{,=>' > assert.pl
$ ./ctags.20151012 --language-force=perl assert.pl
ctags: parsers/perl.c:196: makeTagFromLeftSide: Assertion `e - b + 1 > 0' failed.
ctags: parsers/perl.c:196: parsing assert.pl:1 as Perl
Aborted
```

this is found by afl-fuzz

masatake:
A test case is added.

masatake:
Thank you for reporting this.

PR #631, opened 2015-10-14
Do not assert on malformed input.

dtikhonov:
This fixes issue #627.


masatake:
Could you add the original test case to Units/afl-fuzz.r ?

PR #636, opened 2015-10-14
vera: implement interface kind

### LINKED ISSUE #628: Assertion failed on parsing malformed Vera file (parsers/c.c:1050)
How to reproduce
```
$ echo 'typedef interface c e;' > assert.vr
$ ./ctags.20151012 --language-force=vera assert.vr
ctags: parsers/c.c:1050: veraTagKindFull: Assertion `"Bad Vera tag type" == NULL' failed.
ctags: parsers/c.c:1050: parsing assert.vr:1 as Vera
Aborted
```

this is found by afl-fuzz

masatake:
Close #628.

#628 reports interface kind of vera is implemented
incompletely.

This commit implements it. As the result the assertion
failure reported in #628 is fixed.

To implement interface kind, signal kind is also
defined. (signals are declared inside interface.)

TODO: scope field for signal kind is not implemented yet.

### DIFF
--- a/Units/afl-fuzz.r/github-issue-628-vera.d/input.vr
+++ b/Units/afl-fuzz.r/github-issue-628-vera.d/input.vr
@@ -1,0 +1,1 @@
+typedef interface c e;
--- a/Units/fuzz-hitting-assertions.r/vera-fha-0.d/expected.tags
+++ b/Units/fuzz-hitting-assertions.r/vera-fha-0.d/expected.tags
@@ -1,1 +1,1 @@
-iA	input.vr	/^interface iA {/;"	I
+iA	input.vr	/^interface iA {/;"	i
--- a/Units/fuzz-hitting-assertions.r/vera-fha-0.d/expected.tags
+++ b/Units/fuzz-hitting-assertions.r/vera-fha-0.d/expected.tags
@@ -1,0 +1,1 @@
+iA	input.vr	/^interface iA {/;"	I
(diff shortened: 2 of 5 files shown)


vhda:
I never noticed that we had Vera support in ctags.
It might make sense to move Vera parsing into `verilog.c` because the languages should be similar. For example, the SystemVerilog parser already supports the interface definition you're implementing in this PR.

How urgent are the changes you're implementing? I need to find some free time to see what are the differences between Vera and SystemVerilog and to analyze the source code to determine what is the best course of action.

vhda:
BTW, @masatake could you change the type of input/output/inout signals to "p" (as in "ports)?

masatake:
Sure.

vhda:
And "interface" to "I".

masatake:
Sigh. p is reserved to "program".

masatake:
I would like to commit my version as a temporary solution.
Could you take over after merging?

vhda:
I'm just worried about the kind conflicts, which will bring backward incompatibilities :(
Nonetheless, go ahead and merge the PR. As you have noticed, the tags output is the same as what SystemVerilog is finding, so I don't think there is anything wrong with it.

masatake:
@vhda, thank you for comment.
I have changed the letter for kind interface from 'i' to 'I'.
I don't touch 's', the letter for signal kind now.

vhda:
@masatake I've had the opportunity to look into Vera a bit more in detail and I can see that it's structure takes a lot from C. For example, instead of the `function ... endfunction` pair, it uses standard brackets `function { ... }` to delimit a function implementation. The same is true for other constructs.
Because of this, I think it makes more sense to include it in the C parser.

As such, there is no reason to try to use the same kind letters that are used in the Verilog parser.

masatake:
@vhda, thank you.
I use 'i' for interface kinde again. scope of interface is also implemented.

PR #640, opened 2015-10-20
main: omit patterns for long input lines

### LINKED ISSUE #163: Limit the output pattern entry length
When outputting patterns as excmds, there is no maximum pattern length.  This is OK for normal source files when all lines have a reasonable length (like 120 columns tops and often less), but inappropriate when the lines are overly long, like with minified JavaScript or CSS.

Such files generally contain one single line, line that can be *very* long (like 30k characters).  This results in overly large tag files, yet pretty much useless patterns as all tags have the same one.

I think that the pattern length should be limited to some sane lengths, like 80 characters.  If needed, an option could probably be introduced, but I would believe that 80 characters is unique enough to be useful.

Ideally, and if users of the pattern excmd can use it, the pattern could be truncated around the actual location of the tag on the line rather than only keeping the start, but that's probably only fancier.

masatake:
Too long input lines may make tags file larger.
This commits omits the length of pattern field printed to a tags file.

If a making pattern is longer than the value of Option.patternLengthLimit,
the pattern is omitted; instead of including whole the input line, the
rest of substring after nth byte of Option.patternLengthLimit is is
printed as \.\{N}. Here N is the length of the substring.

As far as testing vim accepets \.\{N} as part of a pattern.

Close #163.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -609,13 +609,14 @@ extern void endEtagsFile (const char *const name)
  *  Tag entry management
  */
 
-static size_t appendSourceLine ( void putc_func (char , void *), const char *const line, void * data)
+static size_t appendSourceLine ( void putc_func (char , void *), const char *const line, void * data, boolean *omitted)
 {
 	size_t length = 0;
 	const char *p;
 
 	/*  Write everything up to, but not including, a line end character.
 	 */
+	*omitted = FALSE;
 	for (p = line  ;  *p != '\0'  ;  ++p)
 	{
 		const int next = *(p + 1);
@@ -624,6 +625,11 @@ static size_t appendSourceLine ( void putc_func (char , void *), const char *con
 		if (c == CRETURN  ||  c == NEWLINE)
 			break;
 
+		if (length >= Option.patternLengthLimit)
+		{
+			*omitted = TRUE;
+			break;
+		}
 		/*  If character is '\', or a terminal '$', then quote it.
 		 */
 		if (c == BACKSLASH  ||  c == (Option.backward ? '?' : '/')  ||
@@ -635,6 +641,7 @@ static size_t appendSourceLine ( void putc_func (char , void *), const char *con
 		putc_func (c, data);
 		++length;
 	}
+
 	return length;
 }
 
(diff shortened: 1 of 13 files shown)


masatake:
This commit doesn't include my work for optimizations reported in #163.

main/entry.c
```diff
@@ -635,6 +640,21 @@ static size_t appendSourceLine ( void putc_func (char , void *), const char *con
...
 		++length;
 	}
+
+	if (length > Option.patternLengthLimit)
+	{
+		char lenpat[32];
+		char *l;
+		int r;
+
+		r = snprintf (lenpat, 32, "\\.\\{%lu}",
```
vhda: Now I understand what you were trying to do using patterns in the tags file.
As a KIS(S) alternative, have you considered simply omitting the remainder of the line and dropping `$` instead of adding this complex pattern?

masatake:
Thank you for comment.

>  have you considered simply omitting the remainder of the line and dropping $ instead of adding this complex pattern?

Do you mean that the pattern terminator  `\{N}` is not needed?
If yes, how about `.*`?

vhda:
Yes, the .* pattern is not required. You can simply break the line at any column you wish as long as you drop the $ terminator.

masatake:
@vhda, thank you. I will update my commit as you wrote.

masatake:
Updated.

main/entry.c
```diff
@@ -635,6 +641,7 @@ static size_t appendSourceLine ( void putc_func (char , void *), const char *con
 		putc_func (c, data);
 		++length;
 	}
+
 	return length;
```
vhda: The length value is returned here, so why not use it at top level to decide if `$` should be appended, instead of passing `omitted` all across the hierarchy?

masatake:
In my understanding the caller cannot know whether the pattern is omitted or not from the length.
For explanation consider  Option.patternLengthLimit = 4.
Let's consider following input line:
```
var\
```
appendSourceLine generates following pattern for the input:
```
var\\
```
The backslash is doubled for escaping purpose.
The length of pattern is 5. It is larger than Option.patternLengthLimit.
However, nothing is omitted in thie pattern generation.
In this case FALSE is set to omitted parameter.

vhda:
But in that scenario you are also setting `omitted` to TRUE, aren't you?

		if (length >= Option.patternLengthLimit)
		{
			*omitted = TRUE;
			break;
		}

masatake:
After processing the last backslash in the input line,  appendSourceLine emits two backslashes.
Then the control goes back to the top of for loop again. However, the control doesn't go to
```
if (length >= Option.patternLengthLimit)
    {
        *omitted = TRUE;
```
because the next character in the input line (*p) is '\0'. The control left the loop. As the result
omitted keeps FALSE.

If I understand my code correctly :-P

vhda:
You're correct, obviously! :-P
I was trying to avoid the extra argument in the functions, but I really do not have any alternatives.
In that case, I have nothing more to had. Sorry for all the fuss.

masatake:
@vhda, thanks.
This is releated to tags file format. So review has been really needed.

PR #643, opened 2015-10-26
main: Don't ignore `.' extra entry when `f' is specified together

masatake:
I extended --extra option(.) for printing the total number of lines of
input file in c5d6954569123.

However, I injected a bug in 98392b01.  When specifying `.' with `f',
the total number of lines is not printed. This commit fixes it.

### DIFF
--- a/Units/extra-total-lines.d/args.ctags
+++ b/Units/extra-total-lines.d/args.ctags
@@ -1,3 +1,4 @@
---extra=+.
+# `f' is optional
+# There was a bug that `.' was disabled when `f' was specified together.
+--extra=+.f
 --excmd=number
-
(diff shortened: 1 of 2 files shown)


Units/extra-total-lines.d/args.ctags
```diff
@@ -1,3 +1,4 @@
---extra=+.
+# `f' is optional
+# There was a bug that `.' was disabled when `f' was specified togeter.
```
vhda: "together" (missing h)

vhda:
LGTM

masatake:
I must run spell checker anytime pushing commits.

vhda:
;)

Issue #140, opened 2014-11-23
Extended regression environment


vhda:
Build a script that is able to download a set of files from the internet, place them under "Units" and generate an expected.tags file using an older version of ctags binary. There should be a configuration file that lists the tar balls that need to be downloaded, using a format like:

```
language;http://link/to/file.tar.gz
```

This is intended only to find regressions that may not be covered by a specific test unit.
Note: a new test unit must be manually created if such a regression is found.

masatake:
I'm working on this issue very hard these days.

https://github.com/masatake/srpmix7

It is not ready for public use. When it is ready, I will report it here.

masatake:
Now srpmix7 can expand dsc/deb source.
After some code clean-up, I will study how to utilize it with ctags development.

masatake:
To think about performance of parsing, we need large code base.
I will revise `--totals` option. I guess I have broken the option in past development.

masatake:
See https://github.com/universal-ctags/codebase
I designed "codebase" for performance measurement.
However, we can extend it to an environement and target for regression test.

PR #644, opened 2015-10-28
Optimize pattern generation

masatake:
Making a pattern takes long time for long input line.  This commit
shortens the time. See the second commit about evaluation of my technique.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -609,7 +609,12 @@ extern void endEtagsFile (const char *const name)
  *  Tag entry management
  */
 
-static size_t appendSourceLine (void putc_func (char , void *), const char *const line, void * data, boolean *omitted)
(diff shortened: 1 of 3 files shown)


main/entry.c
```diff
@@ -88,7 +88,8 @@ tagFile TagFile = {
 	    .queue = NULL,
 	    .length = 0,
 	    .count  = 0
-    }
+    },
+    .patternCacheValid = FALSE,
```
vhda: Strange indentation

vhda: Ignore me! Everything is fine! :)

masatake:
Commit log messages are modified.

masatake:
Now, ctags can handle very long single line JavaScript file.
However, there is not test case for such file. #140 must be solved.

PR #645, opened 2015-10-28
xformat: allow user customizing -x output

masatake:
The output format of -x is statically defined.
This removes the restriction.

TODO:
1. scope field cannot be handled well yet. This needs much more work.
2. no documentation yet.


masatake:
I forgot to add windows releated files.

masatake:
Just rebased on master.

masatake:
Though I  wrote "--xformat" in log messages, "--_xformat" is the actual name of opton.

PR #649, opened 2015-10-29
python: check the enablement of "member" kind in makeFunctionTag

### LINKED ISSUE #648: Bug found in --python-kinds option while generating a tag file
_From @yevhen-m on October 29, 2015 10:5_

Hey, I use ctags to generate tag files for python code.
I get an error, if I use --python-kinds=cf option
f stands for 'function' and is a valid python kind

here is a screenshot
screenshot from 2015-10-28 18 30 45

Will be grateful, if you fix this issue ))

_Copied from original issue: universal-ctags/universal-ctags.github.io#13_

masatake:
In makeFunctionTag function, tags both member and function kinds are
made. If the kinds are disabled, the tags should not be made.

At the head of makeFunctionTag, the enablement of function kind is
checked but that of member is not checked. As the result, an assertion
is made failure as reported in #648 by @yevhen-m.

This commit fixes this; the enablement of "member" kind is also
checked in makeFunctionTag.

Close #648.

### DIFF
--- a/Units/python-disable-member-kind.d/args.ctags
+++ b/Units/python-disable-member-kind.d/args.ctags
@@ -1,0 +1,1 @@
+--python-kinds=f
--- a/Units/python-disable-member-kind.d/input.py
+++ b/Units/python-disable-member-kind.d/input.py
@@ -1,0 +1,6 @@
+class A:
+    def m():
+        pass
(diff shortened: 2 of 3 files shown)


blueyed:
:+1:

PR #650, opened 2015-11-01
c: add signature field to C preprocessor macros

masatake:


vhda:
Nice work @masatake

PR #678, opened 2015-11-15
docs: update the list of supported languages

masatake:
(Use the output of --list-languages directly.)


masatake:
Thank you. Quite comfortable about this work flow.

Issue #402, opened 2015-06-28
line directive and extra language


masatake:
During previewing 
Units/review-needed.r/line_directives.c.d/input.c
I got a coredump.

```
[yamato@x201]~/var/ctags-github% cat /tmp/input.c 
#line 2 f.f
#define x
[yamato@x201]~/var/ctags-github% ./ctags -o - --fields='+l' --line-directives /tmp/input.c 
ctags: main/parse.c:1279: isLanguageKindEnabled: Assertion `kindOpt = langKindOption (language, kind)' failed.
zsh: abort (core dumped)  ./ctags -o - --fields='+l' --line-directives /tmp/input.c
```

With `#line` directive, ctags changes the values of global variables for input file name and its language. In the above example the language is changed from c to fortran.
Though the values are changed, C parser continues parsing and making a tag.
The kind of a tag created by c parser is verified by comparing with fortran's kind table.
`x` has a kind `d`, macro definition while fortran parser doesn't know `d` .
This gap makes the assertion failure.

I was afraid that a code I introduced ago triggers this assertion. However, it does not.
The exuberant ctags doesn't have the assertion line in its source code.
 As the result it generates a bit strange  tags file.

```
[yamato@x201]~/var/ctags-github% ctags -o - --fields='+l' --line-directives /tmp/input.c 
ctags: Warning: Ignoring non-option in ./.ctags

x	/tmp/f.f	2;"	d	language:Fortran	file:
```

Though the language is Fortran, a kind `d`, which Fortran parser doesn't know, is used.

This is a blocker for moving 
Units/review-needed.r/line_directives.c.d to Units/parser-c.r/.

QingmingHe:
Fortran macro definition is much like that in C (or even the same? Gfortran pre-process Fortran code via **--cpp** option). **d kind** should be supported by ctags but not implemented yet. It's on my schedule...

I've no idea how to implemented yet. Maybe through optlib in the first step? But the output tags is duplicate with **--regex-Fortran** options.
```
qmhe@pu:/tmp$ ~/codes/ctags/ctags -o - --fields='+l' --line-directives --regex-Fortran="/^# *define +([[:alnum:]_]+)/\1/d,definitions/" input.c 
x       f.f     /^#define x$/;" d       language:Fortran
x       f.f     /^#define x$/;" d       language:Fortran        file:
```

masatake:
[...]

PR #680, opened 2015-11-15
Add reference tag

masatake:
Traditionally ctags collects the information for locating where an
object having name is DEFINED.

In addition Universal-ctags supports reference tags. If ``r`` extra
tag is enabled, universal-ctags collects the information for locating
where an object having name is REFERENCED. This feature is proposed
by @shigio on #569 for GNU GLOBAL.

Let me show some examples. Here is the target input file named reftag.c.

.. code-block:: c

    #include <stdio.h>
    #include "foo.h"
    #define TYPE point
    struct TYPE { int x, y };
    TYPE p;
    #undef TYPE

Traditionally output:

.. code-block:: console

    $ ./ctags -o - reftag.c
    TYPE	reftag.c	/^#define TYPE /;"	d	file:
    TYPE	reftag.c	/^struct TYPE { int x, y };$/;"	s	file:
    p	reftag.c	/^TYPE p;$/;"	v
    x	reftag.c	/^struct TYPE { int x, y };$/;"	m	struct:TYPE	file:

Output with enabling ``r`` extra tag:

.. code-block:: console

    $ ./ctags --list-extras | grep ^r
    r	Include reference tags	off
    $ ./ctags -o - --extra=+r reftag.c
    TYPE	reftag.c	/^#define TYPE /;"	d	file:
    TYPE	reftag.c	/^#undef TYPE$/;"	d	file:
    TYPE	reftag.c	/^struct TYPE { int x, y };$/;"	s	file:
    foo.h	reftag.c	/^#include "foo.h"/;"	h
    p	reftag.c	/^TYPE p;$/;"	v
    stdio.h	reftag.c	/^#include <stdio.h>/;"	h
    x	reftag.c	/^struct TYPE { int x, y };$/;"	m	struct:TYPE	file:

`#undef X` and two `#include` are newly collected. Reference tags may
have "role" information representing how it is
referenced. Universal-ctags print the role information when `r` field
is enabled with ``--fields=+r``. (If a tag doesn't have no specialized
role, `generic` is used as the name of role.)

.. code-block:: console

    $  ./ctags -o - --extra=+r --fields=+r reftag.c
    TYPE	reftag.c	/^#define TYPE /;"	d	file:
    TYPE	reftag.c	/^#undef TYPE$/;"	d	file:	role:undef
    TYPE	reftag.c	/^struct TYPE { int x, y };$/;"	s	file:
    foo.h	reftag.c	/^#include "foo.h"/;"	h	role:local
    p	reftag.c	/^TYPE p;$/;"	v
    stdio.h	reftag.c	/^#include <stdio.h>/;"	h	role:system
    x	reftag.c	/^struct TYPE { int x, y };$/;"	m	struct:TYPE	file:

`Reference tag marker` field is specialized to GNU global requirement; D is used
for the traditional definition tags, and R is used for the new reference tags.
The field can be used only in ``--_xformat`` option.

.. code-block:: console

    $ ./ctags -x --_xformat="%R %-16N %4n %-16F %C" --extra=+r reftag.c
    D TYPE                3 reftag.c         #define TYPE point
    D TYPE                4 reftag.c         struct TYPE { int x, y };
    D p                   5 reftag.c         TYPE p;
    D x                   4 reftag.c         struct TYPE { int x, y };
    R TYPE                6 reftag.c         #undef TYPE
    R foo.h               2 reftag.c         #include "foo.h"
    R stdio.h             1 reftag.c         #include <stdio.h>

Though the facility for collecting reference tags is implemented, only
few parsers utilized it now. All available roles can be listed with
``--list-roles`` option:

.. code-block:: console

    $ ./ctags --_list-roles
    C	d	undef	undefined	on
    C	h	system	system header	on
    C	h	local	local header	on
    C++	d	undef	undefined	on
    C++	h	system	system header	on
    C++	h	local	local header	on
    DTS	d	undef	undefined	on
    DTS	h	system	system header	on
    DTS	h	local	local header	on
    Make	I	generic	non-categorized generic role	on
    Make	I	optional	included as an optional makefile	on
    Sh	s	generic	non-categorized generic role	on
    Vera	d	undef	undefined	on
    Vera	h	system	system header	on
    Vera	h	local	local header	on

The first column shows a name of parser.
The second column shows a name of kind.
The third column shows a name of role.
The fourth column shows description of the role.
The first column shows whether the role is enabled or not.
Currently ctags doesn't provide the way for disabling a
specified role.

----

* About documentation more work is needed based on #569.
* reftag code for scheme language is not merged yet.
* reftag for python import directive must be implemented.
* @FOO@ can be reftag. This should be handled in read.c.
* reftag for java import directive will be useful.
* FILE in `#line LINENUM FILE` can be captured as a reftag. See #402.


masatake:
@vhda, I will wait for your ack. As far as I recnognized, there is no serious compatibility issues between exuberant-ctags.

compatibility issues between exuberant-ctags
I'm afraid that I introduced new contradictions in conception level.
But in some area I found reference tags and roles solve
old contradictions. See #402 for example.

vhda:
After reviewing everything I see no problems merging this change.

I do have one question: in future you also intend to support other "reference roles" like variable assignments, for example?

It is strange to me to have ctags parse references to tags because in general if we want to search for a tag we will be **at** the reference location. But then again, GLOBAL appears to use this information, so I see no reason not to provide this functionality when we have someone using it.

Unfortunately, this feature is not going in the direction of #80 as I originally thought when we started discussing the "references" topic.

I will reassign this back to you, as I have not analyzed the code in enough detail to feel comfortable to merge it myself.

masatake:
@vhda, thank you for reviewing.

> I do have one question: in future you also intend to support other "reference roles" like variable assignments, for example?

Yes. However, implementing such fine-grained roles makes developing a parser harder. So I predefined "generic role".

> It is strange to me to have ctags parse references to tags because in general if we want to search for a tag we will be at the reference location. But then again, GLOBAL appears to use this information, so I see no reason not to provide this functionality when we have someone using it.

Your recognition is correct. Reference tags are not useful for classical tools. The classical tools utilize tags files for completion, jumping-to-definition and navigation.

 
Reference tags are for new tools like GLOABL. I image these tools may utilize tags file for querying.
The tool may answer following kind of questions:
* which function changes this global variable?
* which functions calls this function?
Reference tags is for these question. Actually GLOBAL answers these questions.

> Unfortunately, this feature is not going in the direction of #80 as I originally thought when we started discussing the "references" topic.

Reference tags will contribute improving the quality of the "hints".

> I will reassign this back to you, as I have not analyzed the code in enough detail to feel comfortable to merge it myself.

vhda:
Great job @masatake and very good documentation section.

masatake:
Thanks.

bam80:
Thanks for the addition.
Do we have any Vim plugins making use of the new reference tag type?

PR #686, opened 2015-11-17
Roundtrip test

### LINKED ISSUE #685: How to test readtags.c?
I am running into a problem with readtag.c

I use this in my notepad++ plugin but when run it on this file, it crashes my plugin.
https://github.com/notepad-plus-plus/notepad-plus-plus/blob/master/scintilla/test/examples/x.cxx

readtags has a problem parsing the "M" tag that gets generated, most likely because of the "\" at the end of the search pattern. Note that this x.cxx file compiles just fine.

I am still struggling to build readtags.exe with any of the Windows compilers. So debugging the exact problem is a bit hard ATM, but I will get there :-)

I want to know if we have a way to test this readtags parser? If not, any thought how we can create proper tests for readtags?

masatake:
Close #685.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -372,6 +372,7 @@ commit-stamp: FORCE
 # check
 #
 CTAGS_TEST = ./$(CTAGS_EXEC)
+READ_TEST = ./$(READ_CMD)
 ifdef VG
 VALGRIND=--with-valgrind
 endif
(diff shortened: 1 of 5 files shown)


masatake:
@ffes, could you try this?

masatake:
...I should provide the way to run this under valgrind.

masatake:
Whether should I use "--" or "-" as a separator between options and name arguments?

ffes:
I just tried your adjusted readtags.c and it fixes the problem! Thanks!!!

ffes:
Just tested with the new readtags.c on two other (rather large) source trees at the office where I had similar problems, but didn't had/took time to investigate the cause of the problems. They both now work as expected as well, so apparently it was the same problem. So again, thanks! :-)

PR #694, opened 2015-11-27
Repair --line-directive option

### LINKED ISSUE #402: line directive and extra language

masatake:
Close #402.

Universal-ctags --line-directive option is broken as reported in #402.
This patch set fixes it. Though the many areas are changed, the patch set
just makes the behaviour of u-ctags with --line-directive option as the same as
e-ctags. Nothing interesting feature like adding new role for reference tags is added here.

However, in this patch set, macros, types, and functions  in read.h are reorganized.
See docs/internal.rst for more details.


masatake:
@k-takata, I have a question.

Is there way to detect whether a test is run on Appveyor or not from an environment variable or something?

About on Travis TRAVIS is defined. So I can use it in Makefile like:

```
ifdef TRAVIS
SHOW_DIFF_OUTPUT=--show-diff-output
endif
```

I need the same on Appveyor.

masatake:
Maybe I should modify appveyor.bat.

k-takata:
http://www.appveyor.com/docs/environment-variables
`APPVEYOR` is set to `True`.

masatake:
@k-takata, thank you. I will withdraw acd188f when merging.

masatake:
Finally I got the diff output of failed case.

```
+ sed -e 's/./	\0/' /c/projects/ctags/Units/parser-c.r/line_directives.c.d/DIFF.tmp
	--- ./Units/parser-c.r/line_directives.c.d/expected.tags	2015-11-27 07:49:18.369964400 +0000
	+++ /c/projects/ctags/Units/parser-c.r/line_directives.c.d/FILTERED.tmp	2015-11-27 07:53:14.307741300 +0000
	@@ -1,5 +1,5 @@
	-a	a.c	10;"	v	language:C
	-b	b.c	20;"	v	language:C
	-c	c.c	30;"	v	language:C
	-d	OK.zzz	2;"	v	language:NEWLANG
	-e	OK.y	100;"	v	language:YACC
	+a	line_directives.c.d\\a.c	10;"	v	language:C
	+b	line_directives.c.d\\b.c	20;"	v	language:C
	+c	line_directives.c.d\\c.c	30;"	v	language:C
	+d	line_directives.c.d\\OK.zzz	2;"	v	language:NEWLANG
	+e	line_directives.c.d\\OK.y	100;"	v	language:YACC
+ echo
 
+ '[' 5 = 0 ']'
```

masatake:
@k-takata, @ffes, on  "Environment: compiler=msys2, MSYS2_ARCH=x86_64, MSYS2_DIR=msys64, MSYSTEM=MINGW64" what kind of file separator should ctags use when emitting a file path to tags file?
/ or \\ ? 

Search "RAWOUT" in https://ci.appveyor.com/project/masatake/ctags/build/1.0.351/job/jkllof55cpun23cm 
New code it seems that new code I introduced uses both / and \\.

masatake:
old-docs/NEWS.exuberant:
```
* Added compile time option to separate path components with a unix-style path
  separator for sharing tag file across platforms, enabled by defining the
  label UNIX_PATH_SEPARATOR [Win32].

```

masatake:
The separator issue is not new in this patch set.
I decide to inject a filter as a workaround. 
https://github.com/masatake/ctags/blob/repair-line-directive-option/Units/parser-c.r/line_directives.c.d/filter

PR #696, opened 2015-11-27
units: don't remove 'r' suffix from category name when displaying the name

masatake:
Just for improving the appearance of units output, I removed 'r' suffix
from categories of units test cases. This is just inconvenient when
one wants to access to a result of a failed test case.


b4n:
Untested, but I agree stripping the `.r` is cute but inconvenient, and changes look reasonable.

PR #700, opened 2015-12-01
Change the build system to Automake

masatake:
Just for building on travis and appveyor.
misc/dist-test-cases are used for gathering the name of test releated files to  EXTRA_DIST.

Many parts are rearranged from the original @mike-burns 's verison.

Close #132.


masatake:
Checks on ci are passed. We can start more work from here.

masatake:
make dist works fine but make distcheck is failed.

masatake:
As far as checks on travis and appvoyer are passed, I would like to merge changes because this is the bottle neck of merging other PRs.

masatake:
I don't want to install readtags.o and readtags.h in our initial released version(1.0).
(readtags command will be installed.)

Any objection?

masatake:
I will try to merge the translator related commits first.
So Makefile.am can simplified.

I will merge commits up to 14d0731.
The I will try to merge the translator related commits (#614).
Then I will merge commits after 14d0731.

Two issues I would like to fix before merging:
* making make distcheck work, and
* embeding git id if make run in a git repository.

So I would like to you to review up to 14d0731(Use automake) first.

masatake:
I have found more commits after "Use automake" " can move before "Use automake".

masatake:
data/optlib/default.ctags must be generated at build-time.
Anyway commits up to "Use automake" may be ready to merge.

masatake:
Rebased.
TODO:
1. installing readtags.o and readtags.h,
2. cleanup configure.ac, and
3. embedding git id.

Anyway the original automake patch is much simplified.
Soon we can do make dist!

masatake:
I think these are ready to be merged.

masatake:
I don't embed the commit id when building a tagged version as suggested by @k-takata.
I should embed the tag instead but I don't have enough energy to complete it.

masatake:
@k-takata, could you look at c409ec9 ?
I updated my commits based on your comment.

k-takata:
Currently, `PROGRAM_VERSION` is defined as "Development" in ctags.h.
I think it will be changed to "1.0.0" when 1.0.0 is released and we will have a tag "1.0.0" or something like that.
In this case, `./ctags --version` will output `Universal Ctags 1.0.0(1.0.0), ...`, I think.
Is this an intended behavior?

masatake:
Oh, I see. I will try to fix it.

masatake:
@k-takata, updated.
We have to rethink about how and where the version should be defined.
It we just use configure, we should define the version in configure.ac. Very simple.
However, we have to deal with the platforms where configure is not used.
Further more we have to make a tag on git repository when releasing....

masatake:
I compare PACKAGE_VERSION and ctags_repoinfo when printing the version information.

I have to say it is not a good idea. I cannot find a good alternative. This is very related to our release process.

Makefile.am
```diff
@@ -0,0 +1,75 @@
...
+MAIN_HEADS += $(REGEX_HEADS)
+endif
+
+ctags_CPPFLAGS = -I. -I$(srcdir) -I$(srcdir)/main
+ctags_CPPFLAGS+= $(DEBUG_CPPFLAGS)
+ctags_CPPFLAGS+= $(FNMATCH_CPPFLAGS)
+ctags_CPPFLAGS+= $(REGCOMP_CPPFLAGS)
+ctags_CPPFLAGS+= -DHAVE_REPOINFO_H
+
+ctags_CFLAGS   = -g -O2
```
b4n: this probably shouldn't be forced, should it?

Makefile.am
```diff
@@ -0,0 +1,75 @@
...
+ctags_CFLAGS  += $(COVERAGE_CFLAGS)
+ctags_CFLAGS  += $(CGCC_CFLAGS)
+
+nodist_ctags_SOURCES = $(REPOINFO_HEADS)
+BUILT_SOURCES = $(REPOINFO_HEADS)
+CLEANFILES = $(REPOINFO_HEADS)
+$(REPOINFO_OBJS): $(REPOINFO_SRCS) $(REPOINFO_HEADS)
+if BUILD_IN_GIT_REPO
+GEN_REPOINFO = $(srcdir)/misc/gen-repoinfo
+$(REPOINFO_HEADS): $(srcdir)/.git/*
```
b4n: should also probably depend on `$(GEN_REPOINFO)`

Makefile.am
```diff
@@ -0,0 +1,75 @@
...
+	$(GEN_REPOINFO) > $@
+endif
+if !BUILD_IN_GIT_REPO
```
b4n: AFAIK Automake supports `else`

Makefile.am
masatake: I wonder why I forgot this. Thank you.

masatake:
@b4n, thank you. I will make a PR based on your comment.

PR #702, opened 2015-12-03
miscellaneous build script cleanups derived from @mike-burns's automake PR

masatake:
Looking carefully the PR, more than half of commits are independent from
automake; even if we don't introduce automke, the commits are still meaningful.


masatake:
@mike-burns 	gnu_regex requires __USE_GNU

this commit is needed in  Environment: compiler=msys2, MSYS2_ARCH=x86_64, MSYS2_DIR=msys64, MSYSTEM=MINGW64 .

PR #704, opened 2015-12-03
Docs: Use the RTD theme if it is installed locally

ffes:
Based on RTD FAQ and the RTD-theme install instruction
http://read-the-docs.readthedocs.org/en/latest/faq.html
https://github.com/snide/sphinx_rtd_theme (`git or download` section in README)

If the RTD theme is not found the default theme of Sphinx theme is use (as before)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,6 +14,9 @@
 import sys
 import os
 
+# Check if we run on RTD or locally
+on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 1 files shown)


masatake:
@ffes, the change itself is obviously good.

I have a question. I am still updateing docs/*.rst randomly via pull requests. Is it o.k.?
Do you have any requirement? .e.g I wonder you want to set a period for conversion of document system?
In such period should we not commit changes to docs/*rst?

ffes:
When we need to do the big(ger) updates to the docs that are required, we have to have a period of freezing the docs dir, but at the moment just update the docs as well as you do!

masatake:
@ffes, I see. Thank you.

PR #712, opened 2015-12-08
Remove dctags target from all makefiles

k-takata:
maintainer.mak has been already removed with the commit
5dc45359a022626d4f5f3aa89f8d232cda69bcfc, and the dctags target has been
also removed. It is better to remove the dctags target also from
mk_mingw.mak and mk_mvc.mak.

If you want to build debug version, use the following commands:
  MinGW: make -f mk_mingw.mak DEBUG=1
  MSVC:  nmake -f mk_mvc.mak DEBUG=1

### DIFF
--- a/mk_mingw.mak
+++ b/mk_mingw.mak
@@ -12,16 +12,19 @@ OBJEXT = o
 ALL_OBJS += $(REGEX_OBJS)
 ALL_OBJS += $(FNMATCH_OBJS)
 VPATH = . ./main ./parsers
+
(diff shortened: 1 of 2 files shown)


masatake:
Looks good.

Could you add
```
If you want to build debug version, use the following commands:
MinGW: make -f mk_mingw.mak DEBUG=1
MSVC: nmake -f mk_mvc.mak DEBUG=1
```
to docs/windows.rst ?

PR #713, opened 2015-12-10
Add PHP7 support and other fixes

b4n:


masatake:
Thank you. I'm reviewing them.
Could you move new test cases to Units/parser-php.r which I have added just now?

b4n:
> Could you move new test cases to Units/parser-php.r which I have added just now?

I can, but I'd rather move all the relevant ones to the subdiretory afterward.  But I don't really mind, I can do as you prefer.

masatake:
I also don't mind the order;-)
Please, do "the move" any time you are o.k.
I'll review tomorrow JST. I'm sleeply.

b4n:
BTW, I'd like to add https://github.com/b4n/geany/commit/f481587de79ba14f7ee10f5b9d49cf8e0b8245f7 to u-ctags too, but currently it doesn't seem possible to report function return type: we don't have `varType`, and `typeRef`, while similar, is not really usable (as there is nothing to put in `typeRef[0]`, and it's required).

I thought I remembered a discussion about this fields, but I can't seem to find it, so maybe I dreamed it.

masatake:
(I opend #715 for the interesting topic.)

masatake:
I will work on making slot for recording "return type".

PR #719, opened 2015-12-14
Units, C++: add a case reproducing a bug about wrong scope kind

masatake:
"class" is used as the kind of parent scope where "namespace" should
be used.

### DIFF
--- a/Units/parser-cpp.r/namespace-and-scope.b/args.ctags
+++ b/Units/parser-cpp.r/namespace-and-scope.b/args.ctags
@@ -1,0 +1,2 @@
+--kinds-C++=*
+--fields=+K
--- a/Units/parser-cpp.r/namespace-and-scope.b/expected.tags
+++ b/Units/parser-cpp.r/namespace-and-scope.b/expected.tags
@@ -1,0 +1,3 @@
+X	input.cpp	/^namespace X {$/;"	namespace	file:
+f	input.cpp	/^	void f (void);$/;"	prototype	namespace:X	file:
+f	input.cpp	/^void X::f (void) {}$/;"	function	namespace:X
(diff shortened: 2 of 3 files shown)


b4n:
Hum… sorry but I don't get it: the output seem to always use `namespace` scope (which contradicts your statement), and it seems to be correct (at least GCC likes that file just fine).
What's the actual issue here?

masatake:
Thank you for reviewing.

Here is what actually happens:

```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.cpp 
namespace X {
	void f (void);
};
void X::f (void) {}

[yamato@x201]~/var/ctags-github% ./ctags -o - /tmp/foo.cpp
./ctags -o - /tmp/foo.cpp
X	/tmp/foo.cpp	/^namespace X {$/;"	n	file:
f	/tmp/foo.cpp	/^void X::f (void) {}$/;"	f	class:X
```

In scope information of f X is reported as class.

b4n:
@masatake oooh, my bad, of course, it's a failure test case.  Sorry for the noise.

b4n:
Though, this will be very hard to properly fix, as here it's doable because the namespace is declared in the same file, but if it was included from another it wouldn't be possible to know whether `X` is a class, namespace or whatever else.

masatake:
I would like to change "class:" to "UNKNOWN:" or "unknown:".
In some area in ctags, "unknown" kind is needed. It is language independent like 'F", file kind.

cork can be used for finding the declaration in the same file.

`hint` can be used for finding the declaration in another file. See 
https://github.com/universal-ctags/ctags/pull/569 about `hint`.

Anyway recoding a bug in Units is meaningful.

PR #724, opened 2015-12-17
C family: reset AnonymousID for each input file

masatake:
My question in #713:
> In c.c, c parser doesn't reset AnonymousID.
> Do you think resetting it in each new input is better?
> (I don't have no opinion.)

The answer from @b4n:
I'm pretty sure it's better to reset it, because not doing so would
mean different tags being generated depending on the order files are
processed (or if they are processed all at once or separately).


b4n:
LGTM

masatake:
Thank you.

Issue #721, opened 2015-12-15
build on solaris 12 fails


tarzanek:
$ gmake
config.status: creating config.h
config.status: config.h is unchanged
[CC] ctags
Undefined                       first referenced
 symbol                             in file
GdbinitParser                       main/parse.o
CtagsParser                         main/parse.o
M4Parser                            main/parse.o
ld: fatal: symbol referencing errors
collect2: error: ld returned 1 exit status
Makefile:165: recipe for target 'ctags' failed
gmake: *** [ctags] Error 1

tarzanek:
@masatake it seems https://github.com/universal-ctags/ctags/commit/072b121e42420e20f9a657a3e1361f4d6f8896a0 broke it, can you have a look?

tarzanek:
a Solaris 11.3 VM can be fetched here: 
http://www.oracle.com/technetwork/server-storage/solaris11/downloads/vm-templates-2245495.html

after that do pkg install autogen automake gcc
I did ./autorun.sh
then ./configure ...
and gmake

tarzanek:
(pkg install git
if git won't be on the VM to clone the ws)
tia!
L

tarzanek:
D'OH! 
ok, this was a mismerge it seems (though it's weird since I never touched those files ... )
after 
git checkout optlib/gdbinit.c optlib/m4.c optlib/ctags-optlib.c
gmake started to work again! :)

sorry for the bug, closing now

masatake:
How about "make check" ?

tarzanek:
@masatake 
so gmake worked for me, I tried :
$ gmake check
config.status: creating config.h
config.status: config.h is unchanged
./misc/units[1854]: local: not found [No such file or directory]
./misc/units[1856]: local: not found [No such file or directory]
./misc/units[1857]: local: not found [No such file or directory]
./misc/units[1751]: local: not found [No such file or directory]
./misc/units[1752]: local: not found [No such file or directory]
./misc/units[1753]: local: not found [No such file or directory]
./misc/units[1754]: local: not found [No such file or directory]
./misc/units[1756]: local: not found [No such file or directory]
./misc/units[1757]: local: not found [No such file or directory]
./misc/units[1758]: local: not found [No such file or directory]
./misc/units[1760]: local: not found [No such file or directory]
./misc/units[1761]: local: not found [No such file or directory]
./misc/units[1763]: local: not found [No such file or directory]
./misc/units[1764]: local: not found [No such file or directory]
[...]

PR #725, opened 2015-12-17
units: allow to run test cases on zsh

masatake:
Allow to run units and tmain targets on zsh.
Issue reported by @tarzanek on #721.

### DIFF
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -17,6 +17,7 @@ UNITS=
 fuzz: TIMEOUT := $(shell timeout --version > /dev/null 2>&1 && echo 1 || echo 0)
 fuzz: $(CTAGS_TEST)
 	@ \
+	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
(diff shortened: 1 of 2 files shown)


makefiles/testing.mak
```diff
@@ -17,6 +17,7 @@ UNITS=
 fuzz: TIMEOUT := $(shell timeout --version > /dev/null 2>&1 && echo 1 || echo 0)
 fuzz: $(CTAGS_TEST)
 	@ \
+	if [ $$(basename $(SHELL)) = zsh ]; then set -o SH_WORD_SPLIT; fi; \
```
b4n: this way of detecting ZSH seem fragile to me, and I'd expect it to fail if ZSH is linked to /bin/sh (if that's even possible?).

I'd rather try to find something only ZSH defines, e.g. an environment variable like maybe `ZSH_SOMETHING` or alike.

masatake:
@b4n, thank you.

I found a good sentence in configure:
```
if test -n "${ZSH_VERSION+set}" && (emulate sh) >/dev/null 2>&1; then :
```

I will use `ZSH_VERSION`.

b4n:
Looks better, thanks :)

This said, I have no clue about ZSH nor why changes here would fix support for it, so I cant really review further.

PR #726, opened 2015-12-17
Reflect the comments about introducing automake

masatake:


b4n:
LGTM

masatake:
Thank you for suggestion. I've updated the branch.

Issue #727, opened 2015-12-18
Fails to compile when building from tarball instead of git clone


hauleth:
Way to reproduce:

```bash
$ git clone https://github.com/universal-ctags/ctags.git
$ cd ctags
$ mv .git git
$ ./autogen.sh && ./configure && make
```

This results with error:

```
In file included from main/repoinfo.c:4:0:
./main/repoinfo.h:1:1: error: expected identifier or ‘(’ before numeric constant
 0
 ^
```

masatake:
Thank you for reporting.
Could you try #728 ?

masatake:
Closed by #728.

PR #728, opened 2015-12-18
buildsys: make an empty repoinfo.h when building ctags from tarball

masatake:
repoinfo.h was general even when building ctags from tarball.  As
reported by @hauleth in #727, code snippet in Makefile.am made broken
repoinfo.h. This commit improves the situation.

The code snippet in Makefile.am in this commit make an empty repoinfo.h.
Instead CTAGS_REPOINFO is defined as NULL in repoinfo.c if CTAGS_REPOINFO
is not defined.


hauleth:
Looks like it solved problem.

masatake:
@hauleth, thank you again.

PR #730, opened 2015-12-22
Java: implement `imported' role

masatake:
Packages imported with keyword `import' are tagged as reference tags
with this commit. The wild-card(*) used in package specification is
recorded asis.

Example output:

    [yamato@x201]~/var/ctags-github% cat foo.java
    import java.util.Map;
    import java.util.*;

    package X;

    [yamato@x201]~/var/ctags-github% ./ctags --fields=+r --extra=+r -o - foo.java
    X	foo.java	/^package X;$/;"	p
    java.util.*	foo.java	/^import java.util.*;$/;"	p	role:imported
    java.util.Map	foo.java	/^import java.util.Map;$/;"	p	role:imported

### DIFF
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -169,6 +169,7 @@ typedef enum eDeclaration {
 	DECL_NAMESPACE,
 	DECL_NOMANGLE,       /* C++ name demangling block */
 	DECL_PACKAGE,
+	DECL_PACKAGEREF,
 	DECL_PRIVATE,
 	DECL_PROGRAM,        /* Vera program */
 	DECL_PROTECTED,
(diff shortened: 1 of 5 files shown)


parsers/c.c
```diff
@@ -1133,6 +1145,19 @@ static const kindOption *kindForType (const tagType type)
...
+
+	result = ROLE_INDEX_DEFINITION;
+	if (isInputLanguage (Lang_java))
+	{
+		if (type == TAG_PACKAGEREF)
+			result = JAVAR_PACKAGE_IMPORTED;
+	}
+
+	return result;
+}
```
b4n: style: missing newline after the function

b4n:
I'm not familiar with roles, but changes LGTM

masatake:
I'm introducing roles to "include", "use", "import", "source", and so on.
These instructions or directives don't define anything. Instead they refers external file or module.

I wrote about reference tag and role in https://github.com/universal-ctags/ctags/blob/master/docs/news.rst.

PR #733, opened 2015-12-24
buildsys: add --enable-debugging option and disable debugging by default

masatake:
-DDEBUG option was passed by default.  This flag slows ctags down;
whole the buffer area of a vString object is filled with 0 every
vStringClear call.

With this commit -DDEBUG is not passed by default.  To enable it,
--enable-debugging must be passed when running configure.


idning:
it's really slow..

PR #736, opened 2015-12-24
php: update comments

masatake:
Update the comments releated to
2c4744da31d9425da12563f03a820eb5029a98b4.

### DIFF
--- a/parsers/php.c
+++ b/parsers/php.c
@@ -1034,8 +1034,9 @@ static boolean parseClassOrIface (tokenInfo *const token, const phpKind kind,
 	}
 
 	inheritance = vStringNew ();
-	/* skip until the open bracket and assume every identifier (not keyword)
-	 * is an inheritance (like in "class Foo extends Bar implements iA, iB") */
+	/* read every identifiers, keywords and commas, and assume each
+	 *  identifier (not keyword) is an inheritance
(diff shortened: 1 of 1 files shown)


parsers/php.c
```diff
@@ -1034,8 +1034,9 @@ static boolean parseClassOrIface (tokenInfo *const token, const phpKind kind,
 	}
 
 	inheritance = vStringNew ();
-	/* skip until the open bracket and assume every identifier (not keyword)
-	 * is an inheritance (like in "class Foo extends Bar implements iA, iB") */
+	/* Accept only identifiers, keywords, and/or commas; and capture the identifiers.
+	 * This aproach is more robust than "Capture the identifier until an open bracket
+	 * is found". */
```
b4n: the second sentence isn't really important and IMO only reflects what the old (removed) code did, so it shouldn't be there.

I'd myself just alter it like that: 
> read every identifiers, keywords and commas, and assume each identifier (not keyword) is an inheritance (like in "class Foo extends Bar implements iA, iB")

b4n:
Good catch!

masatake:
Thank you. I will update the comment with using your comment and the merge this branch.

b4n:
LGTM

PR #739, opened 2015-12-26
C++: Avoid reporting function prototypes inside a function body

pragmaware:
Avoid reporting function prototypes inside a function body. This fixes cases like:

int something()
{
    foreach(...something)
    {
    }
}

in that foreach() would be reported as a prototype.

Also adds the ability to tag local variable instantiations:

int something()
{
   QString test(some,parameters);
}

### DIFF
--- a/Units/parser-cpp.r/variables-prototypes-2.cpp.b/args.ctags
+++ b/Units/parser-cpp.r/variables-prototypes-2.cpp.b/args.ctags
@@ -1,0 +1,2 @@
+--c++-kinds=+plv
+--fields=SKn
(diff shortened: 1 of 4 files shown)


masatake:
@pragmaware, thank you. Can I ask you to write a test case?
Smaller one is better.

https://github.com/universal-ctags/ctags/blob/master/docs/units.rst
Units/parser-cpp.r/angle_bracket.cpp.d is a simple example.
If you add extra options to ctags command line when running a test case, write args.ctags.

pragmaware:
There it is :)

pragmaware:
OK. I'll close this pull request and open a cleaner one.

PR #740, opened 2015-12-26
xpath metaparser based on libxml2

masatake:
ant.c can be rewritten with xpath metaparser.


masatake:
Remove unnecessary white spaces..

masatake:
For extension ".xml" we need a good selector.

parsers/glade.c
```diff
@@ -0,0 +1,110 @@
...
+};
+
+static void
+findGladeTags (void)
+{
+	xmlDocPtr doc;
+	xmlNode *root;
+	xmlXPathContextPtr ctx;
+
+	findRegexTags ();
```
b4n: ?

parsers/glade.c
```diff
@@ -0,0 +1,110 @@
...
+
+	doc = xmlParseFile(getInputFileName());
```
b4n: It's a bit sad to use the filename, as it makes it a little harder to move to in-memory parsing if we want to at some point.

parsers/glade.c
```diff
@@ -0,0 +1,110 @@
...
+	ctx = xmlXPathNewContext (doc);
+	if (ctx == NULL)
+		error (FATAL, "failed to make a new xpath context for %s", getInputFileName());
+
+	root = xmlDocGetRootElement(doc);
+	if (root == NULL)
+	{
+		verbose ("could not get the root node for %s", getInputFileName());
+		goto out;
+	}
```
b4n: won't all this initialization be the same in all Xpath parsers, and so can't it be moved to some generic place?

parsers/glade.c
b4n: oh, for the doctype extraction in the example.  sounds a bit hacky, couldn't this be extracted with an Xpath that could get installed at runtime?

parsers/glade.c
masatake: Oh, I see. However, we can improve this later.

parsers/glade.c
masatake: Drive the regex parser here.
See Units/simple-glade.d/args.ctags in these commits.

parsers/glade.c
masatake: The interface between glade.c (concrete xpath parser) and lxpath.c (meta xpath parser) is under designed. I have to write more concrete xpath parsers to find the good interface.

The redundancy is for introducing "--xpath-<LANG>=..." option in the future.
As you wrote, I have to refine the interface.

parsers/glade.c
masatake: It is just an example. What I want to say is that calling findRegexTags is meaningful; it allows a user doing something hacky.

parsers/glade.c
masatake: I found `xmlParseMemory'. I will try to use this.

masatake:
Scope is added but the interface between a concrete parser and xpath meta parser becomes ugly.
```
 ./ctags -o - ~/var/ctags-github/Units/simple-dbusinspect.d/input.xml  | sed -e 's|/home/yamato/var/ctags-github/Units/simple-dbusinspect.d/||'
</Units/simple-dbusinspect.d/input.xml  | sed -e 's|/home/yamato/var/ctags-github/Units/simple-dbusinspect.d/||'
ClientAdded	input.xml	/^    <signal name="ClientAdded">$/;"	s	interface:org.gnome.SessionManager
SessionName	input.xml	/^    <property name="SessionName" type="s" access="read">$/;"	p	interface:org.gnome.SessionManager
Setenv	input.xml	/^    <method name="Setenv">$/;"	m	interface:org.gnome.SessionManager
org.gnome.SessionManager	input.xml	/^  <interface name="org.gnome.SessionManager">$/;"	i
```

It seems much more work is needed...

masatake:
ugly and inefficient but I implemented generic selector for xml parser family. (2658334).
Now xpath meta parser can be used in real life. 
After working on ant.c I would like to merge.

@b4n gives me an homework: using xmlParseMemory instead of xmlParseFile.

masatake:
The interface of xpath meta parser is updated as @b4n suggested.

PR #741, opened 2015-12-26
C++: add missing array entry in declString

pragmaware:
Add a missing entry in the declString array. Fixes debug mode runs.


masatake:
Thank you.

PR #743, opened 2015-12-26
Add --maxdepth option for limiting the recursion depth

pragmaware:


pragmaware:
Hum.. no clue about the coverage decrease. Is it something wrong with the code?

masatake:
Good idea.

pragmaware:
Hm.. don't really know how to test this. Should probably create directory hierarchies with files within it and check that the produced tags are correct, but that would require far more code/text than the implementation of the feature itself... not a good idea.

masatake:
Thank you.

PR #746, opened 2016-01-02
Use stderr for MS-DOS/Windows.

adoxa:
There's no longer a reason to use stdout instead of stderr, as djgpp (if you still support DOS at all, I guess this is what you're using) provides `redir` expressly for the purpose of manipulating
standard error and `CMD.EXE` supports `2>` directly.

### DIFF
--- a/main/e_msoft.h
+++ b/main/e_msoft.h
@@ -9,11 +9,6 @@
 #ifndef E_MSOFT_H
 #define E_MSOFT_H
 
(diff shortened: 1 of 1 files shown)


masatake:
Is `stderr` available on the platform?

adoxa:
Yep, both DOS and Windows have stderr.

masatake:
Let's use stderr instead of errout.

PR #754, opened 2016-01-07
passing parserDefinition to its parser call back function

masatake:
pass parserDefinition instance to its parser callback function

In the original ctags architecture, parser callback function of
parserDefinition, the parserDefinition instance, and all the state of
parsing are separated.

This commit is initial trail for integrating them in object oriented way;
parserDefinition is considered as an object, which has the most
of all the sate of parsing as members of the object. The parser callback
is considered as a method of the object. Inheritance and Mix-In can be
introduced in the future.

My analysis of the original architecture:

```
parser callback function of parserDefinition took no argument.  If
more than two parserDefinitions share the same callback function,
isInputLanguage macro must be used to know the owner
parserDefinitions. findCTags function defined in parsers/c.c is
one of such functions. findCTags is shared in CParser, DParser,
CppParser, CsharpParser, JavaParser, and VeraParser.  In addition
all the state of parsing are stored to file local variables.
```

In a parserDefinitions there were two fields specifying the parser
callbacks: one for one path parsing, and the other for multiple path
parsing. This commit integrates the two, too.

Signed-off-by: Masatake YAMATO yamato@redhat.com

### DIFF
--- a/main/lregex.c
+++ b/main/lregex.c
@@ -757,9 +757,11 @@ static int fileReadLineDriver(void)
 	return (readLineFromInputFile () == NULL)? EOF: 1;
 }
 
-extern void findRegexTags (void)
+extern rescanReason findRegexTags (parserDefinition *parser __unused__,
+				   const unsigned int passCount __unused__)
 {
 	findRegexTagsMainloop (fileReadLineDriver);
+	return RESCAN_NONE;
 }
 
 extern boolean hasScopeActionInRegex (const langType language)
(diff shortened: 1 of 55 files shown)


masatake:
After learning more about ctags, I recognize passing whole parserDefinition is not good idea because some fields of parserDefinition are for the main part private purpose. I think parserDefinition should be separated two parts: visible from parser(parserSpec) and invisble from parser(parserDesc). About fields and extras I separated like that way.

PR #757, opened 2016-01-08
Refine win32 makefiles

k-takata:
This PR includes several fixes about win32 makefiles and document updates.


masatake:
LGTM, though I don't know windows well.

PR #767, opened 2016-01-21
Python: fix a bug about multi-line keyword arguments

### LINKED ISSUE #765: Python mult-iline keyword arguments it tag file
Follow up to https://github.com/universal-ctags/ctags/issues/750
Function calls with multi-line keyword arguments results in keyword arguments parsed as variables
Given
```Python
foo(
    arg1=1,
    arg2=2
)
def bar():
    pass
```
Getting following tags
```
arg1    test.py /^    arg1=1,$/;"       v
arg2    test.py /^    arg2=1$/;"        v
bar     test.py /^def bar():$/;"        f
```
Expected output is
```
bar     test.py /^def bar():$/;"        f
```

masatake:
Close #765.

A bug description quoted from #765 reported by @m-novikov:

    Function calls with multi-line keyword arguments results in
    keyword arguments parsed as variables. Given:

    foo(
	    arg1=1,
	    arg2=2
    )
    def bar():
	    pass

In the example arg1 and arg2 are captured as tags.

This commit fixes this bug.

Python parser processes input file line by line.  Line may include a
tag or not. Python parser applies rules for capturing a tag to the
each input line. This commit introduces a fallback rule which calls
skipParens. As the result foo (...) in the above example are skipped.

### DIFF
--- a/Units/parser-python.r/toplevel-funcall-with-keyword-args.d/expected.tags
+++ b/Units/parser-python.r/toplevel-funcall-with-keyword-args.d/expected.tags
@@ -1,0 +1,1 @@
+bar	input.py	/^def bar():$/;"	f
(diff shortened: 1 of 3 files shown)


masatake:
Thank you for testing.

After submitting this PR, I feel this patch is a bit rough for merging.
I will run extra tests.

PR #770, opened 2016-01-23
c++: Fix parsing of C++11 raw string literals

b4n:
See http://en.cppreference.com/w/cpp/language/string_literal

---

This contains a pretty ugly hack to fetch the previous character, in
order not to get fooled by string concatenation hidden behind a macro,
like in `FOUR"five"`, which is not a raw string literal but simply the
identifier `FOUR` followed by the string `"five"`.

While this may sound uncommon, it is not and lead to complaints [2][3]
when Scintilla [1] broke this when they introduced C++11 raw string
literal support themselves.

The implementation here still contains a bug with line continuations: a
raw literal of the form:

```c
const char *str = R\
"xxx(...)xxx";
```

is not properly recognized as such, although it's perfectly valid (yet
probably very uncommon).  For the record, Scintilla has also suffers
from this but nobody complained about it yet.

[1] http://scintilla.org/
[2] https://sourceforge.net/p/scintilla/bugs/1207/
[3] https://sourceforge.net/p/scintilla/bugs/1454/

### DIFF
--- a/Units/parser-cpp.r/cxx11-raw-strings.d/expected.tags
+++ b/Units/parser-cpp.r/cxx11-raw-strings.d/expected.tags
@@ -2,9 +2,21 @@ FOUR	input.cpp	/^#define FOUR /;"	d	file:
 memb1	input.cpp	/^struct typ1 { int memb1; };$/;"	m	struct:typ1	file:
 memb2	input.cpp	/^struct typ2 { int memb2; };$/;"	m	struct:typ2	file:
 memb3	input.cpp	/^struct typ3 { int memb3; };$/;"	m	struct:typ3	file:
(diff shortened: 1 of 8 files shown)


b4n:
See also https://github.com/geany/geany/issues/877

b4n:
Hold that, it dosn't recognize Unicode and wide raw literals (`UR`, `uR`, `u8R` and `LR`)

b4n:
Fixed

b4n:
Also, the hack to lookup a previous character is now moved in read.c, which makes it less of a hack and easier to maintain (as it's next to everything accessing File)

masatake:
LGTM.

PR #772, opened 2016-01-25
C++: handle lesser and greater operators in template parameters

### LINKED ISSUE #769: ctags is unable to recognize less-than sign in templates.
```
void func0() { }
template<int i, bool b = i < 10> void func() { }

void func1() {}
void func2() {}
void func3() {}
void func4() {}
int main() {}
```

It's clear that this piece of code can be compiled with success in a real compiler, but when parsing it using ctags I got an almost empty tag file, namely I lost all functions that follow `func()`:

```
$ ctags -f- test.cc
func0	test.cc	/^void func0() {}$/;"	f
```

masatake:
Close #769.

`<' and `>' operators can be used in parameters list of template
like:

    template<char i, bool B = (i>0)>     void f0(void) { }
    template<char i, bool B = i < 10>    void f1(void) { }

ctags could not handle them well.

This commit tries to improve the way to handle them.

The original checks only the balance of angle brackets.
In addition the original way, this commit considers
following things when checking the balance of angle
brackets:

1. lesser and greater operators between rounded brackets are ignored.

2. When `=` is found, lesser and greater operators are
   ignore till the input satisfies one of following conditions:
   2.1 `,` is found outside rounded brackets,
   2.2 `>` is found outside rounded brackets, or
   2.3 `<` is found outside rounded brackets and it is not the
        front part of `>>`.


b4n:
Not tested, but quick look seems to suggest it does what you say.

Skipping parenthesized expressions is a good idea.  I don't know C++ very well, but the other heuristics seem somehow legitimate, and it's actually very hard to do it right so maybe it's a good enough/a good start.

I though it would fail on e.g. `template<bool b = 0 > 2> voif f(void);`, but apparently it's not valid C++ (according to G++), so it's not a problem :)

So without actual testing, LGTM.

masatake:
@b4n, thank you for reviewing.

masatake:
@Shihira, based on your report, I wrote test cases: https://github.com/masatake/ctags/commit/dd8efdd7b9df3e02759906f75e6bfe90d9a43356

Do you have more bad(unwatend by ctags) cases?
I guess template parameter list can be nested, but I don't know how to write it well.
Nested angle brackets with default values will be able to break ctags parser.

Shihira:
@masatake , thanks for your work and prompt reply. The problem is solved, though little edge cases like `template<bool b = 0 < 2 && 2 < 4>` still bring about parsing failure. It's actually not a problem. For one thing, no one really put so much less than operators in template parameters, and for another, adding a pair of brackets (`template<bool b = (0 < 2 && 2 < 4)>`) works this time, plus making code clearer.

Thanks again. The patch helps a lot.

masatake:
@Shihira, thank you for giving new unwanted input: `(template<bool b = (0 < 2 && 2 < 4)>)`.
I added it to `Units/parser-cpp.r`, our test case repository.
After removing some lines from my original code, all the test cases are passed.

What I did is removing the code about following condition:

    2.3 < is found outside rounded brackets and it is not the front part of >>.

Issue #781, opened 2016-02-03
python: ignores tags after (certain) doc quotes 


blueyed:
```
% cat /tmp/t.py
class Foo():
    bar=dedent("""
               baz
               """)


class Bar():
    pass


% ctags -f - /tmp/t.py
Foo	/tmp/t.py	/^class Foo():$/;"	c	language:Python
bar	/tmp/t.py	/^    bar=dedent("""$/;"	v	language:Python	class:Foo
```

`Bar` is missing.

This is possibly related to the code at https://github.com/universal-ctags/ctags/blob/master/parsers/python.c#L1110-L1118 and/or https://github.com/universal-ctags/ctags/blob/master/parsers/python.c#L866.

blueyed:
Submitted a test case at https://github.com/universal-ctags/ctags/pull/784.

blueyed:
@masatake 
Is this on your radar?
Otherwise I will consider looking into it, given that this causes a fair amount of tags to be missed in certain projects.

masatake:
Though I solved the original issue, I didn't make s pull request because I found another bug releated to the area.

Could you look at #869? 
The fix is not clean enough. I think we need more token oriented parser for solving all the issues.

PR #784, opened 2016-02-03
Failing test for #781: triple quotes used with attrib

blueyed:
From #781:

```
% cat /tmp/t.py
class Foo():
    bar=dedent("""
               baz
               """)


class Bar():
    pass


% ctags -f - /tmp/t.py
Foo	/tmp/t.py	/^class Foo():$/;"	c	language:Python
bar	/tmp/t.py	/^    bar=dedent("""$/;"	v	language:Python	class:Foo
```

`Bar` is missing.

This is possibly related to the code at https://github.com/universal-ctags/ctags/blob/master/parsers/python.c#L1110-L1118 and/or https://github.com/universal-ctags/ctags/blob/master/parsers/python.c#L866.

### DIFF
--- a/Units/parser-python.r/triple-quotes-in-class.b/expected.tags
+++ b/Units/parser-python.r/triple-quotes-in-class.b/expected.tags
@@ -1,0 +1,3 @@
+Bar	input.py	/^class Bar():$/;"	c
+Foo	input.py	/^class Foo():$/;"	c
+a	input.py	/^    a="""$/;"	v	class:Foo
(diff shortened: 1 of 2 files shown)


masatake:
It seems that I injected the bug:
```
findPythonTags
=> skipParens \
   if (variableLineContinuation) \
      if (variable) \
         while ((line = (const char *) readLineFromInputFile ()) != NULL)
   => captureArguments
      => parseArglist
```

parenArglist doesn't considered the tripple quotes In this flow.
Feel free to assign the issue to me. In that case, merge this PR with '.b'.

blueyed:
Pushed as `.b` - please merge it yourself, if you think it makes sense.
Thanks!

PR #795, opened 2016-02-15
build: run optlib2c only if /bin/bash is available

### LINKED ISSUE #794: Build problem on OpenBSD
I'm trying to build on OpenBSD, but it turns out that all the .c files in the optlib directory are empty, which of course means that linking fails with unresolved dependencies. I think this has something to do with the face that the optlib2c script seems to be a bash-script, and I don't have bash installed (and even if I installed it it wouldn't be /bin/bash but /usr/local/bin/bash).

So I guess somehow the shell command which is supposed to create the .c files from the .ctags files somehow gives empty output, which is then echoed to the .c files creating empty files.

I fixed now it by changing the shebang in misc/optlib2c to !#/usr/local/bin/zsh, but it took me a while to figure this out :)

Maybe there should be a check in ./configure for a reasonably "bashy" shell?

masatake:
Close #794.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 misc/dist-test-cases > makefiles/test-cases.mak && \
-autoreconf -vfi && {
+autoreconf -vfi && type /bin/bash > /dev/null && {
     for i in `make -f makefiles/list-translator-input.mak`; do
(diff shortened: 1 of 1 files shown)


masatake:
I withdraw this.

Issue #798, opened 2016-02-18
Syntax errors in tagXPathTable initializations: ant.c, dbusintrospect.c, glade.c, maven2.c


acarlson1029:
Tried to compile with the latest today and ran into the following issue.

The `tagXpathTable` structs being initialized have a minor syntax error:
```c
static tagXpathTable antXpathProjectTable [] = {
    { "target",
      LXPATH_TABLE_DO_RECUR,
      .recurSpec= {
            antFindTagsUnderTask
        }
    },
    { "property/@name",
      LXPATH_TABLE_DO_MAKE,
      .makeTagSpec = {
            K_PROPERTY, ROLE_INDEX_DEFINITION,
            makeTagWithScope,
      }
    },
    { "import/@file",
      LXPATH_TABLE_DO_MAKE,
      .makeTagSpec = {
            K_IMPORT, R_IMPORT_GENERIC,
            makeTagWithScope,
      }
    },
};
```

Where the struct is defined in `path.h` as
```c
typedef struct sTagXpathTable
{
    const char *const xpath;
    enum  { LXPATH_TABLE_DO_MAKE, LXPATH_TABLE_DO_RECUR } specType;
    union {
        tagXpathMakeTagSpec makeTagSpec;
        tagXpathRecurSpec   recurSpec;
    };
    xmlXPathCompExpr* xpathCompiled;
} tagXpathTable;
```

The third argument here is a union, but the initializers are missing the curly braces.
```c
static tagXpathTable antXpathProjectTable [] = {
    { "target",
      LXPATH_TABLE_DO_RECUR,
      { .recurSpec= {
            antFindTagsUnderTask
        }
      }
    },
    { "property/@name",
      LXPATH_TABLE_DO_MAKE,
      { .makeTagSpec = {
            K_PROPERTY, ROLE_INDEX_DEFINITION,
            makeTagWithScope,
        }
      }
    },
    { "import/@file",
      LXPATH_TABLE_DO_MAKE,
      { .makeTagSpec = {
            K_IMPORT, R_IMPORT_GENERIC,
            makeTagWithScope,
        }
      }
    },
};
```

This affects the following files:
 * ant.c
 * dbusintrospect.c
 * glade.c
 * maven2.c

Patch attached.

[0001-Fix-syntax-errors-when-initializing-union-attributes.txt](https://github.com/universal-ctags/ctags/files/137159/0001-Fix-syntax-errors-when-initializing-union-attributes.txt)

[...]

PR #802, opened 2016-02-20
Fix union syntax errors in parser files.

acarlson1029:
Missing curly braces when initializing union members in tagXpathTable structs:
 - recurSpec
 - makeTagSpec

Fix for #798


masatake:
Thank you. After rethining again, I have recongnized your fix is better than mine.

PR #807, opened 2016-02-24
Add another case to handle shift operators in template parameters

focalintent:
Handle this case as well:

```
template <int P> class c {};
c< 8 > bVar;
c< 1<<8 > aVar;
```

### DIFF
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -1777,18 +1777,28 @@ static void skipCppTemplateParameterList (void)
 
 		if (c == '<')
 		{
-			if (roundBracketsLevel == 0)
+			int x = cppGetc ();
+			if(x != '<') 
 			{
-				if (defaultValueExpected == FALSE)
-					++angleBracketsLevel;
+				cppUngetc (x);
+				if (roundBracketsLevel == 0)
+				{
+					if (defaultValueExpected == FALSE)
+						++angleBracketsLevel;
+				}
 			}
 		}
 		else if (c == '>')
 		{
-			if (roundBracketsLevel == 0)
+			int x = cppGetc ();
+			if( x != '>') 
 			{
-				--angleBracketsLevel;
-				defaultValueExpected = FALSE;
+				cppUngetc (x);
+				if (roundBracketsLevel == 0)
+				{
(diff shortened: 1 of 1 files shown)


masatake:
Thank you very much. I will make a test case first.

parsers/c.c
```diff
@@ -1777,19 +1777,33 @@ static void skipCppTemplateParameterList (void)
...
-				if (defaultValueExpected == FALSE)
-					++angleBracketsLevel;
+				cppUngetc (x);
+				if (roundBracketsLevel == 0)
+				{
+					if (defaultValueExpected == FALSE)
+						++angleBracketsLevel;
+				}
 			}
+			else if(CollectingSignature)
```
masatake: > else if (CollectingSignature)

Do we need 'else' here? Only 'if' is not enough?
My understanding is that c must be gather to Signature if CollectingSignature is non-zero.

masatake: Oh, I'm wrong.
You read one more char here. You record it here.

parsers/c.c
```diff
@@ -1777,19 +1777,33 @@ static void skipCppTemplateParameterList (void)
...
+				vStringPut (Signature, c);
```
masatake: Could you change this to

```
vStringPut (Signature, x);
```

It helps code readers.

parsers/c.c
```diff
@@ -1777,19 +1777,33 @@ static void skipCppTemplateParameterList (void)
...
-				defaultValueExpected = FALSE;
+				cppUngetc (x);
+				if (roundBracketsLevel == 0)
+				{
+					--angleBracketsLevel;
+					defaultValueExpected = FALSE;
+				}
 			}
+			else if(CollectingSignature)
+				vStringPut (Signature, c);
```
masatake: Could you change this to

```
vStringPut (Signature, x);
```

It helps code readers.

masatake:
Manually merged. Thank you.

focalintent:
Thank you for catching that!  I changed the c's to x's to be clearer.

masatake:
!!oh, I forgot my comment. I already merged. I will change it by myself.

PR #812, opened 2016-02-26
Refactor, C++: don't use a temporary variable to gather Signature

masatake:

### DIFF
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -1788,7 +1788,7 @@ static void skipCppTemplateParameterList (void)
 				}
 			}
 			else if(CollectingSignature)
-				vStringPut (Signature, x);
+				vStringPut (Signature, c);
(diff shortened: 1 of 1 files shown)


parsers/c.c
```diff
@@ -1803,7 +1803,7 @@ static void skipCppTemplateParameterList (void)
 				}
 			}
 			else if(CollectingSignature)
-				vStringPut (Signature, x);
+				vStringPut (Signature, c);
```
vhda: If I understand the patch correctly the `else` is no longer required.

masatake:
!!! I'm very sorry. My change is really bad idea.
I withdraw this.

masatake:
@vhda, see the original my comment in #807. The comment is hidden because the orignal PR is overwritten by "git push --force".

Issue #787, opened 2016-02-04
PHP parser: include fully qualified class name


sserbin:
Would it be possible to include FQCN in addition to  a class name (or instead) in a tags file? 

Test case:
```php
<?php
namespace foo;

class bar {
    public function __construct() {
    }
}
```

Current output:
```
__construct	a.php	/^    public function __construct() {$/;"	f	class:foo::bar
bar	a.php	/^class bar {$/;"	c	namespace:foo
foo	a.php	/^namespace foo;$/;"	n
```

What I'm looking for:
```
__construct	a.php	/^    public function __construct() {$/;"	f	class:foo::bar
bar	a.php	/^class bar {$/;"	c	namespace:foo
foo\\bar	a.php	/^class bar {$/;"	c	namespace:foo
foo	a.php	/^namespace foo;$/;"	n
```

This would allow you to search/jump to a class definition with both `:ts bar` and `:ts foo\\bar`

masatake:
Thank you for reporting this. You are not only the person who wants this. See #524.
Let's discuss it.

I would like to know all combination of FQCN.

You told me the combination of two classes uses `\\`.
How about a name space and a class? 
How about two namespaces?

Could you tell me all the combinations of kinds which can be part of a FQCN and
separators you expected. If possible, could you tell me your idea how the FQCN and scope are different?

I' have been trying to under stand the critical difference between  FQCN and scope.
If it is possible, I would like to avoid adding new code only for php. 
I would like to introduce language neutral code which serves parsers including php.
So I insist understanding the difference.

`__construct` has `class:foo::bar` as scope information. It looks big hint for me to implement what you want. How about adding `foo::bar` as a tag entry instead of `foo\\bar` ?

(This discussion is applicable to Ruby parser and Python parser.)

sserbin:
Forgot to mention that this is probably useful only for vim (and possibly for other editors which can only use a tag name and not the rest of the fields)

FQCN (fully qualified class name) consists of full namespace + a class name. FQCN is somewhat similar to an absolute file path on a file system. Namespaces can be nested as well. Here are some examples of fqcn's:
`\foo\bar` class "bar" in namespace "foo"
`\foo\bar\baz\bar` class "bar" within namespace "foo\bar\baz"

> You told me the combination of two classes uses `\\`.

Namespace separator is actually one backslash (`\`)  - not sure why universal-ctags are using two (`\\`).

> If possible, could you tell me your idea how the FQCN and scope are different

[...]

PR #819, opened 2016-02-28
main, php: add a root separator to top level items

masatake:
makeQualifiedTagEntry() builds a full qualified tag for
a given tag. It uses scope information; put a separator
between upper item and the current item specified with
the given tag.

This commit handles the case when there is no scope information.  If
such case a root separator must be put as prefix to the current item.

e.g.

input:

    <?php
    namespace foo;
    >

the output with --extra=+q:

    \\foo	input.php	/^namespace foo;$/;"	n
    foo	input.php	/^namespace foo;$/;"	n


masatake:
Related to #787.

Issue #715, opened 2015-12-12
Adding extra fields


masatake:
@b4n wrote in #713 

> BTW, I'd like to add b4n/geany@f481587 to u-ctags too, but currently it doesn't seem possible to report function return type: we don't have varType, and typeRef, while similar, is not really usable (as there is  nothing to put in typeRef[0], and it's required).

> I thought I remembered a discussion about this fields, but I can't seem to find it, so maybe I dreamed it.

masatake:
I have thought about adding more fields. I wrote main/field.[ch] files for the purpose.
However, main/field.[ch] are not enough. tagEntryInfo must be extended for your purpose, as you wrote.

The concept "return type" is so popular among programming languages, so I think we should
add fields for storing it to tagEntryInfo. (If a field is not so popular, we should not extend it. In stead of we should a hashtable to tagEntryInfo. We can store a minor field to the hashtable.)

If you allow me, I would like to exntend tagEntryInfo to store returnType information and submit a PR here. In that case I would like you to review it. After merging my commit you can put your new code derived from  b4n/geany@f481587 on it.

I have more rather minor concerns about adding extra fields. I will write them later.

masatake:
NOTE: Generally we don't have to care about the size of tagEntryInfo; even if it grows there is no impact.
The exception is a parser using cork. Cork keeps copies of all tagEntryInfo captured/created from an input source file to an array. We cannot ignore the size of the array when the number of copies increases.

masatake:
See also #720.

masatake:
Some ideas:

* "returnType:"
* "arity:"
* "entrypoint:" for main functions.
* "nameref:" or "alias:"
* "qufalified: for tags only recorded when `--extra=+q` is given

b4n:
"return type" is nice, but it doesn't cover variable type, in e.g. C's `int foo`.  Actually, `varType` could be used if it allowed the first part (`varType[0]`) to be NULL/not given/whatever when the type is not a reference to a struct/enum/union (or not a known one at least, maybe a typedef).
I'm not sure this would be compatible with the current format though.

masatake:
Introducing two new kinds "unknown" and "builtin" we can reuse the typeref field of tags file.

input.c
```
typedef struct X D;
int a;
struct B b;
C c;
D d;
char e (void);
```

```
a => a ... ;" typeref:builtin:int
b => b...  ;" typeref:struct:B
c => c...   ;" typeref:typedef:C
[...]

PR #824, opened 2016-02-29
New CXX parser review

pragmaware:
Please take a look at the current state of the new CXX parser.

- It is as capable as the old parser was
- It adds several new features:
  - can extract "using" declarations
  - can extract function parameters
  - can extract anonymous structs/unions/classes/enums
  - can make anonymous structure names depend on input file name (to make them persistent across ctags runs)
  - supports C++11 lambdas (as anonymous functions)
  - handles function-level scopes and associates them to local variables (and parameters)
  - extracts local variable declarations which include calls to constructors
  - extracts local variable declarations from inside the for(), while(), if(), switch() parentheses
  - supports function prototypes/declarations with trailing return type
- It passes all the C/C++ unit tests the old parser did and some that the old parser did not.
- It adds a couple of new unit tests and fixes several old ones (especially in relation to the file-scope detection, which is IMHO still debatable)
- Generally it continues working where the old parser did simply bail out
- Is not easily fooled by template angle brackets
- Has an extensive builtin stderr-based debugging scheme which can be enabled with a single #define
- Is commented, uses readable names and has very large room for improvement

I have tried to limit the modifications to the original source code to the bare minimum and keep everything inside the parsers/cxx directory, so merging will be easy. This means that I have kept back several possible improvements that would require the addition of new extensions fields, including:
- identification of the end line of the blocks/scopes
- description of template parameters
- possibly the identification of template specialization parameters
- extraction of at least *some* variable and function return types (not sure if typeref is suitable for this)

Now it's probably time to review it, see if there are any showstopper bugs and identify the steps necessary for merging. The improvements could continue directly inside the universal-ctags source tree.

(please note that the coverage decreases probably because the old parser is now unwired and not being tested anymore)


masatake:
@ffes, @k-takata, can I ask you to help @pragmaware about building ctags with msvc?

masatake:
@pragmaware, grate work!
I will preview the changes in this weekend.

dawikur:
Oh, nice. I will review it in some free time - probably I will split it into smaller pieces of review :)

htfy96:
I noticed that this new parser freezes when parsing the following file:

[mp_wrapper.h.zip](https://github.com/universal-ctags/ctags/files/154336/mp_wrapper.h.zip)

Log:
```
Get file language for mp_wrapper.h
	pattern: mp_wrapper.h
		#candidates: 2
			0: C++ (extension: "h")
			1: ObjectiveC (extension: "h")
		#candidates after sorting and filtering: 2
			0: C++ (extension: "h")
			1: ObjectiveC (extension: "h")
Selector: 0x415c60
	selection: C++
OPENING mp_wrapper.h as C++ language include file
```

htfy96:
Backtrace:
```
Program received signal SIGTSTP, Stopped (user).
0x00000000004232d1 in cxxParserLookForFunctionSignature (pChain=<optimized out>, pInfo=pInfo@entry=0x7fffffffe210, pParamInfo=pParamInfo@entry=0x7fffffffe240)
    at parsers/cxx/cxx_parser_function.c:429
429			while(t != pInfo->pIdentifierEnd);
(gdb) bt
#0  0x00000000004232d1 in cxxParserLookForFunctionSignature (pChain=<optimized out>, pInfo=pInfo@entry=0x7fffffffe210, pParamInfo=pParamInfo@entry=0x7fffffffe240)
    at parsers/cxx/cxx_parser_function.c:429
#1  0x00000000004234e2 in cxxParserExtractFunctionSignatureBeforeOpeningBracket () at parsers/cxx/cxx_parser_function.c:644
#2  0x00000000004229e1 in cxxParserParseBlockHandleOpeningBracket () at parsers/cxx/cxx_parser_block.c:58
#3  0x000000000042261a in cxxParserParseBlock (bExpectClosingBracket=bExpectClosingBracket@entry=TRUE) at parsers/cxx/cxx_parser_block.c:389
#4  0x0000000000422113 in cxxParserParseClassStructOrUnion (eKeyword=eKeyword@entry=CXXKeywordCLASS, eTagKind=eTagKind@entry=CXXTagKindCLASS) at parsers/cxx/cxx_parser.c:598
#5  0x000000000042286c in cxxParserParseBlock (bExpectClosingBracket=bExpectClosingBracket@entry=TRUE) at parsers/cxx/cxx_parser_block.c:180
#6  0x000000000042375b in cxxParserParseNamespace () at parsers/cxx/cxx_parser_namespace.c:111
#7  0x000000000042293b in cxxParserParseBlock (bExpectClosingBracket=bExpectClosingBracket@entry=FALSE) at parsers/cxx/cxx_parser_block.c:145
#8  0x000000000042242f in cxxParserMain (passCount=1) at parsers/cxx/cxx_parser.c:989
#9  0x0000000000412c3a in createTagsForFile (passCount=<optimized out>, language=11, fileName=0x673090 "mp_wrapper.h") at main/parse.c:1812
#10 createTagsWithFallback (language=11, fileName=0x673090 "mp_wrapper.h") at main/parse.c:1835
#11 parseFile (fileName=fileName@entry=0x673090 "mp_wrapper.h") at main/parse.c:1991
#12 0x000000000040bd25 in createTagsForEntry (entryName=0x673090 "mp_wrapper.h") at main/main.c:253
#13 0x0000000000403378 in createTagsForArgs (args=0x673020) at main/main.c:298
#14 makeTags (args=0x673020) at main/main.c:444
#15 main (argc=<optimized out>, argv=0x7fffffffe720) at main/main.c:546
```

pragmaware:
@htfy96: it should be OK now. Thnx for the report.

masatake:
@pragmaware, could you add a test case for the fix for the issue reported by @htfy96?

parsers/cxx/cxx_parser.c
```diff
@@ -0,0 +1,1047 @@
...
+				if(!cxxParserHandleLambda(pParenthesis))
+				{
+					CXX_DEBUG_LEAVE_TEXT("Failed to handle lambda");
+					return FALSE;
+				}
+			} else {
+				if(!cxxParserParseAndCondenseCurrentSubchain(uInitialSubchainMarkerTypes,(uTokenTypes & CXXTokenTypeEOF)))
+				{
+					CXX_DEBUG_LEAVE_TEXT("Failed to parse subchain of type 0x%x",g_cxx.pToken->eType);
+					return false;
```
b4n: should be uppercase `FALSE`.

Units/parser-cxx.r/cxx11-lambdas.d/input.cpp
```diff
@@ -0,0 +1,18 @@
+int test()
+{
+	auto lambda1 = [](auto a,auto && b)
```
dawikur: I think it may be worth adding simplest lambda:
`[] { return 0; }`
for tests.

Units/parser-cxx.r/variables-in-control-statements.d/input.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+	for(int f04=0,f05=5,f06=10;f04<10;f04++)
+	{
+	}
+
+	for(std::string f07 = "test";;)
+	{
+	}
+	
+	for(int f08 : { 1,2,3,4,5 })
```
dawikur: Will 
`for(auto f : ...`
also work?

Units/parser-cxx.r/variables-in-control-statements.d/input.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+	return 0;
+}
+
+// The real test
+
+void test()
+{
+	// Valid variable declarations inside a for loop control structure
+	for(int f01=0;f01<10;f01++)
```
dawikur: What about case:

```cpp
for (int i1; ;) {}
for (int i1; ;) {}
```

here we have two separate `i1` objects.

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -0,0 +1,623 @@
...
+			{
+				vStringPut(t->pszWord,g_cxx.iChar);
+				g_cxx.iChar = cppGetc();
+			}
+		}
+		t->bFollowedBySpace = isspace(g_cxx.iChar);
+		return TRUE;
+	}
+	
+	if(g_cxx.iChar == '"')
```
b4n: as you use `cppGetc()`, isn't this dead code?  it should be returning `STRING_SYMBOL`, not bare quotes

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -0,0 +1,623 @@
...
+			{
+				g_cxx.iChar = cppGetc();
+				break;
+			}
+		}
+		t->bFollowedBySpace = isspace(g_cxx.iChar);
+		return TRUE;
+	}
+
+	if(g_cxx.iChar == '\'')
```
b4n: same, with `CHAR_SYMBOL`

parsers/cxx/cxx_parser.c
pragmaware: Fixed.

Units/parser-cxx.r/signature.cpp.d/input.cpp
```diff
@@ -2,7 +2,8 @@
 
 void foo (int a, char b) {}
 
-int bar (a, b) int a; char b; {}
+/* note that K&R style declarations are not valid in C++ */
+//int bar (a, b) int a; char b; {}
```
b4n: should ideally be check for C.

parsers/cxx/cxx_parser.c
```diff
@@ -0,0 +1,1047 @@
...
+	if(g_bFirstRun)
+		memset(&g_cxx,0,sizeof(CXXParserState));
+
+	g_cxx.eLanguage = language;
+	g_cxx.eCLanguage = language;
+	g_cxx.eCPPLanguage = -1;
+	cxxBuildKeywordHash(language,FALSE);
+}
+
+void cxxParserCleanup()
```
b4n: misses its prototype, and not used at all actually

Units/parser-cxx.r/cxx11-lambdas.d/input.cpp
pragmaware: Added.

Units/parser-cxx.r/variables-in-control-statements.d/input.cpp
pragmaware: Yes, it works.

Units/parser-cxx.r/variables-in-control-statements.d/input.cpp
pragmaware: Interesting question.
Internally the parser will tag i1 twice in the same function.
For some reason, then, a single tag will be emitted.
There must be some kind of duplicate detection mechanism somewhere in the tag file generation code.

The same happens if one declares the same variable twice in a single function.

A new feature could be to add the detection of anonymous code blocks. The local variables would be then attached to their proper scopes.

b4n:
@pragmaware a few small changes can be found  there: https://github.com/b4n/fishman-ctags/commits/b4n/pragmaware/new-cxx-parser. Nothing very fancy though.

b4n:
I didn't perform a thorough review, but the parser code seem to mostly make sense, and the changes to the tests seem also good.  So as it passes the tests, it looks pretty nice :)

parsers/cxx/cxx_parser_tokenizer.c
pragmaware: You're perfectly right.
#ifdef'd the unused code and added handling of STRING_SYMBOL and CHAR_SYMBOL.
Will delete the old code once I'm 100% sure.

Units/parser-cxx.r/signature.cpp.d/input.cpp
pragmaware: Added a test for it.
Later I will look into extracting the parameters from K&R style function definitions too.

parsers/cxx/cxx_parser.c
pragmaware: Hm.. the prototype for cxxCParserInitialize() is cxx_parser.h and it's used in cxx.c.
Maybe I'm missing something?

pragmaware:
@b4n: merged your changes. Thank you :)

parsers/cxx/cxx_parser.c
```diff
@@ -0,0 +1,1047 @@
...
+	if(g_bFirstRun)
+		memset(&g_cxx,0,sizeof(CXXParserState));
+
+	g_cxx.eLanguage = language;
+	g_cxx.eCLanguage = language;
+	g_cxx.eCPPLanguage = -1;
+	cxxBuildKeywordHash(language,FALSE);
+}
+
+void cxxParserCleanup(void)
```
b4n: I was talking about `cxxParserCleanup()`, not initialize ;)

pragmaware: Ow... you're absolutely right.
It is meant to be called in main, after all the files have been processed.
I've lost the call in one of the whole-repository reworks... added it back now.
Thnx! :)

pragmaware:
@masatake: I could add a test case but in this particular instance it would add low value since the problem was caused by a typo in the source code (e.g while(something); <--) and not some kind of input. Not sure we should have tests for such things...

masatake:
@pragmaware, thank you. What I wrote is just optional guideline. When adding a test case prepared by a bug reporter, you can put her/his name to Units/parser-cxx.r/something.d/README or input.cpp.

masatake:
(ALL, I invited @pragmaware to u-ctags organization. This new C++ parser will be biggest work made after forking from e-ctags. We need a maintainer for the new parser.  )

pragmaware:
Thnx for the invitation. I'll maintain this parser :)

pragmaware:
Do you think that this pull request should be merged or we let it stage a bit more?

I would like to try playing with new extension fields but it's probably better to have it as a distinct pull request based on a fresh starting point.

masatake:
@pragmaware, basically I think it is ready to merge the parser. But could you wait until this weekend.
I would like to try other test harnesses(chop, fuzz, and noise with valgrind) than "make units".

pragmaware:
Okie. Maybe I'll also clean it up a little bit more.

.gitignore
```diff
@@ -9,6 +9,7 @@
 *.o
 *.obj
 *.od
+*.ssexwsp
```
masatake: Just a question. What is this? Web search doesn't give me an answer.

Tmain/list-fields.d/stdout-expected.txt
```diff
@@ -73,7 +73,7 @@ Y	input.c	/^struct Y {$/
 i	input.c	/^  int i;$/
 j	input.c	/^  int j;$/
 main	input.c	/^int main(void)$/
-x	input.c	/^  struct X x;$/;"	typeref:struct:Y::X
+x	input.c	/^  struct X x;$/;"	typeref:struct:X
```
masatake: LGTM.

Tmain/list-kind-full.d/stdout-expected.txt
```diff
@@ -13,4 +13,7 @@ t	typedef	typedefs	on	referenceOnly:FALSE	nRoles:0
 u	union	union names	on	referenceOnly:FALSE	nRoles:0
 v	variable	variable definitions	on	referenceOnly:FALSE	nRoles:0
 x	externvar	external and forward variable declarations	off	referenceOnly:FALSE	nRoles:0
-L	label	goto label	off	referenceOnly:FALSE	nRoles:0
+L	label	goto labels	off	referenceOnly:FALSE	nRoles:0
+U	using	using statements	off	referenceOnly:FALSE	nRoles:0
+z	parameter	function parameters inside function definitions	off	referenceOnly:FALSE	nRoles:0
+A	NONE	make anonymous structure names depend on file name	off	referenceOnly:FALSE	nRoles:0
```
masatake: If I understand your code correctly, this `A` is never appeared on tags file.
This is just for controlling emitting tags about anonymous structure. Am I correct?
If yes, it seems that using `--extra` instead of kind is better.

You wrote
```
// FIXME: This is actually an option, but there seems to be no other way
```

(I can make a pull request to your tree if you agree with me.)

masatake:
I'm reviewing the changes in test cases.

.gitignore
pragmaware: Ow... that's a project file for a text editor / mini-IDE I'm writing in my spare time.
Screenshot here: http://labs.pragmaware.net/misc/ssex.shot.1.jpg
And... guess what... it uses ctags for symbol tagging :)
I'll probably release it to the public one day, once it's finished.

I can remove that line from .gitignore, but it shouldn't hurt anyone if it stays there..

masatake: I see. You don't have to remove it. However, it is not popular yet, so could write the description to the log of the commit that changes .gitignore.

Tmain/list-kind-full.d/stdout-expected.txt
pragmaware: Well.. this is used to switch on a particular behaviour on other tags.
The current ctags creates anonymous structure names that do not depend on the file name, they are just like __anonN where N is an increasing number which doesn't seem to be reset when the file changes. So say you have a source tree with several files. If you run ctags multiple times with different file sets the anonymous structure names will generally be different from run to run. This makes it hard to use ctags in "incremental" or "updating" mode.

The +A switch makes the structure names to be something like __anon HASH N where HASH depends on the file name and N is reset at the beginning of each file. This makes anonymous structure names more "stable".

masatake: Thank you for explanation.
I have understand `--extra` is not suitable for purpose.
What we want is mechanism controlling the parser own behavior: `--opt-<LANG>=<OPTION>:<VALUE>` .

e.g.

```
    --opt-C++=anonymousNamingStrategy:filehash
```.

...However, as far as looking at AnonymousID in c.c, it is reset each time a new inpiut file is passed.

Comment (pragmaware) [inline Tmain/list-kind-full.d/stdout-expected.txt]: Having a method to pass additional options to a single parser would be nice.

As for resetting the id... it's even a worse, because the anonymous structure names collide even within a single ctags run (and are still different in different runs). In principle the ctags user can look at the "file" extra field and tweak the structure names by himself but that makes using ctags harder (and almost impossible to do by hand).

On the other side, I could remove the A switch and just make the structure names always depend on the file name. But then it would be a non backward-compatible change...

In the long run having a method for passing other kind of parameters to a single parser is a win.

Comment (pragmaware) [inline Tmain/list-kind-full.d/stdout-expected.txt]: Hm..to be fully compatible without +A I should probably reset the index at each file too...

Comment (masatake) [inline Tmain/list-kind-full.d/stdout-expected.txt]: Let's throw away the compatibility about this issue.
Let's use your naming method.

If we got an bug report, reintroduce the original method.

**Comment by dawikur** [comment]
I will  review it on weekend, as I will have more time for it.

**Comment by akrzyz** [comment]
Hello,
I found case when parser crash.
`int f2()
{
    Bar<int> b;
    auto l = [](auto & p){ return p.template foo<int>();};
    return l(b);
}`

whole file: 
[templateInLambda.zip](https://github.com/universal-ctags/ctags/files/158611/templateInLambda.zip)

**Comment by akrzyz** [comment]
And another one

```
int main()
{
    auto l = []{return Foo::template bar<int>();};
    return l();
}
```

[anotherTemplateInLambda.zip](https://github.com/universal-ctags/ctags/files/158645/anotherTemplateInLambda.zip)

**Comment by pragmaware** [comment]
@akrzyz : should be OK now. Thank you for the report :)

**Comment by masatake** [comment]
(I'm sorry to be late, A bit busy.)


A.
 You use _cxx_ as prefix for avoiding including the same header files twice. e.g. `_cxx_debug_h_`.
Could you change the prefix to `CTAGS_cxx_` or `CTAGS_CXX_` Someone suggested me that  `_` is reserved by compiler or system library ago.


B. name in using 
```
[yamato@x201]~/var/ctags-github% cat input.cpp                          
using namespace x;
using y;
[yamato@x201]~/var/ctags-github% ./ctags -o - --kinds-'*'='*'  input.cpp
namespace x	input.cpp	/^using namespace x;$/;"	U	file:
y	input.cpp	/^using y;$/;"	U	file:
```

About `x` `namespace ` is prefixed. My understaing about C++ is that `namespace ` should be trimmed.
`using y` and `using namespace x` don't define y nor x. It should be tagged as reference. I will work on this after B is solved.

BTW, your debug print facility is very useful.

**Comment by masatake** [comment]
About performance:
```
[yamato@x201]~/var/ctags-github% time ./ctags -R ~/var/linux; wc -l tags
ctags: bad tag in /home/yamato/var/linux/drivers/net/wireless/mediatek/mt7601u/initvals_phy.h
./ctags -R ~/var/linux  3.24s user 0.24s system 99% cpu 3.484 total
281779 tags
[yamato@x201]~/var/ctags-github% time ctags -R ~/var/linux; wc -l tags 

ctags -R ~/var/linux  17.47s user 1.81s system 114% cpu 16.912 total
2927885 tags
```

**Comment by masatake** [comment]
For reviewing parpose, set CXX_TOKEN_POOL_MAXIMUM_SIZE to 2.
Try you will find many memory releated troubles.

```
$ valgrind ./ctags -R ~/var/linux
```

```=3335== LEAK SUMMARY:
==3335==    definitely lost: 4,008 bytes in 167 blocks
==3335==    indirectly lost: 5,344 bytes in 167 blocks
==3335==      possibly lost: 0 bytes in 0 blocks
==3335==    still reachable: 258,184 bytes in 1,934 blocks
==3335==         suppressed: 0 bytes in 0 blocks
==3335== Rerun with --leak-check=full to see details of leaked memory
==3335== 
==3335== For counts of detected and suppressed errors, rerun with: -v
==3335== ERROR SUMMARY: 601 errors from 7 contexts (suppressed: 0 from 0)
```

Here, I put the source code of linux kernel to ~/var/linux.

parsers/cxx/cxx_token.c
```diff
@@ -0,0 +1,165 @@
...
+	
+	if(g_pTokenPool->iCount > 0)
+	{
+		t = cxxTokenChainTakeFirst(g_pTokenPool);
+		CXX_DEBUG_ASSERT(t->pszWord,"The string shouldn't have been destroyed");
+	} else {
+		t = xMalloc(sizeof(CXXToken),CXXToken);
+		t->pszWord = vStringNew(); // we almost always want a string, and since this token is being reused..well.. we always want it
+	}
+
```
masatake: How about filling t with 0 with memset?
When we add a new field to CXXToken, memset will helps us.
Of, course, pszWord must be restored after filling.

pragmaware:
Hm.. it's not faster. I wouldn't expect it to be since its data structures are more complex and often it rescans the token chain two times to extract additional information.
In your test you have hit a bug that made it exit prematurely. I have now fixed it.

A test result on the kernel source is more something like:

Running new ctags...
real    0m25.956s
user    0m23.024s
sys     0m1.696s
Generated 4931468 tags

Running original ctags...
real    0m17.116s
user    0m14.856s
sys     0m1.192s
Generated 4017285 tags

So it's quite slower. But it generates nearly one million of additional tags and provides more informations about the other ones.

pragmaware:
As for the memset, I wanted to avoid it to save some time...

pragmaware:
Have run ctags under valgrind against the kernel sources. 
This is what I get:
[ctags-leak-check.txt](https://github.com/universal-ctags/ctags/files/162284/ctags-leak-check.txt)

Not that bad.
I don't understand why it's activating the xml parser when it isn't selected by --languages...

pragmaware:
Also valgrinded it with CXX_TOKEN_POOL_MAXIMUM_SIZE == 2.
Fixed a pointer read-after-free in that case.

pragmaware:
You're right, the "namespace" token doesn't look well in the using declaration.
However stripping it makes +U emit ambiguous tags since "using namespace a::b" tells the compiler to look for other symbols also inside a::b while "using a::b" tells the compiler to resolve symbol b as a::b.
I have split +U in +N (usingns) and +S (usingsym).
I have also tagged them as reference only (did I get this one right?).

masatake:
> I have split +U in +N (usingns) and +S (usingsym).
> I have also tagged them as reference only (did I get this one right?).

Instead of adding new kind, adding roles to U kind will be better.
See python.c. I will think abot this more, too.

masatake:
O.k. I will make a pull request in which I implement my idea about using statement and roles.

pragmaware:
Heh.. I just did it. Check it out :)

pragmaware:
Roles are nice. Though I didn't even know they existed...
+U as "reference-only" isn't even listed in --list-kinds. Why is that?

pragmaware:
Hm.. attaching roles to the namespace definition looks less attractive to me.
Also because it's not really backward compatibile (there might be tools that already use the namespace kinds without looking at roles).
The using keyword is not really a symbol definition, with the exception of one particular case which is handled as typedef. It's more like a hint to the compiler (and editors using ctags like ssex!) to attempt to resolve symbols in an additional container. For an editor attempting to provide member completions this is quite essential.

pragmaware:
Hm.. on the other side java does it with the "imported" role on the package symbol. So for symmetry an "imported" role on namespace might be an idea. Though that doesn't cover the case of namespace::symbol. C# has a similar statement but it doesn't seem to be extracted (?).

masatake:
Reference tags are still experimental. So you have to use `--_list-...` option to show roles.

```
[yamato@x201]~/var/ctags-github% ./ctags --_list-kinds-full=C++
c	class	classes	on	referenceOnly:FALSE	nRoles:0
d	macro	macro definitions	on	referenceOnly:FALSE	nRoles:1
e	enumerator	enumerators (values inside an enumeration)	on	referenceOnly:FALSE	nRoles:0
f	function	function definitions	on	referenceOnly:FALSE	nRoles:0
g	enum	enumeration names	on	referenceOnly:FALSE	nRoles:0
h	header	included header files	off	referenceOnly:TRUE	nRoles:2
l	local	local variables	off	referenceOnly:FALSE	nRoles:0
m	member	class, struct, and union members	on	referenceOnly:FALSE	nRoles:0
n	namespace	namespaces	on	referenceOnly:FALSE	nRoles:1
p	prototype	function prototypes	off	referenceOnly:FALSE	nRoles:0
s	struct	structure names	on	referenceOnly:FALSE	nRoles:0
t	typedef	typedefs	on	referenceOnly:FALSE	nRoles:0
u	union	union names	on	referenceOnly:FALSE	nRoles:0
v	variable	variable definitions	on	referenceOnly:FALSE	nRoles:0
x	externvar	external and forward variable declarations	off	referenceOnly:FALSE	nRoles:0
z	parameter	function parameters inside function definitions	off	referenceOnly:FALSE	nRoles:0
L	label	goto labels	off	referenceOnly:FALSE	nRoles:0
N	usingns	using namespace statements	off	referenceOnly:TRUE	nRoles:0
S	usingsym	using non-namespace symbol statements	off	referenceOnly:TRUE	nRoles:0
[yamato@x201]~/var/ctags-github% ./ctags --list-roles=C++      
ctags: Unknown option: --list-roles
[yamato@x201]~/var/ctags-github% ./ctags --_list-roles=C++
C++	d	undef	undefined	on
C++	h	system	system header	on
C++	h	local	local header	on
C++	n	used	namespace referenced in using statement	on
```

masatake:
@pragmaware, could you look at the top 3 commits in 
https://github.com/masatake/ctags/commits/adding-roles-for-using-statement-for-new-cxx-parser
?

I choose `member` kind for attaching a role.
I'm not sure this is suitable. Introducing "unknown(X)" could be better.
Anyway I think "using" kind is a bad idea.

Let me explain with an example.
Consider a following line in a shell script.
```
source foo
```
When I don't understand well what is reference tag.
I introduced "source" kind for tagging foo.
After long time thinking, I recognized "source" kind is really wrong name.
Instead, I introduced "script" kind and "loaded" role.
Now I can say:

   foo is a script and loaded with "source foo" statement.

Let's think about C++.

```
    using X;
```
What is X? I would like to say:
 
   X is a member whose name is imported to the current name space.

pragmaware:
Yep, you make sense.

It would be nice to know what is Y when an

   using X::Y;

is seen. But in fact all we can say is that Y is a symbol. It could be a type, it could be a function and I guess (but I'm not 100% sure at the moment) that it could be even a variable.
See http://en.cppreference.com/w/cpp/language/using_declaration for examples of such usages.
That statement only tells something like: hey compiler, when you see Y in the current scope you should really look for X::Y instead, whatever that might be.

Since we don't really know what it is, a new kind is needed. You can call it "using", or "symbol" or whatever. But it probably shouldn't be one of "namespace", "prototype" or "variable"...

For the other case

   using namespace A;

using the "namespace" tag with a different role make sense. Though also "using" with a different role looks more or less the same. The meaning is still quite different from a normal namespace declaration.

masatake:
(I'm bad in English. Tracking the discussion is very hard. So, I would like to summarize my proposals here. If both or one of them are/is acceptable, please, merge/reflect the change associated with the proposal to your source tree.
If not, I would like to explain a bit more. Feel free to show "not agreed with @masatake".)

> Since we don't really know what it is, a new kind is needed. You can call it "using", or "symbol" or whatever. But it probably shouldn't be one of "namespace", "prototype" or "variable"...

1. My proposal is reusing "member" kind, or introducing new kind "unknown" with "imported" kind.
My PR uses "member".

> For the other case

> using namespace A;

> using the "namespace" tag with a different role make sense. Though also "using" with a different role looks more or less the same. The meaning is still quite different from a normal namespace declaration.

2. My proposal is "namespace" kind with "used" role.

pragmaware:
Ok for B. Will merge tomorrow.
For A i would avoid using an existing kind since we don't really know what it is. "unknown" is fine.. or maybe "symbol" ?

pragmaware:
By the way, thanks for the review. It's very useful :)
Too big of a piece of code to find all the issues by myself.

masatake:
@pragmaware, thank you.
> For A i would avoid using an existing kind since we don't really know what it is. "unknown" is fine.. or maybe "symbol" ?

I used "unknown" in python.c. But "symbol" is not bad. Please, choose any you want.
I will review other areas. 

I have one request. You are the first person who tries to add a role/reference tag. 
I wrote document about document about reference tag and role in https://github.com/universal-ctags/ctags/blob/master/docs/news.rst
Feel free reflecting your experience to the document.

masatake:
[Off topic]
>  I don't understand why it's activating the xml parser when it isn't selected by --languages...
xml parser doesn't activated. However, a selector in main/selectors.c related to XML is called.
In additon it accesses uninitialized memory. The later one is fixed in #831. I'm working on the former one; ctags should not run the selectors for disabled parsers.

masatake:
I'm sorry but after rethinging, I recognized "unknown" is really good name for me. In #569, we have discussed multip path parser.  User will run ctags more than twice for the same input. The N th generated tags will be used for generating N + 1 th tags file as hints. The kind name implies the name is still "unknown" in a stage.

akrzyz:
hello,
I found another template issue.

In template.cpp tag for foo1 is made but no further tags are generated.
[template.zip](https://github.com/universal-ctags/ctags/files/167334/template.zip)

And another issue is with sfinae 
In boost_disable_if_c.cpp tag for makeArrayRef is not generated.
[boost_disable_if_c.zip](https://github.com/universal-ctags/ctags/files/167349/boost_disable_if_c.zip)

pragmaware:
@akrzyz: both cases fixed, thnx :)

masatake:
@pragmaware, could you try
```
$ /configure WARNING_CFLAGS='-Wall -Wextra -Wno-missing-field-initializers -Wno-empty-body
$ make
```
You can get more warnings.

masatake:
I'm worry. No warning is reported about your code.

Units/parser-c.r/bug1020715.c.d/expected.tags
```diff
@@ -1 +1 @@
-f	input.c	43;"	f
+f	input.c	43;"	f	signature:()
```
masatake: Good!

Units/parser-c.r/bug1491666.c.d/expected.tags
```diff
@@ -1,6 +1,7 @@
+__anon329ddd9201	input.c	/^typedef struct {$/;"	s	file:
 main	input.c	/^void main (void) {$/;"	f
-my_struct	input.c	/^} my_struct;$/;"	t	typeref:struct:__anon1	file:
-var1	input.c	/^	my_struct var1;$/;"	l
-var2	input.c	/^		var2;$/;"	l
-x	input.c	/^		x;$/;"	m	struct:__anon1	file:
-y	input.c	/^		y;$/;"	m	struct:__anon1	file:
+my_struct	input.c	/^} my_struct;$/;"	t	typeref:struct:__anon329ddd9201	file:
+var1	input.c	/^	my_struct var1;$/;"	l	function:main	file:
```
masatake: Excellent.

pragmaware:
I have some thoughs on the using/namespace/unknown/roles things. These are mainly based on programmatic ctags usage (i.e. an editor that uses ctags to build a symbol hierarchy) as all the ideas came to my mind when working on my editor.

- Adding semantically different roles to existing tag kinds is non backward compatible. I think it should be avoided since it would mislead existing user programs. For instance, "namespace" is an existing tag kind which defines a symbol. "using namespace" is always a reference to an existing symbol and has totally different semantics. This means that any user-program ignoring roles would treat the "using namespace" statement as a symbol defintion which is totally wrong. So "using namespace" and "namespace" should use a different tag kind.

- Since "namespace" can't be used for the reason above, the next best name is "using", which is close to the language and every C++ programmer will understand it without the need of reading documentation. The learning curve for using ctags proficently is quite steep, we should avoid raising the bar further more.

- If roles are really to be used to attach different semantics to a new tag kind then --fields=+r should be on by default and documentation should clearly state that relying on the kind only is not enough to capture semantics associated to the tag. 

- "using X::Y" semantics are different than the ones implied by "using namespace". Since "using" is a new tag we *could* use roles to differentiate between the two usages. But as said above the roles should be clearly visible because the two usages should never be interpreted in the same way. However, since this adds unnecessary complexity to parsing of the tag files I'm generally of the idea that different semantics should imply different tag kinds.

- "unknown" gives the idea that ctags has tagged something completly unidentified. For "using X::Y" this is not really true. It is certainly a compiler level symbol (e.g, not a macro, not a header). Furthermore C++ standard talks about "importing names". So "name" is probably a good kind name here.

- We need to add a "nameref" extension field which will help in resolving the name. I see that nameref is also cited in python.c so it could probably be shared. I can add it, but we need to merge the parser first.

pragmaware:
We need to merge this thing. I need "template" and "nameref" extension fields :)

masatake:
@pragmaware, I have still more option about naming of reference tag and role. That will be the final and biggest issue for merging. (I will merge your change anyway even if you disagree with my option because the new parser is really worthy.)

Before showing my option about the issue, I would like to continue reviewing your changes about test cases. Though the naming of reference tag and role is important, there is no application using on the earth. In other hand, people really want better parser.

pragmaware:
OK. I have a really huge commit ahead: extracting return types of functions.
This is going to modify a very large number of unit tests :)

Units/parser-c.r/bug1458930.c.d/expected.tags
```diff
@@ -1,2 +1,2 @@
-x	input.c	/^char x();$/;"	p	file:
-y	input.c	/^wchar_t y();$/;"	p	file:
+x	input.c	/^char x();$/;"	p
```
masatake: file: field is dropped.
The old parser puts file: to the all prototypes.(qualifyFunctionDeclTag)
How do you think about this?

One of view is that a declaration in a *.c file has no impact on the other files as far as the .c file is not included.
Therefore the declaration should have file:.
Another view is that a declaration is not a definition. Ideally it should be tagged as a reference tag, and have functoin kind with declaration role. Therefore thinking about its scope is meaning less.

masatake:
> OK. I have a really huge commit ahead: extracting return types of functions.
See https://github.com/universal-ctags/ctags/issues/715

pragmaware:
I'm currently using typeRef for this. Though I agree that typeRef being an array of two is kind of improper. It was designed with only class/struct/enum/union in mind.
When there is no class/struct/enum/union I'm using "typename" instead, which is a valid C++ keyword.

I'm also using it for the types of variables (which are already extracted for function parameters).

Units/parser-c.r/macros.c.d/expected.tags
```diff
@@ -1,8 +1,9 @@
+DECL1	input.c	/^DECL1(foo); \/* gcc will accept this as function prototype (with some warnings) *\/$/;"	p	signature:(foo)
+DECL3	input.c	/^DECL3(x, y); \/* gcc will accept this as function prototype (with some warnings) *\/$/;"	p	signature:(x, y)
```
masatake: (I have to look again this. I saw these expressions are used in many places in linux kernel. e.g.MODULE_INIT.
I have to see the impact on these area of new parser.)

Units/parser-cxx.r/bug1575055.cpp.d/expected.tags
```diff
@@ -1,3 +1,3 @@
 MyClass	input.cpp	/^	class MyClass { };$/;"	c	namespace:TheNamespace	file:
 TheNamespace	input.cpp	/^namespace TheNamespace {$/;"	n	file:
-variable	input.cpp	/^	int variable;$/;"	m	namespace:TheNamespace	file:
+variable	input.cpp	/^	int variable;$/;"	v	namespace:TheNamespace
```
masatake: file: is gone. I don't understand the semantics of namespace. Is this a declaration or definition?

About kind, look at README in the same directory. I agree with you; I think this should be v.
If you have your own idea about this, could you update README?

Units/parser-cxx.r/bug1585745.cpp.d/expected.tags
```diff
@@ -3,4 +3,4 @@ Class5	input.cpp	/^class Class5 {$/;"	c	file:
 ~Class2	input.cpp	/^Class2::~ Class2() { }$/;"	f	class:Class2
 ~Class3	input.cpp	/^Class3:: ~Class3() { }$/;"	f	class:Class3
 ~Class4	input.cpp	/^Class4:: ~ Class4() { }$/;"	f	class:Class4
-~Class5	input.cpp	/^    public: ~ Class5() { }$/;"	f	class:Class5
+~Class5	input.cpp	/^    public: ~ Class5() { }$/;"	f	class:Class5	file:
```
masatake: file: is attached here, too.

Units/parser-c.r/bug1458930.c.d/expected.tags
pragmaware: Heh... this is a good question.

The direction I've gone for now is:
   "whenever a symbol/declaration may be visible in any way from another file it's not in file scope".

This means that any variable/function that is not declared static is never limited to file scope because it's enough to declare it as extern somewhere else to be able to access it.

However, as you suggest, a function prototype, especially in a C file is controversial...

Don't know... I can change it if you think that the parser should follow the old rules.

As a side note: it's perfectly possible to include a .c file into another. So this whole "file scope" thing is not 100% reliable. In my editor I just enable all tags and when browsing I let the user choose which one he wants to jump to.

Units/parser-cxx.r/bug1799343-1.cpp.d/expected.tags
```diff
@@ -1,6 +1,7 @@
 C	input.cpp	/^struct C { int x; };$/;"	s	file:
-D	input.cpp	/^ D() { x = 123; }$/;"	f	struct:D
+x	input.cpp	/^struct C { int x; };$/;"	m	struct:C	file:
 D	input.cpp	/^struct D : ::C {$/;"	s	file:
+D	input.cpp	/^ D() { x = 123; }$/;"	f	struct:D	file:
```
masatake: file: is attached, too. (I don't say this is a bug. Just specifying the incompatibility.)

Units/parser-c.r/bug1458930.c.d/expected.tags
pragmaware: Yep.. maybe it's better to try to follow the old rules, so we remain compatible with whatever tool is using that feature.

So all prototypes go in file scope? Or only the ones declared in *.c files?

masatake:
@pragmaware, I want us to concentrate on merging; we should not add extra big changes now.

Basically the most of all things are up to you, the maintainer.
The point of my review here is "compatibility". 
There are two classes of compatibility. One is backward compatibility and another is forward compatibility.
The backward compatibility is more important. (The naming of roles and kinds for reference tags rather belong to forward compatibility.)

For me reviewing the test cases are the only way to detect breaking of the backward compatibility.
So I have been pointing out them incrementally.  You added --sort=no to args.ctags so reviewing is a bit hard, however, I understand very well you want to add  --sort=no. I'm writing a tool for reviewing.

I would like to explain the policy about the backward compatibility.(I should add this to our hacking guide.)

If you, the parser maintainer doesn't have strong opinion about the area, don't introduce the backward compatibility. 
If you have strong option about the area, introduce the change with documentation. See news.rst.
The good news is that universal-ctags has not released yet. I will never work for releasing till ctags.1 is updated well. Introducing big incompatibilities are acceptable only now.
We should know all backward incompatibilities. Unexpected incompatibilities should be avoided. I'm reviewing for the last policy. So I don't say (I cannot say) how we should do in the review comment.
What I can do constantly is just pointing out the existence of backward incompatibility.

Units/parser-c.r/bug1458930.c.d/expected.tags
masatake: As I wrote basically you can decide it; make ctags suitable to your tool.
When you are not sure what to do, keep the compatibility.

Units/parser-cxx.r/bug1575055.cpp.d/expected.tags
pragmaware: v is correct as per our convention. m are members of class/struct/union only.
(though the C++ spec talks also about namespace members and since global variables are in the global namespace then there would be no v variables at all... :D )

Again the same policy.. I've though about the symbol itself and not really the declaration being potentially visible from other files. Maybe my idea is wrong. I can change that.

Units/parser-cxx.r/bug1585745.cpp.d/expected.tags
pragmaware: A class declared in a cpp file is never accessible from other files (well.. unless one includes one cpp file from another). So that's why I have added it. But again, can change.

Units/parser-cxx.r/bug1799343-1.cpp.d/expected.tags
pragmaware: Again, a structure declared in a cpp file. Can't be visible from anywhere else nor can be its members.

Though nitpicking... this is again controversial since the class may be derived from another one that is actually declared in a header file and the virtual overrides *might* be actually stepped into when calling from a different file (say, in a debugger).

pragmaware:
Let's decide a policy for the file: field and I'll try to stick to it :)

Point out the incompatibilities. I'd like to be as compatible as possible (as long as the old output was correct).

pragmaware:
Ah, I'll hold the function return types then (it's really huge).

Units/parser-cxx.r/class-inheritance.cpp.d/expected.tags
```diff
@@ -0,0 +1,10 @@
+A	input.cpp	/^class A$/
+B	input.cpp	/^class B$/
+C	input.cpp	/^template<typename X,typename Y> class C$/
+D	input.cpp	/^class D : public A$/;"	inherits:A
+E	input.cpp	/^class E : public A, public B$/;"	inherits:A,B
+F	input.cpp	/^class F : private A, public B$/;"	inherits:A,B
+G	input.cpp	/^class G : virtual D$/;"	inherits:D
+H	input.cpp	/^class H : public A, virtual B$/;"	inherits:A,B
+I	input.cpp	/^class I : public C<A,B>$/;"	inherits:C<A,B>
+J	input.cpp	/^class J : private A, public C<A,B>$/;"	inherits:A,C<A,B>
```
masatake: This enhancement has impact.
Including readtag's filter engine, some tools may assume "," can be used as a separator for value of inherits field.
Therefore this enhancement and its impact must be written in news.rst.

Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
```diff
@@ -0,0 +1,34 @@
...
+l15	input.cpp	/^	struct Struct1 l15;$/;"	l	function:main	typeref:struct:Struct1	file:
+l16	input.cpp	/^	struct Struct1 * l16, l17, l18[10];$/;"	l	function:main	file:
+l17	input.cpp	/^	struct Struct1 * l16, l17, l18[10];$/;"	l	function:main	file:
+l18	input.cpp	/^	struct Struct1 * l16, l17, l18[10];$/;"	l	function:main	file:
+l19	input.cpp	/^	Struct1 l19 = {};$/;"	l	function:main	file:
+m01	input.cpp	/^	unsigned int m01;$/;"	m	struct:Struct1	file:
+m02	input.cpp	/^	std::string m02;$/;"	m	struct:Struct1	file:
+m03	input.cpp	/^	std::string ** m03, m04;$/;"	m	struct:Struct1	file:
+m04	input.cpp	/^	std::string ** m03, m04;$/;"	m	struct:Struct1	file:
+m05	input.cpp	/^	std::string & (*m05)(int a,int b);$/;"	m	struct:Struct1	file:
```
masatake: About m04, all parser records "struct:Struct1:std" as scope information.
The new parser records "struct:Struct1". It seems that std is dropped. Is it o.k.?

Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
```diff
@@ -0,0 +1,34 @@
...
+m06	input.cpp	/^	std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> > m06[10];$/;"	m	struct:Struct1	file:
+v01	input.cpp	/^} v01, v02[10];$/;"	v
```
masatake: About m06 the old parser records "struct:Struct1" as typefef field.
New parser doesn't. Expected?

masatake:
@pragmaware, I reviewed all the changes in test cases.
I put comments inline.
Again, this is really great one!

I would like to show the steps for merging.
1. About the naming of role, I would like to discuss after merging. At that time I will make a PR and you are the reviewer. I will try to make you accept my idea:-P
2. Like file: field, there are areas where policies are needed. Could you open a new issue?
    The discussion of this issue is too long. If you want we can decide the policies after merging.
3. I added 3 more comments today. If they are not difficult to fix, please, fix before merging. If new policy must be introduced for fixing, goto 2.
4. Write about the new parser on news.rst or new rst file.
5. Merge, but keeping the old parsers for awhile.
6. Close all issues solved by the new parsers.
7. Close this issue.

With these steps, I will not be a show stopper. Many things can be done with your pace.

masatake:
(I will read your comments about file: again tomorrow.)

Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
pragmaware: Struct1::std is something that doesn't exist at all.
The scope of the variable m03 is struct Struct1.
Then the type of m03 is std::string.

Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
pragmaware: Hum... m06 has struct:Struct1 as scope information also in the new parser.
Maybe you are talking about another variable?

pragmaware:
Enabling the old C parser seems to break the --line-directives option, even if that parser is never selected. Line directives that should switch to a different .c file actually do not switch.

Units/parser-cxx.r/class-inheritance.cpp.d/expected.tags
pragmaware: You're right, the old parser did strip template information from base classes.

As a policy I'd like to provide as more information as possible. I'm holding also another change that will add template information to classes in the following form:

template<typename A,typename B> class C : public D<A,B>

C   file.cpp   /whatever/    class     inherits:D<A,B>    template:A,B

So one knows that A and B are template placeholders and can actually substitute real types in place of that (instead of trying to look up A and B somewhere elsee).

Yet more.. it might be nice to even have a

  inherits: public D<A,B>

since the access type of inheritance defines the default access level of the inherited members.

However, all of this is incompatible.

One idea would be to have C/C++ parser specific switches that enable incompatible behaviours.
Something like: --c++-options=+ta. If no option is passed the template and access information is stripped (backward compatibility). If +t is passed then template information is preserved. If +a is passed then access information is preserved.

But all of this requires parser specific options...

pragmaware:
The file scope has other pitfalls.

For instance, a method definition in a *.cpp file may or may not be "visible" from other files depending on wheter the class itself was declared in the same cpp file or not. It's impossible to determine the visibility with a single pass of the file.

I'll write about it in a cxx.rst file now.

pragmaware:
@masatake:

> I would like to show the steps for merging.
> 1. About the naming of role, I would like to discuss after merging. At that time I will make a PR and you are the reviewer. I will try to make you accept my idea:-P

OK

> 2. Like file: field, there are areas where policies are needed. Could you open a new issue?
The discussion of this issue is too long. If you want we can decide the policies after merging.

I have changed the file: field policy to mark prototypes as file scope only if they are declared in a .c or .cpp file. Have written about it in cxx.rst.

> 3. I added 3 more comments today. If they are not difficult to fix, please, fix before merging. If new policy must be introduced for fixing, goto 2.

Replied. See above.

> 4. Write about the new parser on news.rst or new rst file.

Added cxx.rst in docs.

> 5. Merge, but keeping the old parsers for awhile.

I have merged your pull request that re-enables the parsers.

I guess we're ready then. Go?

masatake:
I'm working on `--line-directives` issue.
After that I will merge the new parser.

masatake:
It seems that you are already fixed the `--line-directives` issue. I'm trying to understand the change for it.

masatake:
I forgot writing one thing. All unnecessary whitespaces at the end of lines should be trimmed.
I have done this. I should tell you the first of all. I'm sorry,

masatake:
Merge. Thank you very much.

masatake:
@pragmaware, I update the document a bit. Look at http://docs.ctags.io/en/latest/

pragmaware:
Great! :)

PR #829, opened 2016-03-06
add artifacts to appveyor.yml

kozo2:
https://ci.appveyor.com/project/kozo2/ctags/build/1.0.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,18 @@ The goal of the project is preparing and maintaining common/unified working
 space where people interested in making ctags better can work
 together.
 
+## The latest build and package ##
+
+If you want to try the latest universal-ctags without building it yourself...
+
+### Windows
+- Go to https://ci.appveyor.com/project/kozo2/ctags
+  - Click the ```compiler=msvc, ARCH=x64``` (or ```compiler=msvc, ARCH=x86```) build.
+  - View the *Artifacts* tab and download ```ctags.exe```.
+  - Add the binary folder to your PATH.
+
+### Mac
+See [Homebrew Tap for Universal Ctags](https://github.com/universal-ctags/homebrew-universal-ctags)
 
 ## How to build and install ##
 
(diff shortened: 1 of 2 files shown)


masatake:
Could you tell me the benefit of this change in short?

kozo2:
It seems there is no release and downloads in https://ctags.io/
https://ci.appveyor.com/api/buildjobs/wcujskoqjae8wba8/artifacts/ctags.exe
might be useful for who don't want to build ctags themselves.

masatake:
My understanding is that Windows user can try the latest ctags without building.

With your change  Witndows users can get the latest binary from 
https://travis-ci.org/universal-ctags/ctags/artifacts/ctags.exe?

Am I correct?

kozo2:
Yes, right. > My understanding is...

No, the link is  from appveyor not from travis. > With your change...
https://ci.appveyor.com/project/kozo2/ctags/build/1.0.2

users can get the latest binary from 
https://ci.appveyor.com/project/APPVEYORUSERNAME/ctags/build/FOREACHBUILD

I will update README.md.

masatake:
> I will update README.md.

Thanks in advance.

masatake:
@k-takata, I have a questoin about appveyor.
As far as seeing URL, at travis-ci we have our organization(universal-ctags) own directory like https://travis-ci.org/universal-ctags/ctags/builds/114738509 .

In other hands, at appveyor we don't have such directory and my account is embedded in URL like  https://ci.appveyor.com/project/masatake/ctags/build/1.0.729. 

Is this o.k. ? I wonder what happens if I die today...

vhda:
@masatake don't you dare doing that us! ;)

masatake:
o.k. I will use my account. Using my account is better than using @kozo2's account:-P

masatake:
@kozo2, thank you.

PR #835, opened 2016-03-11
docs: fix document tree

masatake:
I updated toctree and other parts releated to it for generating pdf file.


masatake:
I merged this with very short review period because much more patches in my hand depend on this commit. However, comments about docs are really welcome.

PR #789, opened 2016-02-07
WIP: Minimize entry.h visibility

JonathonReinhart:
*Not yet ready for merge*

I would like to eventually add different storage backends to ctags. First, `entry.h` needs cleaned up to allow for an appropriate level of abstraction.

I'm opening this low-level PR for a few reasons:
- So people can see what's being worked on
- So these improvements can be integrated even if I don't finish the whole project
- So I can take advantage of your CI setup :-)

PR #842, opened 2016-03-13
Revise jonathon reinhart minimize entry visibility

masatake:
Based on the commits proposed in #789, unacceptable changes are fixed.
Close #789.


masatake:
I should do the same thin in input stream so that we can introduce memory input stream.

JonathonReinhart:
Thanks @masatake for continuing this and cleaning up after my PR.

PR #845, opened 2016-03-14
C/C++: Extract type names

pragmaware:
Ready for stress-testing, comments, ideas etc.


main/get.c
```diff
@@ -829,7 +830,8 @@ extern int cppGetc (void)
 				int next = getcFromInputFile ();
 
 				if (next == NEWLINE)
-					continue;
+					// continue; <-- this may exit returning the backslash, which si wrong.
```
masatake: s/si/is/ ? 
//

pragmaware: Ow...
I could even remove that line... have left it there with a comment to remember what I have changed in case it didn't work, but it seems to work so...

masatake:
(This is a general comment. These are not strict rule.)
Could you add the name of parser like C, C++ or C/C++ as prefix for commit title as I did.
For main part "main" can be used. For test cases, "Tmain" or "Units" cane be used.

If you modify both main part and cxx part, could you try splitting the change into more than two commits, some of them for main part and the others for the parser part.

I know well about main part. I will take more time for reviewing changes prefixed with "main: ".
I have to write this in reviewing.rst.

masatake:
Could you wait for merging for awhile?
I'm thinking about "typename".

pragmaware:
Oh sure. I think I'll work more on it later in the week or in the weekend. Have more minor tweaks to function return types to do and at this point we could also add variable types to the pull.

typename is not ideal, but at least it's a language keyword (`typename <some_arbitrary_type>` is valid in several contexts). Though the ideal thing would be to have `type:<type_just_as_it_is_in_the_sources>`. The `A` part in `typeRef:A:B` is unreliable in C++. It's a bit more reliable in C but still useless when talking about arbitrary types (say `const void **`).

I vote for adding `type:<something>` usable by all languages and keep `typeRef:A:B` only in compatibility mode (with --O C++:use-type-ref or --c++-use-type-ref enabled).

masatake:
(Coveralls comes back. I wonder why...)

pragmaware:
It doesn't seem to hit certain code paths.
I'll let it be for now as this state is temporary. I will be committing variable type extraction soon.

masatake:
(@pragmaware, I don't mean I am complaining about the test coverage  of your new code. I'm just wondering the behaviour of coveralls; suddenly it revived.)

pragmaware:
Uff... coveralls.. I don't understand you :/

pragmaware:
It's unclear what coveralls wants. Some code paths might not be hit now but that doesn't mean they won't be hit in the future when other code changes. Some reports also seem to be wrong (?) as certain functions really seem to be called.

pragmaware:
Should we merge this?

parsers/cxx/cxx_parser_variable.c
```diff
@@ -19,19 +19,38 @@
 #include "read.h"
 
 
-
+//
+// Attempt to extract variable declarations from the chain.
+// Returns TRUE if at least one variable was extracted.
+// Returns FALSE if a variable declaration could not be identified.
+//
+// Recognized variable declarations are of the following kinds:
```
masatake: Could you trim the whitespace(s) at the end of line.
(The view on github doesn't render it. But you may see it with `git diff`.

parsers/cxx/cxx_parser_variable.c
```diff
@@ -19,19 +19,38 @@
...
-//  - this function assumes that a function definition or prototype has
-//    been already excluded by other means.
-//  - there is a terminator at the end: one of ; = {
+//  - there is a terminator at the end: either ; or {
 //
-// Returns true if at least one variable was extracted.
+// Notes:
+// - Be aware that if this function returns TRUE then the pChain very likely has been modified
+//   (partially destroyed) as part of the type extraction algorithm.
+//   If the function returns FALSE the chain has not been modified (and
```
masatake: Here, too.

parsers/cxx/cxx_parser_variable.c
```diff
@@ -502,8 +508,31 @@ boolean cxxParserExtractVariableDeclarations(
...
-			//        empty them if they contains something that is not a numeric constant.
+			// Fix square parentheses: if they contain something that is not a numeric
+			// constant then empty them up
+			CXXToken * pPartOfType = t->pPrev;
+			CXX_DEBUG_ASSERT(pPartOfType,"There should be a part of type name here");
+
+			while(pPartOfType && cxxTokenTypeIs(pPartOfType,CXXTokenTypeSquareParenthesisChain))
+			{
+				CXXTokenChain * pAuxChain = pPartOfType->pChain;
+
```
masatake: Here, too.

masatake:
(I'm reviewing now...)

parsers/cxx/cxx_parser_variable.c
pragmaware: Done.

(btw, do trailing spaces break something.. like some source parsing tool?)

masatake: One of the reason I asked is that git diff renders the whitespaces with RED.
It is noise for reviewing task. I may turn it off. However, someone, a potential contributor may make a pull request what trims the whitespaces. We, you, I, and the contributor should not take time for such patch. Only the programmer introducing new code can avoid the patch.

If you were using emacs, `M-x whitespace-cleanup` helps you. However, you have your own editor/IDE.

main/get.c
masatake: It seems that this change is more generic fix than "Extract return types of functions. Correct all unit tests. ".
If possible I would like to you to make a separate commit only for this change.
I would like to see a test case specialized to this change.
Here this change means "changes for main/get.c".

main/get.c is used not only C/C++ but also other parsers.

pragmaware: This is already covered by an existing test which turned out to produce wrong type name output.
Units/readtags.r/backslash-at-the-end-of-pattern.d produced type \int.

How do I remove a single file change from a commit and then readd it as separate?

masatake: It will take your time but I write what I want here.
I guess your change about get.c is for fixing a generic bug.
I would like to you make a separate pull request fixing only the bug. The pull request should
include a test case verifying the fix.
Making a separate pull request takes time. So you may want keep the change in your current working branch. Sometimes I have done so. However, I always make a separate commit when fixing a generic bug in such case. Here I will call the former one "separate-pr method", and the later "separate-commit method". If you don't know I would like you to know "separate-pr method" first.

separate-pr method:

1. checkout master
2. pull upstream (universal-ctags/ctags)
3. make a branch for fixing the generic bug only and checkout the branch
4. fix the bug, and make a commit
5. add a test case, and make a commit
6. push it to your github repo
7. make a PR
8. merge the PR to the upstream. I will do this time, however, you may do in the future.
9. checkout master
10. pull upstream (your commit for fixing the bug will be merged to your local repo)
11 checkout cxx-parser branch
12 rebase the cxx-parser branch on your master
13. you will got conflicts. Solve them.
14. git push --force origin cxx-parser:cxx-parser
15. let me know the pushing

I'm not sure these steps are the best or not. However, I have done so while working on ctags.
If you don't know I recommend you to learn "git rebase -i/--continue/--abort", "git push --force",
and "git add -p".

separate-commit method:
1. git rebase -i master, and mark "edit" on the commi where you introduced the fix for the bug.
2. git add main/get.c
3. write a test case for the fix
4. Do git add the test case
5. make a commit for the fix
6. look at the output of git status
7. do git add the rest of changes
8. git rebase --continue

Good luck!

pragmaware: OK. Take a look at #853 .
There are four commits instead of two because of the upstream merges. But the diff is correct.

masatake: I see. Thank you very much.

masatake:
I'll take over the rest the for work for merging. Thank you very much.

masatake:
Merged after rebasing. Thank you again. These commits provide potential users enough reason updating "ctags".

I arranged some commit log.
Let's try to use "typename" and see what happens. If it is not good, someone will report it to us.
php's return type issue can be fixed based on this @pragmaware 's work. I have to revisit it.

BTW, I know you may want to implement arity: field as the next challenge:-P

PR #846, opened 2016-03-14
ruby: Import Geany changes

b4n:
While trying to merge U-CTags changes back into Geany, I realized we had 2 fixes that didn't make it here.  Here they are now, finally.


masatake:
LGTM. I tested.
Thank you.

masatake:
(After rebuilding on Travis, Coveralls revivied here, too...)

PR #850, opened 2016-03-17
Php withdraw root separator in full qualified tags

masatake:


masatake:
(This should be merged AFTER merging use/namespace patch proposed by @b4n.)

PR #851, opened 2016-03-21
PHP: parse `use` aliases

### LINKED ISSUE #847: Request for PHP Namespace and Use
Reference with [PHP Using Namespace](http://php.net/manual/en/language.namespaces.importing.php)

1. CTags is lacking the ability to decipher PHP namespace alias.
   e.g:
   ```php
<?php
namespace frontend\controllers;

use yii\web\Controller as WebController;

class SiteController extends WebController{
}
```
2. CTags is pointing another class.
   e.g:
   ```php
<?php
namespace frontend\controllers;

use yii\web\Controller;

class SiteController extends Controller{
}
?>
```
   But upon navigation, yii\base\Controller was retrieved instead of yii\web\Controller.

Hope that the team can help to address this issue.

Thank you!

b4n:
Fixes #847.

As it's similar to a C typedef, report the source type in `typeRef[1]`.  `typeRef[0]` is filled sensibly when possible, but most generally not, because it's impossible to know whether it'll be a class, interface or namespace.  In that case, `unknown` is used instead.

### DIFF
--- a/Units/parser-php.r/php-use.d/expected.tags
+++ b/Units/parser-php.r/php-use.d/expected.tags
@@ -1,0 +1,23 @@
+CONST	input.php	/^    const CONST = 1;$/;"	d	class:\\NS\\Cls
+Cls	input.php	/^  class Cls implements Iface {$/;"	c	namespace:\\NS
+Cls	input.php	/^  class Cls implements NSIface {$/;"	c	namespace:\\NS2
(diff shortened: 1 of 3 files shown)


masatake:
(Reviewing now)

masatake:
Merged manually. Thank you.

PR #855, opened 2016-03-26
C++, refactor: introduce enum for specifying elements in g_aCXXAccess…

masatake:
…Strings

In cxxTagBegin, the index of g_aCXXAccessStrings array is hard coded
with integer literal. Using enum improves it in readability a bit.


masatake:
This is wrong fix. I withdraw this.

PR #869, opened 2016-04-06
Python doc quotes handling

### LINKED ISSUE #781: python: ignores tags after (certain) doc quotes

masatake:
Close #781. However, I found  another bug during making this patch set.
The new bug is recorded as triple-quotes-in-class.b in this commits.


blueyed:
Thanks, better!

But it now fails after:

```
ILLEGAL_TABLE_TYPE = """\
    Table type {0!r} not handled by amqp. [value: {1!r}]
"""
```
(which was working before - according to a diff)
The trailing `\` does not seem to be relevant.

blueyed:
(ignore my previous comment about a regression - I got the diff wrong)

Can this be merged then, please?

masatake:
Thank you for testing.
I will keep this one till at the end this week.

Issue #863, opened 2016-04-02
Porting memory I/O from Geany


techee:
In addition to #862 the second big difference in Geany is we need to parse a file loaded into memory instead of parsing a file stored on disk. For this reason there's a small library called MIO:

https://github.com/geany/geany/tree/master/tagmanager/mio

which tries to hide where the file is stored - whether on disk or in memory. One creates a MIO file using either

```
MIO        *mio_new_file            (const gchar *filename,
                                     const gchar *mode);
```

to get a file-based backend or

```
MIO        *mio_new_memory          (guchar        *data,
                                     gsize          size,
                                     MIOReallocFunc realloc_func,
                                     GDestroyNotify free_func);
```

to get a memory-based backend. Afterwards, one can use POSIX-like functions like mio_read(), mio_write(), mio_getc(), mio_ungetc(), mio_setpos(), etc. without having to care of what nature the file really is.

The question now is - would it be acceptable for universal-ctags to use this library? There are two main benefits I can think of:

1. If there's some plan to provide universal-ctags as a shared library (https://github.com/universal-ctags/ctags/issues/63), something like this will be necessary because users of the library will need to parse files from memory.

2. In Geany even for files which we want to parse and which are stored on disk we load them to memory first (if they are not too big) and use the memory back-end. This turned out to be about 30% faster (for the C parser) because the bottle-neck is the getc() call and it's cheaper to do real IO just once at the beginning and then just advance the pointer in the buffer when simulating getc() than calling the real getc() many times.

We can do some things to make the MIO library play more nicely with universal-ctags:

1. It uses some glib types which can be easily converted to ordinary C types.
2. If there's a preference for some other type names, MIO, etc. could be renamed to something else.
3. The whole 4-file library can be easily converted to a single source + header implementation. So instead of library this would just be one additional source file.
4. Also maybe the "virtual" method calls (calling either memory or file back-end) could be just replaced by a simple "if" to have an IO function at a single place instead of one "interface" and two "classes" implementing it. Would save a few LOCs.
5. Make the coding style match the universal-ctags one.

What do you think? 

cc @b4n

masatake:
I'm very interested in MIO.    I have wanted to share more code with geany. However, as far as I am maintaining ctags, I will not provide universal-ctags as a shared library because designing API is too difficult for me. The most of my use case command line interface is enough. However, I have been worked for providing sophisticated command line interface. This is my basic position.

[...]

PR #870, opened 2016-04-06
Add the MIO library

techee:
As discussed here #863  this pull request introduces the MIO library to be able to parse memory buffers similarly to files which is both slightly faster and useful for users using ctags as a library.

The patch is mostly trivial - I just added the mio.c/h files, added MIO instead of FILE to inputFile and the rest is just following and fixing the errors from the compiler (in addition I made a manual check of all FILE, fpos_t, etc. afterwards and converted a few extra cases). There are still a few FILE left - these are mostly in cases where e.g. printing to stdout or in pcoproc.c/h where MIO doesn't make much sense.

I kept the original variable names (which in the case of FILE was in most cases fp) - in Geany they were renamed to mio - I don't know if it's worth the rename to mio, fp is probably fine in this case.

Finally, for files smaller than 1MB I loaded them into memory before parsing and let the parser parse them from memory - it's slightly faster. Tested on linux kernel sources ctags -R . took 40s instead of 44s on my machine.

It would be useful if someone with MS Visual Studio could verify this still compiles and works on Windows.

@b4n I did the stuff I suggested in #863 - I hope I didn't screw up your library too much. What I did:

* replaced the glib types with ordinary C types
* removed the "virtual" calls and replaced them with simple if/else
* made the implementation in a single file
* reformatted the library to more or less match universal-ctags style
* removed the MIO_FORCE_ANSI ifdef as it included some glib file and we don't really need it
* added mio_flush() - of course makes sense just for the file backend (calls fflush())
* made mio_free() return error code from fclose()
* changed mio_new_fp() to return NULL when the passed FILE is NULL (simplifies logic at one place in ctags and makes sense IMO)

Are you fine with the changes I made?

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -296,6 +296,39 @@ static boolean parseLineDirective (void)
  *   Input file I/O operations
  */
 
+#define MAX_IN_MEMORY_FILE_SIZE (1024*1024)
+
+static MIO *getMio (const char *const fileName, const char *const openMode)
+{
+	char buf[4096];
+	MIO *src, *dst;
+	size_t size = 0;
+
+	src = mio_new_file (fileName, openMode);
+	dst = mio_new_memory (NULL, 0, eRealloc, free);
+	while (size < MAX_IN_MEMORY_FILE_SIZE)
+	{
+		size_t num;
+
+		num = mio_read (src, buf, 1, sizeof (buf));
+		mio_write (dst, buf, num, 1);
+		size += num;
+		if (num < sizeof (buf))
+			break;
+	}
+
+	if (!mio_eof (src) || mio_error (src))
+	{
+		mio_free (dst);
+		mio_rewind (src);
+		return src;
+	}
+
+	mio_free (src);
+	mio_rewind (dst);
+	return dst;
+}
+
 /*  This function opens an input file, and resets the line counter.  If it
  *  fails, it will display an error message and leave the File.fp set to NULL.
  */
@@ -321,7 +354,7 @@ extern boolean openInputFile (const char *const fileName, const langType languag
 		File.sourceTagPathHolder = stringListNew ();
 	stringListClear (File.sourceTagPathHolder);
 
-	File.fp = mio_new_file (fileName, openMode);
+	File.fp = getMio (fileName, openMode);
 	if (File.fp == NULL)
 		error (WARNING | PERROR, "cannot open \"%s\"", fileName);
 	else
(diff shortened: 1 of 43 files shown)


techee:
OK, just checked the travis fails - locally I tried just "make check" which worked fine but it seems the travis-check.sh script does something different. Will investigate.

masatake:
How do you think about using  functions/macros defined in routines.[ch] like xMalloc?

techee:
Yeah, can use them. Is there anything else that could be used apart from the memory management functions?

masatake:
@techee, I will put comments inline when I found such functions.

techee:
I used the utils malloc/free functions as suggested. In addition, I noticed there's eStat() which I used to determine the file size so the memory buffer can be allocated directly with the right size. This reduces the linux kernel parse time by another 4s to about 36s.

masatake:
NOTE: `make check VG=1  ` is passed.

masatake:
Thank you very much.

main/read.c
```diff
@@ -296,6 +296,39 @@ static boolean parseLineDirective (void)
...
+#define MAX_IN_MEMORY_FILE_SIZE (1024*1024)
+
+static MIO *getMio (const char *const fileName, const char *const openMode)
+{
+	char buf[4096];
+	MIO *src, *dst;
+	size_t size = 0;
+
+	src = mio_new_file (fileName, openMode);
+	dst = mio_new_memory (NULL, 0, eRealloc, free);
```
masatake: The case src==NULL is not checked here.

main/read.c
```diff
@@ -296,6 +296,39 @@ static boolean parseLineDirective (void)
...
+	while (size < MAX_IN_MEMORY_FILE_SIZE)
+	{
+		size_t num;
+
+		num = mio_read (src, buf, 1, sizeof (buf));
```
masatake: src can be NULL here.

main/read.c
techee: Yes, sorry. But is it any relevant now? The patch was completely rewritten by https://github.com/universal-ctags/ctags/pull/870/commits/c5e3c54615da16d41de82162f1cc2246a2007d5e and this isn't present in the code any more (the same applies to the second comment).

masatake: Oh, sorry, I took a memorandum here. I fixed this in

https://github.com/universal-ctags/ctags/commit/231e40e72126d0c32201fc72ec24a690bb784bc9

Sorry for making noise.

PR #872, opened 2016-04-07
Generalize extended fields

masatake:
Don't merge this. This is just a prototype.

@pragmaware, based on your code(https://github.com/universal-ctags/ctags/pull/857/commits/552bdbcdce8ba6ccdf849d97bc5ef8f0621e827a), I made this patch set.

See  "sectionMarker" in  fdea608. Your idea is integrated to the main part. So the newly defined field
can be listed with `--list-fields` and be enabled/disabled with `--fields=[+|-]`. This is the biggest improvment
of my side. A field can be defined in a parser side. This is a critical idea of yours.

fdea608 is an example which shows you the way to define a field in your parser. A bit ugly, but it works.

This patch set still uses a letter for representing a field.  However, many of letters are  already used.
The letters will run out soon because the name space of letters are global, common between parsers.
However, I can solve the issue later.

Instead of 'custom field' I introduce a new term "parser field'. 
Currently for turning on sectionMarker field, `--fields=+M` is used. My plan is that `--fields=+{reStructuredText/sectionMaker}` can be used. You may say too, long. However, I think having per parser name space is better.

Basically I assume people may use wild card: `--fields=*`.


pragmaware:
Yep, something like this might work. I also like the idea of long names for fields.

Not sure if having field values of different types is really useful. They are all converted into strings for printing after all... so having just a plain const char * pointer might avoid the need for specialized "render" and "copy" functions.

masatake:
Updated.

I decide to allow only long name for parser fields because the name space of short name(letter) is so limited.
As suggested by @pragmaware, a C string is assumed as a value for parser fields. We can extend it later if we need.

TODO:

1. Make --xformat option accept the long name, and
2. Think about prefixing the name of owner language to the long name.

```
-       sectionMarker   off     TRUE     character used for declaring section
```

This can be
```
-       reStructuredText.sectionMarker   off     TRUE     character used for declaring section
```

masatake:
Now `--_xformat` works well with long name.

The last issue is handling multiple languages.

If language A and B define a field having the same name F, what we should do?
I have to handle this case.

masatake:
The most of all things I want about this topic are implemented.

TODO:
More test cases and documentation is needed. 
RenderAsIS can be removed.

pragmaware:
Go for it! :)

masatake:
(Before working on #882, I will finish this first.)

pragmaware:
Can we merge this? I'd like to do some work this weekend :)

masatake:
Sorry to be late:

After passing the testing on CI, I will merge these.

TODO:

* document (news.rst:must, internal.rst:optional),
* test case especially the case more than one parser have a field of the same name (e.g. C and C++),
* pseudo tags field output
* test case for rst and maven2

masatake:
See the  		b3c17a3.  It tells the usage of interface I introduced.

pragmaware:
Great!

PR #880, opened 2016-04-10
Allow typeRef[0] == NULL and generate "unknown" for it in the tags file

techee:
At the moment the value of typeRef[0] when no referenced type exists is
parser-specific. The cxx parser uses "typename" in this case, the php
parser uses "unknown". Instead, allow typeRef[0] to be set to NULL and
generate "unknown" for it when writing out the tag entry. This will make
the output identical for all languages.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -974,12 +974,16 @@ static int addExtensionFields (const tagEntryInfo *const tag)
 	}
 
 	if (getFieldDesc (FIELD_TYPE_REF)->enabled &&
-			tag->extensionFields.typeRef [0] != NULL  &&
 			tag->extensionFields.typeRef [1] != NULL)
+	{
+		const char *typeRef0 = tag->extensionFields.typeRef [0];
+
+		typeRef0 = typeRef0 ? typeRef0 : "unknown";
 		length += mio_printf (TagFile.fp, "%s\t%s:%s:%s", sep,
 				   getFieldName (FIELD_TYPE_REF),
-				   tag->extensionFields.typeRef [0],
+				   typeRef0,
 				   escapeName (tag, FIELD_TYPE_REF));
+	}
 
 	if (getFieldDesc (FIELD_FILE_SCOPE)->enabled &&  tag->isFileScope)
 		length += mio_printf (TagFile.fp, "%s\t%s:", sep,
--- a/parsers/cxx/cxx_tag.c
+++ b/parsers/cxx/cxx_tag.c
@@ -121,12 +121,7 @@ CXXToken * cxxTagSetTypeField(
 {
 	CXX_DEBUG_ASSERT(tag && pTypeStart && pTypeEnd,"Non null parameters are expected");
 
-	const char * szTypeRef0;
-
-	// "typename" is debatable since it's not really
-	// allowed by C++ for unqualified types. However I haven't been able
-	// to come up with something better... so "typename" it is for now.
-	static const char * szTypename = "typename";
+	const char * szTypeRef0 = NULL;
 
 	if(pTypeStart != pTypeEnd)
 	{
@@ -139,11 +134,7 @@ CXXToken * cxxTagSetTypeField(
 		{
 			szTypeRef0 = cxxKeywordName(pTypeStart->eKeyword);
 			pTypeStart = pTypeStart->pNext;
-		} else {
-			szTypeRef0 = szTypename;
 		}
-	} else {
-		szTypeRef0 = szTypename;
 	}
 
 	cxxTokenChainNormalizeTypeNameSpacingInRange(pTypeStart,pTypeEnd);
--- a/Tmain/extra-field.d/stdout-expected.txt
+++ b/Tmain/extra-field.d/stdout-expected.txt
@@ -1,10 +1,10 @@
 X	input.cpp	/^namespace X {$/;"	n	file:	extra:fileScope
 X::Y	input.cpp	/^  extern class Y {$/;"	c	namespace:X	file:	extra:fileScope,qualified
-X::Y::m	input.cpp	/^    int m;$/;"	m	class:X::Y	typeref:typename:int	file:	extra:fileScope,qualified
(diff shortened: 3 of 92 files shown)


masatake:
One of the hard topic in ctags.

Basically I think typeRef[0] should be given by parser own value.

Quoted from your change:
```
-m	input.cpp	/^    int m;$/;"	m	class:X::Y	typeref:typename:int	file:	extra:fileScope
+m	input.cpp	/^    int m;$/;"	m	class:X::Y	typeref:unknown:int		file:	extra:fileScope
```

I think `int` is not unknown. It is really type name.
There are many areas in ctags having such issues.

e.g. separator.  In scope field, names are combined with separators. Which letterswe should use as  separatros ?  Using one common between languages will help developers of upper layer tools make them simple. However, for some language users, a common separator can be unnatural or odd.

Basically I will not provide anything common between languages unless obvious. Finding such common things are too difficult. We don't have much time to know all languages. We cannot predict the syntax of languages not developed yet. Instead, I have extend pseudo tags. Ctags can provide meta information for upper layer tool via pseudo tags. It will be nice if `typename` and `unknown` are explained in pseudo tags.

Try `$ touch foo.php; ./ctags -o - --extra=+p --pseudo-tags='*' foo.php`.

Too complicated? But this is my best.

@pragmaware, do you have comment? 

Anyway these should be written in hacking guide.

main/entry.c
```diff
@@ -974,12 +974,16 @@ static int addExtensionFields (const tagEntryInfo *const tag)
 	}
 
 	if (getFieldDesc (FIELD_TYPE_REF)->enabled &&
-			tag->extensionFields.typeRef [0] != NULL  &&
 			tag->extensionFields.typeRef [1] != NULL)
+	{
+		const char *typeRef0 = tag->extensionFields.typeRef [0];
+
+		typeRef0 = typeRef0 ? typeRef0 : "unknown";
```
masatake: This is acceptable.

parsers/cxx/cxx_tag.c
```diff
@@ -121,12 +121,7 @@ CXXToken * cxxTagSetTypeField(
 {
 	CXX_DEBUG_ASSERT(tag && pTypeStart && pTypeEnd,"Non null parameters are expected");
 
-	const char * szTypeRef0;
-
-	// "typename" is debatable since it's not really
-	// allowed by C++ for unqualified types. However I haven't been able
-	// to come up with something better... so "typename" it is for now.
-	static const char * szTypename = "typename";
+	const char * szTypeRef0 = NULL;
```
masatake: Currently I think this is not acceptable. I think the value should be chosen by the author of the parser.
(Personally "type" is better than "typename".)

techee:
> I think int is not unknown. It is really type name.

My impression was that typeRef[0] contains one of existing kind names, e.g. something like

```
typeRef[0] = PhpKinds[K_FUNCTION].name
```

in which case anything else would be "unknown" (so e.g. "int" would be "unknown" because it's not one of the kinds). But now looking at the new cxx parser it doesn't seem to be the case any more.

> There are many areas in ctags having such issues.
> e.g. separator. In scope field, names are combined with separators. Which letterswe should use as >separatros ? 

Yes, we are keeping this in Geany to get a correct separator:

https://github.com/geany/geany/blob/master/tagmanager/src/tm_tag.c#L1185

But it's not a big problem and we could do something like that here as well. The big problem with the cxx parser is that it uses the keyword "typename" here which really is an existing keyword and can appear in the tag. E.g. for

```
typename boost::disable_if_c<N == 1, ArrayRef<ValueType, SizeType> >::type
makeArrayRef(ValueType (&p_data)[N], SizeType& p_size)
{
    return ArrayRef<ValueType, SizeType>(p_data, p_size);
}
```

the entry is

```
makeArrayRef    input.cpp       /^makeArrayRef(ValueType (&p_data)[N], SizeType& p_size)$/;"    f       typeref:typename:boost::disable_if_c<N==1,ArrayRef<ValueType,SizeType>>::type
```

And here "typename" really means the keyword "typename". But at the same time you get "typename" e.g. for the "int" in your example where the keyword "typename" wasn't used in the source.

In this case we cannot distinguish the two cases in Geany. So I'd suggest renaming "typename" in this case to something other than a real type name that can appear in typeRef[0]. It can be a per-language name, this is not a big problem for us, we can do some mapping for it in Geany.

pragmaware:
Hm.. it's true that "typename" may be either really present in the source or may be "forged" and you don't have a way to tell one case from the other. I didn't think about it.
Well, I'm not sure what to suggest. The first part of typeRef contains mostly irrelevant information and at this point even "unknown" seems redundant to me... I'd use an empty string if I was sure that all editors would accept it :D

techee:
I find the whole thing that a type is split into two parts quite unfortunate. It works only for the most simple types which consist of a single identifier only but it doesn't work very well for the composed ones. For instance with the "typename" you get for a member

```
typedef typename Type :: iterator2 T62;
```

tag

```
T62	input.cpp	/^	typedef typename Type :: iterator2 T62;$/;"	t	class:Container	typeref:typename:Type::iterator2	file:
```

If you have a look at the description of "typeRef", it says:

```
/* type (union/struct/etc.) and name for a variable or typedef. */
const char* typeRef [2];  /* e.g., "struct" and struct name */
```

And this really isn't the case here. "typedef" isn't a type and "Type::iterator2" isn't a "name for a variable or typedef". typeRef seems to have been designed for type declarations

```
typedef Foo Bar;
struct Foo {};
union Foo {};
```

where the first and the second part is clear but not for variable and function types. So the more I think about it the more I believe that there should be a new field similar to Geany's "varType" for variable and function types. What do you think?

masatake:
By reusing typeRef field, I have implemented varType:.  See #882.

@techee, could you fill the help message for "varType" in the pull request?
My idea is "the type of object including variable, function and so on".

@pragmaware, we should avoid "inconsistent" if possible. "redundant" is not good thing, but it is better than "inconsistent".

techee:
Closing, #882 is a better solution.

techee:
@masatake By the way, is there some way to regenerate all unit test expected tag files? For this pull request I did it semi-manually but it would be nice if there was some easier way. Is there some command-line parameter doing this?

PR #884, opened 2016-04-12
misc: add "review" script for reviewing Units results

masatake:
Usage:

	misc/review list
	misc/review inspect

"list" shows all failed result under *.d and *.t marked cased.
*.b is not listed.

"inspect" can be used for inspecting the results of each Units
test cases. It is menu driven.


masatake:
Untested well.

@techee, I guess this is what you want.

techee:
Haven't tried but from what I've seen it's actually better than what I imagined - I had in mind just a simple non-interactive script that would replace all expected tag files with the actual tag files (and the review would be done when committing the changes). Maybe this would still make sense in addition to the interactive mode for cases which affect lots of test cases so every single file doesn't have to be approved.

masatake:
Though the batch mode you mentioned will be useful but I don't add such feature in this commit.
It looks dangerous for me.

PR #886, opened 2016-04-16
CXX: Add some new fields to the cxx parser

pragmaware:
- Add template parameter support in class/struct/union declarations and function prototypes/definitions.
- Add "end" field for marking end lines of classes, unions, structs, namespaces, enums function declarations and lambdas
- Add support for extracting lambda capture lists

### DIFF
--- a/Units/parser-c.r/bit_field.c.d/args.ctags
+++ b/Units/parser-c.r/bit_field.c.d/args.ctags
@@ -1,2 +1,2 @@
 --sort=no
---fields=+{c.end}
+--fields=+{*.end}
--- a/Units/parser-c.r/bit_field.c.d/args.ctags
+++ b/Units/parser-c.r/bit_field.c.d/args.ctags
@@ -1,1 +1,2 @@
 --sort=no
+--fields=+{c.end}
--- a/parsers/cxx/cxx_tag.c
+++ b/parsers/cxx/cxx_tag.c
@@ -228,14 +228,14 @@ void cxxTagSetCPPField(CXXTagCPPField eField,const char * szValue)
 }
 
 void cxxTagSetCorkQueueCPPField(
-		tagEntryInfo * pTag,
+		int iIndex,
 		CXXTagCPPField eField,
 		const char * szValue
 	)
 {
 	CXX_DEBUG_ASSERT(g_aCXXCPPFields[eField].enabled,"The field must be enabled!");
 
-	attachParserField(pTag,g_aCXXCPPFields[eField].ftype,eStrdup(szValue));
+	attachParserFieldToCorkEntry(iIndex,g_aCXXCPPFields[eField].ftype,szValue);
 }
 
 void cxxTagSetCField(CXXTagCField eField,const char * szValue)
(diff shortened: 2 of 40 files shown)


Units/parser-c.r/bit_field.c.d/args.ctags
```diff
@@ -1 +1,2 @@
 --sort=no
+--fields=+{c.end}
```
masatake: Could you try to use wildcard notation `{end}` here? I hope this notation works well.
`{end}` means `{*.end}`. And `*` means "any language".

I implemented but I don't have a test case for the wildcard notation.
I would like to reuse this as a test case for the notation.

masatake:
13677b4 looks good.
I would like to take time for reviewing  ee27cda.
`end' is very impressive.

Units/parser-c.r/bit_field.c.d/args.ctags
pragmaware: Seems to work :)

parsers/cxx/cxx_tag.c
```diff
@@ -189,14 +221,44 @@ void cxxTagSetCPPField(CXXTagCPPField eField,const char * szValue)
...
+
+int cxxTagCommit(void)
 {
 	if(g_oCXXTag.isFileScope)
 	{
 		if (isXtagEnabled(XTAG_FILE_SCOPE))
 			markTagExtraBit (&g_oCXXTag, XTAG_FILE_SCOPE);
 		else
-			return;
+			return SCOPE_NIL; // FIXME: why the "invalid" cork queue index is SCOPE_NIL?
```
masatake: I agree. CORK_NIL is better?

masatake:
LGTM. make units LANGUAGE=C++ VG=1 is passed.

masatake:
Please, wait for merging. I found a critical bug in my code(main part).

masatake:
You understand corkAPI very well.

cxxTagSetCorkQueueCField and cxxTagSetCorkQueueCPPField are needed.
I think this should be part of main side.

Please, merge the commits in this PR.
I will fix main part later.

masatake:
What I should provide is
```
extern int attachFieldToCorkEntry (fieldType type, int index,
				   const char *value)
{
	tagEntryInfo * tag = getEntryInCorkQueue(index);

	if (index == SCOPE_NIL)
		return -1;

	return attachField (type, tag, eStrdup (value);
}
```

masatake:
Macros can be handled well based on your work.
```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.c 
cat /tmp/foo.c 
#define F(X) X
#define G
#include <S>
#define Z(Y)					\
int main (void)					\
{						\
  return 0;					\
}
[yamato@x201]~/var/ctags-github%  u-ctags --fields='+{end}{line}' -o - /tmp/foo.c 
 u-ctags --fields='+{end}{line}' -o - /tmp/foo.c 
F	/tmp/foo.c	/^#define F(/;"	d	line:1	file:	end:1
G	/tmp/foo.c	/^#define G$/;"	d	line:2	file:	end:2
Z	/tmp/foo.c	/^#define Z(/;"	d	line:4	file:	end:8
```

masatake:
Let's merge them!

PR #887, opened 2016-04-18
PHP: Check if alias kind is enabled.

Maners:
Otherwise, it is not possible to disable generating this kind of tags
via the --php-kinds option.


b4n:
LGTM

masatake:
Thank you.

I wonder the main part should verify a kind is enabled/disabled.

Verifying in the parser has still meaningful regarding optimization.

PR #892, opened 2016-04-19
misc/review: add command to show args.ctags

masatake:

### DIFF
--- a/misc/review
+++ b/misc/review
@@ -126,11 +126,29 @@ inspect_cmd_skip ()
     echo "SKIPPING $1" 1>&2
 }
 
+inspect_cmd_show_genenric ()
+{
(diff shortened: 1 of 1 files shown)


misc/review
```diff
@@ -126,11 +126,29 @@ inspect_cmd_skip ()
     echo "SKIPPING $1" 1>&2
 }
 
-inspect_cmd_show_diff ()
+inspect_cmd_show_genenric ()
```
vhda: generic

masatake: Thanks. I will decide to write a script for checking spells and integrate it to "make check".

vhda: Haven't been of much help lately. The least I can do is warn you of a misspelling when I spot it!

PR #893, opened 2016-04-19
main: implement --tag-relative=never|always

masatake:
Proposed in #868.

"always" and "never" as an argument for --tag-relative

```

Even if "yes" is specified as an option argument for --tag-relative,
absolute paths are used in tags output if an input is given as
an absolute path. This behavior is expected in exuberant-ctags
as written in its man-page.

In addition to "yes" and "no", universal-ctags takes "never" and "always".

If "never" is given, absolute paths are used in tags output regardless
the path representation for input file(s). If "always" is given,
relative paths are used always.

```


blueyed:
In the commit message: s/regardless the/regardless of the/

docs/news.rst
```diff
@@ -349,6 +349,21 @@ For specifying only one pseudo tag, specify it without sign:
...
+
+Even if "yes" is specified as an option argument for --tag-relative,
+absolute paths are used in tags output if an input is given as
+an absolute path. This behavior is expected in exuberant-ctags
+as written in its man-page.
+
+In addition to "yes" and "no", universal-ctags takes "never" and "always".
+
+If "never" is given, absolute paths are used in tags output regardless
+the path representation for input file(s). If "always" is given,
```
blueyed: s/regardless the/regardless of the/

main/options.c
```diff
@@ -367,8 +367,10 @@ static optionDescription LongOptionDescription [] = {
  {1,"       Define regular expression for locating tags in specific language."},
  {0,"  --sort=[yes|no|foldcase]"},
  {0,"       Should tags be sorted (optionally ignoring case) [yes]?"},
- {0,"  --tag-relative=[yes|no]"},
+ {0,"  --tag-relative=[yes|no|always|never]"},
  {0,"       Should paths be relative to location of tag file [no; yes when -e]?"},
+ {0,"       always: be relative even if input files are passed in absolute paths" },
+ {0,"       never:  be absolute even if input files are passed in relative paths" },
```
blueyed: s/passed in/passed in with/

masatake:
@blueyed, thank you for comments.

blueyed:
@masatake 
I assume I should test and merge it?  Or at least provide feedback?

masatake:
@blueyed, I would like you to test my change.
I'm trying to understand the result of the test case on "Environment: compiler=msys2, MSYS2_ARCH=x86_64, MSYS2_DIR=msys64, MSYSTEM=MINGW64".

masatake:
It seems that absoluteFilename and relativeFilename on the platform don't work as we expected.

masatake:
@k-takata san, I need your help.

As far as seeing the test result it seems that relativeFilename and absoluteFilename are not implemented well on environment "Environment: compiler=msys2, MSYS2_ARCH=x86_64, MSYS2_DIR=msys64, MSYSTEM=MINGW64". I would like to correct the implementation of
these two functions on the platform.

I got Windows-8.1 (kvm guest on Fedora), and I installed msys2-64. `autogen.sh && ./configure && make ` works fine. I got ctags.exe for the branch I proposed in this PR. Surprisingly the test case failed in approver is passed successfully on my environment. I wonder what is the difference between my environment and msys2 of appveyor. Any idea?

masatake:
I have found the reason why my code doesn't work on msys2: both `/` and `\\` are used in a path.

masatake:
On msys2 the output of pwd of bash is different from the value provided by getcwd function.

masatake:
The runtime environment is the same between bash where pwd is called and ctags which calls getcwd.

masatake:
I will add code do as follows:
On the platform where separators used in getcwd function and pwd command are different, `--tag-relative=never` and --tag-relative=always` are disabled.

masatake:
@blueyed, do you use windows?
I decide to disable `--tag-relative=never|always` on Windows. Is it o.k.?

masatake:
@k-takata,  in the environment "Environment: compiler=msys2, MSYS2_ARCH=x86_64, MSYS2_DIR=msys64, MSYSTEM=MINGW64" , pwd command used in a test harness script and getcwd function called in ctags return paths with different separators. I don't have actual output of the two now.

I introduced window8.1 and set up a msys2 environment for reproducing what I see in locally, but I could not  get the same result.

masatake:
I will merge this after merging #1230.

blueyed:
@masatake 
I am not using Windows - sorry for the late reply.

Thanks a lot for getting this in!

PR #895, opened 2016-04-21
CXX: Yet more work on the cxx parser

pragmaware:
- Extract a set of properties for functions and variables: static, inline, extern, mutable, virtual, pure...
- Try to detect template specializations in various forms
- Fix a couple of corner cases in variable extraction (i.e. do NOT extract variables from things like if(a & b)).


masatake:
LGTM. (make units VG=1 is passed).

PR #897, opened 2016-04-23
eiffel: Add support the across loop

dkearns:
This adds support for Eiffel's [across loop](https://www.eiffel.org/doc/eiffel/ET%3A%20Instructions#A%20closer%20look%20at%20the%20%27%27iteration%27%27%20form) instruction.


masatake:
Thank you!

PR #903, opened 2016-04-27
Fixed "duplicate const" compiler warnings when building on OS X using clang.

siegel:
When building on OS X 10.11 using clang-3.9, there were a bunch of warnings about "duplicate const" qualifiers. This PR resolves those warnings.


masatake:
Thank you!
Files under `optlib2c/` is generated by `misc/optlib2c` perl script.
If possible, could you also fix a line which emits duplicated `const` in ``emit_regexs` subroutine of the file?
If you fix the script, `optlib2c/`*.c will be updated when you run `make`.

siegel:
Certainly, I will look into that.

masatake:
You initialized `startOfLine` in readtags.c with `0`.
This is wrong. If you really need, you have to do like `memset(&startOfLine, 0, sizeof(startOfLine));`.
As far as loop condition, `startOfLine` is initialized any condition. If `fgetpos` is failed, we cannot use the value. However, filling the variable with 0 is not solution for the issue.

If you inspect more this area about fgetpos usage, I'm happy but your pull request is acceptable even if you don't touch `startOfLine`. In other word not in short time, not-initializing `starfOfLine` is acceptable.

Feel free to ask me if my english doesn't make sense.

masatake:
It seems that fgetpos issue you are spotting is something to do with https://github.com/universal-ctags/ctags/pull/904 which I'm working on.

masatake:
I'm sorry. #904 is nothing to do with this.

masatake:
However, I think we can replace fsetpos/fgetpos with fseek/ftell. As I did in #904, it seems that tags file should be opened in binary mode. If fsetpos and fgetpos are replaced, initializing `startofLine` is not needed.

siegel:
The initializations of fpos_t were done to resolve compiler and clang static analyzer warnings. Since it's a scalar (perhaps by accident), initializing an fpos_t with zero is syntactically and semantically correct. Using memset() or bzero() to do the initialization is correct as well, of course, so it comes down to a matter of style, and I have no objection either way. :-)

siegel:
(Also, I apologize for mixing up this issue with the "duplicate const" PR; this is clearly unrelated but somehow I pushed the wrong button and the commit for my readtags fixes got attached to this PR rather than creating a new PR.)

b4n:
> Since it's a scalar (perhaps by accident), initializing an fpos_t with zero is syntactically and semantically correct. 

It's not defined as scalar, so it's indeed some kind of accident it works.  According to the Linux Programmer's Manual fseek(3) manpage (emphasis mine):

> The `fgetpos()` and `fsetpos()` functions are alternate interfaces equivalent to `ftell()` and `fseek()` (with `whence` set to `SEEK_SET`), setting and storing the current value of the file offset into or  from  the  object referenced by `pos`.  **On some non-UNIX systems, an `fpos_t` object may be a complex object and these routines may be the only way to portably reposition a text stream.**

b4n:
I also fail to see any use for any explicit initialization: both locations you changed have an unconditional `fgetpos()` before any other use of the memory, so there's no extra initialization required.  If CLang is complaining, it's a bug somewhere in CLang (either it's implementation of `fgetpos()` requiring a pre-initialized argument (why?), or, more likely, in the sequence flow analyzer wrongly assuming the `fgetpos()` call may not happen, or not knowing it's an initialization function).

masatake:
@siegel, this is not a "style" issue. Look at the result on travis, with your new code, the building ctags is failed on some platforms, including GNU/Linux. I cannot merge your change.

masatake:
@b4n, I guess if we deal with the value returned from fgetpos properly, clang may say nothing.

siegel:
The point about fpos_t is valid and well formed :-), I'll make corrections here and submit a new PR. Thanks for the feedback.

b4n:
@masatake oh, good point!  Yeah this should check the return value of `fgetpos()` -- although I highly doubt it'll ever fail unless the stream is not seekable (typical `EBADF` from seeking functions), which is highly unlikely here, but still worth checking for safety.

masatake:
With checking I guess we can suppress the warning. I will do it after merging new PR from @siegel.

masatake:
I will merge your changes other than `2526de6`.

masatake:
I will cherry-pick 
	8849b9c and 9434eb3 after squashing.

masatake:
See #911.
@siegel, thank you for contribution. I'm waiting for your PR about readtags.c.
None maintains readtags.c well. So your contribution is really needed.

PR #904, opened 2016-04-28
readtags: use mio in query dsl implementation

masatake:
open_memstream and fmemopen were used to parse a query string.

Quoted from man fmemopen(3):

       fmemopen(), open_memstream(), open_wmemstream():
           Since glibc 2.10:
               _POSIX_C_SOURCE >= 200809L
           Before glibc 2.10:
               _GNU_SOURCE

It seems that open_memstream and fmemopen become parts of POSIX in
2008. MIO may be more portable than them.

### DIFF
--- a/readtags.c
+++ b/readtags.c
@@ -472,7 +472,7 @@ static tagFile *initialize (const char *const filePath, tagFileInfo *const info)
 		result->fields.max = 20;
 		result->fields.list = (tagExtensionField*) calloc (
 			result->fields.max, sizeof (tagExtensionField));
-		result->fp = fopen (filePath, "r");
+		result->fp = fopen (filePath, "rb");
 		if (result->fp == NULL)
 		{
 			free (result);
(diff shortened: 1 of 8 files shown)


readtags.c
```diff
@@ -472,7 +472,7 @@ static tagFile *initialize (const char *const filePath, tagFileInfo *const info)
 		result->fields.max = 20;
 		result->fields.list = (tagExtensionField*) calloc (
 			result->fields.max, sizeof (tagExtensionField));
-		result->fp = fopen (filePath, "r");
+		result->fp = fopen (filePath, "rb");
```
masatake: What does this mean?

b4n: open binary, e.g. no line end conversion on Windows (only real meaning I know from the top of my head).
But you changed that, so maybe you know :)

masatake: @b4n, thank you for the comment.
Without this, fsetpos/fgetpos(and fseek/ftell) don't work as I expected. Looks very strange for me.
I will write smaller test program to understand what really happens.

masatake: As far as trying on my environment where msys2 is installed, readtags works fine without "b" flag of fopen.

masatake:
Now I understand what happens.

PR #908, opened 2016-04-30
CXX: Support new and delete operators.

pragmaware:
Add support for C++ new, new[], delete and delete[] operators.


masatake:
"operator" is included in the name. 
Instead of including "operator" to the names, how about introducing "operator" kind?



About prototype for operator, I'm not sure what is the best. Ideally, I would like to convert all "prototype" kind for function to reference tag:

```
foo ... kind:prototype
```
`=>`

```
foo ... kind:function role:prototype
```

However, this breaks compatibility. However, we should know what we can do if we could ignore the compatibility issues.

pragmaware:
Yep, we could add a new kind for the operators.

Though, as you noted, it would force us to find a new way to differentiate between prototypes and definitions: additional complexity. On the other side users can already figure out that a function or prototype is an operator by looking at the "operator" keyword at the beginning of the name.

Backward compatibility is also a concern.

So in the end, weighting pros and cons, maybe the way it is now is not that bad.

masatake:
I see. Please, merge this.

PR #918, opened 2016-05-06
Iniconf metaparser

masatake:


b4n:
I don't know if I see the real benefit of multiple parsers here instead of a generic one (which my be missing, BTW).

For example, the YumRepoParser is a generic parser simply not emitting tags for keys.  This could be achieved easily with a generic one and simply disabling the `key` kind, couldn't it?

Similarly, the PythonLoggingConfig parser also mostly filters its output.  And here, while I don't know those configs, it looks like it's a stripped down version of a generic thing that could be exploited just as fine by users if it was generic: just search for `^logger_` tags and `^logger_[^.]*\.qualname` (assuming some regex-ish matching on FQN).
And couldn't the other sections and key be useful?  As an editor developer, I use CTags parser also as a TOC, so it'd be odd it would miss the majority of the entries like this -- yet again, I don't know this specific tool's configuration.

The SystemdUnitParser does add something, XRefs.  That's probably nice and warrant a specific parser, so no objection here.

In the end, I think there should be only 2 parsers (for now): a generic one that simply emits every section and key, and the SystemdUnit one that adds XRefs.  This makes me wonder if we really need this kind of flexibility, or if it wouldn't be enough to allow adding extra stuff, for XRef's sake?  Basically emit the tags in the generic code and let the specific callbacks add what they want.
What do you think?

masatake:
@b4n, thank you for comment. You are correct. Really good point.

(Your comment is applicable to xml parser. `id=` should be tagged in any parsers using xpath meta parser.)

I understand the needs for generic iniconf parser.
However, I also want specialized parsers. They provides higher level semantics of kinds.

Ideally I should provide the way to stack parsers. e.g. genericIniconfParser and  YumRepoParser.
genericIniconfParser emits all sections and keys. YumRepoParser just emis the name of repository.
(This is applicable to SystemdUnit parser, too.)

Think about following input (input.repo):
```
[fedora]
name=Fedora $releasever - $basearch
failovermethod=priority
```

What is the ideal output?

```
fedora /tmp/input.repo	/^[fedora]$/;"	kind:section	line:1	language:genericIniconf
name /tmp/input.repo	/^name=Fedora \$releasever - \$basearch$/;"	kind:key	line:2	language:genericIniconf
input.repo	/tmp/input.repo	3;"	kind:file	line:3	language:genericIniconf	extra:inputFileWithEndline
failovermethod /tmp/input.repo	/^failovermethod=priority$/;"	kind:key	line:3	language:genericIniconf
fedora	/tmp/input.repo	/^[fedora]$/;"	kind:repoid	line:1	language:YumRepo
input.repo	/tmp/input.repo	3;"	kind:file	line:3	language:YumRepo	extra:inputFileWithEndline
```
I think tagging `fedora` and `name` twice is acceptable. If you don't want it you can disable genericIniconfParser or YumRepoParser.

What I wonder is about "input.repo". It is recorded twice. 
Basically I would like to introduce bottom level parser named "fundamental" (taking from fundamental mode of emacs.)

So parser stack in my idea becomes "fundamentalParser" + genericIniconfParser + YumRepoParser.
This parser stack may generates:
```
fedora /tmp/input.repo	/^[fedora]$/;"	kind:section	line:1	language:genericIniconf
name /tmp/input.repo	/^name=Fedora \$releasever - \$basearch$/;"	kind:key	line:2	language:genericIniconf
failovermethod /tmp/input.repo	/^failovermethod=priority$/;"	kind:key	line:3	language:genericIniconf
fedora	/tmp/input.repo	/^[fedora]$/;"	kind:repoid	line:1	language:YumRepo
input.repo	/tmp/input.repo	3;"	kind:file	line:3	language:fundamental	extra:inputFileWithEndline
```

Do you find large gap between your ideal output and what I wrote here?

I'm just writing about tags output. I will have to think about effective implementation.

masatake:
After studying in implementing m4 and autoconf parsers, I update this commits.

I reintroduced iniconfi parser which tagges all sections and keys in .ini and .conf files as suggested by @b4n.

An upper layer parser is called from the iniconf parser. The upper layer parser picks up the subset of sections and keys that the parser is interested in.

A user can disable the iniconf parser if the one wants to get only the higher level tags of the input.
A user can disable the upper layer parser if the one wants to get only the generic tags of the input.

These study can be applied to json, sexp based language, yaml, and xml.

masatake:
An example:
```
[yamato@x201]~/var/ctags-github% cat example.repo 
[fedora]
name=Fedora $releasever - $basearch
failovermethod=priority
#baseurl=http://download.fedoraproject.org/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
enabled=1
#metadata_expire=7d
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
skip_if_unavailable=False
[yamato@x201]~/var/ctags-github% ./ctags --fields=+lK -o - example.repo 
enabled	example.repo	/^enabled=1$/;"	key	language:Iniconf
failovermethod	example.repo	/^failovermethod=priority$/;"	key	language:Iniconf
fedora	example.repo	/^[fedora]$/;"	repoid	language:YumRepo
fedora	example.repo	/^[fedora]$/;"	section	language:Iniconf
gpgcheck	example.repo	/^gpgcheck=1$/;"	key	language:Iniconf
gpgkey	example.repo	/^gpgkey=file:\/\/\/etc\/pki\/rpm-gpg\/RPM-GPG-KEY-fedora-$releasever-$basearch$/;"	key	language:Iniconf
metalink	example.repo	/^metalink=https:\/\/mirrors.fedoraproject.org\/metalink?repo=fedora-$releasever&arch=$basearch$/;"	key	language:Iniconf
name	example.repo	/^name=Fedora $releasever - $basearch$/;"	key	language:Iniconf
skip_if_unavailable	example.repo	/^skip_if_unavailable=False$/;"	key	language:Iniconf
[yamato@x201]~/var/ctags-github% ./ctags --languages=-Iniconf --fields=+lK -o - example.repo 
fedora	example.repo	/^[fedora]$/;"	repoid	language:YumRepo
```

PR #921, opened 2016-05-07
Record python entry point

masatake:


masatake:
After thinking over the entrypoint all last night, I am not sure it should be a kind.
In c `main` is a function. So entrypoint can be a field like:
```
main	foo.c	/^main(void)$/;"	f	typeref:typename:int	entrypoint:
```

In make, the first target of the makefile can be an entry point.
In these cases entrypoint field can be used for entrypoints.

Using entrypoint field in python is o.k. But what kind should be used for the line which the entry field is attached? `condition` kind is one of idea.

masatake:
Adding `entrypoint:` field is enough for named definitions like `main` in C. Quite natural.
Python has no such definitions.  

If we have parser own `extra`, what we can do?
1. adding `line` kind to python,
2. adding `entrypoint:` field to `line` kind of Python, and
3. adding `entrypoint` extra to Python.

`--fields=+{entrypoint}` makes the C parser attach `entrypoint:` field to `function` kind if a tag has a name `main`.

How about python?
`--fields=+{entrypoint} --extra-Python={entrypoint} --kinds-Python=+{line}` makes the Python parser do the same as C.

Too long. However, consistent. In relaxng parser, `<start>` is an entry point.
How about relaxng?
`--field=+{entrypoint} --extra-RelaxNG={entrypoint} --kinds-RelaxNG=+{tag}`.

What we need to implement the entrypoint? For C, no extra effort is needed in the main part.
For python and relaxng
A. parser own extra,
B. long name notation in --kinds
are needed. (B is optional.)

b4n:
IMO recording an "entry point" is kind of artificial, and very subtle.  Yes, in Python it's common, or even customary, to guard the "main" (as in C's `main()`) part in `if __name__ == "__main__"`, but many simple scripts don't do that and just write the "main" part in the root scope.  That's just fine for anything that isn't also supposed to be usable as a module.

So to record the "entry point" correctly, you'd have to find the first executed instruction.  Basically, the first instruction at the root level that isn't a definition, import or variable declaration.  But any function call, or any other thing that "looks like code that is supposed to be run as the script body".  What to do with i.e. `foo += 2`?  It's uncommon in non-"main", yet it's totally valid for a module too.

masatake:
Such fancy things can be done in a regex based subparser.

PR #929, opened 2016-05-12
Add guide for contributors

masatake:


masatake:
To show this to the author of Myrddin parser, I commit this.
Reviewing or fixing directly are really welcome.

PR #932, opened 2016-05-13
Fix tests on *BSD

oridb:
/bin/sh is not bash. misc/units contains bashisms. The simplest solution is to invoke it using bash.


masatake:
Debian people may want removing dependency to bash.

masatake:
Could you try #935?

oridb:
#935 seems to work on FreeBSD. I'll try on a couple of other /bin/sh implementations later, but if removing bashisms, I'll close this.

masatake:
Thank you for the verification.

Issue #933, opened 2016-05-13
 Test cases under review-needed.r (language: Eiffel)


dkearns:
Reviewing old Exuberant Ctags Eiffel tests under review-needed.r.

masatake:
Yes, I know the need of reviewing them.
However, I don't know Eiffel. It seems that you are an Eiffel hacker. Could you help me?

masatake:
#861 is also available:-P.

masatake:
The original report solves this:-P.

PR #934, opened 2016-05-13
Units: review and fix Eiffel tests under review-needed.r

dkearns:
Moved all tests to parser-eiffel.r

Fixes issue #933.


Units/parser-eiffel.r/obsolete.e.d/input.e
```diff
@@ -1,4 +1,4 @@
-class OBSOLETE obsolete "Use a different one"
+class OBSOLETE_ obsolete "Use a different one"
```
masatake: I don't understand the intent of the original author of this test input.
1. `OBSOLETE` is an key, so this should not be tagged.
2. `OBSOLETE` is a odd but a class name, so this should be tagged.

If this should not be tagged, we should not change the input. We must keep as is.
If this should be tagged, we have to fix the parser.

I checked the log of subversion but I don't find the original intent.

How do you think?

masatake:
Thank you very much. If possible I would like you to write one test case which verifies `local` kind.
To pass an extra option to a test case, use `args.ctags` file.
See http://docs.ctags.io/en/latest/units.html

Units/parser-eiffel.r/obsolete.e.d/input.e
dkearns: Eiffel is actually case-insensitive so a class name matching a keyword is a syntax error.

As indexing.e.d has the same problem I assume it was a misunderstanding of the original author.  Assuming this I interpreted the test as simply attempting to ensure that the keyword "obsolete" doesn't break the parser.  Perhaps it did in the past?

dkearns:
I just fixed the bare minimum here for the initial review of the old tests.

I'll look at adding more in the coming weeks.  I think there's a few bugs in the parser that I hope to fix when I can find the time.

Units/parser-eiffel.r/obsolete.e.d/input.e
masatake: I see. I read change logs of exuberant ctags. However, I cannot find any information about this.
Your mention looks reasonable for me.

masatake:
Thank you very much.

PR #936, opened 2016-05-14
Test cases under review needed.r language vim

dkearns:
These tests are OK; they reflect the current state of the parser. 

However, I did notice that variables seem to be missed by the parser in a couple of situations. I'll file a bug report or fix them when I've had a better look.


masatake:
`*.b` suffix can be used for marking/filling a bug. 
See `Gathering test cases for known bugs` in http://docs.ctags.io/en/latest/units.html .

masatake:
Thank you very much!

dkearns:
As an aside, thanks for all the documentation. I haven't read it all yet but it is appreciated.

masatake:
Comments for documentation is really welcome. The documentation is critical bottle-neck for releasing.

Issue #931, opened 2016-05-12
extern "C" declarations/definitions not recognised in C++


Corax26:
In C++, `extern "C"` declarations and definitions are not recognised, except when an `extern "C"` scope is used:

```c++
extern "C" void f(void);
extern "C" void g(void) {}

extern "C"
{
  void h(void);
}
```
> `ctags --c++-kinds=+p`

Output:
```
h	test.cpp	/^  void h(void);$/;"	p	typeref:typename:void	file:
```

This is a regression, compare with the output of ctags 5.8:
```
f	test.cpp	/^extern "C" void f(void);$/;"	p	file:
g	test.cpp	/^extern "C" void g(void) {}$/;"	f
h	test.cpp	/^  void h(void);$/;"	p	file:
```

pragmaware:
Fixed via #937

Corax26:
Great thanks!

PR #937, opened 2016-05-14
Fix upstream bug #931

pragmaware:
This fixes bug #931 and adds a test case.

It can be improved as declarations appearing in blocks declared as extern aren't marked as extern in properties. Need to think a bit about this.


masatake:
Just a question. The original report is about C++. However, as far as reading args.ctags and the name of input file(input.c), you may assume the issue is about C. I wonder why.

pragmaware:
You're right, I have mixed it up a bit.
extern is a C keyword and should appear as a property in the C language. However extern "language" is a C++ extension. Will sort it out.

This week I'm on holidays with limited access to computer and connectivity so I'll probably work on it in the next week.

pragmaware:
I'll try to make this fix better in a new pull request.

PR #939, opened 2016-05-15
Units, Ruby: review tests under review-needed.r

dkearns:
Moved all tests to parser-ruby.r

universal-ctags now appears to do better on simple.rb.d than ripper-tags.


masatake:
Thank you!

In next time when reviewing is done, please, change the suffix `*.t` to `*.d`.

dkearns:
Sorry about that I thought I had...

PR #944, opened 2016-05-20
make: fix a typo in parenthesis handling

masatake:


masatake:
@b4n, this should be fixed in geany if it is not fixed yet.

b4n:
@masatake indeed, good catch, thanks!

PR #960, opened 2016-05-25
Fix bug #931

pragmaware:
Fixes handling of extern "C" declarations.

### DIFF
--- a/parsers/cxx/cxx_parser_block.c
+++ b/parsers/cxx/cxx_parser_block.c
@@ -162,6 +162,8 @@ boolean cxxParserParseBlock(boolean bExpectClosingBracket)
 	{
 		if(!cxxParserParseNextToken())
 		{
+found_eof:
+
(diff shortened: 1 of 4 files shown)


parsers/cxx/cxx_parser_block.c
```diff
@@ -342,7 +347,43 @@ boolean cxxParserParseBlock(boolean bExpectClosingBracket)
...
+
 						cxxTokenChainDestroyLast(g_cxx.pTokenChain);
+
+						if(!cxxParserParseNextToken())
+							goto found_eof;
+
+						if(cxxTokenTypeIs(g_cxx.pToken,CXXTokenTypeStringConstant))
+						{
+							// assume extern "language"
+
```
masatake: In the future we can record the "language" to "linkage:" parser field:-P

masatake:
LGTM.

Issue #945, opened 2016-05-21
-I option handling in C parser


masatake:
input.c:
```
struct packet_type edsa_packet_type __read_mostly = {
	.type	= cpu_to_be16(ETH_P_EDSA),
	.func	= edsa_rcv,
};
```

New C parser output:

```
$ ./ctags -o - -I __read_mostly foo.c
edsa_packet_type	foo.c	/^struct packet_type edsa_packet_type __read_mostly = {$/;"	s	file:
```

Old C parser output:

```
$ ./ctags --languages=+OldC,-C -o - -I __read_mostly foo.c
edsa_packet_type	foo.c	/^struct packet_type edsa_packet_type __read_mostly = {$/;"	v	typeref:struct:packet_type
```

OldC parser tags edsa_packet_type as a variable. This is my expected behaviour.

pragmaware:
Fixed via #962

PR #962, opened 2016-05-25
Fix bug #945.

pragmaware:
Properly handle struct variable initialization.
Fixes bug #945.


masatake:
I would like to say "thank you" as a user.
This feature is really needed for my daily job.

tags.sh in linux kernel (https://git.kernel.org/cgit/linux/kernel/git/mmarek/kbuild.git/tree/scripts/tags.sh) uses this feature aggressively.

PR #963, opened 2016-05-26
Units, Vim: fix vimball.vim valgrind test failure

dkearns:
This is just a minimal fix but I believe it's in keeping with the general style of the file.


masatake:
LGTM.

PR #967, opened 2016-05-31
Fix #930

### LINKED ISSUE #930: Handling .h file in relaxed syntax
I found this case in linux kernel.

Consider a C language header having .h suffix.
In C language `private` is not a keyword.
New Cxx parser doesn't record a variable named `private` because 
the parser recognized it as C++ keyword.``
Old Cxx parser 

```
[yamato@x201]/tmp% cat foo.h
extern int private;
static inline int f(void)
{
  return private;
}
typedef int int32;


[yamato@x201]/tmp% u-ctags -o - /tmp/foo.h 
# nothing is captured.

[yamato@x201]/tmp% u-ctags --languages=+OldC++ --language-force=OldC++ -o - /tmp/foo.h
f	/tmp/foo.h	/^static inline int f(void)$/;"	f
int32	/tmp/foo.h	/^typedef int int32;$/;"	t

```

We can avoid this ignorance with modifying language map, however I hope u-ctags handling this situation well because this situation may be so popular.

pragmaware:
Be smarter about public/protected/private keywords when parsing .h files as C++.
Targets issue #930.


parsers/cxx/cxx_parser.c
```diff
@@ -563,7 +563,7 @@ boolean cxxParserParseClassStructOrUnion(
 
 	// enable "final" keyword handling
 	g_cxx.bParsingClassStructOrUnionDeclaration = TRUE;
-
+
```
masatake: Newly introduced white space. Could you remove it when merging?

masatake:
Thank you. It works fine.

PR #977, opened 2016-06-05
CXX: Refactor and cleanup [not for merging yet]

pragmaware:
In order to hack-in Objective-C support the C/C++ kind options must be splitted. Well, unless we want to have also "interface" / "implementation" / "protocol" and similar stuff mixed with C/C++ kinds.

The shared kind options were a sort of anomaly since in many (most?) other places C and C++ are treated as separate languages. On the other side this change brings a subtle side effect which is not fully backward compatible: one no longer can rely on C++ kinds being set by using --kinds-c=something. So people/tools that were launching ctags on a mixed source tree with only one set of flags might end up having unexpected or missing tags in the output.

### DIFF
--- a/Tmain/list-kinds-full.d/stdout-expected.txt
+++ b/Tmain/list-kinds-full.d/stdout-expected.txt
@@ -1,13 +1,11 @@
 #LETTER	NAME	ENABLED	REFONLY	NROLES	DESCRIPTION
-c	class	on	FALSE	0	classes
 d	macro	on	FALSE	1	macro definitions
 e	enumerator	on	FALSE	0	enumerators (values inside an enumeration)
 f	function	on	FALSE	0	function definitions
 g	enum	on	FALSE	0	enumeration names
 h	header	off	TRUE	2	included header files
 l	local	off	FALSE	0	local variables
-m	member	on	FALSE	0	class, struct, and union members
-n	namespace	on	FALSE	0	namespaces
+m	member	on	FALSE	0	struct, and union members
 p	prototype	off	FALSE	0	function prototypes
 s	struct	on	FALSE	0	structure names
 t	typedef	on	FALSE	0	typedefs
@@ -16,5 +14,3 @@ v	variable	on	FALSE	0	variable definitions
 x	externvar	off	FALSE	0	external and forward variable declarations
 z	parameter	off	FALSE	0	function parameters inside function definitions
 L	label	off	FALSE	0	goto labels
-N	name	off	FALSE	0	names imported via using scope::symbol
-U	using	off	TRUE	0	using namespace statements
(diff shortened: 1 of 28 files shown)


masatake:
Interesting topic.

> On the other side this change brings a subtle side effect which is not fully backward compatible
Can I ask you to write this change to docs/parser-cxx.rst?

I support your approach.  I would like to call it "copy" approach here.
 `--kinds-*=` notation can be used for brevity to enable/disable common kinds between two parsers.

I took "copy" approach in automake parser.
I dealt with similar case when I worked on automake parser, which is derived from make parser. 
https://github.com/universal-ctags/ctags/commit/980eddd6c29c735d1a45062395d95e2d8da4f806#diff-0ca96ab6cd8a5da2e38290252b631269R86

In the automake parser, I took following approach in the stage initializing stage of automake: 

#. copy the kind array of make parser to the kind array of automake parser, then
#. append automake specific kinds to the kind array of automake parser

With this approach, I can reuse the kind index of make parser in automake parser.

There is another approach. We can assume C++ parser is a parser extended from C parser.
I use an example for explanation. `function` kind is defined C parser. When C++ parser captures a function, the parser records it with `function` kind of C language, not C++ language. I would like to call it "shared" approach.

However, currently, there is no facility for implementing "shared" approach in main part.
To provide the facility, I have to learn more about requirements.

pragmaware:
I'll rework this one to use the shared approach as much as possible.

masatake:
Currently kindOption object is directly passed to `initTagEntry`.
This is the neck for implementing the shared approach.

I will explain this more in with an example.
The super language C defines kind "function."
The super language C++ wants to use the kind in sharing style.
Here sharing means if a ctags user disables "function" kind of C, it affects C++ parser; C++ parser also doesn't record "function" kind in C++ input.

It means C++ needs the way to access the kindOption struct of "function" of C parser.
How? (Here we assume C and C++ parsers are implemented in different .c files.)

Introducing per language header file (e.g. c.h) and putting enum CKind, which defines
names for indexes for each kind of the array of kindOptoin struct, looks good idea.
The implementation of C++ can include it. For recording a function in C++ parser,
C++ parser specifies both "function" kind with enum member and "C" as the owner of the kind.

`initTagEntry` requires kindOption struct itself. So not suitable for this purpose.
New interface may be needed.

```
extern void initTagEntryIndirect (tagEntryInfo *const e, const char *const name,
				  langType kindOwner, int kind_index);
```

Another apraoch is introducing CAbstractParser.
In the example I derrived C++ parser directly from C parser.
How about derriving C, C++, ObjC from CAbstractParser?
The common kinds can be defined CAbstractParser.

I can implement the idea I wrote here and apply it to make/automake parsers.
Cxx parser is not small, so prototyping on make/automake is good idea, I think.

@pragmaware, if you are interested in `initTagEntryIndirect`, tell me. I will start implementing it.
This is very related to #918.

PR #983, opened 2016-06-09
Multiple parsers for nested input

masatake:
If you are interested in performance of ctags, see 
https://github.com/universal-ctags/ctags/commit/c7adb37b4ad441e33295453cc64d224c94546ec4 .

If you are interested in tagging javascript code inside html code, see
https://github.com/universal-ctags/ctags/commit/e1e639e438e867f3aeef6efb3a25d85f54f0fd56

There are multiple ways to combine parsers for one input file.  These commit for just one of them.

Enjoy!


masatake:
As always, I forgot updating vcxproj files.

PR #985, opened 2016-06-11
CXX: Split parser kinds

pragmaware:
Split the C and C++ parser kinds. 
Use the "index sharing" splitting technique also for the fields.

### DIFF
--- a/Tmain/cxx-how-kinds-defs-are-shared-or-copyed.d/run.sh
+++ b/Tmain/cxx-how-kinds-defs-are-shared-or-copyed.d/run.sh
@@ -1,0 +1,20 @@
+# Copyright: 2016 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
+printf "[C] enabling: "
+${CTAGS} --quiet --options=NONE \
+	 --kinds-C=+f --list-kinds=C | grep ^f
+
+printf "[C] disbaling: "
+${CTAGS} --quiet --options=NONE \
+	 --kinds-C=-f --list-kinds=C | grep ^f
+
+printf "[C++] enabling: "
+${CTAGS} --quiet --options=NONE \
+	 --kinds-C=+f --list-kinds=C++ | grep ^f
+
+printf "[C++] disbaling: "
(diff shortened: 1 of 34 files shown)


masatake:
I don't understand what "shared" means.

In my understanding there are two aproach: "share" and "copy".
I will try to explain my idea here.

If two language shares a kind (e.g. `f`),  enabling the kind in one language affects the other language.
Exuberant-ctags uses "shared" model in C and C++ parsers.
Let's see it:
```
[yamato@x201]~/var/ctags-github% ctags --C-kinds=-f --list-kinds=C++ | grep ^f 

f  function definitions [off]
[yamato@x201]~/var/ctags-github% ctags --C-kinds=+f --list-kinds=C++ | grep ^f

f  function definitions
```

Enabling/disabling f kind of C parser is observed in C++ language.

In copy approach,  enabling a kind in one language doesn't affect the other language.
It looks for me that your new commit chooses copy approach.
```
[yamato@x201]~/var/ctags-github% u-ctags --C-kinds=+f --list-kinds=C++ | grep ^f
f  function definitions
[yamato@x201]~/var/ctags-github% u-ctags --C-kinds=-f --list-kinds=C++ | grep ^f
f  function definitions
```

Enabling/disabling f kind of C doesn't affect f kind of C++ language.


Thank you for reading. I would like to agree with you about what is "share" and "copy" means.
We, developers, use the same term for different meaning is horribly bad. Basically I think you
dont decline my models I wrote here:-)

Then I would like to hear,"share", or "copy", which is suitable for C/C++/ObjC parsers.

pragmaware:
Right, from this point of view the approach in this pull request is "copy" while the current master code is still "shared". I copied the kinds that are used in both parsers.

With my "shared" term I meant that a part of the kind indices is now shared between the parsers while in my previous attempt to split the kinds all the indices were separate (and I used variables to store them).

masatake:
@pragmaware, I see. I would like to take more time for this commits. Please, wait for awhile. My interests are (1) kinds and roles related to c preprocessor , and (2) and incompatiblity between exuberant-crags.

masatake:
LGTM.

PR #990, opened 2016-06-16
Cxx: fix debuggging build

masatake:
Building with setting  CXX_DEBUGGING_ENABLED 1, gcc reports an error when
compiling cxx_scope.c.


pragmaware:
Ow.. missed that. Thnx!

Issue #63, opened 2014-08-07
make ctags a library


masatake:
"Library" in ctags may have 3 aspects:
1. reading tags file,
2. backend language parser, and
3. running ctags.

The scope is this issue is "3. running ctags."
We have to research Geany.

Decreasing global variables will be the initial step.

dtikhonov:
@b4n, why not `fmemopen(3)` instead of MIO?

b4n:
@dtikhonov Well, first, I didn't know about it :)  Then, and probably a reason why I didn't find about it at the time (2011), it's not very portable (in POSIX.1-2008 but not before, otherwise seems to be `_GNU_SOURCE`).

So it'd be very nice, but I'm really not sure it's an acceptable dependency (would have to be checked on all platforms, also for consistency as if it's pre-POSIX, it's likely to have slight inconsistencies -- and we can't really reimplement it where missing as FILE is totally opaque).

elextr:
@b4n, isn't another reason for not parsing in another thread to prevent having to make a copy of the Scintilla buffer.  ATM the parser blocks Scintilla, so it can't move the buffer (by re-allocating it) or the gap during a parse.

masatake:
libctags.a is introduced. Though it is not a generic library, but I think I can close this now.
Refining the API for Geany is the next step.

PR #994, opened 2016-06-21
Add interactive `--json` command mode

tmm1:
This PR adds a new interactive mode which can be used to invoke ctags programatically. In this mode, the ctags binary receives json commands and outputs json responses.

If libjansson is not available during build time, the behavior is unchanged:

```
$ ./ctags --json
ctags: Unknown option: --json
```

With libjansson available during build, a new interactive mode can be started with the `--json` option:

```
$ echo | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
```

At the moment, only one command named `generate-tags` is implemented. It can take a filename to read from disk:

```
$ echo "class Test; def foobar() end; end" > test.rb

$ echo '{"command":"generate-tags", "filename":"test.rb"}' | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
{"_type": "tag", "line": 1, "name": "Test", "kind": "class", "path": "test.rb", "language": "Ruby", "pattern": "/^class Test; def foobar() end; end$/"}
{"completed": "generate-tags"}
```

or it can be fed file contents directly over stdin by specifying the size in the preceeding json command:

```
$ wc -c test.rb
  34 test.rb

$ (echo '{"command":"generate-tags", "filename":"fake.rb", "size":34}' ; cat test.rb) | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
{"_type": "tag", "line": 1, "name": "Test", "kind": "class", "path": "fake.rb", "language": "Ruby", "pattern": "/^class Test; def foobar() end; end$/"}
{"completed": "generate-tags"}
```

Multiple commands can be fed into a single session, allowing for example an IDE to keep a long-running instance of `ctags` around to generate tags for files as they are changed by the user.


tmm1:
Errors and warnings are also reported in json in this mode:

```
$ echo '{}' | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
{"error": "command name not found", "fatal": true}

$ echo '{"command":"fake"}' | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
{"error": "unknown command name", "fatal": true}

$ echo '{"command":"generate-tags", "filename":"nosuchfile"}' | ./ctags --json
{"name": "Universal Ctags", "version": "0.0.0"}
{"warning": true, "error": "cannot open input file \"nosuchfile\"", "errno": 2, "perror": "No such file or directory"}
{"completed": "generate-tags"}
```

masatake:
Interesting.
I'm reading your changes.
If would like to change main part before merging your code.
After changing I would like you to rebase your changes on my changes.

main/main.c
```diff
@@ -546,7 +609,11 @@ extern int main (int __unused__ argc, char **argv)
 	verbose ("Reading initial options from command line\n");
 	parseCmdlineOptions (args);
 	checkOptions ();
-	makeTags (args);
+	if (Option.json)
```
masatake: (masatake should do): main part should provide pluggin interface.

main/routines.c
```diff
@@ -202,11 +206,45 @@ extern const char *getExecutablePath (void)
...
+	free (reason);
+}
+#endif
+
 extern void error (
 		const errorSelection selection, const char *const format, ...)
 {
 	va_list ap;
 
+#ifdef HAVE_JANSSON
```
masatake: (masatake should do): main part should provide pluggin interface for replacing error handler.

main/entry.c
```diff
@@ -1361,7 +1423,9 @@ static void writeTagEntry (const tagEntryInfo *const tag)
 		return;
 
 	DebugStatement ( debugEntry (tag); )
-	if (Option.xref)
+	if (Option.json)
```
masatake: (masatake should do): main part should provide pluggin interface for replacing output handler

masatake:
See #995.  I hope your commits can be simplified a bit with #995.

I have not read all your changes, yet but I have some comments.

0. new plug able interfaces

  Could you look at #995?
  If the intention of the changes is not clear, ask me a question.

1. About json tag writer

`--json` option change both output routine (tag writer) and the main loop.

However, it is better to provide the way using one of two separately.
For example, I would like to use json output with normal batch mode.
In #995, I provide a new option `--output-format`. You can adapt your json
tag writer to the option. An error message printer is also plug able in
#995. You can replace the error message printer when `--output-format=json`
is given.

I think the output format is not stable enough. So I would like you to
include the version of json output to the output like `{"json-output-version": "0.0"}`
or something. (This is just an example. I don't know well about json.)

If it is possible I would like you to merge the json version output and pseudo
tags facility. This is optional required. Basically this is my homework. However,
any prototype helps me to make a direction.

Test cases under Tmain directory is needed.
See Tmain/output-format-option.d of #995. You can extend it for json output.

It possible I would like you to update docs/*.rst for this new feature.

2. About interactive main loop

This is really interesting. I don't understand the issues in this area yet.
So basically I will accept your changes. Could you make a PR after I merge
"json tag writer" ?

Some issues I have recognized.

* Like the version of json output format I proposed above, having the version of json
  input format is also nice.

* Run parser in a sub process. Many parts of ctags expect `error(FATAL,...` will stops
  the current process; no cleanup is done. Your json error handler breaks the assumption.
  Memory leaks and other troubles will be triggered with your json main loop.
  For avoiding them, run a parser in a sub process for each input file.

* Test cases under Tmain are needed for the json main loop.

* Updating documentation is needed.

Thank you for reading.

masatake:
One more comment.


See `Features` array in `main/options.c`.  You should add `json`. 
So users (and scripts written by them) can detect whether the json feature is avaiable or not in stable way: `--list-features`.

tmm1:
> See Features array in main/options.c. You should add json. 
> So users (and scripts written by them) can detect whether the json feature is avaiable or not in stable way: --list-features.

Thanks. Can I add jansson as a required dependency for ctags, or should I make it optional? If jansson is not available during build time, `--list-features` should not list `json`?

tmm1:
> Can I add jansson as a required dependency for ctags, or should I make it optional?

Never mind, I left it optional. I have opened https://github.com/universal-ctags/ctags/pull/1010 which adds `--output-format=json`

masatake:
The output side is done. The next is input side.
I'm look into the following sub issue for input:

> Run parser in a sub process. Many parts of ctags expect error(FATAL,... will stops
the current process; no cleanup is done. Your json error handler breaks the assumption.
Memory leaks and other troubles will be triggered with your json main loop.
For avoiding them, run a parser in a sub process for each input file.

Before implementing facility to run a parser in a process, I think I should clean up the main parser loop.
This will be the hardest refactoring I have ever tried in this project because it is related to too many areas and issues. Give me time.

tmm1:
> Before implementing facility to run a parser in a process, I think I should clean up the main parser loop.
> This will be the hardest refactoring I have ever tried in this project because it is related to too many areas and issues. Give me time.

Sure, if I can help in some way let me know.

> Many parts of ctags expect error(FATAL,... will stops
> the current process; no cleanup is done. Your json error handler breaks the assumption.

I understand this is a very tricky problem as ctags was not written to be a library. I think for v1 of the json input feature, we can force the process to exit after printing the fatal error message. Since ctags is in C, it is very fast to spawn new instances of it so long-lived process with proper error handling is not a strong requirement.

tmm1:
One option for the interactive feature is to use `--json-rpc` and conform to http://www.jsonrpc.org/specification for request and response payloads. I do not have much experience with jsonrpc, but it looks simple to try and see if it is a good fit.

tmm1:
> I do not have much experience with jsonrpc, but it looks simple to try and see if it is a good fit.

Note that in my original design source code "can be fed file contents directly over stdin by specifying the size in the preceeding json command". Such an option is not available as part of the jsonrpc specification, so we cannot ever be 100% jsonrpc compliant. Feeding source code over stdin is a strong requirement for my application (for example, to generate tags for IDE buffer before the user saves to file or to process source code stored in a database).

codebrainz:
> for example, to generate tags for IDE buffer before the user saves to file or to process source code stored in a database

IMO, it would be way more generally useful to IDEs to implement #63 than to require them to implement JSON readers and subprocess/socket IO. If #63 existed, the JSON/socket stuff could easily be layered ontop of it too. Also, if it requires running a separate process (and blocking on it), it will be too slow to work in real-time as-you-type in an editor.

tmm1:
I've reimplemented the remaining features on top of master in a new branch, as `ctags --interactive`: https://github.com/universal-ctags/ctags/compare/master...tmm1:interactive-mode

However, It's not quite working yet.. I'm calling `createTagsForEntry()` from the new `interactiveLoop()`, but no tags are being printed to stdout.

Ideally we would also print out all the ptags once on boot.

masatake:
I got

```[jet@localhost ctags]$ cat input
{"command":"generate-tags", "filename":"test.rb"}[jet@localhost ctags]$ cat input;echo
{"command":"generate-tags", "filename":"test.rb"}
[jet@localhost ctags]$ cat input | ./ctags  --interactive=json
{"completed": "generate-tags"}
{"_type": "tag", "name": "Test", "path": "test.rb", "pattern": "/^class Test; def foobar() end; end$/", "kind": "class"}
```

with following patch

```
[jet@localhost ctags]$ git diff  | cat
index c7f0b47..83b44d3 100644
--- a/main/entry.c
+++ b/main/entry.c
@@ -592,6 +592,15 @@ static void writeEtagsIncludes (MIO *const fp)
 	}
 }
 
+extern void flushTagFile (void)
+{
+  TagFile.numTags.added = 1;
+  mio_flush (TagFile.fp);
+  abort_if_ferror (TagFile.fp);
+  sortTagFile ();
+  remove (tagFileName ());
+}
+
 extern void closeTagFile (const boolean resize)
 {
 	long desiredSize, size;
diff --git a/main/main.c b/main/main.c
index ff813d5..c8554e3 100644
--- a/main/main.c
+++ b/main/main.c
@@ -531,14 +531,16 @@ void interactiveLoop (cookedArgs *args, void *user __unused__)
       if (size == -1) { /* read from disk */
         createTagsForEntry (filename);
       } else {			/* read nbytes from stream */
-        uint8_t *data = eMalloc (size);
+        unsigned char *data = eMalloc (size);
         size = fread (data, 1, size, stdin);
         MIO *mio = mio_new_memory (data, size, eRealloc, eFree);
         parseFileWithMio (filename, mio);
       }
-
+      
       fprintf (stdout, "{\"completed\": \"generate-tags\"}\n");
-      fflush (stdout);
+      // fflush (stdout);
+      extern void flushTagFile (void);
+      flushTagFile ();
     } else {
       error (FATAL, "unknown command name");
       goto next;
```

masatake:
```
+  TagFile.numTags.added = 1;
```
This hack is not needed.

masatake:
```
$ cat input | ./ctags  --extra=+p --interactive=json
{"completed": "generate-tags"}
{"_type": "ptag", "name": "TAG_FILE_FORMAT", "path": "2", "pattern": "extended format; --format=1 will not append ;\" to lines"}
{"_type": "ptag", "name": "TAG_FILE_SORTED", "path": "1", "pattern": "0=unsorted, 1=sorted, 2=foldcase"}
{"_type": "ptag", "name": "TAG_PROGRAM_AUTHOR", "path": "Universal Ctags Team", "pattern": ""}
{"_type": "ptag", "name": "TAG_PROGRAM_NAME", "path": "Universal Ctags", "pattern": "Derived from Exuberant Ctags"}
{"_type": "ptag", "name": "TAG_PROGRAM_URL", "path": "https://ctags.io/", "pattern": "official site"}
{"_type": "ptag", "name": "TAG_PROGRAM_VERSION", "path": "0.0.0", "pattern": "360887e"}
{"_type": "tag", "name": "Test", "path": "test.rb", "pattern": "/^class Test; def foobar() end; end$/", "kind": "class"}
```
ptags are rendered well.

masatake:
We have to think more about what is `flushTagFile`.

masatake:
O.k. I will implement something like flushTagFile.


I wonder how to handle `{"completed": "generate-tags"}`.

tmm1:
> O.k. I will implement something like flushTagFile.

Thanks.



> I wonder how to handle {"completed": "generate-tags"}.

I think this can remain inside the `interactiveLoop`. It should be emitted _after_ the flush, so that the receiver knows it is possible to send another command and all the response has been sent.



> ptags are rendered well.

For ptags, I would prefer if they were only emitted once automatically when the program is started in interactive mode. So `echo "" | ctags --interactive` will print all the ptags, and then everytime you send a `"command":"generate-tags"` it does not print them again and again.

PR #995, opened 2016-06-22
Make more areas plug able

masatake:
This is for #994.

Interface for three hooks (plug able areas) are introduced.

* handler for printing a tag entry
```
typedef int (* writeEntryFunc) (MIO * mio, const tagEntryInfo *const tag, void *data);
typedef void * (* preWriteEntryFunc) (MIO * mio);
typedef void (* postWriteEntryFunc)  (MIO * mio, const char* filename, void *data);

extern void setTagWriter (writeEntryFunc func,
			  preWriteEntryFunc preFunc,
			  postWriteEntryFunc postFunc);

```

* function for main loop
```
typedef void (* mainLoopFunc) (cookedArgs *args, void *data);
extern void setMainLoop (mainLoopFunc func, void *data);
```

* handler for printing an error message
```
typedef boolean (* errorPrintFunc) (const errorSelection selection, const char *const format,
				    va_list ap, void *data);
extern void setErrorPrinter (errorPrintFunc printer, void *data);
```

An example:
```
extern int main (int __unused__ argc, char **argv)
{
	cookedArgs *args;

	setErrorPrinter (stderrDefaultErrorPrinter, NULL);
	setMainLoop (batchMakeTags, NULL);
	setTagWriter (writeCtagsEntry, NULL, NULL);


```

In addition `--output-format` option is introduced.

### DIFF
--- a/main/routines.c
+++ b/main/routines.c
@@ -17,8 +17,6 @@
 #endif
 #include <ctype.h>
 #include <string.h>
-#include <stdarg.h>
-#include <errno.h>
 #include <stdio.h>  /* to declare tempnam(), and SEEK_SET (hopefully) */
 
 #ifdef HAVE_FCNTL_H
@@ -149,7 +147,7 @@
 /*
  *  Miscellaneous macros
  */
-#define selected(var,feature)	(((int)(var) & (int)(feature)) == (int)feature)
+
 
 /*
 *   DATA DEFINITIONS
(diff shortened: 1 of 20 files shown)


tmm1:
:+1: this is a very nice refactoring. I will rebase my PR on top of these new APIs.

main/routines.c
```diff
@@ -149,7 +147,7 @@
 /*
  *  Miscellaneous macros
  */
-#define selected(var,feature)	(((int)(var) & (int)(feature)) == (int)feature)
+
```
vhda: No need for this extra space

vhda:
Nice work once again @masatake

main/routines.c
masatake: How do I say... My recognition is that the region from where I removed the macro is a section in which macros are defined.  Though I don't write about the style rule in this project, using two lines for representing the existence of a section is acceptable.

(Sometimes I also made comments about whitespaces. It is basically because git diff reported them with printing them in red.)

PR #998, opened 2016-06-23
get: Avoid unnecessary and hard-to-read use of the comma operator

b4n:
PR on behalf of @erikd, forwarding geany/geany#1074


masatake:
@erikd and @b4n, thank you.

Issue #992, opened 2016-06-20
cxx can't recognize namespace alias


idning:
namespace a = b;

will cause the whole file unable to get parsed.
```

$ cat a.cpp && ctags a.cpp && cat tags
#include <iostream>
namespace a = b;
int main(){
    std::cout << "helloworld" << endl;
    return 0;
}

ctags: Warning: Language "go" already defined
!_TAG_FILE_FORMAT       2       /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED       0       /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
!_TAG_PROGRAM_NAME      Universal Ctags /Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL       https://ctags.io/       /official site/
!_TAG_PROGRAM_VERSION   0.0.0   /c11568b/

```

idning:
and the version `8213b2c` works well.

masatake:
Reproduced.

pragmaware:
Should be fixed via #1000 .

PR #1000, opened 2016-06-25
CXX: Improve namespace handling, add support for namespace aliases.

pragmaware:
- Fixes issue #992
- Adds support for namespace alises
- Adds support for inline namespaces (marked in properties)

### DIFF
--- a/Units/parser-cxx.r/namespace.cpp.d/input.cpp
+++ b/Units/parser-cxx.r/namespace.cpp.d/input.cpp
@@ -1,10 +1,55 @@
-namespace {
-	void anon_f() { }
-}
 
-namespace a {
-	void a_f() { }
-	namespace b {
-		void a_b_f() { }
+namespace {
+	void anon_f() { };
+	
+	namespace {
+		void anon_anon_f() { };
 	}
 }
+
+namespace a1 {
+	void a1_f() { }
+
+	namespace a2 {
+		void a1_a2_f() { }
+		
+		namespace a3 {
+			void a1_a2_a3_f() { };
+		}
+	}
+	
+	namespace {
+		void a1_anon_f() { };
+	}
+}
+
+namespace b1::b2 {
+	void b1_b2_f() { };
+
+	namespace b3::b4 {
+		void b1_b2_b3_b4_f() { };
+	}
+}
+
+inline namespace c1 {
+	namespace c2 {
+		void c1_c2_f() { };
+	}
+}
+
+inline namespace d1::d2 {
+	void d1_d2_f() { };
+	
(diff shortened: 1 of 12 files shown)


Units/parser-cxx.r/namespace.cpp.d/input.cpp
```diff
@@ -1,10 +1,55 @@
...
+}
+
+namespace e1 = a1::a2::a3;
+
+namespace e2 = b1::b2 __attribute__((abi_tag("blah")));
+
+namespace z1 { };
+
+
+
```
masatake: These lines are not needed.

masatake:
Thank you.
Many unnecessary whitespaces are introduced. Could you trim them? You can find them with `git diff`.

parsers/cxx/cxx_parser_namespace.c
```diff
@@ -54,105 +55,234 @@ boolean cxxParserParseNamespace(void)
...
+
 		if(!cxxParserParseNextToken())
 		{
 			// syntax error, but we tolerate this
-			CXX_DEBUG_LEAVE_TEXT("Implicit EOF in cxxParserParseNextToken");
+			CXX_DEBUG_LEAVE_TEXT("EOF in cxxParserParseNextToken");
 			return TRUE; // EOF
 		}
-
+
```
masatake: This is an example of whitespace, which is not needed.
Before making a PR, please, check your change with git diff. The white spaces are printed with red in git diff output.
See https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#Formatting-and-Whitespace

pragmaware:
Ops... merged before you actually reviewed it as I wanted to work on __attribute__ which involves modifying the same files. It was waiting since some time and I've thought it was OK.
Anyway, I'll trim whitespaces in a commit attached to pull #1005.

masatake:
I'm sorry for taking time. I have small troubles in my health.

I'm reworking on --field option now. During it I found critical issue about these commits.
You intrdouced "name" field for C++. Please, change the name. It is already used as a common field name.
```
% ./ctags --list-fields | grep 'name'
N       name            on      NONE             TRUE     tag name (fixed field)        
K       NONE            off     NONE             TRUE     Kind of tag as full name      
t       typeref         on      NONE             TRUE     Type and name of a variable or typedef
Z       scope           off     NONE             TRUE     Include the "scope:" key in scope field (use s) in tags output, scope name in xref output
p       scopeKind       on      NONE             TRUE     Kind of scope as full name    
-       name            on      C++              TRUE     aliased names                 
```

PR #1003, opened 2016-06-25
Python: rewrite as a tokenizing parser

### LINKED ISSUE #997: Python parser stops after finding `Class(` in a string
Given the following test.py

```python
class Foo:
    def method(self):
        meh = 'Foo('

class NotFound:
    def notfound(self):
        pass
```

`ctags -f- test.py` results in only this:
```
Foo	test.py	/^class Foo:$/;"	c	language:Python
method	test.py	/^    def method(self):$/;"	m	language:Python	class:Foo
```

That's apparently of the string `Foo(` used in a class method, where `Foo` is the class itself.

I've tried to bisect it, bad it's in there since at least ~1 year ago already (cf9350b).

b4n:
**This is an *experimental* PR.**

It imports a token-based Python parser I started a while ago.  It used to work fine, but a lot of changes happened to the current Python parser and U-CTags as a whole since then, so it requires catching up.  I patched it with *most* of the changes that happened since, but it's not yet perfect -- and probably will require a fair bit of rebasing.

- [x] The main problem I have is fixing the imports, among other things because I have no idea what all the roles and separate kinds are supposed to be for -- it doesn't seem to make any sense to me, but I probably am just missing something.

- [x] Also, I didn't want (still) to import the FQN tag generation, because I still don't get why a parser needs to do it on its own -- especially if it uses Cork.  I may also miss something, or maybe it was just implemented manually here because it's not yet done some way else and people wanted it?

- [x] I didn't yet try to import the block end tagging, for 2 reasons:

    1. it was added on top of imports parsing, and this still ain't right as stated above.
    2. It looks wrong to me to manually write a number out (as text, even), while I would imagine this should use the same facility as tag positions, and then follow the pattern/number options.  Is that expected to happen at some point, or is there a reason not to?

- [x] ~~Finally, I don't get why i.e. *parser-python.r/simple.py* fails.  The output doesn't have the same fields at all, but I don't understand why.  If anyone got a clue, I'm all ears.~~ Scratch that, I somehow messed up that file in the merge or somewhere (probably because it was modified from both sides, and I didn't have a building parse to test it at that time).

- [x] Decide whether inheritance should be emitted empty or not at all when there is none.
    i.e. with something like
    ```python
class Foo:
    pass
```
   whether to emit
   ```console
$ ./ctags --fields=+i -o- input.py
Foo	input.py	/^class Foo:$/;"	c
```
    or
    ```console
$ ./ctags --fields=+i -o- input.py
Foo	input.py	/^class Foo:$/;"	c	inherits:
```

But it should fix #997 :)

### CODE REVIEW

**Comment by masatake** [comment]
Thank you very much. I will look into the commits.

> The main problem I have is fixing the imports, among other things because I have no idea what all the roles and separate kinds are supposed to be for -- it doesn't seem to make any sense to me, but I probably am just missing something.

The intent for the roles is for tracking the importing/imported relation between python modules.
Such information may help users to understand where a name comes from.
Howerver, ignore the roles. Tagging definition correctly is the most important. Roles handling can be added after all.

> Also, I didn't want (still) to import the FQN tag generation, because I still don't get why a parser needs to do it on its own -- especially if it uses Cork. I may also miss something, or maybe it was just implemented manually here because it's not yet done some way else and people wanted it?

Yes, requestAutomaticFQTag field of parserDefinition generates FQN. So a parser using CORK should not care about it.

> I didn't yet try to import the block end tagging, for 2 reasons:
>
>    it was added on top of imports parsing, and this still ain't right as stated above.
>    It looks wrong to me to manually write a number out (as text, even), while I would imagine this should use the same facility as tag positions, and then follow the pattern/number options. Is that expected to happen at some point, or is there a reason not to?

I'm working on converting `end` parser own field to a common field; I'm adding `endLine` field to tagEntryInfo structure now. The change touches the area where @pragmaware changed in #1000. So I will push my change after merging #1000.  I will review (and merge) #1000 first.

**Comment by b4n** [comment]
> The intent for the roles is for tracking the importing/imported relation between python modules.
Such information may help users to understand where a name comes from.

Yeah I understood it was supposed to add extra info, but I didn't get what it tried to output for what.  But I re-tried, and I think now I got it.
I got some dumb issues on that area, it's way better now.

---

I still need to get back inheritance report, `end` tags, support for parenthesized imports (`from foo import (a, b as c, d)`), FQN tags and fix *parser-python.r/toplevel-funcall-with-keyword-args*.

I'll work on that later today.

**Comment by b4n** [comment]
> I still need to get back inheritance report

Hum actually I have it just fine, but the old parser reports it empty when there is none, and I don't report it at all.  What do we prefer?

**Comment by b4n** [comment]
> Yes, requestAutomaticFQTag field of parserDefinition generates FQN. So a parser using CORK should not care about it.

Nice!  I converted the parser to really use Cork, and that did indeed make the trick super easily :)
It's also a lot more reliable than doing it manually, as the older parsed show it by missing several cases (mostly, variables as members).

**Comment by masatake** [comment]
I'm suprised that `requestAutomaticFQTag` works fine:-P.
( I have to take more time to improve fields option and to make end field common field.)

**Comment by b4n** [comment]
So, I think I'm about done catching up.  The only remaining question is whether I should generate `inheritance` info if it's empty, or not.  Current parser did, mine here doesn't, but either way is real easy so it's just a matter of deciding which is better.

Also, *parser-python.r/simple.py* still needs for extra fixing, but that's only either irrelevant differences (e.g. how newlines are recorded in argument lists -- space or nothing?), or missing FQ tags in the expected output.
I'm just waiting on the decision about inheritance to fix it completely, so I can have a nicely succeeding test case when all is fixed.

You're welcome to start reviewing this, give feedback and all that.

**Comment by b4n** [comment]
> I'm suprised that `requestAutomaticFQTag` works fine:-P.

Well, looks like you did write it correctly even if not sure, it really seems to do the right thing, congrats :)

**Comment by b4n** [comment]
I went ahead and for the moment restored compatibility with previous parser's output regarding empty inheritance lists, so I could fix up the last failing test.  We still can change this, nothing's written in stone.

**Comment by masatake** [comment]
(`make noise LANGUAGES=Python` is passed)

**Comment by masatake** [comment]
I will convert the cstring based "end" parser own field to an integer based common field AFTER merging this new parser. I have written the code for conversion already. However, the change about it is not small.

**Comment by masatake** [comment]
I'm reading the changes. I have not found any spot till now. The new parser excites me for adding new features: handling decoration for an example.

I found some "FIXME" markers in the code. If possible, could you add `*.b` test cases for the FIXMEs.

I'm looking into the code area where variable kinds are captured:

```
[yamato@x201 wildcard-in-lang-of-fields-option.d]$ cat /tmp/foo.py 
cat /tmp/foo.py 
class F:
    x = 0

z, F.x, y = 1, 1, 1

[yamato@x201 wildcard-in-lang-of-fields-option.d]$ u-ctags -o - /tmp/foo.py 
u-ctags -o - /tmp/foo.py 
F	/tmp/foo.py	/^class F:$/;"	c
x	/tmp/foo.py	/^    x = 0$/;"	v	class:F
z	/tmp/foo.py	/^z, F.x, y = 1, 1, 1$/;"	v
```

Here `y` is not captured. Is this known issue?

**Comment by b4n** [comment]
> I found some "FIXME" markers in the code. If possible, could you add `*.b` test cases for the FIXMEs.

Sure.

> […]
> Here `y` is not captured. Is this known issue?

No, it wasn't.  But now it is, and I'll try and fix it.
Though, it's kinda tricky: should it output `F.y` too, or not?  It's very possible to add members to foreign classes like that, so maybe it should.  And if it should output it, it should also give it a proper parent, which might be a little tricky.

**Comment by b4n** [comment]
I added tests for the 2 relevant FIXMEs in the code.
However, the test for `u`/`r`/`b` string is kind of pointless, because no matter how it's handled it would still result in the same output.  This particular FIXME in the code is an actual question, not a known limitation: I actually have no idea of what not handling prefixes as such could break.  So if anybody got any idea of how this could affect parsing, please tell me and/or provide a test case.

The last FIXME is related to what output we actually want, so I can't write a test :)

**Comment by b4n** [comment]
> I will convert the cstring based "end" parser own field to an integer based common field

Is an integer good enough?  We currently provide both line number and file position for the tag location, wouldn't the same be required for end location?

**Comment by b4n** [comment]
Fixed support for multiple variable declarations on the same line (the most relevant FIXME there was), and support for semicolons.

File: parsers/python.c [hunk 1]
```diff
@@ -107,1237 +132,1052 @@ static kindOption PythonKinds[] = {
...
- * We don't return the lambda name since it is useless for now since we already
- * know it when we call this function, and it would be a little slower. */
-static boolean varIsLambda (const char *cp, char **arglist)
+static boolean parseVariable (tokenInfo *const token)
 {
-	boolean is_lambda = FALSE;
+	/* In order to support proper tag type for lambdas in multiple
+	 * assignations, we first collect all the names, and then try and map
+	 * an assignation to it */
+	tokenInfo *nameTokens[8] = { NULL };
```
b4n: 8 name tokens is an arbitrary limit I made up, but I believe it's way large enough for non-crazy code.

However, we can easily augment this by simply changing this value here as the rest of the code only assumes it's at least 1, and otherwise relies on the actual element count.

masatake:
> Is an integer good enough? We currently provide both line number and file position for the tag location, wouldn't the same be required for end location?

As far as I can remeber the position is not printed in tags format. It uses only in TAGS output.
The position is typed MIOPos in tagsEntryInfo. That means, if I understand correctly, the value cannot be printed and be used as a value for position calculation.  So I think, I cannot do interesting things even if we record the end position. When I designed 'promise' api, I was frustrated with fpos_t.

masatake:
> Though, it's kinda tricky: should it output F.y too, or not? It's very possible to add members to foreign classes like that, so maybe it should. And if it should output it, it should also give it a proper parent, which might be a little tricky.

Let's forget about `F.y`. We cannot say it as a definition strongly. Rather we can capture it as a reference tag in the future. Capturing definition tags and making consensus within developers about what is rerefence tag are more important.

b4n:
> As far as I can remeber the position is not printed in tags format. It uses only in TAGS output.
The position is typed MIOPos in tagsEntryInfo. That means, if I understand correctly, the value cannot be printed and be used as a value for position calculation. So I think, I cannot do interesting things even if we record the end position. When I designed 'promise' api, I was frustrated with fpos_t.

`fpos_t` (and `MIOPos` which internally uses it for the file backend) represent an abstract position in a file, not something you can print.  The advantage over the line number is mostly that it's a lot more precise: it represents a byte position in the file, not merely a line number (which might be the same for a lot of bytes).
So the advantage of `fpos_t` would be that you can i.e. get a position offset out of it, seek to that location and get the text around, etc.  Basically, have a byte-level granularity instead of a line-level one.

Alternatively maybe we could record a line+column (or byte offset on the line) position and that would be enough/better.

b4n:
> Let's forget about `F.y`. We cannot say it as a definition strongly. Rather we can capture it as a reference tag in the future. Capturing definition tags and making consensus within developers about what is rerefence tag are more important. 

Currently the parser doesn't differentiate any variable definitions strongly.  It just detects assignations, which might not be the same as definitions.  The only thing that make it mostly right is that they are only recorded for top/class level, so it's unlikely to contain too much mutable code.

But indeed, it's harder to get it right for the `F.y` case, because it might have been declared outside the current scope, so it wouldn't be a mere matter of seeing if it's already defined in the current or parent scope, but if would require checking if it's not set *anywhere* else.

But anyway, something should probably be done in the parser for that, it being recording it or ignoring it explicitly, because currently it would break parsing of other assignations on the same line.

b4n:
I just added a few more features (that might be questionable, and if so please question them):

* tagging function parameters (like the new C parse can do), as `z` (disabled by default)
* tagging of local variables, as `l` (disabled by default)
* always tag local lambdas, as we for local functions.  this is questionable on two grounds:
  * are functions and lambdas really the same from a tagging perspective?
  * should we really tag local functions by default if tagging local variables is disabled by default?  (I believe it makes sense, simply because local variable will generate a lot of tags that might be considered noise, whereas local functions are less common and more useful to tag)

b4n:
And here goes even more fancy things.  Sorry I'm now playing with it and it's fairly satisfying :)

@masatake 3c8dcf126acf218bd484efa410c868f98e1454e8 and here you go: support for reporting decorators :)  Comments on expected output welcome.

masatake:
About "inherits:" we should add an empty one if `--fields={inherits}' is given because with an empty "inherits:", people and their tools can know the following facts:
1. --fields={inherits} is given when generating the tags file, and
2. no parent class exists

If the field is removed from the output because there is no parent class, people cannot designate above two facts.

If one says a tags file is too large, I will say you can use 'sed' for trimming unwanted fields.

masatake:
Could you add "python" to "Fully rewritten parsers" of docs/news.rst?
If you have an extra node about design and implementation, write them to docs/parser-python.rst.

masatake:
I made a PR about keyword installation to your tree.

About decorator, I have wondered which is better keeping arguments of a decoration or dropping them.
Your choice is dropping. After thinking dropping will be better. If we need arguments, we should handle it in a reference tag for the decorator.

I have no more suggestions. Please, merge this when you are o.k.
I would like to say once as a co-developer of u-cutags; and say again as a user of u-ctags.

b4n:
> About "inherits:" we should add an empty one if `--fields={inherits}' is given because with an empty "inherits:", people and their tools can know the following facts:
1. --fields={inherits} is given when generating the tags file, and

How is this interesting?  And if it is, shouldn't it rather be reported using a pseudotag listing enabled kinds/fields?

> 2. no parent class exists

Fair point.

Should I apply the same rationale to new `decorators` field?  it's kind of similar, but would appear a lot more often without a value, as decorators are far less common.

> I made a PR about keyword installation to your tree.

Cherry-picked, thanks.

>About decorator, I have wondered which is better keeping arguments of a decoration or dropping them.
Your choice is dropping. After thinking dropping will be better. If we need arguments, we should handle it in a reference tag for the decorator.

Hum, my choice was keeping them, as [seen in the new test](https://github.com/universal-ctags/ctags/pull/1003/files#diff-fd99a075efa233098208dd6619e1f49fR2); if you have an example where it drops them, please show it to me so I can try and fix it :)
Though, it's a fair question: do we actually want the argument list for decorators?  I would think we do, because it changes the meaning of the decorator, but maybe it has drawbacks or should be reported some other way (reference tag as you mentioned?).

> I have no more suggestions. Please, merge this when you are o.k.

I still need to do something about `x, F.y = 1, 2`, which currently leads to ignoring the whole declaration (including `x`).  If we don't make a decision on what should be done for those, I'll at least make the parser ignore them so the rest of the declarations will be properly tagged.
Though, this case should be fairly rare in practice, but who knows.

Also, should I squash some commits?  Some of those make sense (like 5eb1888, *fixup for e2a894e*), but most don't.  Even the "merge <id>* ones probably shouldn't be squashed, because they are real changes to the new parser, and squashing them in the merge commit would make the merge commit awful, and merging them before wouldn't help the merge commit.  So if anything, those should be reworded, or left alone.

b4n:
About dotted variable assignments, I added a commit explicitly ignoring them not to break other variables in the statement.  What we should actually do with those is still in question, but at least they don't break other things.

masatake:
> Should I apply the same rationale to new decorators field? it's kind of similar, but would appear a lot more often without a value, as decorators are far less common.

hmm...Could you apply the same rule to the new decorators field?
One of possible objection for such empty fields may be too noisy. However, whether printing or not printing empty fields should be controlled in the main part. I would like to ask parser developers to pass the information to the main part. If the we got too many objections, we can add `--suppress-empty-fields` option.

> Hum, my choice was keeping them, as seen in the new test; if you have an example where it drops them, please show it to me so I can try and fix it :)

Oh, sorry. I misread the lines. `skipOverPair`` is not just skipping. The skipped bytes are stored to `repr`.

> Though, it's a fair question: do we actually want the argument list for decorators? I would think we do, because it changes the meaning of the decorator, but maybe it has drawbacks or should be reported some other way (reference tag as you mentioned?).

I agree with you. Let's record arguments, too.

If you didn't record just decorator names as I misread I will record arguments for a decorator in a reference tag like:
```
     adecorator foo /^@adecorator('a', 'b')/;" kind:decorator role:decorating-class target:aClass arguments:('a', 'b')
```
We don't have serious application, so it is not easy to tell what is the best.

>>   I have no more suggestions. Please, merge this when you are o.k.

>I still need to do something about x, F.y = 1, 2, which currently leads to ignoring the whole declaration (including x). If we don't make a decision on what should be done for those, I'll at least make the parser ignore them so the rest of the declarations will be properly tagged.
Though, this case should be fairly rare in practice, but who knows.

How about merging the current version? We can rethink this after getting a bug report.

> Also, should I squash some commits? Some of those make sense (like 5eb1888, fixup for e2a894e), but most don't. Even the "merge * ones probably shouldn't be squashed, because they are real changes to the new parser, and squashing them in the merge commit would make the merge commit awful, and merging them before wouldn't help the merge commit. So if anything, those should be reworded, or left alone.

Yes, I would like to squash the commits including my commits sent to you because this is a new parser.

masatake:
s/I would like to squash/I would like you to squash/.

b4n:
> Yes, I would like to squash the commits including my commits sent to you because this is a new parser.

I'm having trouble rebasing this cleanly, because there was several merges in the way and they introduce incompatibility.
I can probably fairly easily just make one big commit of it all, but that wouldn't be great either.  Maybe I can make a fat commit of the ~early ones and try and keep newer relevant ones, but it'll be some work.

masatake:
Oh, I see. One big patch is acceptable of course. 

How about two patches? One introduces the new parser and modifies existing test cases. Another introduces new test cases for the new parser.

masatake:
Yesterday, I proposed "two patches" aporach.


I tried to make the first patch: https://github.com/masatake/ctags/tree/new-python-parser-rebased-on-master .

What I did in the first patch are (0) making a branch on master and checkout it, (2) copy new python.c to the branch, (3) running make check, (4) reviewing the test result and accepting all of them.

If you are o.k. about what I did in my branch, could you clone the branch, rewrite my temorarory commit log and cherry-pick new test cases in your branch?

(I should call my proposal 1+n patches, not two patches.)

masatake:
@b4n, how about this one?
Should I take over this?

b4n:
@masatake nah, I just need to find a little time to deal with it, but I will.

masatake:
Sorry for hurrying you up. Sure I will wait the merge.

b4n:
don't worry, it's good to remember me from time to time if I still didn't do it, because some other stuff might have got in the way and shadowed this one more than it should have :)

b4n:
@masatake When I finish the cleanup, should I force-push here or make another PR? 
There's a fair bunch of comments here, and maybe it'd be easier to review for you if you can have both branches?  Just tell me.

b4n:
(I just re-pushed a version without conflicts and up2date with master to easy reviewing the rebased version)

b4n:
First rebase proposal at https://github.com/b4n/fishman-ctags/tree/python-rewrite-rebase.  It only merges commits that as required to have a fully buildable history: you can successfully build and test each commit (thanks to `git rebase --exec` it was fairly easy to check).

I find this nice as it gives us some of the history, yet removes the ugly merges and incremental non-working catch-up with master state.

But now I rebased the hard part, I can very easily do whatever you like (so long as it has *less* granularity).

PS: please not make *too* many changes to *python.c* in master until this gets merged, you'd be all very kind :)

b4n:
Somehow the `python-decorators` tests didn't get tested with `units LANGUAGES=Python` so I missed that it failed.  Now fixed, I will backport in the rebase branch

PR #1005, opened 2016-06-27
CXX: Support gcc __attribute__(()) syntax

### LINKED ISSUE #1001: A problem with __attribute__
Hello,

The universal ctags (using f6dafc0609db8a478c9791eb39a65494ec0ac921) failed to detect a function that use \_\_attribute__ statement for a parameter.

In the header libcryptsetup.h from cryptsetup-1.4.2.tar.xz package:

    int crypt_repair(struct crypt_device *cd,
		 const char *requested_type,
		 void *params __attribute__((unused)));

This command returns empty output:

    ctags -x --c-kinds=fpvx libcryptsetup.h | grep crypt_repair

What -I flag should I provide to detect this function?

Thank you.

pragmaware:
Add support for the gcc __attribute__(()) syntax. Try to extract some information from it.
Fixes #1001


masatake:
I would like to review this. Could you wait for awhile?

masatake:
(I'm reviewing this.)

parsers/cxx/cxx_parser_internal.h
```diff
@@ -220,7 +220,9 @@ typedef enum _CXXParserKeywordState
 	// "const" has been seen at block level
 	CXXParserKeywordStateSeenConst = (1 << 9),
 	// "volatile" has been seen at block level
-	CXXParserKeywordStateSeenVolatile = (1 << 10)
+	CXXParserKeywordStateSeenVolatile = (1 << 10),
+	// __attribute__((deprecated)) has been seen
+	CXXParserKeywordStateSeenAttributeDeprecated = (1 << 11)
```
masatake: I recommend put `,` at the end of newly introduced entry. So the next time introducing new one, you don't have to append `,` at the end of its previously line. That makes changes smaller. Smaller changes help us to understand them.

parsers/cxx/cxx_tag.h
```diff
@@ -132,7 +132,9 @@ typedef enum _CXXTagProperty
 	// Template specialization a<x>()
 	CXXTagPropertyTemplateSpecialization = (1 << 13),
 	// Template specialization of scope a<x>::b() (which implies TemplateSpec too)
-	CXXTagPropertyScopeTemplateSpecialization = (1 << 14)
+	CXXTagPropertyScopeTemplateSpecialization = (1 << 14),
+	// __attribute__((deprecated)) has been seen
+	CXXTagPropertyDeprecated = (1 << 15)
```
masatake: Ditto. 

```
+	CXXTagPropertyDeprecated = (1 << 15),
```

masatake:
LGTM. Thank yo.
I added minor comments for making future changes smaller.

parsers/cxx/cxx_parser_internal.h
codebrainz: It's not [legal C89](http://port70.net/~nsz/c/c89/c89-draft.html#3.5.2.2) (I assume uctags uses C89 since it doesn't use `AC_PROG_C99` in `configure.ac` to get C99 features). Kind of pedantic though, just sayin :)

masatake: I would like to use C99. 
Can I ask make a PR adding AC_PROG_C99 to configure.ac?

PR #1006, opened 2016-06-28
Improve reStructuredText coverage

b4n:
…and add a note that heading width computation isn't actually accurate (according to Docutils' *rst2html* at least), but that's way too tricky to fix just now, and possibly ever (requires full Unicode character width understanding).

### DIFF
--- a/Units/iso8859-1-restructuredtext.d/expected.tags
+++ b/Units/iso8859-1-restructuredtext.d/expected.tags
@@ -1,0 +1,4 @@
+Chàptèr 1	input.rst	/^Chàptèr 1$/;"	c
+Subsubsection 1.1.1.1	input.rst	/^Subsubsection 1.1.1.1$/;"	t	subsection:Sùbs
+Séçtiön 1.1	input.rst	/^Séçtiön 1.1$/;"	s	chapter:Chàptèr 1
+Sùbs	input.rst	/^Sùbs$/;"	S	section:Séçtiön 1.1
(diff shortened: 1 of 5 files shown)


masatake:
Thank you. LGTM.

All issues will be solved by the developers of galactic-ctags in the 22 century.

vhda:
:laughing:

b4n:
> All issues will be solved by the developers of galactic-ctags in the 22 century. 

Hehe :]

Though, that's something we could actually fix if we had any Unicode-aware library, and that would be the case if you really want to introduce GLib dependency, it could help here with something like this (untested in the context of RST):
```C
#include <stdio.h>
#include <glib.h>

static gint utf8_display_width (const gchar *str)
{
  gint width = 0;
  
  if (! g_utf8_validate (str, -1, NULL))
    return -1;
  
  for (; *str; str = g_utf8_next_char (str))
  {
    gunichar ch = g_utf8_get_char (str);

    if (g_unichar_iswide_cjk (ch))
      width += 2;
    else if (! g_unichar_iszerowidth (ch))
      width++;
  }
  
  return width;
}

int main (int argc,
          char **argv)
{
  while (*(++argv))
    printf ("'%s' is %d wide\n", *argv, utf8_display_width (*argv));
  
  return 0;
}
```
Though, properly supporting this is not necessarily so useful, as the only case where it'd break our parsing is if with zero-wide characters, and if those aren't compensated by wide ones.  So yeah it can happen, but how common is that.

PR #1007, opened 2016-06-28
Vim: fix matching of global variables in functions

dkearns:
Don't match local variables without a scope prefix named g* as global
variables.

### DIFF
--- a/Units/parser-vim.r/vim-let-in-function.d/expected.tags
+++ b/Units/parser-vim.r/vim-let-in-function.d/expected.tags
@@ -1,0 +1,2 @@
+Func	input.vim	/^function Func()$/;"	f
+g:global_var	input.vim	/^  let g:global_var = 42$/;"	v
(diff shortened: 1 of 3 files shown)


Units/parser-vim.r/vim-let-in-function.d/expected.tags
```diff
@@ -0,0 +1,8 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
+!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
+!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
+!_TAG_PROGRAM_VERSION	0.0.0	/56612cf/
```
masatake: You can trim the pseudo tags from the expected.tags.
units scripts automatically trims them when comparing actual output.

BTW, if you want to test pseudo tags, you may have to use tmain.

vhda:
:+1:

dkearns:
Sorry about that.  I forgot.

I've added a note to remind myself and fixed this PR.

PR #1008, opened 2016-06-28
Vim style fixes

dkearns:


dkearns:
Are tabs supposed to be used only for indenting the start of a line or anywhere there is four consecutive spaces?  There seems to be a mix in the files I surveyed.

Apparently the GitHub 'viewer' respects http://editorconfig.org/.  Is it worth adding?

vhda:
What has been done so far is if a new file is added it should follow: https://github.com/universal-ctags/ctags/blob/master/.indent.pro (configuration file for `indent`). Otherwise indentation is only corrected, compared to the output of `indent, if that line is modified by a functional correction.

masatake:
I have not defined a style rule yet.

Basically, what I, an emacs user, do is:

1. Using `linux` style which can be enabled with `C-c .`,
2. Checking red markers in git diff

I think above rules may be enough for the main part which I mainly maintained.
About the each parser, the maintainer of the parser should/can define if the one wants.

Not a few books about programming explain about the importance about the style used in a project.
However, ... it is not so attractive for me. Using variety style in ctags source code is good for testing our C parser:-P.

@dkearns, I don't know well about I don't know well about vim script, so I would like you to maintain it.
In such case, you can rearrange the style of vim.c as you want. You can define your own rule.
Idealy it should be documented or be defined as a config of editors. Splitting the commits like this pull request are REALLY nice. Thank you. I would like to avoid random changes in a commit without strong reason.

In addition, if you are interested in defining the rule used in whole universal ctags, I will not object your rule of style. However, some developers may use their own IDE. So I'm not sure all of them agree with the rule.

I'm not sure what is the idea. For example, I got a comment about using `_`(under score) in names.
I would like to use it. I don't like came case. However, I don't want to make "big diff" for changing the names from camel style to underscore style. For making smaller diff I accept camel case in older code.
In new code sometimes I used under score style. Because in new code, style doesn't have impact on the size of diff. 

Readable diff is much important than style. I think you may know well this policy because you made this pull request from the two commits 00638e2 and be57b9b.

We should write down the policy somewhere in docs. However, again, it is not attractive.

I wonder I make sense.

masatake:
LGTM.

vhda:
@masatake we've had this discussion so many times already and nothing was ever decided. Every time I edit code in Ctags (which hasn't been that common lately, unfortunately) I have to modify my Vim configurations depending on the file I'm looking at.
I agree with the proposal of @dkearns: let's just import `indent` configurations to EditorConfig and start using it once and for all. We can create a branch for that where configuration changes can be added.

Again, having `git blame` returning the correct commits is, for me, brings an added value. It helps a lot when debugging and I've seen other projects (git, for one) following this rule: don't make changes to indentation alone.

masatake:
Can I ask this basic rule to docs/style.rst or somewhere you prefer?
I'll follow the style and rule.

dkearns:
Sorry, is there anything more annoying than the 'new guy' complaining about formatting?

To be clear I don't think EditorConfig has much in the way of features beyond the absolute basics and I was suggesting it specifically to help with the GitHub file viewer which makes a mess of files with the non-standard tabstop value of 4.  The only other way to solve that that I'm aware of is to add a ts=4 query parameter.

I haven't used `indent` for a long time so I'm not sure how reliable it is but if it does a good job it's probably the best idea.  I'm happy to follow the principle of only updating the formatting of otherwise changed lines and diffing against the `indent`ed version though it's probably a bit of a pain for casual committers.

While I personally fall into the grand formatting update camp because I prefer to read consistently formatted code above all else I understand the arguments in favour of less churn so if that's the general principle I can happily remove the whitespace fixes commit.  I think it's probably better to have a project wide style guide. 

@masatake I'm happy to take informal maintainership of the Vim parser.  Feel free to ping me if I don't notice any related issues.  Though it sounds like @vhda is also a Vim user so you're not alone.

masatake:
Sorry to be late. Let's look into this.

@vhda,

> Again, having git blame returning the correct commits is, for me, brings an added value. It helps a lot when debugging and I've seen other projects (git, for one) following this rule: don't make changes to indentation alone.

I would like to confirm the meaning of the sentence.
Do you mean we should not add just-whitespace-cleanup commit?
(if yes...)

I understand the keeping the output of git balme is important. However, I don't want to review a patch including both real code change and whitespace-cleanup; the change about whitespace-cleanup is just noise. I don't like it.

My proposal:

* we will not accept just style only pull request,
* in a pull request commits for whitespace-cleanup and commits for real changes are separated.

> @masatake we've had this discussion so many times already and nothing was ever decided. 

So I wonder what I should do next. 

> I agree with the proposal of @dkearns: let's just import indent configurations to EditorConfig and start using it once and for all. We can create a branch for that where configuration changes can be added.

Is this the what I should do next?

I don't understand the critical requirement and trade-off of this issue.

masatake:
@vhda, can I ask you to review this?
@dkearns is the maintainer of vim.c so, basically @dkearns should merge anything he wants.
However, the changes are related to the style. I'm not sure whether it conflicts with your vision.

The vision should be realized in document or a tool. So people can follow the vision. 
I don't have any vision so this topic will not progress as far as I lead.

dkearns:
git-blame does support ignoring whitespace changes so simple indent corrections and similar shouldn't be a problem in that regard.

masatake:
> 

> git-blame does support ignoring whitespace changes so simple indent corrections and similar shouldn't be a problem in that regard.

Oh, great. In that case I think we should make a separate commit, shouldn't we?

dkearns:
Yes.

vhda:
Unfortunately, free time hasn't been abundant lately.
I wasn't aware @dkearns was maintaining the vim parser, so my apologies.

In my "vision" all files would also have the same indentation... so I guess it will stay a vision ;)

@dkearns feel free to rebase this with master, my vision should not stop your work!

dkearns:
@vhda I'm a bit confused. I was just trying to contribute to your dream before I even knew you dared.

The whitespace patch essentially just :retab-ed the file with tabstop=4 which seems to me to be the dominant project style and that specified in the modeline.

If git-blame -w addresses your previous concern regarding history then you can feel free to reindent the entire project.

I don't mind how it's indented as long as it's consistent. I've seen too much bloodshed over the years to fight in that war again.

masatake:
Of course, Having common indentation style is better. However, about this project, human resource is important; a parser maintainer should be able to do what s/he wants in the parser. So @dkearns, feel free to merge the changes for vim.c.

About the other code we don't have a maintainer, you can lead the direction. These should be done by person who has both vision and time.  

If anyone does nothing in an area, I have to look into the area. However, it will never get higher priroty in my task queue.

dkearns:
OK, I'll look into cleaning up the basics in the next couple of days.

PR #1009, opened 2016-06-28
Vim: add support for :map-<nowait>

dkearns:


vhda:
Are you sure you don't mean `<nowait>` in the subject?

dkearns:
Indeed, thanks.

I'd just found a bug in the map-<special> syntax highlighting...

masatake:
LGTM.

PR #1010, opened 2016-06-28
Add --output-format=json

tmm1:
cc #994 #995 @masatake


masatake:
I reviewed them and made a PR to @tmm1 's tree. (https://github.com/tmm1/ctags/pull/1).

However, there are still two issues.

1. add `jansson` to .travis.yml.
   I'm not sure where I should add `jansson` to.

2. the json format
    Deeeep consideration is needed. Especially about version information. It takes time. However, I'm busy in python parser. Please, wait for awhile.

main/output-json.c
```diff
@@ -0,0 +1,136 @@
...
+
+	if (isFieldEnabled (FIELD_LINE_NUMBER))
+		json_object_set_new (response, getFieldName (FIELD_LINE_NUMBER), json_integer (tag->lineNumber));
+
+	if (isFieldEnabled (FIELD_LANGUAGE)  &&  tag->language != NULL)
+		json_object_set_new (response, getFieldName (FIELD_LANGUAGE), escapeName (tag, FIELD_LANGUAGE));
+
+	if (isFieldEnabled (FIELD_SCOPE) || making_fq_tag)
+	{
+		if (isFieldEnabled (FIELD_SCOPE))
```
masatake: I cannot remember why FIELD_SCOPE must be checked twice.

masatake: I just remembered the reason. I'll make a PR.

tmm1:
> add jansson to .travis.yml.

jansson tests are now passing on travis CI

> the json format

I am open to any suggestions here. I agree adding some type of version field is a good idea, but it seems repetitive to include `"json-version":1.0` in every single line of output.

Currently I have a `"_type":"tag"`. Perhaps it should simply be `"type"`, but I wasn't sure if that would conflict with internal fields. Also maybe we should call it `"type":"definition"` and later we can add `"type":"reference"`?

> more test cases. I will work on this.

I added some more simple tests. If you have ideas for more I can write them.

> Do you have any real application for the json input/output?

Not yet, I am working on some prototypes using this new format and interaction model.

masatake:
Welcome back!

>
> 
>     the json format
> 
> I am open to any suggestions here. I agree adding some type of version field is a good idea, but it seems repetitive to include "json-version":1.0 in every single line of output.

The protocol version should be printed only once when ctags is invoked with --output-format=json option.

Basicaly I think we can use pseudo tags as facility for the purpose.

I'm working on 2 areas:

(1) clean up addExtensionFields function of output-json.c, 
(2) add hook points where printing psuedo tags

Please, wait for my counter pull requests about these topics.

In addition I'll add a draft version of document explaining the json output.


> Currently I have a "_type":"tag". Perhaps it should simply be "type", but I wasn't sure if that would conflict with internal fields. Also maybe we should call it "type":"definition" and later we can add "type":"reference"?

"_type" :"tag" is really good. Definition tag and reference tag should be distinguished with whether "role:" field is avaiable or not; a definition tag doesn't have the field. It implies `role:definition`.

masatake:
I have a question about your planed application.

ctags uses letters for specifying something in wide area.  Kind is typical one.
```
% ./ctags --list-kinds-full=Ruby
#LETTER NAME            ENABLED  REFONLY NROLES MASTER     DESCRIPTION                   
c       class           on       FALSE   0                 classes                       
...
```

Here `c` is a letter, I'm talking about. All letter has an associated name: "class" is for `c`.

Do you want the letter in your application?  Providing just name from ctags to your application is not enough?

The name is much more self descriptive and I think it is suitable for modern file format.

tmm1:
I agree name is more elegant and makes sense for json format. I do not need
the letter abbreviation for my use-case.

masatake:
Thank you for the response. I see.

masatake:
`addExtensionFields` field of output-json.c becomes much simple:
```
static void addExtensionFields (json_t *response, const tagEntryInfo *const tag)
{
	int k;

	/* FIELD_KIND has not name; getFieldName (FIELD_KIND) returns NULL.
	   FIELD_KIND_LONG does, too.
	   That cannot be changed to keep the compatibility of tags file format.
	   Use FIELD_KIND_KEY instead */
	if (isFieldEnabled (FIELD_KIND) || isFieldEnabled (FIELD_KIND_LONG))
		enableField (FIELD_KIND_KEY, TRUE, FALSE);

	/* FIELD_SCOPE has not name; getFieldName (FIELD_KIND_KEY) returns NULL.
	   That cannot be changed to keep the compatibility of tags file format.
	   Use FIELD_SCOPE_KEY and FIELD_SCOPE_KIND_LONG instead. */
	if (isFieldEnabled (FIELD_SCOPE))
	{
		enableField (FIELD_SCOPE_KEY, TRUE, FALSE);
		enableField (FIELD_SCOPE_KIND_LONG, TRUE, FALSE);
	}

	for (k = FIELD_EXTENSION_START; k <= FIELD_BUILTIN_LAST; k++)
		renderExtensionFieldMaybe (k, tag, response);
}
```

masatake:
Currently json output goes "tags" file by default.
How about using stdout by default like -x output?

tmm1:
> Currently json output goes "tags" file by default. (So you may specify -o - or -f - when testing your code.)
> How about using stdout by default like -x output?

Good idea, I updated the PR use stdout in json mode.

> addExtensionFields field of output-json.c becomes much simple:

This looks great! Do you have a branch which adds `FIELD_EXTENSION_START`?

tmm1:
> "_type" :"tag" is really good. Definition tag and reference tag should be distinguished with whether "role:" field is avaiable or not; a definition tag doesn't have the field. It implies role:definition.

Okay, I understand now. I've added a `"_type":"header"` which is emitted at the beginning of the json output:

```
{"_type": "header", "program_name": "Universal Ctags", "program_author": "Universal Ctags Team", "program_url": "https://ctags.io/", "program_version": "0.0.0", "file_format": "1.0"}
```

masatake:
Can I take over this PR ?
I also implemented stdout redirection and "_type":"header" alike implementation.
(I use "_type":"ptag".)

tmm1:
> Can I take over this PR ?

Sure, please do!

masatake:
Oh, joining to your repository is good idea. I have never thought of.
My change may need "git rebase & git push --force". So your branch will become broken. Is it o.k.?
My original thought was opening my own PR based on this ( tmm1:output-format-json ) branch.

tmm1:
> My original thought was opening my own PR based on this ( tmm1:output-format-json ) branch.

That sounds simplest. I can close this one and link back to it from the new PR.

PR #1011, opened 2016-06-28
Fix whitespace and warning

tmm1:
```
main/parse.c:1638:33: warning: duplicate 'const' declaration specifier [-Wduplicate-decl-specifier]
        const langType language, const const char *kindLong, const boolean mode)
                                       ^
1 warning generated.
```

### DIFF
--- a/main/parse.c
+++ b/main/parse.c
@@ -1635,7 +1635,7 @@ static boolean enableLanguageKind (
 }
 
 static boolean enableLanguageKindLong (
-	const langType language, const const char *kindLong, const boolean mode)
+	const langType language, const char *kindLong, const boolean mode)
 {
 	boolean result = FALSE;
 	kindOption* const opt = langKindLongOption (language, kindLong);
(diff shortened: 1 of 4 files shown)


main/parse.c
```diff
@@ -1635,7 +1635,7 @@ static boolean enableLanguageKind (
 }
 
 static boolean enableLanguageKindLong (
-	const langType language, const const char *kindLong, const boolean mode)
+	const langType language, const char *kindLong, const boolean mode)
```
b4n: seeing the other arguments, I would think it was supposed to be `const char *const kindLong`

b4n:
3a524e4 meh, looks like a noise commit to me (that will clutter blame output), but kinda whatever

main/parse.c
masatake: Yes. @tmm1, can I ask you to change the line as @b4n said?
Then I will merge your changes.

masatake:
@b4n, I will merge this "whitespace" related change.

... hmm, if one wants just removing  "whitespace" , I may make the changes from the one pend. However, @tmm1 is working on json i/o.  I guess the changes submitted here is the side products of the work.

masatake:
I opend #1059 for "const const". As soon as verifying on CIs are passed, I will merge #1059.

k-takata:
It seems that all changes in this PR have been already fixed by other commits. So closing.

PR #1013, opened 2016-07-01
[SELF INCOMPATIBLE] Revise --field option

masatake:
I would like to change the notation for specifying parser own fields.
@pragmaware, I would like you to know the changes.

I introduced `--fields=[+-]{<LANG>.<FIELD>}` notation.

However, it is not consistent with `--kinds-<LANG>=<KIND>' and planed `--extra-<LANG>=<EXTRA>`
notation. Therefore I would like to change the field notation as follows:

1. `--field=[+-]<COMMON-FIELD>` is for enabling/disabling common fields.
2. `--field=*` is for enabling all common fields.
3. `--field-<LANG>=[+-]<PARSER-OWN-FIELD>` is for enabling/disabling parser own fields.
4. `--field-<LANG>=*` is for enabling/disabling parser own fields of <LANG>.
5. `--field-*=<FIELD>` is for enabling/disabling both parser own fields and common fields named <FIELD> at once.

e.g.

	$ ./ctags '--fields-C=+{properties}'

this was

	$ ./ctags '--fields=+{C.properties}'

About `--_xformat` we will continue to use `%{<LANG>.<FIELD>}` notation.
This is inconsistent with `--field-<LANG>` notation. However, the consistency
within --kind, --field and --exttra is more important than consistency between
`--_xformat` and `--field`.

I will merge this after merging new python parser.


masatake:
@b4n, I would like to merge these changes.
If you got a trouble in test cases related to your new python parser, call me.
I can fix it quickly.

Typical change can be found in https://github.com/universal-ctags/ctags/pull/1013/commits/5946654bf9421ae45063d3bb1e1b90c1f2104efc .

Issue #1015, opened 2016-07-02
Make long line breaking configurable


t-b:
Since 7b47ff23 (main: omit patterns for long input lines, 2015-10-15) long tag lines are cut off.
I'm using ctags as something parser-like for C-style function declarations and generate code from the tag lines. The above mentioned change broke that as now the function signatures in the tag file are not complete anymore.

Something like `--pattern-length-limit=0` to turn that feature off would be nice.

If I get some pointers where to add stuff, I can have  a try myself.

masatake:
How about this one?
```
diff --git a/main/entry.c b/main/entry.c
index 25490b3..24de761 100644
--- a/main/entry.c
+++ b/main/entry.c
@@ -676,7 +676,7 @@ static size_t appendInputLine (int putc_func (char , void *), const char *const
 		if (c == CRETURN  ||  c == NEWLINE)
 			break;
 
-		if (length >= Option.patternLengthLimit)
+		if (Option.patternLengthLimit != 0 && length >= Option.patternLengthLimit)
 		{
 			*omitted = TRUE;
 			break;

```

t-b:
@masatake But the option `--pattern-length-limit` is not read, or?
I've created PR #1018 which tries to add that option.

masatake:
> @masatake But the option --pattern-length-limit is not read, or?

Oh!, sorry. I forgot that I just made a stub. I will take a look at the pull request. Thank you.

t-b:
Fixed per #1018 .

PR #1018, opened 2016-07-03
Add pattern length limit

t-b:
As proposed in #1015.

I don't know the codebase at all, so a careful look is needed.
I've tested locally and it works.


main/options.c
```diff
@@ -2417,7 +2417,7 @@ static parametricOption ParametricOptions [] = {
 	{ "_list-roles",            processListRolesOptions,        TRUE,   STAGE_ANY },
 	{ "maxdepth",               processMaxRecursionDepthOption, TRUE,   STAGE_ANY },
 	{ "options",                processOptionFile,              FALSE,  STAGE_ANY },
-	{ "output-format",          processOutputFormat,            TRUE,   STAGE_ANY},
+	{ "output-format",          processOutputFormat,            TRUE,   STAGE_ANY },
```
masatake: I wonder why this change is needed. All entries has a whitespace after STAGE_ANY.

main/options.c
```diff
@@ -348,6 +348,8 @@ static optionDescription LongOptionDescription [] = {
 #endif
  {0,"  --output-format=ctags|etags|xref"},
  {0,"      Specify the output format. [ctags]"},
+ {0,"  --pattern-length-limit=N"},
+ {0,"      Cutoff tag entries after N characters. Disable by setting to 0. [96]"},
```
masatake: > "      Cutoff tag entries after ..."

I think this should be "      Cutoff patterns of tag entries after ...".
Am I wrong?

main/options.c
```diff
@@ -2363,6 +2365,18 @@ static void processMaxRecursionDepthOption (const char *const option, const char
 	Option.maxRecursionDepth = atol(parameter);
 }
 
+static void processPatternLengthLimit(const char *const option, const char *const parameter)
+{
+	if (parameter == NULL || parameter[0] == '\0')
+		error (FATAL, "A parameter is needed after \"%s\" option", option);
+
+	if (atol (parameter) < 0)
```
masatake: Could you consider to use strtol?

I tried following small program:

```
[yamato@x201 tmp]$ cat foo.c
#include <stdlib.h>
int main(void) {return atoi("abc");} 
[yamato@x201 tmp]$ gcc foo.c
[yamato@x201 tmp]$ ./a.out ; echo $?
0
```

Atoi returns 0 when a unconvertable string is given.
I would like to reject such value here. A user giiving such value  is wrong. However, in some aspects, this limit is introduced to protect ctags from a kind of DoS attach. So I would like to deal with the option value carefully.

masatake:
"excuberant tags" in the log of the second commit may be "exuberant ctags".

masatake:
I made a counter pull request. (https://github.com/t-b/ctags/pull/1).

main/options.c
t-b: Looking at https://github.com/universal-ctags/ctags/blob/master/main/options.c#L2420 there is no space after STAGE_ANY. I can of course remove the commit if you like

main/options.c
t-b: Thanks. Fixed.

main/options.c
t-b: > Atoi returns 0 when a unconvertable string is given.

strtol does also return 0 for an unconvertable string. At least if you pass `NULL` as second argument.

Using strtol is not so straightforward one has to do something like

```c
#include <stdlib.h>
#include <stdio.h>
#include <string.h>

int main(void)
{
  char *ptr;
  const char input[] = "123";
  long val = strtol(input, &ptr, 10);

  printf("val = %d, ptr = %d\n", val, ptr);

  if(!*ptr)
    return val;

  return 1;
}
```
but currently no callers of strtol in options.c do that.

I've switched to `strtol` now.

main/options.c
masatake: Oh, I see. Thank you.

main/options.c
```diff
@@ -2363,6 +2365,22 @@ static void processMaxRecursionDepthOption (const char *const option, const char
...
 
+static void processPatternLengthLimit(const char *const option, const char *const parameter)
+{
+	char* end;
+	long patternLengthLimit;
+
+	if (parameter == NULL || parameter[0] == '\0')
+		error (FATAL, "A parameter is needed after \"%s\" option", option);
+
+	patternLengthLimit = strtol(parameter, &end, 0);
```
masatake:
```
RETURN VALUE
       The strtol() function returns the result of the conversion, unless the value would underflow
       or  overflow.   If  an  underflow occurs, strtol() returns LONG_MIN.  If an overflow occurs,
       strtol() returns LONG_MAX.  In both cases, errno is set to ERANGE.  Precisely the same holds
       for strtoll() (with LLONG_MIN and LLONG_MAX instead of LONG_MIN and LONG_MAX).
```

Check the value of errno.

t-b: @masatake Thanks, done.

main/options.c
```diff
@@ -2363,6 +2366,22 @@ static void processMaxRecursionDepthOption (const char *const option, const char
...
+	long patternLengthLimit;
+
+	if (parameter == NULL || parameter[0] == '\0')
+		error (FATAL, "A parameter is needed after \"%s\" option", option);
+
+	errno = 0;
+	patternLengthLimit = strtol(parameter, NULL, 0);
+
+	if (errno != 0 || patternLengthLimit < 0)
+		error (FATAL, "-%s: Invalid pattern length limit", option);
```
b4n: that's not really enough as it would accept `foo` just fine still, and  `12foo` too.  Correct usage is to check *both* errno and whether endptr points somewhere else than the end of the string.

```C
boolean str_to_long(const char *string, long *value, int base)
{
    char *endptr;

    errno = 0;
    *value = strtol (string, &endptr, base);
    return (*endptr) == 0 && string != endptr &&
           (errno != EINVAL && errno != ERANGE);
}

/* ... */
    long val;
    if (string_to_long(option, &val, 0) && val >= 0)
        patternLengthLimit = val;
    else /* ... */
```

or something like that

b4n: BTW in this case you should rather use `strtoul()`

```C
boolean str_to_ulong(const char *string, int base, unsigned long *value)
{
    char *endptr;

    errno = 0;
    *value = strtoul (string, &endptr, base);
    return *endptr == 0 && string != endptr && errno == 0;
}

/* ... */
    unsigned long val;
    if (string_to_long(option, &val, 0) && val <= UINT_MAX)
        patternLengthLimit = (unsigned int) val;
    else /* ... */

Comment (t-b) [inline main/options.c]: @b4n Thanks for the hints. I guess it makes sense to add both `str_to_ulong` and `str_to_long` for future callers. Where should I put them?

Comment (b4n) [inline main/options.c]: Yeah I suggested extra functions because it's likely to be useful in several places; but if it's not it can be inlined.

Not sure where to put it, I'd wait for @masatake

Comment (masatake) [inline main/options.c]: routine.[hc] is the place.
Could you use CamelCase for naming about public (extern'ed) names: strToULong and stroToLong ?

Thank you.

Comment (t-b) [inline main/options.c]: @masatake @b4n I've added the requested changes.

@b4n If you like to have more acknowledgement in df7365d I'm happy to adapt the commit message.

File: main/options.c [hunk 6]
```diff
@@ -2215,9 +2218,9 @@ static void processEchoOption (const char *const option, const char *const param
 static void processForceQuitOption (const char *const option __unused__,
 				    const char *const parameter)
 {
-	long s = 0;
-	if (parameter != NULL && parameter[0] != '\0')
-		s = strtol (parameter, NULL, 0);
+	long s;
+	if (parameter == NULL || parameter[0] == '\0' || !strToLong(parameter, 0, &s))
+		s = 0;
 	exit (s);
```
b4n: Now there is `strToInt()` it would probably better to use an int argument to exit() as it's what it expects.  Not that it changes much, but it'd be cleaner.

main/options.c
b4n: @t-b it's fine like that, don't worry

b4n:
From travis:
```
main/routines.c: In function ‘strToUInt’:
main/routines.c:393:59: error: ‘UINT_MAX’ undeclared (first use in this function)
main/routines.c:393:59: note: each undeclared identifier is reported only once for each function it appears in
main/routines.c: In function ‘strToInt’:
main/routines.c:404:56: error: ‘INT_MAX’ undeclared (first use in this function)
main/routines.c:404:80: error: ‘INT_MIN’ undeclared (first use in this function)
```

you need to include *limits.h*

t-b:
@b4n `limits.h` is already included in routines.c. Some configure.ac magic was missing, which I added.

I've also added the appropriate guard to `#include <errno.h>`.

masatake:
How about this one?

```
diff --git a/main/e_msoft.h b/main/e_msoft.h
index 9efcf6d..d421e48 100644
--- a/main/e_msoft.h
+++ b/main/e_msoft.h
@@ -14,6 +14,7 @@
 #define MSDOS_STYLE_PATH 1
 #define HAVE_FCNTL_H 1
 #define HAVE_IO_H 1
+#define HAVE_LIMITS_H 1
 #define HAVE_STDLIB_H 1
 #define HAVE_SYS_STAT_H 1
 #define HAVE_SYS_TYPES_H 1
```

t-b:
@masatake Thanks, that is what I needed.

masatake:
Could you apply the following change to 	b84e47d 
```
diff --git a/main/lxcmd.c b/main/lxcmd.c
index 70737ab..31b7d48 100644
--- a/main/lxcmd.c
+++ b/main/lxcmd.c
@@ -604,12 +604,12 @@ extern void freeXcmdResources (void)
 }
 
 #ifdef HAVE_COPROC
-static void xcmd_flag_not_avaible_status_long (const char* const s __unused__, const char* const v, void* data)
+static void xcmd_flag_not_avaible_status_long (const char* const s, const char* const v, void* data)
 {
        xcmdPath *path = data;
 
        if(!strToInt(v, 0, &path->not_available_status))
-               error (FATAL, "Could not parse availability status: %s", v);
+               error (FATAL, "Could not parse the value for %s flag: %s", s, v);
 }
 #endif

```

masatake:
Following example of a session can be used for Tmain test and we can gain the coverage a bit.
```
% ./ctags --quiet --langdef=ZZZ --xcmd-ZZZ=ls'{notAvailableStatus=x}'  
<-langdef=ZZZ --xcmd-ZZZ=ls'{notAvailableStatus=x}'  
ctags: Could not parse the value for notAvailableStatus flag: x
```

t-b:
> Could you apply the following change to b84e47d

Done.

I've also added a test for notAvailablestatus as requested.

t-b:
@masatake I've fixed the wrong year, thanks. I urgently need a morning tea ;)

masatake:
Thank you very much. I will merge after for awhile.

PR #1019, opened 2016-07-04
Sharing kind descriptors

### LINKED ISSUE #1002: Broken detection of functions and prototypes
Hello,

The latest ctags (56612cfbcd1baaa4e8f04ded6aac57cac01fd136) can't detect any functions.

Using this command to list functions and prototypes:

    ctags -x --c-kinds=fpvx libcryptsetup.h

The tool lists macroses, enumerators, enums, structs and typedefs instead of functions and prototypes.

Seems that something was broken (or changed incompatibly) between 56612cfbcd1baaa4e8f04ded6aac57cac01fd136 (doesn't work) and e5b7a2508db6e1374ea78a300cc9b45f4b26b02a (works) commits.

Thank you.

masatake:
Close #1002.

@pragmaware, I would like you to see my change on Cxx(7f010fc) and the description about newly introduced API called "dependency" (0ea3e52).

I will append one more commit, a test case reported in #1002.


masatake:
Hmm... it seems that VC doesn't support  variadic macro parameter.

b4n:
> Hmm... it seems that VC doesn't support variadic macro parameter.

The C99 syntax for variadic macros is
```C
#define FOO(ARG1, ...) { ARG1, __VA_ARGS__ }
```
not using `NAME...`

masatake:
@b4n, thank you.

codebrainz:
> and note that it doesn't allow for 0 parameters as the variadic arguments, you need at least 1

0ec9fe2 doesn't pass the required argument for `...` unless I misread [the usage](https://github.com/universal-ctags/ctags/pull/1019/commits/0ec9fe29c0f52ff7993bfe77c119d7582725d3e3#diff-72c43358b7fa1d3abef0275855104b15R62).

masatake:
@codebrainz, thank you. It seems that I should not use variadic macro here.

Further more "sharing kinds" is wrong name for this PR.  It should be "synchronizing kinds". I will rebuild this branch.

PR #1024, opened 2016-07-06
javascript: capture function in arguments list

### LINKED ISSUE #1014: JavaScript: functions not recognized if wrapped
Hi everybody.
If I have some JavaScript literal functions with one wrap level inside a global function, universal-ctags works perfectly.
With test1.js file:
```
function (A){
  var Test;
  Test.func1 = function func1(){ console.log('func1'); };
  Test.func2 = function func2(){ console.log('func2'); };
}
```
`ctags test1.js` produces
```
!_TAG_PROGRAM_VERSION	0.0.0	/5ab7c69/
func1	test1.js	/^  Test.func1 = function func1(){$/;"	f
func1	test1.js	/^  Test.func1 = function func1(){$/;"	f	function:Test
func2	test1.js	/^  Test.func2 = function func2(){$/;"	f
func2	test1.js	/^  Test.func2 = function func2(){$/;"	f	function:Test
```
But if I add another wrap level, having these definitions passed like a method parameter (test2.js):
```
AUI().use(
  function funcGen(A){
    var Test;
    Test.func1 = function func1(){ console.log('func1'); };
    Test.func2 = function func2(){ console.log('func2'); };
  }
);
```
universal-ctags doesn't work anymore (`ctags test2.js`):
```
!_TAG_PROGRAM_VERSION	0.0.0	/5ab7c69/
```
Is there some special configurations or tricks to get things works in the second case?
Thanks.

masatake:
Close #1014.

### DIFF
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -745,7 +745,10 @@ static void skipArgumentList (tokenInfo *const token, boolean include_newlines, 
 		while (nest_level > 0 && ! isType (token, TOKEN_EOF))
 		{
 			readTokenFull (token, FALSE, repr);
-			if (isType (token, TOKEN_OPEN_PAREN))
+			if (isType (token, TOKEN_KEYWORD) && token->keyword == KEYWORD_function &&
+			    repr == NULL)
+				parseFunction (token);
+			else if (isType (token, TOKEN_OPEN_PAREN))
 				nest_level++;
 			else if (isType (token, TOKEN_CLOSE_PAREN))
 				nest_level--;
--- a/Units/parser-javascript.r/js-function-in-arglist.d/expected.tags
+++ b/Units/parser-javascript.r/js-function-in-arglist.d/expected.tags
@@ -1,0 +1,5 @@
+func1	input.js	/^    Test.func1 = function func1(){ console.log('func1'); };$/;"	f
+func1	input.js	/^    Test.func1 = function func1(){ console.log('func1'); };$/;"	f	function:Test
+func2	input.js	/^    Test.func2 = function func2(){ console.log('func2'); };$/;"	f
+func2	input.js	/^    Test.func2 = function func2(){ console.log('func2'); };$/;"	f	function:Test
+funcGen	input.js	/^  function funcGen(A){$/;"	f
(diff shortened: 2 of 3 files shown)


masatake:
@b4n, can I ask you to review my change?

I wonder whether `funcGen` in my test case should be captured or not.

b4n:
I'll take a look.

parsers/jscript.c
```diff
@@ -745,7 +745,10 @@ static void skipArgumentList (tokenInfo *const token, boolean include_newlines,
 		while (nest_level > 0 && ! isType (token, TOKEN_EOF))
 		{
 			readTokenFull (token, FALSE, repr);
-			if (isType (token, TOKEN_OPEN_PAREN))
+			if (isType (token, TOKEN_KEYWORD) && token->keyword == KEYWORD_function &&
+			    repr == NULL)
```
b4n: not handling in case of `repr` makes the behavior probably a little too unpredictable.  Is there a reason not to, apart missing stuff in the `repr`?  Does it lead to some problem?

parsers/jscript.c
```diff
@@ -745,7 +745,10 @@ static void skipArgumentList (tokenInfo *const token, boolean include_newlines,
...
+				parseFunction (token);
+			else if (isType (token, TOKEN_OPEN_PAREN))
```
b4n: you could avoid changing this line by adding an `else if` to the end of the chain rather than prepending it.   It wouldn't change anything as the other types handled are different from `KEYWORD` anyway.

b4n:
I find this kinda ugly, but why not.

However, it will fail parsing anonymous functions.  And well, it's kinda questionable what to do here anyway: it's an argument for a function, so the scope is highly volatile and unknown, and what to do with its body's content is also questionable.  But well, I guess the fix here is fine at least for now; it's kinda harmless I guess if it pleases a user it's good.

parsers/jscript.c
b4n: BTW, there probably should never actually be anywhere we collect the argument list representation where a function definition is valid.

b4n:
I suggest 78a90f0351faa8459cb79651a5061492564db94a from #1045 to supersede this PR.

parsers/jscript.c
masatake: > not handling in case of repr makes the behavior probably a little too unpredictable. Is there a reason not to, apart missing stuff in the repr? 

Sorry but I cannot remember why I did so.

PR #1028, opened 2016-07-09
Improve review tool

masatake:
@b4n, have you ever tried my misc/review script?
I hope this one helps your work.

It can be run with "misc/review inspect" from the at top src dir.
This tool helps us in following work flow:

1. edit c code
2. run "make units"
3. see DIFF.TMP 
4. move FILTERD.TMP to expected.tags
5. remove the both TMP files
6. go to 3 with another test case or exit.

The script focus on steps from 3 to 6.

Currently it supports only Units test cases and their result.
When trying, I recommend you to install colordiff and pygments but the script works well without them.

I will merge this as soon as possible.

### DIFF
--- a/misc/review
+++ b/misc/review
@@ -130,10 +130,11 @@ inspect_cmd_show_generic ()
 {
     local tcase=$1
     local file=$2
+    shift 2
     local f=./Units/${tcase}/${file}
 
     if [[ -r "$f" ]]; then
-	cat "$f"
+	"$@" "$f"
     else
 	echo "No ${file}" 1>&2
     fi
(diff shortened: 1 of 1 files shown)


masatake:
An example session:

```
[jet@localhost ctags]$ misc/review inspect
1) <a>ccept	      3) <d>iff		    5) show <i>nput
2) <s>kip	      4) show args.<c>tags  6) <q>uit
[[ parser-python.r/py-skipped-string.d ]]? =
--- ./Units/parser-python.r/py-skipped-string.d/expected.tags	2016-02-10 01:28:31.201664911 +0900
+++ /home/jet/var/ctags/Units/parser-python.r/py-skipped-string.d/FILTERED.tmp	2016-07-09 12:33:11.349331417 +0900
@@ -1 +0,0 @@
-f1	input.py	/^def f1():$/;"	f
@@ -3,0 +3 @@
+zf1	input.py	/^def zf1():$/;"	f
---
#  ./Units/parser-python.r/py-skipped-string.d/DIFF.tmp
[[ parser-python.r/py-skipped-string.d ]]? i
0001: # triple start string immediately after a normal string not detected
0002: 
0003: def zf1():
0004:     ''"""
0005:     The string above was not detected as triple start string,
0006:     but the one below instead.
0007:     """
0008:     print "f1"
0009: 
0010: def f2():
0011:     ''"""
0012:     The string above was then detected as end string,
0013:     and the one below as start string again.
0014:     """
0015:     print "f2"
0016: 
0017: def f3():
0018:     """
0019:     The string below is prepared so that ctags with the bug does not start a
0020:     new triple string. For a clean precondition for the next test.
0021:     ''"""
0022:     print "f3"
0023: 
0024: # normal string immediately after a normal string not detected
0025: 
0026: ''" import os\
0027: "
0028: 
0029: ""' def fX():\
0030: '
0031: 
---
#  ./Units/parser-python.r/py-skipped-string.d/input.py
[[ parser-python.r/py-skipped-string.d ]]? 
1) <a>ccept	      3) <d>iff		    5) show <i>nput
2) <s>kip	      4) show args.<c>tags  6) <q>uit
[[ parser-python.r/py-skipped-string.d ]]? 
```

PR #1030, opened 2016-07-10
Eiffel: add support for Liberty Eiffel's "insert"

dkearns:
Liberty Eiffel uses the "insert" keyword for non-conforming inheritance.

### DIFF
--- a/parsers/eiffel.c
+++ b/parsers/eiffel.c
@@ -149,6 +149,7 @@ static const keywordTable EiffelKeywordTable [] = {
 	{ "indexing",       KEYWORD_indexing   },
 	{ "infix",          KEYWORD_infix      },
 	{ "inherit",        KEYWORD_inherit    },
+	{ "insert",         KEYWORD_inherit    },
 	{ "inspect",        KEYWORD_inspect    },
 	{ "invariant",      KEYWORD_invariant  },
 	{ "is",             KEYWORD_is         },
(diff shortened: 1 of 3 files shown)


masatake:
LGTM.

parsers/eiffel.c
```diff
@@ -149,6 +149,7 @@ static const keywordTable EiffelKeywordTable [] = {
 	{ "indexing",       KEYWORD_indexing   },
 	{ "infix",          KEYWORD_infix      },
 	{ "inherit",        KEYWORD_inherit    },
+	{ "insert",         KEYWORD_insert     },
```
b4n: if `insert` is an alias of `inherit`, you probably should rather simply map `"insert"` to `KEYWORD_inherit` and not modify anything else

b4n: basically, it would mean the only change to this file would be ``` +	{ "insert",         KEYWORD_inherit     },``` and all the rest is left alone.

dkearns: Updated thanks.

I was considering capturing redefinitions and the difference between inherit and insert but the distinction is not necessary now.

masatake:
Let's merge it!

masatake:
Thanks. Do you have a chance to use Eiffel?

dkearns:
I do periodically. I'm one of the hundred... :)

PR #1031, opened 2016-07-12
Fix empty field handling in xformat

masatake:
This is fixing for following crash.

```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.py 
class Foo:
    def doIt():
        pass
[yamato@x201]~/var/ctags-github% ./ctags -x --_xformat=%p /tmp/foo.py
zsh: segmentation fault (core dumped)  ./ctags -x --_xformat=%p /tmp/foo.py
```


masatake:
Sorry for making the review period of this change short.
I need this in json-output. So I will merge in soon.

Freel free to give comments after merging.
If you would like to review but you don't have time, say so in the comment.

PR #1032, opened 2016-07-13
json output format (take 3)

masatake:
This PR is a successor of #1010.

What I did from #1010:

* simplify the code for emitting extension fields (addExtensionFields in output-json.c),
* allow to emit pseudo tags (--extra=+p)
* use stdout as the default destination
* introduce JSON_OUTPUT_VERSION, a new pseudo tag
* write draft version of document

@tmm1, I would like you to review following two aspects.
* docs/output-json.rst, this is just a draft version. 
* the format of JSON_OUTPUT_VERSION. Ideally it should be an integer. But I use a string to reuse existing code.  Basically I would like to merge current code I submitted here. However, we can improve later. In such case I would like to discuss how JSON_OUTPUT_VERSION should be before merging.

```
% ./ctags --extra='p' --pseudo-tags=JSON_OUTPUT_VERSION  --output-format=json /tmp/foo.py
```

This is the current output.  As I wrote in the output we are "in development" stage. So we can change after merging this branch.


tmm1:
I reviewed the new commits and the changes and new docs all look great. I agree this is a good base to merge and improve further with additional PRs. Thanks for your help with this feature!

masatake:
@tmm1, thank you for contributing and reviewing.

PR #1039, opened 2016-07-15
Units: add a test case about #930

masatake:


masatake:
Updated.

masatake:
... Instead of making c++ parser accept c input, switching to C parser from C++ parser in upper layer may be better in clean design. The issue is performance. For switching fro C+ parser to C parser, the contests of .h file must be parsed for detecing the language of input (pre parse.)

masatake:
Updated. bool is a keyword both in C and C++.

codebrainz:
> bool is a keyword both in C and C++.

`bool` isn't a keyword in C, it's a standard library macro defined in `stdbool.h`, if it matters for your test.

masatake:
@codebrainz, thank you. I will update the commits again.

PR #1041, opened 2016-07-16
Eiffel add "note" keyword support

dkearns:


parsers/eiffel.c
```diff
@@ -44,23 +44,68 @@
...
+	KEYWORD_separate,
+	KEYWORD_strip,
+	KEYWORD_then,
+	KEYWORD_true,
+	KEYWORD_undefine,
+	KEYWORD_unique,
+	KEYWORD_until,
+	KEYWORD_variant,
+	KEYWORD_when,
+	KEYWORD_xor
```
masatake: Good!

masatake:
Are all test cases for indexing gone?
If you was "I won't remove support for it.", we have to keep some of test cases for indexing.
Forgive me I missed something.

dkearns:
You don't miss a trick. :)

I've got to update the tests in general as the some of the syntax is still outdated.  I was going to add a "legacy syntax" test case when I do that.

However, I'll add a simple one for "indexing" and squash it into a275580.

Thanks.

dkearns:
Done.

masatake:
Fine, thank you.

PR #1043, opened 2016-07-20
docs: copy editing

dkearns:
This is just a first pass.

### DIFF
--- a/docs/contributions.rst
+++ b/docs/contributions.rst
@@ -13,7 +13,7 @@ Contributions
 You are welcome.
 
 
-These are what we would like potential contributors to know.  In this
+This is what we would like potential contributors to know. In this
 section "you" means a contributor, and "we" means reviewers. "I" means
 Masatake YAMATO, the author of this section.
 
@@ -26,29 +26,31 @@ Origin of changes and license
 
 Make clear where the patches come from and who wrote them.
 
-If you backport patches from Geany or somewhere other projects, their
-commit ids should be logged, too.
+If you backport patches from Geany or some other project, their
+commit IDs should be logged, too.
 
-Put copyright notice when adding a new ``{parsers,main}/*.[ch]`` file.
+Include a copyright notice when adding a new
+``{parsers,main}/*.[ch]`` file.
 
-About new file, license notice is needed to at the head of source
-file.
+A new file also requires a license notice at the head of the file.
 
-We expect your change (or new code) is provided under term of General
-Public License version 2 or any later version. We would like you to
-express "version 2 or any later version".
+We expect your change (or new code) to be provided under the terms of
+the General Public License version 2 or any later version. We would
+like you to express "version 2 or any later version".
 
 Commit log
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-(About a new parser following criteria is not applicable.)
+(For new parsers the following criteria is not applicable.)
 
-Make clear the original motivation and/or impact on tags file.
-If you fix a bug reported somewhere on the web, its URL should
+Make clear the original motivation for the change and/or the impact
+on the tags file.
+
+If you fix a bug reported somewhere on the web, its URL should be
 logged, too.
 
-If the bug is reported on the tracker of sourceforge web site
-of exuberant ctags project, log it as sf-bugs:N, sf-patches:N,
+If the bug is reported in the Exuberant-ctags tracker on the
+SourceForge web site, log it as sf-bugs:N, sf-patches:N,
 sf-support-requests:N, or sf-feature-requests:N.
 docs/tracking.rst also should be updated.
 
@@ -63,17 +65,19 @@ Testing
 
 Add test cases, and run both existing cases and your new cases.
 
-If you add a new parser or modify existing parser, add new test cases
-to "Units". If you modify main part, add new test cases to
-"Tmain". The way to write and run test cases are written in "Testing
-ctags" of this document. Other than tmain test harness, you can
-specify VG=1 for running test cases under valgrind memory debugger.
+If you add a new parser or modify an existing parser, add new test
+cases to "Units". If you modify the core, add new test cases to
+"Tmain". The way to write and run test cases is described in the
+"Testing ctags" section of this guide.
+
+With the exception of the tmain test harness, you can specify VG=1
+for running test cases under the Valgrind memory debugger.
 
 A parse should not enter an infinite loop for bad input.
 A parse should not crash for bad input.
-A parse should return the control to its caller for bad input.
+A parse should return control to its caller for bad input.
 
-Write what kind of tests are passed in a commit log.
+Describe what kind of tests are passed in the commit message.
 e.g. ::
 
   make units LANGUAGES=TTCN VG=1 is passed.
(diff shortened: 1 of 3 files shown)


docs/news.rst
```diff
@@ -117,116 +118,118 @@ code "wildcard"(``*``) in option is introduced.
...
+.. TODO: Review...
+
+Japanese programmers sometimes use the Japanese language in source
+code comments. Of course, it is not limited to Japanese. People may
+use their own native language and in such cases encoding becomes an
+issue.
 
-ctags didn't care it. ctags just reads input as just bytes sequence and
-use them as is when writing tags entries.
+ctags doesn't consider the input encoding it just reads input as a
```
masatake: I think this should be:
```
ctags doesn't consider the input encoding; it just reads input as a
```
(semicolon is inserted.) Am I wrong?

masatake:
Thank you very much. I'm reading/reviewing them now.

docs/news.rst
dkearns: Yes, that's right thanks. Hopefully I didn't introduce too many errors myself.

dkearns:
I noticed there's some inconsistencies in how the project/program name is formatted in the docs.

Should it be "universal-ctags", "Universal-ctags", "Universal Ctags", "Universal CTags" or something else?  The documentation currently uses a few different styles and I haven't attempted to normalise them yet.

It seems that "Exuberant Ctags" is used by the 'parent' project.

I don't think it really matters which is chosen but it would be nice to be a little more consistent.

dkearns:
Perhaps the phrase "upper layer tools" could be simplified to "client tools" in the docs?

docs/index.rst
```diff
@@ -1,32 +1,32 @@
...
+`Universal-ctags`_ was created to continue the development of Darren
+Hiebert's `Exuberant-ctags`_ after activity on that project
+unfortunately stalled.
 
-The goal of the project is preparing and maintaining common/unified
-space where people interested in making ctags better can work
-together.
+Reza Jelveh <reza.jelveh@gmail.com> initially created a personal fork
+on GitHub and as interest and participation grew it was decided to
+move development to a dedicated GitHub organization.
```
dkearns: If you want to include this bit of history it might be worth expanding upon it with your role.

masatake:
Sorry to be late. 
Let's use 'Universal-ctags', 'Exuberant-ctags'. u-ctags is e-ctags can be used as abbreviations. The abbreviations can be used for distinguishing commands. 

I agree with you that we should use "clients" or "client tools". I'll put thse terms on index.rst as comments.

docs/index.rst
masatake: No, I don't want more. This is enough.  Thank you.

docs/news.rst
```diff
@@ -461,57 +475,59 @@ used to enable it.
...
+unconstrained pattern length. For example, running ctags on a
+minified JavaScript source file often exhibits this behaviour.
 
 Reference tags
 ---------------------------------------------------------------------
 
-Traditionally ctags collects the information for locating where an
-object having name is DEFINED.
+Traditionally ctags collects the information for locating where a
+language object is DEFINED.
```
masatake: "language object", this is the term I have been looking for. We should use this term all in the documents. Thank you.

docs/news.rst
```diff
@@ -461,57 +475,59 @@ used to enable it.
...
-Let me show some examples. Here is the target input file named reftag.c.
+Here are some examples. Here is the target input file named reftag.c.
 
 .. code-block:: c
 
     #include <stdio.h>
     #include "foo.h"
     #define TYPE point
-    struct TYPE { int x, y };
+    struct TYPE { int x, y; };
```
masatake: Thank you.

docs/news.rst
```diff
@@ -576,19 +598,20 @@ few parsers utilized it now. All available roles can be listed with
     DTS	d	undef	undefined	on
     DTS	h	system	system header	on
     DTS	h	local	local header	on
-    Make	I	generic	non-categorized generic role	on
-    Make	I	optional	included as an optional makefile	on
-    Sh	s	generic	non-categorized generic role	on
+    Make	I	included	included	on
+    Make	I	optional	optionally included	on
+    Sh	s	loaded	loaded	on
     Vera	d	undef	undefined	on
```
masatake: Thank you.

dkearns: That's obviously still only a partial list. It might be worth making that clear.

I didn't update it to the full list as it seemed a bit long now.

docs/news.rst
```diff
@@ -726,151 +750,159 @@ Example output:
...
+	-       properties      off     C++              TRUE     properties (static, virtual, inline, mutable,...)
 	-       sectionMarker   off     reStructuredText TRUE     character used for declaring section
-	-       version         off     Maven2          TRUE     version of artifact
+	-       version         off     Maven2           TRUE     version of artifact
 
-e.g. `reStructuredText` is the owner of `sectionMarker` field. Like
-`end` field owned by `C` and `C++`, more than one parsers have fields
-with the same name.
+e.g. `reStructuredText` is the owner of the `sectionMarker` field.
+Like the `end` field is owned by `C` and `C++`, more than one parser
```
masatake:
```
Like the `end` field is owned by `C` and `C++`
```

Is `is` needed ?

docs/news.rst
```diff
@@ -943,40 +976,39 @@ elements of the format.
...
 readtags has ability to find tag entries by name.
 
-The concept filtering is inspired from display filter of wireshark.
-You can give more complex condition for searching. Currently this
-feature is available only on platforms where `fmemopen` is available
-as part of libc. Filtering in readtags command is an
+The concept of filtering is inspired by the display filter of
+Wireshark. You can specify more complex conditions for searching.
+Currently this feature is available only on platforms where
+`fmemopen` is available as part of libc. Filtering in readtags is an
```
masatake: (@masatake's TODO: fmemopen is replaced with mio.)

docs/contributions.rst
```diff
@@ -63,17 +65,19 @@ Testing
 
 Add test cases, and run both existing cases and your new cases.
 
-If you add a new parser or modify existing parser, add new test cases
-to "Units". If you modify main part, add new test cases to
-"Tmain". The way to write and run test cases are written in "Testing
-ctags" of this document. Other than tmain test harness, you can
-specify VG=1 for running test cases under valgrind memory debugger.
+If you add a new parser or modify an existing parser, add new test
+cases to "Units". If you modify the core, add new test cases to
```
masatake: I know we should call "core" part. However, using the name `core` may be conflicted with `core` file in meaing of `coredump`. So the our choice was `main`. Generally I use `main part` in discussions.

So I wonder the senten can be `If you modify the main part`. Can I ask you to reconsider this?

docs/contributions.rst
```diff
@@ -63,17 +65,19 @@ Testing
...
+"Tmain". The way to write and run test cases is described in "Testing
+ctags" of this document.
+
+You can specify VG=1 for running test cases under the Valgrind memory
+debugger.
```
masatake: "Tmain" test harness doesn't support VG=1.

docs/contributions.rst
```diff
@@ -84,72 +88,75 @@ e.g. ::
...
 global variables.
 
 Remove unused variables and types. If you want to keep them in your
-source code, put enough description as comments.
+source code, include a descriptive comment.
 
-Use available facility that ctags main part provides.  If the
-facility is not enough for writing a parser, consider extending the
-main part first.
+Use the available facilities provided by the ctags core. If the
```
masatake: `main part` is unacceptable expression?

docs/contributions.rst
```diff
@@ -84,72 +88,75 @@ e.g. ::
...
 
-Trim unnecessary white space at the end of line. In emacs `M-x
-whitespace-cleanup`. Don't mix `whitespace cleanup` fix and the other
-improvements into one commit when changing the existing code. Style,
-including `whitespace cleanup` should be in a separate commit.
-Mixing makes reviewing harder.
+Don't mix `whitespace cleanup` fixes and other improvements in one
+commit when changing the existing code. Style fixes, including
+`whitespace cleanup`, should be in a separate commit. Mixing
+functional changes with style fixes makes reviewing harder.
```
masatake: BTW, @vhda, @dkearns, how do you think this rule?
As I wrote before, this rule helps reviewers.

docs/contributions.rst
```diff
@@ -281,14 +291,15 @@ Quoted from @steveno in #393:
...
-Add your `.c` file to somewhere in `source.mak`.
+Add your `.c` file to `source.mak`.
 
-In addition update, `win32/ctags_vs2013.vcxproj` and
-`win32/ctags_vs2013.vcxproj.filters`. So our CI process
-run on Appveyor will be failed.
+In addition, update `win32/ctags_vs2013.vcxproj` and
+`win32/ctags_vs2013.vcxproj.filters`.
+.. TODO: I'm not sure what this means -- @dkearns
+So our CI process run on Appveyor will be failed.
```
masatake: Sorry. This should be:

```
If you forget updating these files, our CI process run on Appveyor will be failed.
```

b4n: `will be failing`, or better `will fail`; but not `will be failed` :)

masatake:
Thank you very much again for fixing my mistakes in multi level.
It looks much bettern than before.


Only one critical gap I found is that "the core" and "the man part". Of couse what we specify with the terms are the same.  If "the main part" is not seriously strange as in English, I would like to continue to use "the main part". If it is really strange(or wrong), we have to find new name. As I wrote in inline comment, we should avoid using "core" as file or directory name because it implies "core" of "core dump".

docs/contributions.rst
masatake: Thank you.

codebrainz:
@masatake as a (subjective) data point, in my native tongue (Canadian English), "core" sounds more natural/descriptive than "main part" in this context. With "main part" I would probably first assume it refers to a `main.c` file or wherever `main()` entry point is located. In any case, there are so many variations in English, I think I'd probably figure either one out eventually :)

masatake:
@codebrainz, thank you.


How about the following policy:

* use "core" or "core part" (as @dkearns changed) in docs/*.rst,
* don't change the directory name ("main") of the source tree, and
* describe the above relation in docs/internal.rst.

docs/news.rst
dkearns: Yes, if you remove it then it reads as "Like the `end` field (owned by `C` and `C++`), more than one parser..." shifting the subject to the field rather than the concept of field ownership.

It's still pretty awkward though so how about just simplifying it to something like: "e.g. `reStructuredText` is the owner of the `sectionMarker` field and both `C` and `C++` own the `end` field."

While it's not enforced I assume it is expected that fields with the same name will be used for the same purpose across languages?

docs/contributions.rst
dkearns: OK, sorry I'll restore the warning.  I just saw it was included under the tmain: target in Makefile and assumed it was now supported.

docs/contributions.rst
dkearns: That's the way I always do it.  I usually even split whitespace only style fixes from other types of style fixes.

docs/contributions.rst
dkearns: That was mostly a note for me to learn about Appveyor.  I understood the gist of the sentence. :)

dkearns:
@masatake as @codebrainz said it's not that confusing. It's just a bit awkward to read and caused me to pause and check it was actually referring to the concept I assumed it was.  I also changed a reference to "parser part" as you may have noticed.

"core part" isn't a lot better than "main part".

Even "if you modify files under main/ ..." would be a bit clearer I think.

I don't think there's a need to change any names in the source tree even if "core" is used.

It never would have occurred to me that it could be mistaken, in context, for a reference to a core dump file but that may be my native speaker bias showing through.

If you prefer using "the main part". You could use something like "the main part (the ctags core)", "the main part (the ctags engine)", "the main part (files under main/)" or something similar the first time it is referenced.

masatake:
People may not be confused with the name. I'm talking about software/program/tools.

Could you look at following discussions?

https://github.com/universal-ctags/ctags/pull/295
https://github.com/universal-ctags/ctags/issues/294

Do I make sense? 

How about "the main part (code driving parsers)" ?

docs/news.rst
masatake: > Yes, if you remove it then it reads as "Like the end field (owned by C and C++), more than one parser..." shifting the subject to the field rather than the concept of field ownership.

Oh, I see. Thank you very much. 

> It's still pretty awkward though so how about just simplifying it to something like: "e.g. reStructuredText is the owner of the sectionMarker field and both C and C++ own the end field."

Let's use your sentence.

> While it's not enforced I assume it is expected that fields with the same name will be used for the same purpose across languages?

Yes. However, if it is used for the same purpose, we should move the field to the main part.
The parser own field is still very new concept.  So I'm not sure what kind of fields will be introduced/proposed in the future.

masatake:
O.k. About the source tree we will use "main" for the present.
I understand the explanation as suggested by @dkearns. I will do it in the future.

@dkearns, could you merge this? 
I'm afraid conflicts with the other merges.

dkearns:
I think I may have misinterpreted your final comment as suggesting we backtrack to the initial idea of using "core" for the general reference in the docs (not the source name).

If so let me know and I'll update the references.

As for, #294 I didn't realise that it was originally named core/.  It makes more sense now, thanks.

PR #1047, opened 2016-07-26
Eiffel: add support for self-initializing attributes

dkearns:
Erroneous "attribute" tags were being created and the rest of the file
then consumed in most cases.

### DIFF
--- a/Units/parser-eiffel.r/obsolete.e.d/input.e
+++ b/Units/parser-eiffel.r/obsolete.e.d/input.e
@@ -3,6 +3,8 @@ class OBSOLETE_ obsolete "Use a different one"
 feature
 
 	obsolete_attribute: INTEGER obsolete "Use something else"
+		attribute
+		end
 
 	obsolete_function (one: INTEGER): INTEGER obsolete "Use something else"
 		do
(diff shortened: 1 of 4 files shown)


Units/parser-eiffel.r/obsolete.e.d/input.e
```diff
@@ -3,6 +3,8 @@ class OBSOLETE_ obsolete "Use a different one"
 feature
 
 	obsolete_attribute: INTEGER obsolete "Use something else"
```
masatake: Do you mean the origin test input is broken?

dkearns: Yes, without the "attribute...end" it's not valid Eiffel syntax though the Ctags parser handles it.

I assume I missed it the first time around when eyeballing it due to the the unusual (for Eiffel) one-line formatting.

I'm 'linting' the input files now after changes since my skills as the human syntax parser are obviously on the wane.

masatake: Oh, I see.

masatake:
LGTM.

masatake:
(You don't need my ack for merging changes about your maintained parser.)

Issue #825, opened 2016-02-29
Replace m4 parser


masatake:
Replace opt/m4.ctags with better implementation: https://github.com/b4n/geany/blob/wip/master/tagmanager/ctags/m4.c

masatake:
@b4n, I would like to import your m4.c. Can I ask you to make the version of GPL clear?
If you allow, I would like to import it to ctags with "the GNU General Public License version 2 or (at your option) any later version." .

b4n:
@masatake sure, it's now done.

There likely is a typo at https://github.com/b4n/geany/blob/wip/master/tagmanager/ctags/m4.c#L308, I guess I meant ```(c == '"' || c == '\'' || c == '`')```, but I'm not totally sure right now so I didn't change it yet.  Pretty sure though esp. as this happened in a larger change.

masatake:
Thanks. I'm working on parsers for understanding build system and scripts.
Autoconf/m4 parsers are the last pieces.

masatake:
The m4 parser is very interesting. It changes the language/parser dynamically to autoconf parser when the m4 parser finds pattern like AC_. I have to think more to find good way to adapt the parser to universal-ctags.

masatake:
Finally I found a good way.
I will introduce two new functions: pushLanguage and popLanguage.

masatake:
I have a question.

In https://github.com/b4n/geany/blob/wip/master/tagmanager/ctags/m4.c#L328
you calls `skipBlanks`. I think this can be removed because  `readMacroArgument` is called 
in https://github.com/b4n/geany/blob/wip/master/tagmanager/ctags/m4.c#L330.

At the head of `readMacroArgument`, `readMacroArgument` does the same as what `skipBlanks` does.

@b4n, can I ask you give me comments?

b4n:
@masatake you seem to be right indeed.  Also, as it's the only usage of `skipBlanks()`, it could be removed altogether.

masatake:
@b4n, thank you.

b4n:
@masatake fixed in https://github.com/b4n/geany/commit/ce7d77b73f1c71163f3378c5ef7bbd7e3d620735

b4n:
> The m4 parser is very interesting. It changes the language/parser dynamically to autoconf parser when the m4 parser finds pattern like AC_. I have to think more to find good way to adapt the parser to universal-ctags.

[...]

Issue #821, opened 2016-02-28
parser for configure.ac


masatake:

b4n:
I started a generic M4 parser with quoting switching support some time ago, allowing to dynamically switch to `[` and `]` quoting, and a specialized version for Autoconf with those as default. https://github.com/b4n/geany/blob/wip/master/tagmanager/ctags/m4.c

Not sure if it's what you're thinking about here, or if you meant real handling of the dual M4/shell?

masatake:
@b4n, thank you for notifying. 

(I don't understand what I want yet.)

I would like to capture at least
AC_DEFINE,
AC_SUBST
AH_TEMPLATE,
AC_ARG_WITH, and
AC_ARG_ENABLE.

They are rather high level(?) items than items m4.c captures.

m4.c is good place to start.

masatake:
Based on @b4n's work I have added code for capturing Autoconf level language objects.

```
[yamato@x201]~/var/ctags-github% ./ctags --list-kinds-full=Autoconf
./ctags --list-kinds-full=Autoconf
#LETTER NAME            ENABLED  REFONLY NROLES MASTER     DESCRIPTION                   
p       package         on       FALSE   0                 packages                      
t       template        on       FALSE   0                 templates                     
m       macro           on       FALSE   0                 autoconf macros               
W       optwith         on       FALSE   0                 options specified with --with-...
E       optenable       on       FALSE   0                 options specified with --enable-...
s       subst           on       FALSE   0                 substitution keys             
C       condition       on       FALSE   0                 automake conditions           
d       definition      on       FALSE   0                 definitions                   
[yamato@x201]~/var/ctags-github% ./ctags --list-kinds-full=m4
./ctags --list-kinds-full=m4
#LETTER NAME            ENABLED  REFONLY NROLES MASTER     DESCRIPTION                   
d       macro           on       FALSE   1                 macros                        
I       macrofile       on       TRUE    2                 macro files                 
[yamato@x201]~/var/ctags-github% ./ctags -x -o - ./configure.ac 
./ctags -x -o - ./configure.ac 
BUILD_IN_GIT_REPO condition   318 ./configure.ac   AM_CONDITIONAL([BUILD_IN_GIT_REPO], [test "x${in_git_repo}" = "xyes"])
CASE_INSENSITIVE_FILENAMES definition  348 ./configure.ac   AC_DEFINE(CASE_INSENSITIVE_FILENAMES)
[...]

PR #1049, opened 2016-07-27
The separation of Automake and make parsers

masatake:
pushLanguage/popLanguage functions are introduced in this PR.

This is mainly designed for importing Autoconf/m4 parsers written by @b4n in geany.
The m4 parser pretends the Autoconf parser when the m4 parser detects Autoconf specific macros are used in the input stream. In the implementation the two parsers are rather strongly coupled.

I'm working on providing API that helps decoupling them. The pushLanguage/popLanguage is the part of the incomplete API.

In this PR, I applied the two function for decopuling to existing Make and Automake parsers. They are also coupled.

#918, #825, #821, all .xml related parsers many need these functions.

I will write documents after completing the API.


masatake:
A change that should be in `main:` was in `Automake, Make`.  I moved the change to  `main:`.

PR #1050, opened 2016-07-27
docs: copy editing "Parsers"

dkearns:
This covers the "Parsers" section of the guide.

It think it's easier for review purposes to break these up into smaller chunks and guide sections seem small enough.


masatake:
LGTM. Thank you very much.

PR #1056, opened 2016-07-29
docs: copy editing "Choosing a proper parser in ctags"

dkearns:


docs/guessing.rst
```diff
@@ -63,49 +64,48 @@ input as a zsh script:
       #compdef ....
 
 
-NOTE: This feature takes a performance hit: it opens the input file
+NOTE: This feature comes with a performance hit: it opens the input file
```
masatake: Should I use `;` here instead of `:' ?

```
NOTE: This feature comes with a performance hit; it opens...
```

steveno: Yes. The semicolon is better. 

Thank you,
Steven N. Oliver 

Sent from Outlook Mobile




On Sun, Jul 31, 2016 at 6:11 AM -0400, "Masatake YAMATO" <notifications@github.com> wrote:












In docs/guessing.rst:
> @@ -63,49 +64,48 @@ input as a zsh script:
>        #compdef ....
>  
>  
> -NOTE: This feature takes a performance hit: it opens the input file
> +NOTE: This feature comes with a performance hit: it opens the input file



Should I use ; here instead of `:' ?

NOTE: This feature comes with a performance hit; it opens...

dkearns: Fixed, thanks.

masatake:
Thank you. LGFM.

PR #1058, opened 2016-07-30
Python: rewrite as a tokenizing parser (rebased version)

b4n:
This is the rebased (cleaned up history) version of #1003.

It has a very slight difference: the indentation of *Units/parser-python.r/bug1809024.py.d/input.py* is different, because I didn't re-change it and kept the version currently in master (in the other PR, I used the version I fixed myself).

The other difference is obviously the history.

I can leave it like that, or further squash it.  This state is as incremental as possible (well, easily at least) without introducing regressions in the middle of the way.


Units/parser-python.r/newlines-cr.d/README
```diff
@@ -0,0 +1,2 @@
+Tags are right, but location are not recorded correctly.
+Likely due to a bug in main part line number recording with CRs alone.
```
masatake: In tthat case, could you change the suffix of this directory to `.b` from `.d`?
So we can remember there is a bug.

masatake:
I'm very sorry but could you remove `04dde62354b92496d0c45811ac06cbd45586939c`?

I asked add the empty decoration fields. However,  I feel they are noisy after reading real tags output which includes empty decoration: fields.

masatake:
Other than two things I pointed out(about `.b` suffix and the empty decorator fields), Looks good.
Could you merge these?

Units/parser-python.r/newlines-cr.d/README
b4n: OK

b4n:
> I asked add the empty decoratorfields. However, I feel they are noisy after reading real tags output which includes empty decorator: fields.

Should I then do the same for inheritance?

Units/parser-python.r/newlines-cr.d/README
b4n: done

masatake:
Thank you for updating.

> Should I then do the same for inheritance?

Sorry, but I don't understand the "same" in this context.
`fieldSpec` of "inheritance" field has `isValueAvailable` method.
With this method the main part can know whether the value should be rendered or not.

If you are talking about this method, the answer is no because this method can be used only in common fields; decorator field is not common field.

b4n:
> Sorry, but I don't understand the "same" in this context.

I mean, should I make the parser write the `inheritance` field only when it found something to put there, or should it leave it unset when there's nothing to put there.

But your answer suggests that there is a mechanism in core for hiding the field in some situations, so maybe it's not important.

masatake:
> I mean, should I make the parser write the inheritance field only when it found something to put there, or should it leave it unset when there's nothing to put there.

Oh, I see.

O.k. I have no more comment. Could you merge this? Let's get feedbacks from real users.

masatake:
Thank you.

tek:
this seems to have removed the possiblity to omit imports as tags. With `--python-kinds=-Ix`, i get a tag for `Foo` at `from mod import Foo`, which was previously ignored with `--python-kinds=-i`.

b4n:
@tek AFAIK this didn't change from the previous parser as it was in master: the `Foo` you see generated a tag of `unknown` kind, so you'd have to disable it with `--kinds-python=-x`.

If it's indeed a change from the *latest version* of the previous parser, please open a separate report.

tek:
never mind – it seems I had updated ctags yesterday before this PR, updating again now fixed the issue: `--python-kinds=-x` works as expected.

Anyway, there is still a tag created for the module from which is being imported; while I have no quarrel with that, it seems to be disabled by `--python-kinds=-i`, which is not listed by `--list-kinds=python`. Thought that might be a mistake.

b4n:
@tek Hum.  Indeed there is a difference between current master and previous pareser's results.  Apparently reference tags are always emitted, even without `--extra=+r` -- and even with `--extra=-r`.

I however am really not sure what's doing this… @masatake?

b4n:
#1060

b4n:
@tek behavior regarding reference tags should now be fixed.

tek:
looks good!

PR #1059, opened 2016-08-01
Main: remove double const

masatake:
Suggested by @tmm1 and @b4n in #1011


tmm1:
:+1: thanks!

Issue #1062, opened 2016-08-03
lregex parser broken?


tarzanek:
Hi guys
It seems one of last commits in 
https://github.com/universal-ctags/ctags/commits/master/main/lregex.c
broke the selection of type for matched line
Expected(from ctags 5.9svn ):
```
Sample  Sample.pas      /^unit Sample;$/;"      unit    line:22
uses    Sample.pas      /^uses$/;"      Section line:23
interface       Sample.pas      /^interface$/;" Section line:26
TSample Sample.pas      /^    TSample = class$/;"       Class   line:28
GetId   Sample.pas      /^      function GetId : integer;$/;"   function        line:33
SetId   Sample.pas      /^      procedure SetId(const aValue : integer);$/;"    procedure       line:34
GetClassName    Sample.pas      /^      class function GetClassName: string;$/;"        function        line:36
GetIdAndDescriptionString       Sample.pas      /^      function GetIdAndDescriptionString: string; virtual;$/;"        function        line:37
GetUser Sample.pas      /^      function GetUser: TUser;$/;"    function        line:38
Id      Sample.pas      /^      property Id: integer read GetId write SetId;$/;"        property        line:40
Description     Sample.pas      /^      property Description: string read FDescription write FDescription;$/;"  property        line:41
implementation  Sample.pas      /^implementation$/;"    Section line:44
uses    Sample.pas      /^uses$/;"      Section line:45
GetId   Sample.pas      /^function TSample.GetId: integer;$/;"  function        line:48
TSample.GetId   Sample.pas      /^function TSample.GetId: integer;$/;"  function        line:48
SetId   Sample.pas      /^procedure TSample.SetId(const aValue: integer);$/;"   procedure       line:53
TSample.SetId   Sample.pas      /^procedure TSample.SetId(const aValue: integer);$/;"   procedure       line:53
TSample.GetClassName    Sample.pas      /^class function TSample.GetClassName: string;$/;"      function        line:58
GetUser Sample.pas      /^function TSample.GetUser: TUser;$/;"  function        line:63
TSample.GetUser Sample.pas      /^function TSample.GetUser: TUser;$/;"  function        line:63
```

What I get using universal ctags:
```
Sample  Sample.pas      /^unit Sample;$/;"      Section line:22
uses    Sample.pas      /^uses$/;"      Section line:23
interface       Sample.pas      /^interface$/;" Section line:26
TSample Sample.pas      /^    TSample = class$/;"       Class   line:28
[...]

PR #1064, opened 2016-08-05
main: warn when a kind letter is reused in a regex parser definition

masatake:
u-ctags doesn't allow using a same letter for different log names.

e.g. (taken from #1062)

    --langdef=pascal
    --regex-pascal=/(uses|interface|implementation)[[:space:]]*/\1/s,Section/
    --regex-pascal=/unit[[:space:]]+([a-zA-Z0-9_<>,][a-zA-Z0-9_<>, ]+)[;(]/\1/s,unit/

In both "Section" and "unit" `s' is used as a kind letter. u-ctags
threw later one, "unit", away silently.

As reported in #1062, user can reuse a letter accidentally.
With this commit ctags can make warnrnings when detecting such reuses.

### DIFF
--- a/Tmain/warn-reusing-kind-letter.d/run.sh
+++ b/Tmain/warn-reusing-kind-letter.d/run.sh
@@ -1,0 +1,13 @@
+# Copyright: 2016 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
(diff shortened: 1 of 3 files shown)


masatake:
Squashed.

PR #1065, opened 2016-08-05
Style: add EditorConfig file to normalise indentation

dkearns:
There's some annoying inconsistencies in the source code indentation resulting from the use of mixed spaces and tabs for indentation, intraline tab usage and assumptions about the tab stop size when aligning lines. See #1008 for the latest discussion.

This is intended to provide a better base configuration for editors and matches what seems to be the predominant style.

EditorConfig is also recognised by GitHub so files should look a little nicer in their file viewer too.

See http://editorconfig.org/ for details.

Obviously this only addresses very basic formatting issues.

There is a configuration file ( `.indent.pro`) in the repository for use with GNU Indent but in my limited testing it tends to make a bit of a mess of the existing files.  Uncrustify appears to do a much better job so I'll  try and capture the current style in a config file for that as well.


b4n:
> and assumptions about the tab stop size when aligning lines.

IMO spaces should be used for the alignment part (e.g.
```
\tfoo(1,
\t    2);
```
Unfortunately it's not so commonly supported properly.

Apart that, looks fine for what I understand of it.  Out of curiosity, what's the meaning of `root=true`?

dkearns:
> Unfortunately it's not so commonly supported properly.

Right, that will have to be handled by Uncrustify, or similar, if it's to be enforced.  EditorConfig just handles the absolute basics but it seems worthwhile given the extensive editor support.

> Out of curiosity, what's the meaning of root=true?

You can override settings with a config file per directory. `root=true` informs the plugins that it is the topmost config file.

dkearns:
The current tab-based indenting was inherited from Exuberant-ctags so if there's some silent consensus that we should find our way back to the light I'd be happy enough to detabify the tree too.

masatake:
I have no objection. I just installed editorconfig.el.

masatake:
@pragmaware, how about you?

masatake:
It seems that @pragmaware is in busy. So let's merge this.

PR #1066, opened 2016-08-07
Include clojurescript and conditional clj files

tfoldi:
Add clojurescript files (cljs) and conditional clojure files (cljc) to the extension list.


masatake:
Thank you.

PR #1067, opened 2016-08-08
M4 and autoconf parsers

### LINKED ISSUE #825: Replace m4 parser

masatake:
#Close #825 and #821.

This is study to know how to stack parsers for the same input file in bi-directional way.
See  ac24bf3.
As the case study I chose M4 and Autoconf.
The newly introduced APIs exported from the main part to parsers can be used for solving #918.


parsers/m4.c
```diff
@@ -0,0 +1,648 @@
...
+	if (probe_data->found)
+		return;
+
+	if (client->host != probe_data->client)
+		return;
+
+	if (client->probeLanguage && client->probeLanguage (probe_data->token,
+							    probe_data->client->data))
+	{
+		prepareForNewInput (client);
```
masatake: @masatake's TODO: this should be done in maySwitchLanguage.

masatake:
Rebased.

PR #1071, opened 2016-08-09
Add --interactive mode

tmm1:
Re-implementation of #994 

``` sh
$ (echo '{"command":"generate-tags","filename":"test.rb","size":13}'; echo "def foo() end") | ./ctags --interactive
{"_type": "tag", "name": "foo", "path": "test.rb", "pattern": "/^def foo() end/", "kind": "method"}
{"completed": "generate-tags"}
```


tmm1:
@masatake I've merged master and got the tests passing again. Would love to get this merged.. have you had a chance to think about the json interface? Do you have any pending concerns about this approach?

masatake:
I'm sorry to be late. I'm still struggling the output format.

tmm1:
I have added some tests for the new `--interactive` mode.

masatake:
It seems that stat command implementation is different in GNU and MacOSX.

```
diff --git a/Tmain/interactive-mode.d/run.sh b/Tmain/interactive-mode.d/run.sh
index 85e7180..4dc9117 100644
--- a/Tmain/interactive-mode.d/run.sh
+++ b/Tmain/interactive-mode.d/run.sh
@@ -41,7 +41,7 @@ echo =======================================
 echo
 echo generate tags from data
 echo =======================================
-size=$(stat -f %z test.rb)
+size=$(stat -c %s test.rb)
 (
   echo '{"command":"generate-tags", "filena
```

I recommend to add a small shell utility function. Here is a sketch:

```
make_json_request()
{
    local input=$1
    local size

    if ( stat --help | grep -q GNU ) 2>&1 > /dev/null; then
	if ! size=$(stat -c %s "$input"); then
	    size=
	fi
    else
	if ! size=$(stat -f %z "$input"); then
	    size=
	fi
    fi

    if [ -n "$size" ]; then
	echo '{"command":"generate-tags", "filename":"$input", "size":'$size'}'
	cat "$input"
	return 0
    else
	return 1
    fi
}
```

masatake:
Which is better as a name, interactive or stream?

tmm1:
> It seems that stat command implementation is different in GNU and MacOSX.

I switched to using `wc` to get the filesize instead.



> Which is better as a name, interactive or stream?

I don't think stream makes sense as a name. Maybe "command mode" instead of "interactive mode"?

tmm1:
@masatake Is there any existing option to return the `pattern` field without regex encoding?

masatake:
> @masatake Is there any existing option to return the pattern field without regex encoding?

No, but why do you ask it to me?
Is it needed to pass the test on Travis?

tmm1:
> No, but why do you ask it to me?
> Is it needed to pass the test on Travis?

No, unrelated request. For my use-case I wanted to display the matching line to user, so I need to convert it from regex back to original line.

I think maybe the test failures are because of too-old version of jannson. I will check.

main/main.c
```diff
@@ -484,6 +491,66 @@ static void batchMakeTags (cookedArgs *args, void *user CTAGS_ATTR_UNUSED)
...
+      goto next;
+    }
+
+    if (!strcmp ("list-parsers", json_string_value (command))) {
+            /* todo */
+    } else if (!strcmp ("generate-tags", json_string_value (command))) {
+      json_int_t size = -1;
+      const char *filename;
+
+      if (json_unpack (request, "{s?I ss}", "size", &size, "filename", &filename) == -1) {
```
tmm1: The `?` syntax means optional key, and is new to jansson 2.8. Since some distributions do not have this version available, I will need to find a backwards-compatible way to achieve the same result.

masatake:
> No, unrelated request. For my use-case I wanted to display the matching line to user, so I need to convert it from regex back to original line.

I see. If you want seriously, open an issue. However, I cannot implement it in soon. This PR should be merged first. More over json output should be improved first.

tmm1:
> I see. If you want seriously, open an issue. However, I cannot implement it in soon. This PR should be merged first. More over json output should be improved first.

Understood. I can work around it for now. I looked at the implementation for `makePatternString` so I can try to write a patch also.

I agree first we should improve json output and finalize this PR. I will fix the CI as first priority.

tmm1:
All the tests are passing now!

masatake:
Fine!

Two more requests:

*  commit log "compiles"  for 360887e doesn't make sense. This is just an example. I would like you to reivese commits. Some of   them should be squashed. Some of them can use "interactive: " as prefix for commits.

* documents. Could you add docs/interactive-mode.rst ? At least Basic idea(typical application) and spec of input must be written.

tmm1:
I have rebased and squashes the commit history.

Next I will write some documentation.

tmm1:
I have added some documentation for the new feature.

Tmain/utils.sh
```diff
@@ -12,6 +12,11 @@ remove_commit_id()
     sed -i -e '/!_TAG_PROGRAM_VERSION.*/s#/[^/]*/#//#' $1
 }
 
+filesize()
+{
+    wc -c "$1" | awk '{ print $1 }'
```
masatake: How about
```
wc -c < "$1"
```

This can remove awk invocation.

Tmain/interactive-mode.d/test.c
```diff
@@ -0,0 +1,9 @@
+#include <stdio.h>
+
+int main(int argc, char **argv) {
+  say_hello();
+}
+
+void say_hello() {
+  printf("hello world\n");
+}
```
b4n: better to have valid test source files, so put the `say_hello()` declaration after its first use (e.g. above `main()`)

Tmain/utils.sh
tmm1: Currently the `awk` is also removing the leading whitespace.

```
$ wc -c < test.rb
      56
```

I guess in my usage the whitespace doesn't matter, so I can remove the `awk`.

Tmain/interactive-mode.d/run.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo '{"command":"generate-tags", "filename":"test.rb"}'
+  echo '{"command":"generate-tags", "filename":"test.c"}'
+) | ${CTAGS} --interactive
+
+echo
+echo generate tags from data
+echo =======================================
+size=$(filesize test.rb)
+(
+  echo '{"command":"generate-tags", "filename":"test.rb", "size":'$size'}'
```
tmm1: Good point.

Tmain/interactive-mode.d/test.c
tmm1: Fixed.

tmm1:
I have addressed all the review comments on this PR.

Tmain/interactive-mode.d/stdout-expected.txt
```diff
@@ -0,0 +1,46 @@
...
+
+error on invalid command
+=======================================
+{"name": "Universal Ctags", "version": "0.0.0"}
+{"error": "unknown command name", "fatal": true}
+
+error on missing arguments
+=======================================
+{"name": "Universal Ctags", "version": "0.0.0"}
+{"error": "invalid generate-tags request", "fatal": true}
```
tmm1: Perhaps in error cases, we should still output a `completed` object? Like `{"completed":"invalid"}`. Might make it simpler to integrate, if you can always expect a completed object before you know to send the next command.

Tmain/interactive-mode.d/stdout-expected.txt
```diff
@@ -0,0 +1,46 @@
...
+
+error on invalid file
+=======================================
+{"name": "Universal Ctags", "version": "0.0.0"}
+{"error": "cannot open input file \"test.foo\"", "warning": true, "errno": 2, "perror": "No such file or directory"}
+{"completed": "generate-tags"}
```
tmm1: Since we have `_type:"tag"` for tag outputs, maybe we should use something like `_type:"program"` for the welcome message on boot, and `_type:"error"` + `_type:"completed"` for the other tags. This should again make it simpler to integrate if a user can always expect a `_type` entry in returned objects.

tmm1:
I updated the output so that all json objects returned always have a `_type` field. So far we have `program`, `error`, `completed`, `tag` and `ptag`.

masatake:
It seems that one more library version related issue is reported by travis.

masatake:
It seems that my nose was broken. Your code is good. However, pre existing function names don't match to the new feature you introduced. I will work on renaming and refactoring the pre existing function.

masatake:
Merged via #1244.

Issue #999, opened 2016-06-24
two errors in plist parser during build


JohnLunzer:
```
  CC     parsers/ctags-plist.o
parsers/plist.c:46: error: unknown field ‘recurSpec’ specified in initializer
parsers/plist.c:48: warning: missing braces around initializer
parsers/plist.c:48: warning: (near initialization for ‘plistXpathMainTable[0].<anonymous>.makeTagSpec’)
parsers/plist.c:48: warning: initialization makes integer from pointer without a cast
parsers/plist.c:48: error: initializer element is not computable at load time
parsers/plist.c:48: error: (near initialization for ‘plistXpathMainTable[0].<anonymous>.makeTagSpec.kind’)
parsers/plist.c:55: error: unknown field ‘makeTagSpec’ specified in initializer
parsers/plist.c:56: warning: missing braces around initializer
parsers/plist.c:56: warning: (near initialization for ‘plistXpathTextTable[0].<anonymous>.makeTagSpec’)
```
I fixed the braces locally and the build was successful. Seems like maybe the braces accidentally got removed...

masatake:
I don't understand the rrors. (Not reproduced my PC.) Could you make a pull request?

masatake:
Maybe this issue is very related to your environment. We run building-test in multiple CI environemnt.
As faras I can remember we have never introduced changes which didn't pass the test.

Feel free to reopen this issue.

PR #1075, opened 2016-08-12
Fix compilation error in CentOS 6.

amosbird:
According to c11 6.7.9p9 http://www.iso-9899.info/n1570.html#6.7.9p9
[Initialization] Except where explicitly stated otherwise, for the
purposes of this subclause unnamed members of objects of structure and
union type do not | participate in initialization. Unnamed members of
structure objects have indeterminate value even after initialization.

This is also mentioned in #999.


k-takata:
Why the indents are changed?

amosbird:
Why does the test fail?

masatake:
Thank you.
Could you do git-squash the two commits into one and git-push --force ?

I took the same mistake ago. We should have centos[67] as test environments on travis.

amosbird:
@masatake done.

Issue #1068, opened 2016-08-09
no tags for exported variables in JavaScript ES6 code


hoschi:
Example code:

```
let localLet = "";
function localFunction () {

}
export let namedExport = {};
export function functionExport () {

}

let defaultExport = {};
export default defaultExport;
```

Generated tags file:

```
!_TAG_FILE_FORMAT       2       /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED       1       /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
!_TAG_PROGRAM_NAME      Universal Ctags /Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL       https://ctags.io/       /official site/
!_TAG_PROGRAM_VERSION   0.0.0   /e7cdc72/
localFunction   test.js /^function localFunction () {$/;"       f
localLet        test.js /^let localLet = "";$/;"        v
```

I expect to get tags also for:
- namedExport
- functionExport
- defaultExport

masatake:
(It seems that u-ctags has not supported ES6 yet. Any volunteers are welcome:)

This is a workaround:

```
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 7540205..1336dfe 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -71,7 +71,9 @@ typedef enum eKeywordId {
 	KEYWORD_catch,
 	KEYWORD_finally,
 	KEYWORD_sap,
-	KEYWORD_return
+	KEYWORD_return,
+	KEYWORD_default,
+	KEYWORD_export,	
 } keywordId;
 
 typedef enum eTokenType {
@@ -158,7 +160,9 @@ static const keywordTable JsKeywordTable [] = {
 	{ "catch",		KEYWORD_catch				},
 	{ "finally",	KEYWORD_finally				},
 	{ "sap",	    KEYWORD_sap    				},
-	{ "return",		KEYWORD_return				}
+	{ "return",		KEYWORD_return				},
+	{ "default",		KEYWORD_default				},
+	{ "export",		KEYWORD_export				},	
 };
 
 /*
@@ -1940,6 +1944,10 @@ static void parseJsFile (tokenInfo *const token)
 	{
 		readToken (token);
 
+		if (isType (token, TOKEN_KEYWORD)
[...]

Issue #780, opened 2016-01-31
JavaScript object functions (ECMAScript 6)


marlun:
If I've got the following file

```
import Foo from './foo';

const Bar = Foo.create({
  jump() {
    return 123;
  }
})

export default Bar;
```

Bar is found but jump is not. I get the following tags file

```
!_TAG_FILE_FORMAT   2   /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED   1   /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
!_TAG_PROGRAM_NAME  Universal Ctags /Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL   https://ctags.io/   /official site/
!_TAG_PROGRAM_VERSION   0.0.0   /b6da9a1/
Bar test.js /^const Bar = Foo.create({$/;"  C
```

If I've got the following file

```
import Foo from './foo';

class Bar2 extends Foo {
  jump2() {
    return 123;
  }
}

export default Bar2;
```

Nothing is found where I would like to get Bar2 and jump2. My tags file look like the following

```
!_TAG_FILE_FORMAT   2   /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED   1   /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
!_TAG_PROGRAM_NAME  Universal Ctags /Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL   https://ctags.io/   /official site/
!_TAG_PROGRAM_VERSION   0.0.0   /b6da9a1/
```

I found the following configuration somewhere on the web which makes ctags find the functions but now it doesn't find Bar, it does however find Bar2.

```
--langdef=js
--langmap=js:.js
--langmap=js:+.jsx

--regex-js=/[ \t.]([A-Z][A-Z0-9._$]+)[ \t]*[=:][ \t]*([0-9"'\[\{]|null)/\1/n,constant/

--regex-js=/\.([A-Za-z0-9._$]+)[ \t]*=[ \t]*\{/\1/o,object/
--regex-js=/['"]*([A-Za-z0-9_$]+)['"]*[ \t]*:[ \t]*\{/\1/o,object/
[...]

PR #1076, opened 2016-08-14
Javascript ES6 methods, test for #780

hoschi:
As said in #1068 I want to contribute tests. For #1068 there is already a test in the commit of @b4n. So, I created a test for #780 instead.

### DIFF
--- a/Units/parser-javascript.r/js-methods.b/expected.tags
+++ b/Units/parser-javascript.r/js-methods.b/expected.tags
@@ -1,0 +1,3 @@
+Obj     input.js        /^var Obj = {$/;"       c
+myMethod     input.js        /^    myMethod(a, b) {$/;"      m       class:Obj
+myGenerator     input.js        /^    *myGenerator(a, b) {$/;"      m       class:Obj
--- a/Units/parser-javascript.r/js-methods.b/input.js
+++ b/Units/parser-javascript.r/js-methods.b/input.js
@@ -1,0 +1,9 @@
+// methods (functions of objects)
+// see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions
+// http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions
+var Obj = {
+    myMethod(a, b) {
+    },
+	*myGenerator(a, b) {
+	}
+}


masatake:
Thank you.

I'm surprised at the notation ` A.getPropName() + 'test'`.
I wonder how this should be recorded in tags file. Obviously this is not a name. This is code generating a name of method.  I wonder such code should be recorded or not. If we should decide to record such code, I think we should provide the way to distinguish the code from the name to client tools that read tags file.

One of the approach is introducing new kind. Here  G(generatedMethod) is used as an example:
```
A.getPropName() + 'test'     input.js        /^    [A.getPropName() + 'test']() {$/;"      G       class:Obj
'gen' + 2     input.js        /^    *['gen' + 2]() {$/;"      G       class:Obj
```

How do you think about this?

hoschi:
Good point! I thought not much about it, because I thought ctags has already a way to handle this dynamic properties for other languages. I see it is a problem, at least for editors (client tools example). The editor needs to recognize the whole string "'gen' + 2" as "a tag", for e.g. jumping to its definition/usage.

In my opinion your suggestion is sound. Gather the information, but don't treat it as a normal tag. Should I change the test cases to match your example? I can also remove the special syntax from this PR and create a second one, if you need more time to think about the problem/solution.

masatake:
> The editor needs to recognize the whole string "'gen' + 2" as "a tag", for e.g. jumping to its definition/usage.

I don't have much chance to use Javascript. However, I cannot image the situation that a user of the editor specifies "''gen' + 2" as "a tag" for jumping.

In my limited experiencel "'gen' + 2"  tag can be used only for navigating and summarizing.
See http://docs.ctags.io/en/latest/contributions.html#specific-to-add-new-parser-and-or-new-kind-role
In the page, my assumption about how client tools utilizes tags file is written.

Do you think a kind of tool can utilize "'gen' + 2" for auto completion and jumping? 


Above I wrote is only a question coming from my curiosity. The question is never the reason to reject this pull request. However, the question is quite important for writing a good parser.

Splitting a test case into two is good idea. Let's use `G` as a kind letter for the generated names.
It is nice to add a short document about test case like README (https://github.com/universal-ctags/ctags/blob/master/Units/parser-python.r/newlines-cr.b/README) or DESC.rst(
https://github.com/masatake/ctags/blob/70f3db4f0c9d9ed862bda1077e8f3a218dc336ec/Units/parser-m4.r/m4-quotestring.b/DESC.rst).

I'm planning to generage  docs/known-issues.rst by merging all DESC.rst under Units directory.

hoschi:
> Do you think a kind of tool can utilize "'gen' + 2" for auto completion and jumping?

I thought about something like that:
```
var Obj = {
    myMethod(a, b) {
    },
	*myGenerator(a, b) {
	},
	[A.getPropName() + 'test']() {
	},
	*['gen' + 2]() {
	}
}

var result = Obj['gen' + 2]();
```
Where the user wants to jump from the last line to the definition of the method. But in 99% of cases I can image, the developer would store the (dynamic) name of a method in constant or variable:
```
const dynamicMethodName = 'gen' + 2;
var Obj = {
    myMethod(a, b) {
    },
	*myGenerator(a, b) {
	},
	[A.getPropName() + 'test']() {
	},
	*[dynamicMethodName]() {
	}
}

var result = Obj[dynamicMethodName]();
```

> In my limited experiencel "'gen' + 2" tag can be used only for navigating and summarizing.

This would be a more realistic usage of the generated tags of this kind. E.g. an code outliner.

> It is nice to add a short document about test case like README or DESC

What would you see in this document for this PR? Is moving the comment lines from input.js with references to spec/examples to DESC.rst enough?

My next actions gathered from your response are:
* [ ] move generated method names into own PR, change tags file as mentoined above
* [ ] create a DESC.rst which points to this discussion for new PR
* [ ] create DESC.rst for this PR

masatake:
Hm...ctags is a tool for tagging *names*. We are talking about expressions generating names.
Basically I think ctags should not deal with the expressions.

Do you have an interest in being maintainer of the javascript parser?
(I'm always looking for good parser maintainers.)

hoschi:
> Hm...ctags is a tool for tagging names. We are talking about expressions generating names.
Basically I think ctags should not deal with the expressions.

I lost the context here.

> Do you have an interest in being maintainer of the javascript parser?
(I'm always looking for good parser maintainers.)

I'm pretty busy with my day job and unfortunately don't have the time to be a maintainer, at least yet. Would that involve writing C(++) code? Thanks anyway for thinking about me as parser maintainer :)

Can you answer my question about the content in the DESC file, so I can start tackling the todos?

masatake:
> 
>
>>   Hm...ctags is a tool for tagging names. We are talking about expressions generating names.
>>    Basically I think ctags should not deal with the expressions.
>
> I lost the context here.
I'm sorry. 
How do I say... What you expected to Javascript parser of ctags is very different from what ctags can do now. ctags can record names and extra informatoins.

```
A.getPropName() + 'test'
```

This it not a name. This is an expression. ctags cannot record an expression.

I have an example which tells how difficult to handle an expression in ctags.

```
A.getPropName () + /* comment */ + 'test'
```
This is a valid expression. In this case what ctags should record as a tag?

* `A.getPropName () /* comment */ + 'test'`,
* `A.getPropName () + 'test'`,
* `A.getPropName ()+'test'`, or
* `A.getPropName()+'test'`.

To make tags file useful, the expression must be normalized when recording.
I don't know well about Javascript, but I guess there is no standard rule for normalizing an expression.
If we just deal with names, we don't have to think about normalization. (An exception is ObjectiveC.)

This is a just one minor example. The main part of ctags may need to be extended for recording expressions. 
How? I don't know. Someone who really want to record expressoins in ctags has to study how to extend ctags for recording expressions.

If we cannot find such person, I think we should abort recording expressions.

So I would like to merge shrunken version of the orignal test case:

input.js:
```
var Obj = {
    myMethod(a, b) {
    },
    *myGenerator(a, b) {
    },
    [A.getPropName() + 'test']() {
    },
    *['gen' + 2]() {
    }
}
```

expected.tags may contain only myGenerator and myMethod. In such case, the intent of test case is so obvious. Therefore no extra documentation is needed.

masatake:
I'm sorry. I wrote too many things.
Let's start from adding small things. Complicated things can be dealt with later.

hoschi:
Thanks for the deep explanation. I rebased and changed the PR to only contain myMethod and myGenerator.

masatake:
Thank you.

PR #1079, opened 2016-08-16
Token-based HTML parser

techee:
This is a new token-based HTML parser. Apart from the usual advantages like comment detection and EOL independence the main goals of creating it were:

- support of additional tags we have in Geany (h1, h2, h3) 
- speed - for universal-ctags it brings the parsing speed of the boost library html documentation from 5.5s to 3.2s (in Geany the speed difference is even bigger because for h1, h2, h3 tags we have additional 3 regexes which make the parsing speed about 2x slow in the original regex-based parser)
- good error recovery when run on invalid input - this is important for Geany where using real XML parser would be probably too strict during typing and would end with an error so the symbol list would be probably empty most of the time

I didn't really want to spend much time on parsing embedded javascript code so the parser just simulates the original regex for the javascript functions with all its limitations.

It should be quite easy to further extend the parser and detect additional tags and attributes.

cc @b4n


parsers/html.c
```diff
@@ -8,35 +8,434 @@
...
+						int f = ' ';
+						do
+						{
+							d = e;
+							e = f;
+							f = getcFromInputFile ();
+						}
+						while (f != EOF && ! (d == '-' && e == '-' && f == '>'));
+
+						if (skipComments || f == EOF)
```
b4n: this would skip the comment and return EOF when `skipComments == TRUE` and `f == EOF`, is that really wanted?  I'd rather see returning `TOKEN_COMMENT`, and let the next read return `TOKEN_EOF`.

parsers/html.c
```diff
@@ -8,35 +8,434 @@
...
+			}
+			else if (d == '?')
+				token->type = TOKEN_OTHER;
+			else if (d == '/')
+				token->type = TOKEN_TAG_START2;
+			else
+			{
+				ungetcToInputFile (d);
+				token->type = TOKEN_TAG_START;
+			}
```
b4n: doesn't handle `<[CDATA[`, which might lead to some confusion (?)

parsers/html.c
```diff
@@ -8,35 +8,434 @@
...
+
+		default:
+		{
+			do
+			{
+				vStringPut (token->string, tolower (c));
+				c = getcFromInputFile ();
+			}
+			while (c != EOF && c != ' ' && c != '\t' && c != '<' && c != '/' &&
+				   c != '=' && c != '>' && c != '\'' && c != '"');
```
b4n: shouldn't this include NL and CR -- and any whitespace, actually?

parsers/html.c
techee: But it's an invalid TOKEN_COMMENT in this case as the comment wasn't terminated by --> but rather by EOF.

b4n: sure, so it might depend on how it's used; but here it drops the comment content if the comment is unterminated, while it doesn't if it is.

parsers/html.c
techee: <! should return TOKEN_OTHER - see the few lines above where the comment is handled. At top level the parser skips everything until it finds TOKEN_TAG_START so it should work correctly.

parsers/html.c
```diff
@@ -8,35 +8,434 @@
...
+		}
+		while (token->type != TOKEN_TAG_END && token->type != TOKEN_TAG_END2 &&
+			   token->type != TOKEN_EOF);
+
+		if (token->type == TOKEN_TAG_END)
+		{
+			readTagContent (token, text, startTag == KEYWORD_script);
+
+			readToken (token, TRUE);
+			if (textCreated && text != NULL)
```
b4n: meh, that's a rather obfuscated way of meaning "if this was a heading"

b4n: …which makes the body a lot less obvious

parsers/html.c
techee: Yep, isspace() will probably be better here.

parsers/html.c
b4n: My point was that something like `<![CDATA[<h1>]]>` might confuse the parser, or at least that `<h1><![CDATA[Pears > apples]]></h1>` will not generate something too nice.

b4n:
Looks mostly good at first read.

parsers/html.c
techee: It means something a little different - it tells whether it's the right heading tag end. In this case it's related to an invalid html like

```
<h1>something<h1>is</h1>here</h1>
```

The check makes sure the tag gets generated at the second </h1>, i.e. the collected text will contain "something is here" instead of "something is". Maybe too paranoid but I just wanted to make sure we know where we get tag generated.

parsers/html.c
techee: OK, will have a look at it.

parsers/html.c
b4n: Yeah ok, but my point is that this refers to [L347](https://github.com/universal-ctags/ctags/pull/1079/files/3048c15bf2917579a3e81e0cfff8d1ca3ad52ea7#diff-e619353ef58a45cb65c0674de4ebdb9fR347) in a really not obvious way, which made me wonder at first if the heading kind selection below couldn't get fooled and generate `H3` tags for a `A` tag.

If it was something like `isHeading` or maybe simply `/* if we were reading a heading */`, it would be a lot clearer to the casual reader.

parsers/html.c
```diff
@@ -8,35 +8,434 @@
...
+
+			readToken (token, TRUE);
+		}
+	}
+
+	if (textCreated)
+		vStringDelete (text);
+}
+
+static void findHtmlTags ()
```
b4n: missing `void` in the argument list :)

parsers/html.c
techee: It shouldn't matter. I introduced TOKEN_COMMENT because I was fighting with cases like

<foo>abcdefghi</bar>

Inside readTokenText() I don't know whether the < is tag start or comment start but at this point I cannot read more characters because I'd need multi-level-ungetc() which we don't have. On the other hand if the following readToken() just silently skipped the comment, I wouldn't know I can expect another text behind it.

If the comment token is at the end of input, it doesn't matter much if TOKEN_COMMENT or EOF is returned because no other text is behind it.

masatake:
@techee, I have a request about the new html parser.

Currently the html parser tags javascript functions  embedded in a html file.
Could you remove the code for tagging them from your parser?
Instead, I would like you to use real Javascript parser implemented in javascript.c.
Promise API (http://docs.ctags.io/en/latest/internal.html#promise-api) I designed and implmented can be uesd for run a subparser(Javascript) from a parser(html). yacc.c is a good example to know the Promies API.

This is applicable to inline stylesheets; I would like you to run CSS parser from your html parser.

b4n:
About `CDATA`, I propose this test case:
```xml
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
  <title>CDATA tests</title>
</head>
<body>
  <div>

  <h1><![CDATA[Pears & apples]]></h1>
  <p>
  <![CDATA[<h1>Bug</h1>]]>
  </p>

  <![CDATA[<h1>]]>
  <h1>Cherries</h1>
  <h1>Strawberries</h1>

  </div>
</body>
</html>
```

However, it might not be so important to support `<![CDATA[` as it's apparently not valid in HTML, but only in XML and SGML (which includes XHTML served as XML, but not XHTML served as `text/html`).  And actually if you just open this file in Firefox or Chromium they won't display the expected content.
So it might not be worth the work implementing `<![CDATA[` support for true XHTML.

parsers/html.c
techee: OK, will try to improve the readability.

techee:
> However, it might not be so important to support <![CDATA[ as it's apparently not valid in HTML, but only in XML and SGML (which includes XHTML served as XML, but not XHTML served as text/html). And actually if you just open this file in Firefox or Chromium they won't display the expected content.

In this case I'll probably skip the support of CDATA for the first version of the parser - it can be done later.

techee:
@masatake Sure, will have a look at it.

techee:
@masatake Just had a brief look at the Promise API and have two questions:

1. Is it possible to specify position on the line where the input for the embedded language starts? E.g. can it handle something like:

```
<script>function foo(){}</script>
```

where both languages are mixed on the same line?

2. Can the host language pre-process the input for the embedded language? Within the <script> block there could be a HTML comment removing parts of the javascript code.

masatake:
> 
    Is it possible to specify position on the line where the input for the embedded language starts? E.g. can it handle something like:...

Yes. 
```
int  makePromise   (const char *parser,
                    unsigned long startLine, int startCharOffset,
                    unsigned long endLine, int endCharOffset,
                    unsigned long sourceLineOffset);
````

You can specify it to `startCharOffset`.
The byte offset counted from the start of line should be specified to `startCharOffset`.
(Not from the start of file.)
For specifying the end position of the area, you can use `endCharOffset`.

>  Can the host language pre-process the input for the embedded language? Within the

(Maybe an incomplete comment is submitted.)

No, could you tell me more about this topic?

**Comment by b4n** [comment]
Pre-processing could work, but it would require a fairly more subtle API, writing to a (memory) stream and then using that as the input for the next parser.  That could work just fine, but it might require some non-trivial changes I imagine.  Though probably not much more than what would be required to have a full in-memory library API :)

**Comment by techee** [comment]
I was thinking about something simple like the possibility for the hosting parser to register a callback like

```
boolean filter_character(int c, void *user_data);
```

which would decide whether the input character should be passed to the embedded language's parser or not. But after thinking about it we couldn't use it in this form for HTML because we learn whether something is a comment too late (after reading 4 characters) and we couldn't ungetc() them so the embedded parser can read them.

Anyway, not parsing HTML comments within `<script>` isn't a big problem, everyone's using javascript comments there.

**Comment by masatake** [comment]
As an experiment I'm adding two hooks to rearrange input(char or string) passed to a parser.
They change the result of calling `getcFromInputFile` and `readLineFromInputFile`.

**Comment by masatake** [comment]
input-text-stream

There is a weakness; the regex match is run after "char preprocessing" is done. Ideally regex match should be run before the char level preprocessing.

**Comment by masatake** [comment]
> Anyway, not parsing HTML comments within <script> isn't a big problem, everyone's using javascript comments there.

I see. I will not implmenet char based preprocessor for awhile. (I will implement line based preprocessor for different purpose.)

I you have a question about makePromise, fell free to ask me. None has eaten the dog food ever.
If it is seriously trouble-full, change the role with me. I will try to add makePromise with your hints about new html parser.

**Comment by techee** [comment]
@masatake Done (and it works, cool!). I added a new function to read.c to get the offset from the beginning of the line because computing it in the parser would be annoying (and any other parser using the promise API would have to do the same). Please let me know if it's OK this way or if it should be modified some way.

@b4n Hopefully all your comments are addressed. If the patches are OK, I could squash some of the fixes to the initial commit.

File: main/read.c [hunk 6]
```diff
@@ -135,6 +135,13 @@ extern unsigned long getInputLineNumber (void)
 	return File.input.lineNumber;
 }
 
+extern int getInputLineOffset (void)
+{
+	unsigned char *base = (unsigned char *) vStringValue (File.line);
+	int ret = File.currentLine - base - File.ungetchIdx;
+	return ret >= 0 ? ret : 0;
```
b4n: unfortunately if `ungetchIdx` is greater than `File.currentLine - base` (i.e. is on the previous line), both `getInputLineNumber()` and `getInputLineOffset()` will be wrong…

parsers/html.c
```diff
@@ -8,35 +8,393 @@
...
+
+						makeSimpleTag (text, HtmlKinds, headingKind);
+					}
+				}
+				else if (startTag == KEYWORD_script)
+				{
+					keywordId endTag = lookupKeyword (vStringValue (token->string), Lang_html);
+					if (startTag == endTag)
+						makePromise ("JavaScript", startLineNumber, startLineOffset,
+									 endLineNumber, endLineOffset, startSourceLineNumber);
```
b4n: technically script could be something else than JavaScript, but I won't be that annoying :)

parsers/html.c
```diff
@@ -8,35 +8,393 @@
...
+				}
+				else if (startTag == KEYWORD_style)
+				{
+					keywordId endTag = lookupKeyword (vStringValue (token->string), Lang_html);
+					if (startTag == endTag)
+						makePromise ("CSS", startLineNumber, startLineOffset,
+									 endLineNumber, endLineOffset, startSourceLineNumber);
```
b4n: Same for style sections AFAIK

b4n:
Test results fail, maybe you forgot to include an *args.ctags*?

b4n:
Better :)  LGBI

main/read.c
techee: Right, I noticed that too but it shouldn't be a practical problem as one typically ungec()s either real code characters (new token) or whitespace characters (EOLs to stay at the previous line) but not both of them at the same time. So there shouldn't be any forgotten non-whitespace characters from the previous line and we'll only miss some whitespace characters which don't matter.

b4n: Agreed it's unlikely to really matter (even more so as it's likely to be 1 byte at most), but we never know until it bites :)

Anyway, not a blocker, it's a mere comment on a current limitation.

parsers/html.c
techee: What do you mean?

b4n: That they [can technically be something else than CSS](https://www.w3.org/TR/html5/document-metadata.html#attr-style-type), but again, forget that, nobody actually use anything else.

But well, if we wanted to be totally strict, we'd have to check whether the `type` attribute is provided and then if it's the CSS MIME type.

techee: Ah, OK, I didn't connect this case with the javascript case in my brain so I missed the meaning of the original sentence.

Yeah, these are rather theoretical uses right now.

masatake:
@techee, I sent an invitation for joining to Universal-ctags organization. Can I ask you to maintain html parser?

techee:
> @techee, I sent an invitation for joining to Universal-ctags organization. Can I ask you to maintain html parser?

Sure. I can also maintain the Go parser as I wrote some patches for it in the past and I'm quite familiar with it.

techee:
Should I squash the commits before it gets merged?

masatake:
Following tests are passed.

```
make units VG=1 LANGUAGES=HTML
make chop VG=1 LANGUAGES=HTML
make fuzz LANGUAGES=HTML VG=1
make noise LANGUAGES=HTML 
```

masatake:
I'm surprised that the API works well. None has tested it. 

I read the test case. CSS, JavaScript and HTML are included. Excellent! Thank you, @techee.
@b4n, thank you for reviewing the changes.

From my view, a code reader, capturing a `<link>`  as a reference tag will be next step.


Could you merge `c90cd07` first? When merging `c90cd07`, could you change the prefix of commit title from `read` to `main`?

The rest of changes about new html parser, could you use `HTML`?
They should be squashed into one. In addition could you write about docs/news.html?
The new parser should be categorized to `Fully rewritten parsers`.

I'm quite happy if you update the Promise API (http://docs.ctags.io/en/latest/internal.html#promise-api).

techee:
@masatake Yep, can make the changes you propose. What do you mean by

> The rest of changes about new html parser, could you use HTML?

?

techee:
> From my view, a code reader, capturing a <link> as a reference tag will be next step.

What is a "reference tag"?

masatake:
> > @masatake Yep, can make the changes you propose. What do you mean by

>    The rest of changes about new html parser, could you use HTML?
> ?

I'm sorry. What I should write is "could oyu use HTML as PREFIX for the commit headers?".
e.g. `HTML:  	Return TOKEN_COMMENT also when EOF is reached`.

> What is a "reference tag"?

See http://docs.ctags.io/en/latest/news.html#reference-tags

```
<head>
<link rel="stylesheet" type="text/css" href="theme.css">
</head> 
```

This can be tagged as
```
theme.css ... /^<link rel="stylesheet" type="text/css" href="theme.css/;" kind:link role:stylesheet
```

This is just an idea. The kind and role design should be done by people who knows well this area:-P

I forgot writing one thing. You can add a document about the HTML parser to http://docs.ctags.io/en/latest/parsers.html .

masatake:
@techee, the reference tag is just an idea for future development. Of couse you can merge the current change.

techee:
> @techee, the reference tag is just an idea for future development. Of couse you can merge the current change.

Yes, I understood it that way and won't do it right now. I'll just write some documentation, push the squashed commit here so you can have a look if everything is alright and it can get merged then.

techee:
@masatake Done (without the <link> implementation). Does it look OK like this?

parsers/html.c
```diff
@@ -8,35 +8,405 @@
...
+		case '>':
+		{
+			token->type = TOKEN_TAG_END;
+			break;
+		}
+
+		case '=':
+			token->type = TOKEN_EQUAL;
+			break;
+
```
masatake: For the case `'>'`, you put `{` and `}` around the statements after the case label.
However, the case `'='`, you don't.  Could you do the same (putting or not putting)?

masatake:
About the commit log:
```
- javascript function parsing is still based on regular expression
      matching with all its limitations
```

This is incorrect now:-)

masatake:
I pointed two, a style issue and an item in the commit log.
Other than the two, LGTM.

masatake:
@techee, thank you for updating internal.rst.

techee:
@masatake Fixed. OK to merge?

masatake:
@techee, yes.

masatake:
Thank you.

PR #1081, opened 2016-08-19
python: Fix crash with dotted variables missing an explicit initializer

### LINKED ISSUE #1080: ctags crashes on Python dotted tuple destructuring
Following Python code crashes the ctags I just built from git master source:

a, x.a = y

```gdb
Program received signal SIGSEGV, Segmentation fault.
0x0000000000441dac in initPythonEntry (e=0x7fffffffe500, token=0x0, kind=K_VARIABLE) at parsers/python.c:223
(gdb) where
#0  0x0000000000441dac in initPythonEntry (e=0x7fffffffe500, token=0x0, kind=K_VARIABLE) at parsers/python.c:223
#1  0x0000000000441fc8 in makeSimplePythonTag (token=<optimized out>, kind=<optimized out>) at parsers/python.c:309
#2  0x00000000004436e4 in parseVariable (kind=K_VARIABLE, token=0x687180) at parsers/python.c:1154
#3  findPythonTags () at parsers/python.c:1233
#4  0x0000000000413815 in createTagsForFile (passCount=1, language=52) at main/parse.c:2122
#5  createTagsWithFallback1 (language=52) at main/parse.c:2146
#6  0x0000000000414ff7 in createTagsWithFallback (mio=<optimized out>, language=52, fileName=0x6874f0 "/tmp/test.py") at main/parse.c:2206
#7  parseFile (fileName=0x6874f0 "/tmp/test.py") at main/parse.c:2366
#8  0x000000000040bc87 in createTagsForEntry (entryName=0x6874f0 "/tmp/test.py") at main/main.c:259
#9  0x000000000040bf60 in createTagsForArgs (args=0x687480) at main/main.c:304
#10 batchMakeTags (args=0x687480, user=<optimized out>) at main/main.c:461
#11 0x000000000040b9ce in runMainLoop (args=0x687480) at main/main.c:433
#12 main (argc=<optimized out>, argv=0x7fffffffe950) at main/main.c:569
```

I think it might be related to the (relatively new) code in parsers/python.c:

		if (token->type == '.')
		{
			/* FIXME: what to do with dotted names?  We currently ignore them
			 *        as we need to do something not to break the whole
			 *        declaration, but the expected behavior is questionable */
			deleteToken (name);
			name = NULL;

This leaves name NULL, and it later gets added to the nameTokens array:

		nameTokens[nameCount++] = name; /// can be NULL

b4n:
Fixes #1080.

### DIFF
--- a/Units/parser-python.r/dotted-variable-leftovers.d/expected.tags
+++ b/Units/parser-python.r/dotted-variable-leftovers.d/expected.tags
@@ -1,0 +1,3 @@
+X	input.py	/^class X(list):$/;"	c
+a	input.py	/^a, x.a = x$/;"	v
+x	input.py	/^x = X()$/;"	v
--- a/Units/parser-python.r/dotted-variable-leftovers.d/input.py
+++ b/Units/parser-python.r/dotted-variable-leftovers.d/input.py
@@ -1,0 +1,8 @@
+class X(list):
+    pass
+
+x = X()
+x.append(1)
+x.append(2)
+
+a, x.a = x
--- a/parsers/python.c
+++ b/parsers/python.c
@@ -1150,8 +1150,11 @@ static boolean parseVariable (tokenInfo *const token, const pythonKind kind)
 
 		/* if we got leftover to initialize, just make variables out of them.
 		 * This handles cases like `a, b, c = (c, d, e)` -- or worse */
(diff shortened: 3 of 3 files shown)


masatake:
LGTM. Thank you.

PR #1083, opened 2016-08-20
Geany sync, part 1

techee:
These are small fixes brought from Geany which I noticed when working on https://github.com/geany/geany/pull/1160 or which resulted as a reaction to review comments. Mostly just variable renames, whitespace changes and comment fixes.

The last patch however is something we need for Geany. The macro `__unused__` unfortunately clashes with the same macro defined in glib which we use in Geany and we need to rename it to something else. In Geany UNUSED and PRINTF were used so I kept these but I can rename them to whatever else.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -1,4 +1,3 @@
-
 /*
 *   Copyright (c) 1996-2002, Darren Hiebert
 *
--- a/main/entry.c
+++ b/main/entry.c
@@ -78,7 +78,7 @@
 typedef struct eTagFile {
 	char *name;
 	char *directory;
-	MIO *fp;
+	MIO *mio;
 	struct sNumTags { unsigned long added, prev; } numTags;
 	struct sMax { size_t line, tag; } max;
 	vString *vLine;
@@ -146,9 +146,9 @@ extern const char *tagFileName (void)
 *   Pseudo tag support
 */
 
-extern void abort_if_ferror(MIO *const fp)
+extern void abort_if_ferror(MIO *const mio)
 {
-	if (mio_error (fp))
+	if (mio_error (mio))
 		error (FATAL | PERROR, "cannot write tag file");
 }
 
@@ -212,7 +212,7 @@ extern void makeFileTag (const char *const fileName)
 }
 
 static void updateSortedFlag (
-		const char *const line, MIO *const fp, MIOPos startOfLine)
+		const char *const line, MIO *const mio, MIOPos startOfLine)
 {
 	const char *const tab = strchr (line, '\t');
 
@@ -224,7 +224,7 @@ static void updateSortedFlag (
 		{
 			MIOPos nextLine;
 
-			if (mio_getpos (fp, &nextLine) == -1 || mio_getpos (fp, &startOfLine) == -1)
+			if (mio_getpos (mio, &nextLine) == -1 || mio_getpos (mio, &startOfLine) == -1)
 				error (WARNING, "Failed to update 'sorted' pseudo-tag");
 			else
 			{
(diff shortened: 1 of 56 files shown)


parsers/objc.c
```diff
@@ -23,17 +23,6 @@
...
-/* To get rid of unused parameter warning in
- * -Wextra */
-#ifdef UNUSED
-#elif defined(__GNUC__)
-# define UNUSED(x) UNUSED_ ## x __attribute__((unused))
-#elif defined(__LCLINT__)
-# define UNUSED(x) /*@unused@*/ x
-#else
-# define UNUSED(x) x
-#endif
```
b4n: these definitions of UNUSED aren't the same as the one you imported, so callers should be fixed (if any)

parsers/ocaml.c
```diff
@@ -21,16 +21,6 @@
...
-/* To get rid of unused parameter warning in
- * -Wextra */
-#ifdef UNUSED
-#elif defined(__GNUC__)
-# define UNUSED(x) UNUSED_ ## x __attribute__((unused))
-#elif defined(__LCLINT__)
-# define UNUSED(x) /*@unused@*/ x
-#else
-# define UNUSED(x) x
-#endif
```
b4n: ditto

main/entry.c
```diff
@@ -505,13 +505,13 @@ static int replacementTruncate (const char *const name, const long size)
...
-	MIO *fp;
+	MIO *mio;
 
 	/*  Open/Prepare the tag file and place its lines into allocated buffers.
 	 */
 	if (TagsToStdout)
 	{
-		fp = TagFile.fp;
+		mio = TagFile.mio;
 		mio_seek (fp, 0, SEEK_SET);
```
b4n: Travis tells us you forgot to rename `fp` in the `mio_seek()` call :)

b4n: and there's more below in this function

codebrainz:
To avoid future clashes, you could rename them to something like `CTAGS_UNUSED` and `CTAGS_PRINTF` or something :)

parsers/objc.c
techee: The only possibly valuable thing I see here is the `__LCLINT__ ` part - should I add it to the global definition of UNUSED? Otherwise it tries to do the same as the global UNUSED.

b4n: no, its API is `UNUSED(x)`, not `UNUSED`.  So yeah, in most cases here it's no a problem because you apparently can parenthesize argument definitions, but it still leads to a different (unexpected) expansion.

about `__LCLINT__` I don't know.

techee: Ah, I haven't noticed that, will fix.

But that means we had it wrong in Geany because there the global macro is called UNUSED too so all these different definitions get skipped by the first ifdef.

b4n: And we have the ObjectiveC parser, indeed.  Well, we're lucky that C allows `type (param)` just as well as `type param` apparently.

techee:
> To avoid future clashes, you could rename them to something like CTAGS_UNUSED and CTAGS_PRINTF or something :)

Yeah, maybe a good idea. Any other suggestions regarding the name?

b4n:
> Yeah, maybe a good idea. Any other suggestions regarding the name?

maybe drop some concept of `ATTRIBUTE` in somewhere not to get anyone confuse `CTAGS_PRINTF` with a printf-like macro.

techee:
Could be

```
CTAGS_ATTR_UNUSED
CTAGS_ATTR_PRINTF
```

or we could also keep the lowercase letters

```
__ctags_attr_unused__
__ctags_attr_printf__
```

It's getting somewhat long though.

codebrainz:
> or we could also keep the lowercase letters

You'd have to remove the leading double-underscores for it to be valid C code.

codebrainz:
> maybe drop some concept of ATTRIBUTE in somewhere not to get anyone confuse CTAGS_PRINTF with a printf-like macro

Or just call that one `CTAGS_PRINTF_FORMAT` or such. Some compilers have no concept of `__attribute__()`.

b4n:
> ```
CTAGS_ATTR_UNUSED
CTAGS_ATTR_PRINTF
```

I like that

> Or just call that one `CTAGS_PRINTF_FORMAT` or such. Some compilers have no concept of `__attribute__()`.


Bah, how it's implemented doesn't matter, so the name should just be clear enough for the developers, no need to worry about whether it's implemented with `__attribute__` or some other magic.

**Comment by codebrainz** [comment]
> Bah, how it's implemented doesn't matter, so the name should just be clear enough for the developers, no need to worry about whether it's implemented with __attribute__ or some other magic.

I just meant "attribute" is kind of GCC-specific, so it might not make sense to every developer whereas something generic like `CTAGS_PRINTF_FORMAT` resolves the ambiguity with a possible `CTAGS_PRINTF` macro and adds more useful info to the symbol (namely that the printf _format_ string will be checked).

**Comment by techee** [comment]
I've repushed the patches with the MIO variable fix squashed and with the new attr names. I used

```
CTAGS_ATTR_UNUSED
CTAGS_ATTR_PRINTF
```

as these seem to be most descriptive to me.

masatake:
I made a pull request for updating figures related to `fp` and `mio` fields.

LGTM. Feel free to merge the chages.

techee:
@masatake Merged, thanks for the figure updates.

PR #1087, opened 2016-08-22
Support JSX files

### LINKED ISSUE #1086: Add JSX support
JSX is a variant of JS that looks like this:

```jsx
var CommentBox = React.createClass({
  render: function() {
    return (
      <div className="commentBox">
        Hello, world! I am a CommentBox.
      </div>
    );
  }
});
ReactDOM.render(
  <CommentBox />,
  document.getElementById('content')
);
```

Documentation is here: https://facebook.github.io/jsx/ and files typically have a .jsx extension.

ctags could probably get away with just treating .jsx files as .js files.

Wilfred:
Fixes #1086. I've tested it and as far as I can see, the jscript parser Just Works with .jsx files.

### DIFF
--- a/Units/simple-jsx.d/expected.tags
+++ b/Units/simple-jsx.d/expected.tags
@@ -1,0 +1,1 @@
+foo	input.jsx	/^function foo() {$/;"	f
--- a/Units/simple-jsx.d/input.jsx
+++ b/Units/simple-jsx.d/input.jsx
@@ -1,0 +1,7 @@
+function foo() {
+  var x = <div>
+    <Menu />
+  <div>;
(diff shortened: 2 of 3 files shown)


masatake:
Could you put  the url `https://facebook.github.io/jsx/` as comment ?
I would like you to add a test case to Units/parser-javascript.r/jsx-simple.d.
How to write a test case is described in `http://docs.ctags.io/en/latest/units.html`.

Wilfred:
@masatake I've added tests and a comment, does this look good now?

masatake:
Thank you very much.

PR #1089, opened 2016-08-23
manual: switch the base file format of manual page to rst

masatake:
The format conversion from nroff to rst is done by @ffes.
I take over the work.

Changes done by Universal-ctags project have been documented in
a file named "hacking guide" with rst format. The man page (ctags.1)
has not been updated well. The man page is written in nroff format.

Using different formats in the documents makes hard to share contents
between two files.

This commit utilizes rst2man, a tool for converting the format of
input files from rst to nroff.  It means contents can share between
two documents.

Here after man page should be written rst format (ctags.1.in.rst).
You will get the nroff formatted man page(ctags.1) with following two
steps:

    1. autogen.sh (for bootstrap from code just checked out from git
    repository) and/or make convert the rst file to a nroff template
    (ctags.1.in).

    2. configure script converts the nroff template to the nroff
    file. (ctags.in).  In the environment where rst2man is available,
    make runs config.status to do the both step 1 and 2 when the
    command detects ctags.1.in.rst is modified.

For users who are in platforms where rst2man is not available,
when you change ctags.1.in.rst, you should commit the both changes of
ctags.1.in.rst and ctags.1.in.


masatake:
In the future we should merge docs and man directories.

configure.ac
```diff
@@ -301,6 +301,8 @@ AC_SYS_LARGEFILE
 AC_CHECK_PROGS([perl_found], [perl], [yes], [no])
 AM_CONDITIONAL([RUN_OPTLIB2C], [test "${perl_found}" = "yes"])
 
+AC_CHECK_PROGS([rst2man_found], [rst2man], [yes], [no])
```
codebrainz: According to [the docs](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html) the 3rd parameter is `value-if-not-found` and the 4th is the search path, unlike the singular `AC_CHECK_PROG` macro.

I usually do like this:

```
AC_PATH_PROG([RST2MAN], [rst2man], [no])
AM_CONDITIONAL([HAVE_RST2MAN], [test "x$RST2MAN" != "xno"])
```

Then in the `Makefile.am` you can call `$(RST2MAN)` to use the same binary Autoconf found, or even use `$(AM_V_GEN)$(RST2MAN)` if you want to also support silent-rules.

Makefile.am
```diff
@@ -96,3 +96,8 @@ include $(top_srcdir)/makefiles/test-cases.mak
 
 include $(top_srcdir)/makefiles/help.mak
 
+if HAVE_RST2MAN
+%.1.in: %.1.in.rst
```
codebrainz: This might not be portable to non-GNU make. I'm not sure of the exact portable syntax for suffix rules with Automake and these complicated extensions, but I think it's something like:

```
SUFFIXES = .1.in .1.in.rst
.1.in.rst.1.in:
    rst2man $< $@
```

Also it might cause a weird dependency problem if this rule produces files that Autoconf substitutes in, where you need to run `make` once first to produce the Autoconf input file and then run `configure` again to produce the final output, or it even might not work at all if you tried with a clean source tree. It's also entirely possible I just misunderstood what's happening here :)

ffes:
Not sure about storing both files, although I understand the reason behind the idea.

On Cygwin their is no `rst2man` but there is `rst2man.py`, and it works fine. So creating the nroff file is possible but can not be done automatically at the moment. Not sure why it is named `rst2man.py`. I will ask the maintainer of that package.

The problem with merging the man directory into the docs is the `.in` part of the file. But it is quite important to have the man page on the documentation site.

configure.ac
masatake: Thank you.

Makefile.am
masatake: Thank you for comment.


The rule is for helping a developer who writes the man page.
I expect such person may install both rst2man and GNU make.

`.in` file is recorded to our git repository. The `.in` file is for people who doesn't modify the file.
In such case the rule will never used; `configure` just generates the man page from the `.in` file.

codebrainz: Wouldn't it be reasonable to assume that someone checking out from Git either has (or can install) the tool, or if not, that file just won't be generated?

For the proper release source tarball, it would of course include the generated file so nobody would need the tool unless they wanted to edit the RestructuredText file.

b4n:
Hum, why not use `ctags.1.rst.in` and avoid any hack, instead of using `ctags.1.in.rst` and having hacks all over it?  I mean, just let Autoconf do the replacements in the rst, rather than after parsing the rst.  Or did I miss something?

b4n:
Like https://github.com/b4n/fishman-ctags/commit/fe9428e63671f209b89e57802bdae3b8d8a2c4f9.  Of course, this one *requires* rst2html, but if it's a problem *man/ctags.1* could be tracked, which IMO isn't really worse than what's here already.

b4n:
(I also took the opportunity to make the `.1.rst` -> `.1` rule use suffix rules, because `$<` is only portable in that context)

Makefile.am
b4n: > I expect such person may install both rst2man and GNU make.

That's not really a reason not to make a very simple modification for the rule to work everywhere, is it?

masatake:
@b4n, your one is better than mine. I wil rearrange this pull request with including your change.

masatake:
Updated.

PR #1091, opened 2016-08-25
main: Implement iFileGetLine() using mio_gets()

techee:
This is an experiment I tried to improve performance of the reading code - more below. Comments are welcome.

This avoids many calls of mio_getc() and in addition eliminates many
calls of vStringPut() which get expensive when called for every
single character.

When tested on the new HTML parser, this change increases the speed to
85MB/s from 70MB/s on my machine. When modified the parser so it just
calls getcFromInputFile() until EOF, the speed gets up to about 160MB/s
from about 110MB/s. When profiled, iFileGetLine() almost disappears from
the profile and the CPU usage is distributed between getcFromInputFile()
and mio_gets() (because these two have to read every character). Other
parser's speed improvement depends on how much they are optimized -
the faster the parser is, the more it will benefit from this change.

The change requires a slight modification of mio_gets() to handle also
CR characters - not sure if it is compatible with gets() implementation.
If not, the function could be renamed to something else.

The greatest part of the patch is the change of the code dealing with
#line directives to read characters from a string instead of MIO input
but this change is mostly straightforward.

For some reason the c-heapoverflow-sh-bug-363 test doesn't generate the
tag now - I'm not sure if it's OK this way but this test case seems to
test something different and the generated tag is strange anyway.


main/read.c
```diff
@@ -676,61 +671,6 @@ static void fileNewline (void)
...
-		File.newLine = TRUE;
-		mio_getpos (File.mio, &StartOfLine);
-	}
-	else if (c == CRETURN)
-	{
-		/* Turn line breaks into a canonical form. The three commonly
-		 * used forms if line breaks: LF (UNIX/Mac OS X), CR (Mac OS 9),
-		 * and CR-LF (MS-DOS) are converted into a generic newline.
-		 */
-#ifndef macintosh
```
techee: macintosh doesn't seem to be defined anywhere so it's as if this check wasn't here.

b4n: isn't it supposed to be defined by the compiler on OS9 and below?

main/mio.c
```diff
@@ -990,7 +990,10 @@ char *mio_gets (MIO *mio, char *s, size_t size)
 				mio->impl.mem.pos++;
 				if (s[i] == '\n')
 				{
-					i++;
+					if (i > 0 && s[i-1] == '\r')
+						s[i-1] = '\n';
+					else
+						i++;
```
b4n: hum, IIUC this just normalizes line endings to `\n`, right? seems wrong to me to do it on MIO's side.

But maybe MIO should consider `\r` a line end though, indeed.  and then also eat a following `\n` probably.
```diff
 				if (s[i] == '\n')
 				{
 					i++;
 					break;
 				}
+				else if (s[i] == '\r')
+				{
+					i++;
+					if (s[i] == '\n')
+						i++;
+					break;
+				}
 			}
 			if (i > 0)
			{
```

main/read.c
```diff
@@ -397,27 +397,27 @@ static boolean setSourceFileName (vString *const fileName)
...
+	skipWhite (str);
+	s = *str;
+	while (*s != '\0' && isdigit (*s))
 	{
-		lNum = (lNum * 10) + (c - '0');
-		c = mio_getc (File.mio);
+		lNum = (lNum * 10) + (*s - '0');
+		s++;
 	}
-	mio_ungetc (File.mio, c);
```
b4n: not sure, but shouldn't you `s--`?

main/mio.c
b4n: I'm not 100% sure of what is the "correct" behavior regarding naked `\r` on non-Macintosh OSes.  It would be interesting to check how different libcs implement gets() in that regard -- the manpage on my system only talks about `\n`, so the behavior here was correct, but maybe it's different with `\r` sequences depending on the OS.

That would warrant a test case in MIO's test suite… oh wait, it's not imported here :]  (j/k about that, but should be sorted out at some point).
Someone could maybe test that on Windows and see what the libc does?  and same on Linux, but I can also do that when I get the time.

b4n: BTW, CTags did something odd (for bad reasons if you ask me) with `\r`s without following `\n`: it handed them as bare characters, not terminating the line.  However, various tools do consider them line ends on GNU/Linux at least.

main/read.c
techee: I didn't quite understand why they did the ungetc() here as the character gets tested below. It should behave the same because this is followed by readLineNumber() which skips spaces at the beginning. This way just looks more correct to me.

main/read.c
techee: OK, no idea, will check. The question is whether it matters these days...

main/mio.c
techee:
```
+               else if (s[i] == '\r')
+               {
+                   i++;
```

This will require reading something into s[i] before the '\n' check happens but yeah, my version didn't treat \r only as a newline.

```
+                   if (s[i] == '\n')
+                       i++;
+                   break;
+               }
```

techee: > I'm not 100% sure of what is the "correct" behavior regarding naked \r on non-Macintosh OSes. It would be interesting to check how different libcs implement gets() in that regard -- the manpage on my system only talks about \n, so the behavior here was correct, but maybe it's different with \r sequences depending on the OS.

Yeah, this is the only possibly "unclean" thing about this patch because we should handle \r somehow and I'm not sure whether this will be compatible with libc (will have a look).

b4n: No, what I don't like is mostly the line end normalization here.

Especially, I just tested and `fgets()` from the my glibc does put both the `\r` and `\n` in the returned string, and I truly want MIO to keep having the same behavior whether run on memory or file backend.

Results of my tests on glibc:
* LF is put in the string
* CR/LF is put in the string
* CR is not handled specially.

So the current MIO code before these changes present the same behavior as glibc on GNU/Linux, so it's good.  Don't change that :)

techee: The problem is that for CR-only line endings the whole file will be treated as a single line. It's true there shouldn't be much of them these days so maybe it's alright-ish but not correct.

b4n: well, it is regarding the glibc's behavior.  If you want to handle it differently, I think it should be done outside of MIO -- although I get your point.

BTW, small program very easy to check how `fgets()` behaves:
```c

#include <stdio.h>
#include <string.h>

int main(void)
{
  char buf[32];

  if (fgets(buf, sizeof buf, stdin)) {
    size_t len = strlen(buf);
    size_t i;

    printf("read %lu bytes:\n", len);
    for (i = 0; i < len; i++) {
      printf("  buf[%lu] = ", i);
      switch (buf[i]) {
        case '\r': puts("CR"); break;
        case '\n': puts("LF"); break;
        case '\t': puts("LT"); break;
        case '\v': puts("VT"); break;
        /* ... */
        default:
          putc(buf[i], stdout);
          putc('\n', stdout);
          break;
      }
    }
  }
  return 0;
}
```
and feed it whatever you wanna check.

b4n: > The problem is that for CR-only line endings the whole file will be treated as a single line. It's true there shouldn't be much of them these days so maybe it's alright-ish but not correct.

BTW, CTags currently does read `\r` line endings as plain characters everywhere `macintosh` isn't defined, so not handling CR has current CTags behavior on most situations.

I don't *really* think it's fully sensible behavior as it seems inconsistent with many other tools (including GCC), but it's how it is since f158c872a72d9ba5042258abefdc07a47e797c5c (as a result of https://sourceforge.net/p/ctags/bugs/1773926/).
Also, how actually common are CR-only files anyway?

techee: > BTW, CTags did something odd (for bad reasons if you ask me) with \rs without following \n: it handed them as bare characters, not terminating the line. However, various tools do consider them line ends on GNU/Linux at least.

Indeed! But that's actually great for me because I can keep it this way and just check if what I get from gets() ends with CRLF and convert it to LF and keep CRs within the string without having to do anything.

Great bug! :-)

techee:
Should be fixed now - the extra handling of newlines has minimal impact on performance - just about 1%.

b4n:
> For some reason the c-heapoverflow-sh-bug-363 test doesn't generate the tag now - I'm not sure if it's OK this way but this test case seems to test something different and the generated tag is strange anyway.

I don't think any regression is acceptable until clearly understood, because it means some behavior changed.  I guess here it's due to the `\0` inside the input file.  It might be interesting to see whether we do want to support them or accept them being a newline -- I'm not sure what actually happens now.

b4n:
76c40043247c66730d70a611ab016ec0a0fb1872 wow, is that right?? but well, maybe it's not so extraordinary if the various things don't fit in the cache and the compiler assumes some side effect might happen.

main/mio.c
```diff
@@ -976,18 +976,21 @@ char *mio_gets (MIO *mio, char *s, size_t size)
 		if (size > 0)
 		{
 			size_t i = 0;
+			size_t pos = mio->impl.mem.pos;
+			size_t file_size = mio->impl.mem.size;
```
b4n: `buf_size` would probably be a better name here

techee:
> 76c4004 wow, is that right?? but well, maybe it's not so extraordinary if the various things don't fit in the cache and the compiler assumes some side effect might happen.

Yep, it really makes a difference. I was playing with both getcFromInputFile() (no success) and mio_gets() and this surprised me a bit - I would expect the compiler would do something like that by itself but at least with my gcc 5.4 and standard compiler options I get these results.

main/mio.c
techee: OK

techee:
> I don't think any regression is acceptable until clearly understood, because it means some behavior changed. I guess here it's due to the \0 inside the input file. It might be interesting to see whether we do want to support them or accept them being a newline -- I'm not sure what actually happens now.

Ah, OK, there's \0 - I think what happens now is that anything behind it is dropped till the end of the line because of vStringSetLength().

techee:
> I don't think any regression is acceptable until clearly understood, because it means some behavior changed. I guess here it's due to the \0 inside the input file. It might be interesting to see whether we do want to support them or accept them being a newline -- I'm not sure what actually happens now.

It will be hard to do anything about it - fgets() returns a null terminated string and nothing like number of bytes read so everything after the first \0 gets dropped. Previously \0s were silently skipped by vStringPut().

But I think this shouldn't matter much.

main/read.c
techee: Well, haven't found anything but I don't think it's necessary to maintain support for the old platform (no idea whether it actually builds there).

techee:
I tried also the file-based MIO backend (where I discovered the slight difference in mio_gets() behaviour I fixed in the last patch) and the speed difference is even more dramatic - it's now almost as fast as when using the memory backend.

I also played a bit with getcFromInputFile() - one could rewrite it as a macro so the most common case is inlined:

```
#define getcFromInputFile() \
	(File.ungetchIdx == 0 && File.currentLine != NULL && *File.currentLine != '\0' ? \
		*File.currentLine++ : getcFromInputFileInternal())
```

where getcFromInputFileInternal() is what our getcFromInputFile() looks like now. I got about 10% speed increase but it requires to make the File variable global plus moving all the data structures into the header and I'm not sure if it's worth it.

I also tried various compiler options: -O3 increased the speed to 99MB/s from 96MB/s (-O2) for the HTML parser so it might be worth considering. Nothing dramatic though.

techee:
Hmm, one of the Windows targets gives me

```
Testing union.f as Fortran                                  failed (unexpected exit status: 1)
Shrinking ./Units/parser-fortran.r/union.f.d/input.f as Fortran
[0        537]...ok(0)
the target command line exits normally against the original input
```

Any idea how to reproduce it?

dkearns:
> Any idea how to reproduce it?

The generated test files are usually left after a failure. Are there no *.tmp files in ./Units/parser-fortran.r/union.f.d/ ?

techee:
> The generated test files are usually left after a failure. Are there no *.tmp files in ./Units/parser-fortran.r/union.f.d/ ?

The problem is I haven't seen it locally - it only happens in one of the AppVeyor configurations.

techee:
Is the test somehow randomized? I can imagine if it puts \0 somewhere sometimes to the input the output will be different now as the handling of \0 changed.

techee:
@masatake Any idea what the error I mentioned above means and how to reproduce it? What is the

```
[0        537]...ok(0)
```

part?

masatake:
I will take a look it.

masatake:
Look at the end of log output.

```
#FAILED (unexpected-exit-status):       1
	parser-fortran.r/union.f
	............................................................
	    adding exclude pattern: .bzr
	    adding exclude pattern: .bzrignore
	    adding exclude pattern: CVS
	    adding exclude pattern: .cvsignore
	    adding exclude pattern: _darcs
	    adding exclude pattern: .deps
	    adding exclude pattern: EIFGEN
	    adding exclude pattern: .git
	    adding exclude pattern: .gitignore
	    adding exclude pattern: .gitattributes
	    adding exclude pattern: .hg
	    adding exclude pattern: .hgignore
	    adding exclude pattern: PENDING
	    adding exclude pattern: RCS
	    adding exclude pattern: RESYNC
	    adding exclude pattern: SCCS
	    adding exclude pattern: .svn
	    adding exclude pattern: *~
	    adding exclude pattern: .*.swp
	    adding exclude pattern: *.o
	    adding exclude pattern: *.a
	    adding exclude pattern: *.so
	    adding exclude pattern: *.obj
	    adding exclude pattern: *.lib
	    adding exclude pattern: *.dll
	    adding exclude pattern: *.exe
	    adding exclude pattern: *.gcno
	    adding exclude pattern: *.gcda
	    adding exclude pattern: *.class
	    adding exclude pattern: *.pyc
	    adding exclude pattern: *.pyo
	Reading initial options from command line
	  Option: --libexec-dir=./libexec
	Reset DriversPathList
	Append ./libexec\drivers to DriversPathList
	  Option: --libexec-dir=+./Units/parser-fortran.r/union.f.d
	Prepend ./Units/parser-fortran.r/union.f.d\drivers to DriversPathList
	  Option: --data-dir=+./Units/parser-fortran.r/union.f.d
	Prepend ./Units/parser-fortran.r/union.f.d\optlib to OptlibPathList
	  Option: -o -
	Reading command line arguments
	Get file language for ./Units/parser-fortran.r/union.f.d/input.f
		pattern: input.f
			#candidates: 1
				0: Fortran (extension: "f")
			#candidates after sorting and filtering: 1
				0: Fortran (extension: "f")
	OPENING ./Units/parser-fortran.r/union.f.d/input.f as Fortran language file
	Initialize parser: Fortran
	ctags.exe: out of memory
```

It seems that the memory is exhausted. I will inspect more.

masatake:
```
$ make units LANGUAGES=Fortran VG=1
...
Category: parser-fortran.r
Testing array-spec.f90 as Fortran                           failed (valgrind-error)
Testing array_spec.f90 as Fortran                           failed (valgrind-error)
Testing auto.f as Fortran                                   failed (valgrind-error)
Testing bug565813.f90 as Fortran                            failed (valgrind-error)
Testing bug620288.f as Fortran                              failed (valgrind-error)
...
```

masatake:
```
==14224== Invalid read of size 1
==14224==    at 0x41714E: iFileGetLine (read.c:702)
==14224==    by 0x4175BF: getcFromInputFile.part.3 (read.c:754)
==14224==    by 0x432817: getLineType (fortran.c:590)
==14224==    by 0x432817: getFixedFormChar (fortran.c:697)
==14224==    by 0x432817: getChar.part.7 (fortran.c:817)
==14224==    by 0x4330E4: getChar (fortran.c:1023)
==14224==    by 0x4330E4: readToken (fortran.c:996)
==14224==    by 0x434E6F: parseProgramUnit (fortran.c:2506)
==14224==    by 0x434E6F: findFortranTags (fortran.c:2545)
==14224==    by 0x4144B1: createTagsForFile (parse.c:2124)
==14224==    by 0x4144B1: createTagsWithFallback1 (parse.c:2145)
==14224==    by 0x41610C: createTagsWithFallback (parse.c:2206)
==14224==    by 0x41610C: parseFile (parse.c:2366)
==14224==    by 0x40C9F6: createTagsForEntry (main.c:259)
==14224==    by 0x40CEDF: createTagsForArgs (main.c:304)
==14224==    by 0x40CEDF: batchMakeTags (main.c:461)
==14224==    by 0x403417: runMainLoop (main.c:433)
==14224==    by 0x403417: main (main.c:569)
==14224==  Address 0x6304a37 is 7 bytes after a block of size 64 alloc'd
==14224==    at 0x4C2DC57: realloc (vg_replace_malloc.c:785)
==14224==    by 0x4186EA: eRealloc (routines.c:237)
==14224==    by 0x41A870: vStringResize (vstring.c:34)
==14224==    by 0x41A870: vStringAutoResize (vstring.c:52)
==14224==    by 0x4170C3: iFileGetLine (read.c:711)
==14224==    by 0x4175BF: getcFromInputFile.part.3 (read.c:754)
==14224==    by 0x432817: getLineType (fortran.c:590)
==14224==    by 0x432817: getFixedFormChar (fortran.c:697)
==14224==    by 0x432817: getChar.part.7 (fortran.c:817)
==14224==    by 0x4330E4: getChar (fortran.c:1023)
==14224==    by 0x4330E4: readToken (fortran.c:996)
==14224==    by 0x434E6F: parseProgramUnit (fortran.c:2506)
==14224==    by 0x434E6F: findFortranTags (fortran.c:2545)
==14224==    by 0x4144B1: createTagsForFile (parse.c:2124)
==14224==    by 0x4144B1: createTagsWithFallback1 (parse.c:2145)
==14224==    by 0x41610C: createTagsWithFallback (parse.c:2206)
==14224==    by 0x41610C: parseFile (parse.c:2366)
==14224==    by 0x40C9F6: createTagsForEntry (main.c:259)
==14224==    by 0x40CEDF: createTagsForArgs (main.c:304)
==14224==    by 0x40CEDF: batchMakeTags (main.c:461)

```

masatake:
units target with VG=1 is very helpful for debugging.

masatake:
@techee, can I ask you to continue debugging based on the valgrind output?

masatake:
`Running under valgrind and timeout` subsection in http://docs.ctags.io/en/latest/units.html will help you to know about `units VG=1`.

techee:
@masatake Thanks a lot, will check. I suspect it's an off-by-one error when I'm resizing the string.

techee:
Fixed. All tests pass with VG=1. I've also made it behave more like the original version so the only difference is handling of \0s.

There's a failure in AppVeyor again but this time it seems just something strange happened on their side as the binary cannot be found.

main/mio.c
```diff
@@ -999,8 +1004,10 @@ char *mio_gets (MIO *mio, char *s, size_t size)
 				s[i] = 0;
 				rv = s;
 			}
-			if (mio->impl.mem.pos >= mio->impl.mem.size)
+			if (!newline && pos >= buf_size)
```
b4n: hum, why?

b4n: Anyway: please do not change MIO memory implementation if it doesn't match the FILE implementation.

techee: @b4n That's why I introduced the change - it _didn't_ match the FILE implementation - see the comment or just try it. If you have a file with:

```
abcd\nEOF
```

and do gets() on it, it should first read `abcd\n` without setting EOF (because EOF hasn't been reached yet) and set EOF only on the subsequent call.

techee: You can just simply set MAX_IN_MEMORY_FILE_SIZE to say 1 so getMio() always returns the file backend and you'll see different unit test results without this patch.

masatake:
Any remaning issue in this PR?
I will merge this before working on #1135.

techee:
I'm not aware of any problem with this (apart from the slightly different behaviour of \0 which now terminates line but I think we don't have to worry about \0s in input source files much).

Would be nice to get it in, thanks.

techee:
I rebased this on top of master to fix merge conflicts after merging my other pull requests. Should be ready to merge.

main/read.c
masatake: This ungetc exists in exuberant-crags. Even in the initial commit of the subversion repository. I read the code, too. And I also think it works fine without this ungetc.

main/read.c
```diff
@@ -669,68 +658,12 @@ static void fileNewline (void)
...
-	{
-		File.newLine = true;
-		mio_getpos (File.mio, &StartOfLine);
-	}
-	else if (c == CRETURN)
-	{
-		/* Turn line breaks into a canonical form. The three commonly
-		 * used forms if line breaks: LF (UNIX/Mac OS X), CR (Mac OS 9),
-		 * and CR-LF (MS-DOS) are converted into a generic newline.
-		 */
```
masatake: Could you move this comment to XYZ ?

main/read.c
```diff
@@ -743,25 +676,54 @@ extern void ungetcToInputFile (int c)
...
-		if (c == '\n'  ||  (c == EOF  &&  vStringLength (File.line) > 0))
+		bool newLine;
+		bool eof;
+
+		mio_gets (File.mio, str, size);
+		vStringSetLength (File.line);
+		haveLine = vStringLength (File.line) > 0;
+		newLine = haveLine && vStringLast (File.line) == '\n';
+		eof = mio_eof (File.mio);
+		if (newLine && vStringLength (File.line) > 1 &&
```
masatake: To here, XYZ.

main/mio.c
```diff
@@ -984,18 +984,21 @@ char *mio_gets (MIO *mio, char *s, size_t size)
 		{
 			size_t i = 0;
 			bool newline = false;
+			size_t pos = mio->impl.mem.pos;
```
masatake: Could you put comment here?
Like: "Using local vairable here is for optimization.".

We don't have a test mechanism for detecting performance degrading.
So someone can remove your crafted code without understanding the intention for optimization when refactoring.

main/read.c
techee: Done. I just modified the comment a bit because the part about normalizing CR (Mac OS 9) newlines wasn't quite correct (it was done only when macintosh macro was defined which I guess was defined under Mac OS 9 but the normalization didn't happen on other platforms).

masatake:
Thank you.

masatake:
For working #1135, I will merge this.

PR #1095, opened 2016-08-27
Fix case insensitive keyword lookup use and optimize it

techee:
Noticed when I tried to optimize the new Python parser as it seemed a bit slow.

### DIFF
--- a/Units/review-needed.r/jsFunc_tutorial.js.t/expected.tags
+++ b/Units/review-needed.r/jsFunc_tutorial.js.t/expected.tags
@@ -26,7 +26,7 @@ addSalaryFunction	input.js	/^function addSalaryFunction(addition)$/;"	c
 addSalaryFunctionDT9	input.js	/^function addSalaryFunctionDT9(addition)$/;"	c
 ball0	input.js	/^var ball0=new DT1(); \/\/ ball0 now points to a new object$/;"	v
 ball1	input.js	/^var ball1=new DT2("creating new Ball");  \/\/ creates object & $/;"	v
-ball2	input.js	/^var ball2=new Object();$/;"	v
+ball2	input.js	/^var ball2=new Object();$/;"	c
 ball3	input.js	/^var ball3=new DT3(); \/\/ ball0 now points to a new instance of type Ball$/;"	v
 ball4	input.js	/^var ball4=new DT3();$/;"	v
 ball5	input.js	/^var ball5=new DT3();$/;"	v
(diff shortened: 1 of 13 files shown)


b4n:
LGBI.  And indeed, I never suspected `analyzeToken()` performed a case-insensitive analysis.

masatake:
The signatures of lookupKeywordLowercase and lookupKeyword are very different.
lookupKeywordLowercase takes a vString instance.
In other hand lookupKeyword takes `char*`.

My understanding is that `lookupKeyword*` is low level interface that takes `char*`. `analyzeToken` is high level interface stacked on the lower interface and takes 'vString`.
I agree with you the name `analyzeToken` is not good. However, the name implies the level of interface.

So I would like you to change the signature of `lookupKeywordLowercase`. I expect `lookupKeywordLowercase` takes `char*` as input.

Approach:
vStringCopyToLower is used in the current implementation.
You can introduce vStringCopySToLower. (Here `S` means C string.)
vStringCopyToLower resizes a vString specified with DEST for optimizing.
In vStringCopySToLower it is not needed because DEST vString is allocated in `lookupKeywordLowercase` and it grows constantly.

Using I thought strcasecmp was better. However, as you may know, it changes semantics of `lookupKeywordLowercase` a bit.

techee:
In that case I'd prefer factoring-out common things to a separate function with an extra boolean argument saying if it's case-sensitive or insensitive lookup and use `strcasecmp` or `strcmp` depending on the mode. Then we could have

```
lookupKeyword(char*);
lookupCaseKeyword(char*);
```

The semantics of lookupCaseKeyword() will be slightly different than lookupKeywordLowercase() but all the languages that use it want is case-insensitive comparison which will work just fine.

Is it OK do it this way?

masatake:
@techee, it is ok. Thank you.

techee:
I forgot about the hash value which I couldn't do case-insensitively. However, I solved it simply by adding another vString variable - the performance is still fine despite the extra copy.

masatake:
> I forgot about the hash value which I couldn't do case-insensitively. However, I solved it simply by adding another vString variable - the performance is still fine despite the extra copy.

How about changing hash function?

```
static unsigned int hashValue (const char *const string, langType language)
{
        const signed char *p;
        unsigned int h = 5381;

        Assert (string != NULL);

        /* "djb" hash as used in g_str_hash() in glib */
        for (p = (const signed char *)string; *p != '\0'; p++)
                h = (h << 5) + h + *p;

        /* consider language as an extra "character" and add it to the hash */
        h = (h << 5) + h + language;

        return h;
}

```

Instead of referring `*p` directory,  you can use `tolower(*p)`.

techee:
@masatake Yes you are right, it is more elegant. Done.

masatake:
LGTM.

Issue #1101, opened 2016-08-31
Python: "member" kind vs. "method"


blueyed:
With the Python parser, `K_METHOD` is used for class methods only (https://github.com/universal-ctags/ctags/blob/55d068620ed8f74c8e706cb7fbaeab78dcd0fc69/parsers/python.c#L243-L245).

It is referred / mapped to as 'member' though (https://github.com/universal-ctags/ctags/blob/55d068620ed8f74c8e706cb7fbaeab78dcd0fc69/parsers/python.c#L127).

Whereas non-function class members are referred to as "variables".

This seems to have changed recently (probably with the big refactor of the Python parser), since I used to exclude the "v" kind (which referred to local variables then before probably)?!

Also the long description for "variables" should be "class variables" to make it clearer.
I suggest changing the "member" in `PythonKinds` to "method" / "class method".

blueyed:
How does universal-ctags' list of kinds compare to Exuberant ctags' list here?

b4n:
> This seems to have changed recently (probably with the big refactor of the Python parser), since I used to exclude the "v" kind (which referred to local variables then before probably)?!

I'm not totally sure, but it's very possible that all variables were tagged as `variable`, where now there's a distinction between `variable` and `local`.  However, `local` is disabled by default.

> Also the long description for "variables" should be "class variables" to make it clearer.

Mmmyeah, but it's also used for root variables AFAIK, isn't it?

> I suggest changing the "member" in `PythonKinds` to "method" / "class method".

Wile it sounds better indeed, I didn't do it because it changes the kinds for not much reason (as `member` is the old name).

If we were to change all that, I'd suggest
 * `variable`: global variable
 * `member`: class variable
 * `local`: (function) local variable

blueyed:
> I'm not totally sure, but it's very possible that all variables were tagged as `variable`, where now there's a distinction between `variable` and `local`.  However, `local` is disabled by default.

Ok.

>> Also the long description for "variables" should be "class variables" to make it clearer.

> Mmmyeah, but it's also used for root variables AFAIK, isn't it?

Ah, forgot about that.

> If we were to change all that, I'd suggest
 * `variable`: global variable
 * `member`: class variable
 * `local`: (function) local variable

Makes sense.  But let's then not change it for now, since it will break the API for no good reason.

blueyed:
FWIW: exuberant-ctags has:
```
c  classes
f  functions
m  class members
v  variables
i  imports
```

[...]

PR #1102, opened 2016-08-31
Python: "member" => "method", "variables" => "class variables"

blueyed:
Fixes https://github.com/universal-ctags/ctags/issues/1101.

### DIFF
--- a/parsers/python.c
+++ b/parsers/python.c
@@ -122,17 +122,17 @@ static roleDesc PythonUnknownRoles [] = {
 };
 
 static kindOption PythonKinds[COUNT_KIND] = {
-	{TRUE, 'c', "class",    "classes"},
-	{TRUE, 'f', "function", "functions"},
-	{TRUE, 'm', "member",   "class members"},
-	{TRUE, 'v', "variable", "variables"},
-	{TRUE, 'I', "namespace", "name referring a module defined in other file"},
-	{TRUE, 'i', "module",    "modules",
-	 .referenceOnly = TRUE,  ATTACH_ROLES(PythonModuleRoles)},
-	{TRUE, 'x', "unknown",   "name referring a classe/variable/function/module defined in other module",
-	 .referenceOnly = FALSE, ATTACH_ROLES(PythonUnknownRoles)},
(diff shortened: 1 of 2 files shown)


parsers/python.c
```diff
@@ -122,17 +122,17 @@ static roleDesc PythonUnknownRoles [] = {
...
-	{TRUE, 'v', "variable", "variables"},
-	{TRUE, 'I', "namespace", "name referring a module defined in other file"},
-	{TRUE, 'i', "module",    "modules",
-	 .referenceOnly = TRUE,  ATTACH_ROLES(PythonModuleRoles)},
-	{TRUE, 'x', "unknown",   "name referring a classe/variable/function/module defined in other module",
-	 .referenceOnly = FALSE, ATTACH_ROLES(PythonUnknownRoles)},
+	{TRUE, 'c', "class",      "classes"},
+	{TRUE, 'f', "function",   "functions"},
+	{TRUE, 'm', "method",     "class methods"},
+	{TRUE, 'v', "variable",   "class variables"},
```
b4n: This is probably not true, AFAIR it is used for any non-local variables, e.g. also root global variables.

parsers/python.c
```diff
@@ -122,17 +122,17 @@ static roleDesc PythonUnknownRoles [] = {
...
-	{TRUE, 'm', "member",   "class members"},
-	{TRUE, 'v', "variable", "variables"},
-	{TRUE, 'I', "namespace", "name referring a module defined in other file"},
-	{TRUE, 'i', "module",    "modules",
-	 .referenceOnly = TRUE,  ATTACH_ROLES(PythonModuleRoles)},
-	{TRUE, 'x', "unknown",   "name referring a classe/variable/function/module defined in other module",
-	 .referenceOnly = FALSE, ATTACH_ROLES(PythonUnknownRoles)},
+	{TRUE, 'c', "class",      "classes"},
+	{TRUE, 'f', "function",   "functions"},
+	{TRUE, 'm', "method",     "class methods"},
```
b4n: Sounds better, but breaks kinds with the old parser for not much, so I don't know.

PR #1109, opened 2016-09-05
Add pointer array and pool data structures

techee:
More details are in the individual commits.

To demonstrate the use of pool, I used it in the python parser for strings and tokens. The performance increase isn't much visible at this point because the bottlenecks are somewhere else (#1091, #1095) but once these are in, the pool increases the parser speed from 40MB/s to 46MB/s on my machine for django sources.

### DIFF
--- a/main/objpool.h
+++ b/main/objpool.h
@@ -1,0 +1,39 @@
+/*
+*   Copyright (c) 2016, Jiri Techet
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   Defines generic pool for object reuse reducing the amount of allocations
+*   and deallocations.
+*/
+#ifndef CTAGS_MAIN_OBJPOOL_H
+#define CTAGS_MAIN_OBJPOOL_H
+
+/*
+*   INCLUDE FILES
+*/
+#include "general.h"  /* must always come first */
+
+#include "ptrarray.h"
+
+/*
+*   DATA DECLARATIONS
+*/
+typedef void * (*objPoolCreateFunc) (void);
+typedef void (*objPoolDeleteFunc) (void *data);
+typedef void (*objPoolClearFunc) (void *data);
+
+struct sObjPool;
+typedef struct sObjPool objPool;
+
+/*
+*   FUNCTION PROTOTYPES
+*/
+extern objPool *objPoolNew (unsigned int size,
+	objPoolCreateFunc createFunc, objPoolDeleteFunc deleteFunc, objPoolClearFunc clearFunc);
+extern void objPoolDelete (objPool *pool);
+extern void *objPoolGet (objPool *pool);
+extern void objPoolPut (objPool *pool, void *obj);
+
+#endif  /* CTAGS_MAIN_OBJPOOL_H */
--- a/main/ptrarray.c
+++ b/main/ptrarray.c
@@ -1,0 +1,152 @@
+/*
+*   Copyright (c) 1999-2002, Darren Hiebert
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   This module contains functions managing resizable pointer arrays.
+*/
+
+/*
+*   INCLUDE FILES
+*/
+#include "general.h"  /* must always come first */
+
+#include <string.h>
+
+#include "debug.h"
+#include "ptrarray.h"
+
+/*
+*   DATA DECLARATIONS
+*/
+
+struct sPtrArray {
+	unsigned int max;
+	unsigned int count;
+	void **array;
+	ptrArrayDeleteFunc deleteFunc;
+};
+
+/*
+*   FUNCTION DEFINITIONS
+*/
+
+extern ptrArray *ptrArrayNew (ptrArrayDeleteFunc deleteFunc)
+{
+	ptrArray* const result = xMalloc (1, ptrArray);
+	result->max = 8;
+	result->count = 0;
+	result->array = xMalloc (result->max, void*);
+	result->deleteFunc = deleteFunc;
+	return result;
+}
+
+extern void ptrArrayAdd (ptrArray *const current, void *ptr)
+{
+	Assert (current != NULL);
+	if (current->count == current->max)
+	{
+		current->max *= 2;
+		current->array = xRealloc (current->array, current->max, void*);
+	}
+	current->array [current->count++] = ptr;
+}
+
+extern void ptrArrayRemoveLast (ptrArray *const current)
+{
+	Assert (current != NULL);
+	Assert (current->count > 0);
+	--current->count;
+}
+
+/* Combine array `from' into `current', deleting `from' */
+extern void ptrArrayCombine (ptrArray *const current, ptrArray *const from)
+{
+	unsigned int i;
+	Assert (current != NULL);
+	Assert (from != NULL);
+	for (i = 0  ;  i < from->count  ;  ++i)
+		ptrArrayAdd (current, from->array [i]);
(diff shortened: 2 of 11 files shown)


main/ptrarray.c
```diff
@@ -0,0 +1,116 @@
...
+*   FUNCTION DEFINITIONS
+*/
+
+extern ptrArray *ptrArrayNew (ptrArrayDeleteFunc deleteFunc)
+{
+	enum { initialSize = 8 };
+	ptrArray* const result = xMalloc (1, ptrArray);
+	result->max = initialSize;
+	result->count = 0;
+	result->array = xMalloc (initialSize, void*);
```
b4n: could be `current->max` like on resize and drop the enum

main/ptrarray.c
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/* Combine array `from' into `current', deleting `from' */
+extern void ptrArrayCombine (ptrArray *const current, ptrArray *const from)
+{
+	unsigned int i;
+	Assert (current != NULL);
+	Assert (from != NULL);
+	for (i = 0  ;  i < from->count  ;  ++i)
+		ptrArrayAdd (current, from->array [i]);
```
b4n: `memcpy()` maybe?

main/strlist.h
```diff
@@ -15,16 +15,15 @@
...
 
 /*
 *   DATA DECLARATIONS
 */
 typedef struct sStringList {
-	unsigned int max;
-	unsigned int count;
-	vString    **list;
+	ptrArray *array;
 } stringList;
```
b4n: this could be merely `typedef ptrArray stringList`, and would avoid the intermediate allocation.

main/ptrarray.h
```diff
@@ -0,0 +1,57 @@
...
+*   INCLUDE FILES
+*/
+#include "general.h"  /* must always come first */
+
+/*
+*   MACROS
+*/
+#define ptrArrayCount(current) ((current)->count)
+#define ptrArrayItem(current, indx) ((current)->array[indx])
+#define ptrArrayLast(current) ((current)->array[(current)->count - 1])
```
b4n: dangerous if `count == 0`

b4n:
Why not.  Though, wouldn't a slice allocator be better?  Sure it would not keep the member pointers around, but it could reuse *any* block of the right size, not only in a specific pool.  And I'd guess we allocate many things of the same size all over the place.

BTW, 03050a51e302977cf4dca275a3367a288f48bf4a seem suboptimal to me: if it shows significant performance gain, shouldn't simply `vString`s have their pool and use that transparently?

techee:
>Why not. Though, wouldn't a slice allocator be better? Sure it would not keep the member pointers around, but it could reuse any block of the right size, not only in a specific pool. And I'd guess we allocate many things of the same size all over the place.

Possibly but making a generic one would be quite a few more lines of code, at least judging from

https://github.com/GNOME/glib/blob/master/glib/gslice.c

And once the objects get allocated and are in the pool, the pool should be faster because there's no need for any complicated lookup of a free slice.

Also the question is whether GSlice is faster these days:

http://stackoverflow.com/questions/28406339/why-slice-allocator-is-slower-than-malloc-in-glib2-0

> BTW, 03050a5 seem suboptimal to me: if it shows significant performance gain, shouldn't simply vStrings have their pool and use that transparently?

Good idea, I'll try to experiment with that. We should just make sure that too long strings aren't kept in the pool because over time they might consume more and more memory.

main/ptrarray.c
techee: OK.

main/ptrarray.c
techee: OK

main/strlist.h
techee: Yeah, good point.

main/ptrarray.h
techee: I maybe overdid it with the macros - they probably aren't necessary for the pointer array (they make sense for vString because one can really see performance difference between the macro implementation of vStringPut and the function one).

main/ptrarray.c
techee: Well, one would have to make sure the size of dest is enough which ptrArrayAdd() does for us and I'm a bit lazy to do that ;-). Especially when this function doesn't appear in the profiles.

b4n:
>Also the question is whether GSlice is faster these days:
>
>http://stackoverflow.com/questions/28406339/why-slice-allocator-is-slower-than-malloc-in-glib2-0

I don't know the answer, but as the accepted answer points out, the test case in the question are totally irrelevant for what GSlice is: I definitely would expect it to be slower when not re-using existing memory.  So yeah maybe GSlice is slower nowadays, maybe it could be improved -- or not --, and maybe glibc's malloc() is good enough nowadays.  But that's a lot of maybes I wouldn't draw any conclusion from :)

techee:
I tried using the global pool for vString - it didn't bring any extra performance. But I noticed one thing - cxx uses token pool of size 8192 each of which contains a string. This means in order to make the string pool functional, it's size would have to be more than 8192 and I'm a bit worried the size might get out of control. So I decided not to make this change.

In addition, I dropped the string pool in the python parser completely - the performance of the parser improves just by 1MB/s thanks to this and it's probably not worth it in this case. If some other parser needs to allocate strings more, it can consider using the pool.

masatake:
@techee, have you even think about introducing reference counting  to vString?
The places where vsting is copied are not few. I guess in some of the places, `_ref` function can be used  instead of copy function. That will gain performance.

techee:
@masatake I don't think it would help much - for instance in the Python parser strings are created to collect some information, more strings are concatenated to it and when finished, the string is deleted. I don't see where I would use ref here. Also I don't see anything dramatic related to string creation in the profiles.

Anyway, token creation seems to be more frequent and more expensive here.

masatake:
I would like to review this.

main/ptrarray.h
```diff
@@ -0,0 +1,50 @@
...
+
+typedef void (*ptrArrayDeleteFunc) (void *data);
+
+typedef struct sPtrArray {
+	unsigned int max;
+	unsigned int count;
+	void **array;
+	ptrArrayDeleteFunc deleteFunc;
+} ptrArray;
+
```
masatake: Is it possible to move this struct definition to ptrarray.c?
The other code should not see the internal of the data type.

main/objpool.h
```diff
@@ -0,0 +1,44 @@
...
+typedef void (*objPoolClearFunc) (void *data);
+
+typedef struct sObjPool {
+	ptrArray *array;
+	unsigned int size;
+	objPoolCreateFunc createFunc;
+	objPoolDeleteFunc deleteFunc;
+	objPoolClearFunc clearFunc;
+} objPool;
+
```
masatake: Could you move this to objpool.c?
If you could do it, you can remove ptrarray.h includsion, too.

(If circular reference of type definition is an issue in your work, main/types.h will help you.)

masatake:
I should review this early. I'm afraid confliction with the code introduced/update in cxx parser these days.

I'm very sorry to be late to reply.

main/ptrarray.h
techee: I can forward-declare the struct in the header and move this to the source. Similarly with the pool.

techee:
> I should review this early. I'm afraid confliction with the code introduced/update in cxx parser these days.

The change in the cxx parser is really trivial and I can rewrite the cxx part if there's some bigger change.

techee:
@masatake Done. I squashed the changes with the corresponding commits.

masatake:
Let's merge this.

masatake:
I'm afraid a conflict occurs  in another area. So I will merge this now.

PR #1110, opened 2016-09-05
Various minor optimizations

techee:
The biggest optimization from these is the isIdentChar() conversion to macros. The second (though much smaller improvement) is the conversion of vStringClear() to macro.

The remaining two patches should improve performance too but the improvement is almost invisible and these are more cleanups than anything else.

This should be the last of my general performance patches - the rest should be done parser by parser.

### DIFF
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -103,22 +103,6 @@ extern void vStringPut (vString *const string, const int c)
 }
 #endif
 
-extern void vStringCatS (vString *const string, const char *const s)
-{
-#if 1
-	const size_t len = strlen (s);
-	while (string->length + len + 1 >= string->size)/*  check for buffer overflow */
-		vStringAutoResize (string);
-	strcpy (string->buffer + string->length, s);
-	string->length += len;
-#else
-	const char *p = s;
-	do
-		vStringPut (string, *p);
-	while (*p++ != '\0');
-#endif
-}
-
 extern vString *vStringNewCopy (const vString *const string)
 {
 	vString *vs = vStringNew ();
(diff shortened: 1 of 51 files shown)


codebrainz:
Out of curiosity, why use function-like macros with all their problems when you can use C99 inline functions?

techee:
>Out of curiosity, why use function-like macros with all their problems when you can use C99 inline functions?

Well, I haven't seen them anywhere in the code so I'm not sure if C99 is supported by ctags in all environments. Is it?

codebrainz:
Not sure, but the C compiler is put into C99 mode in 25808ebec037938bb9eb3ecca43d61db554e7e80 (requested by @masatake).

techee:
I've added one more patch which improves string resizing. This basically obsoletes

https://github.com/universal-ctags/ctags/pull/1114

main/vstring.c
```diff
@@ -90,7 +83,7 @@ extern vString *vStringNew (void)
 extern void vStringPut (vString *const string, const int c)
 {
 	if (string->length + 1 == string->size)  /*  check for buffer overflow */
-		vStringAutoResize (string);
+		vStringResize (string, string->size * 2);
```
b4n: I'd rather just ask for the extra byte required here, and let the `vStringResize()` implementation do the doubling.  Here it's kind of misleadingly suggesting `vStringResize()` gives the exact requested size, and as it's not the case, adds an extra location where the doubling happens, which means it'd have to be changed if resizing policy ever changes.

main/vstring.c
```diff
@@ -115,8 +108,8 @@ extern vString *vStringNewInit (const char *const s)
 static void stringCat (
 		vString *const string, const char *const s, const size_t length)
 {
-	while (string->length + length + 1 >= string->size)
-		vStringAutoResize (string);
+	if (string->length + length + 1 > string->size)
+		vStringResize (string, string->length + length + 1);
```
b4n: that's better IMO: it requests the exact size it wants, and lets `vStringResize()` do what it wants.

main/vstring.c
techee: OK, will do (also at other places).

techee: On the other hand it may look strange from the caller's perspective requesting just 1 byte more - if you e.g. have a look at the use in read.c requesting 1 byte more seems very inefficient at the place the function is called. Maybe the doubling should be left as an implementation detail of vStringResize() and callers should request the size that makes sense at the particular place (which is more or less the case now).

techee:
@masatake Could this one be merged too? I'd like to evaluate if the pool implementation from #1109 improves things for other parsers too (at least from what I saw in the profiles, I think so) and add it to more parsers. It would be good to have this one in so I can see better where the other performance bottlenecks are.

masatake:
Sorry to be late. If you think my review is needed for a patch, assign it to me explicitly. Sorry for incontinence.
Next three months I will be in busy.

techee:
> Sorry to be late. If you think my review is needed for a patch, assign it to me explicitly. 

Not really, @b4n is the best static analyzer I've ever used :-). Just asking if you find the changes OK in general.

masatake:
> Not really, @b4n is the best static analyzer I've ever used :-). Just asking if you find the changes OK in genera.

Strongly agree with you.

masatake:
Anyway, I will take a look.

masatake:
@codebrainz, a PR converting macro to inline function is welcome:-).

techee:
> @codebrainz, a PR converting macro to inline function is welcome:-).

That would be trivial, the bigger question is whether we should require C99-capable parser or not.

codebrainz:
@techee if you mean the parser from a C99 compiler, that is already required by [this](https://github.com/universal-ctags/ctags/blob/07541c5491e714820e94a1f8f10705d7ecf98dc2/configure.ac#L277) as well as GNU extensions to C99 forced on [here](https://github.com/universal-ctags/ctags/blob/07541c5491e714820e94a1f8f10705d7ecf98dc2/configure.ac#L479).

masatake:
@techee, I'm very sorry, but could you solve the conflicitons?
Maybe about #1144.

I would like to get changes about `inline` in another PR.
I would like to merge this PR first.

techee:
Alright, I rebased this and other my pull requests on top of master - the boolean->bool change made conflicts almost everywhere which was a bit annoying to fix. Hopefully I didn't introduce some other bugs on the way.

masatake:
Thank you. I will merge this. (Basically I think the original author should do merge-operation if s/he is the member of the organization. This is one of exceptions. I would like to update the base line of code to avoid potential conflicts.)

Issue #1111, opened 2016-09-06
Segmentation fault on large projects


zegervdv:
When trying to run `ctags -R` on large projects it segfaults.

I have run a valgrind check:
~~~
==15076== Memcheck, a memory error detector
==15076== Copyright (C) 2002-2012, and GNU GPL'd, by Julian Seward et al.
==15076== Using Valgrind-3.8.1 and LibVEX; rerun with -h for copyright info
==15076== Command: /usr/local/bin/ctags -R
==15076== 
==15076== Warning: set address range perms: large range [0x16439028, 0x26439058) (noaccess)
==15076== Warning: set address range perms: large range [0x7aef0040, 0x9aef0040) (undefined)
==15076== Warning: set address range perms: large range [0x3aeef028, 0x5aeef058) (noaccess)
==15076== Invalid write of size 1
==15076==    at 0x42AEA2: cxxParserParseNextToken (cxx_parser_tokenizer.c:1231)
==15076==    by 0x4285E4: cxxParserParseBlock (cxx_parser_block.c:164)
==15076==    by 0x427D43: cxxParserMain (cxx_parser.c:1304)
==15076==    by 0x4142D1: createTagsWithFallback1 (parse.c:2124)
==15076==    by 0x415A76: parseFile (parse.c:2206)
==15076==    by 0x40C6C6: createTagsForEntry (main.c:259)
==15076==    by 0x40C64C: recurseIntoDirectory (main.c:145)
==15076==    by 0x40C71F: createTagsForEntry (main.c:255)
==15076==    by 0x40C64C: recurseIntoDirectory (main.c:145)
==15076==    by 0x40C71F: createTagsForEntry (main.c:255)
==15076==    by 0x40C64C: recurseIntoDirectory (main.c:145)
==15076==    by 0x40C71F: createTagsForEntry (main.c:255)
==15076==  Address 0x9aef0040 is 0 bytes after a block of size 1,073,741,824 alloc'd
==15076==    at 0x4A06C20: realloc (vg_replace_malloc.c:662)
==15076==    by 0x418ECA: eRealloc (routines.c:237)
==15076==    by 0x41A70C: vStringAutoResize (vstring.c:34)
==15076==    by 0x42AEB4: cxxParserParseNextToken (cxx_parser_tokenizer.c:1231)
==15076==    by 0x4285E4: cxxParserParseBlock (cxx_parser_block.c:164)
[...]

PR #1114, opened 2016-09-07
Be verbose when allocating too large string

techee:
Previously if the string got too large, string resizing just silently
didn't happen. End with a fatal error in this case so it's clear what
happens.

We could relay on the xMalloc() behaviour here but it's better to
terminate earlier than running out of memory (before the system tries
to find enough memory and starts swapping). Strings should never be
too large and it always means there's some bug in ctags so the patch
lowers the maximum string memory to 100MB which should be more than
enough for anything non-buggy.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -882,11 +882,9 @@ extern char *readLineRaw (vString *const vLine, MIO *const mio)
 					 *pLastChar != '\n'  &&  *pLastChar != '\r')
 			{
 				/*  buffer overflow */
-				reReadLine = vStringAutoResize (vLine);
-				if (reReadLine)
-					mio_seek (mio, startOfLine, SEEK_SET);
-				else
-					error (FATAL | PERROR, "input line too big; out of memory");
+				vStringAutoResize (vLine);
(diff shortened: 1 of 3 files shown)


techee:
This is why #1111 terminated with segmentation fault.

techee:
Now thinking about it maybe we should just remove the check in the string and try to realloc (which could fail and write an error message). Since whole source lines are read, I can imagine e.g. some minified json export with a huge single line where this could unnecessarily fail.

b4n:
Mmmmyeah I'm not a fan of an arbitrary limit like that.

> Now thinking about it maybe we should just remove the check in the string and try to realloc (which could fail and write an error message). Since whole source lines are read, I can imagine e.g. some minified json export with a huge single line where this could unnecessarily fail.

Good remark, so even more, it doesn't seem a good idea to limit the input line length to 100Mb.

techee:
This one could be dropped if #1110 gets merged.

techee:
Closing as #1110 contains a better fix.

PR #1116, opened 2016-09-07
Don't compare foreign values to enumeration type

### LINKED ISSUE #1115: Scary warnings with Clang from Travis
I see warnings like this:

```
parsers/eiffel.c:696:9: warning: comparison of constant -1 with expression of
      type 'keywordId' (aka 'enum eKeywordId') is always false
      [-Wtautological-constant-out-of-range-compare]
                                if (isKeyword (token, KEYWORD_NONE))
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/eiffel.c:37:59: note: expanded from macro 'isKeyword'
#define isKeyword(token,k)    (boolean) ((token)->keyword == (k))
                                         ~~~~~~~~~~~~~~~~ ^  ~~~
```

If really the compiler optimizes the test out because it assumes it can't happen, it'll result in absolutely unexpected -- and wrong -- behavior.
Affected parsers are: c eiffel flex fortran go jscript php python rpmspec sql tex vhdl

Should I prepare a patch that makes sure a plain integer is used to store the keyword, instead of the enumeration type?

b4n:
Clang warns when comparing an enumeration type with a value not found
in this enumeration:

```clang
warning: comparison of constant VALUE with expression of type 'TYPE' is
      always false [-Wtautological-constant-out-of-range-compare]
```

If then the compiler then decides to optimize the test away because it
assumes the test is indeed always false, it can lead to pretty subtle
and nasty issues.

Closes #1115.

### DIFF
--- a/parsers/c.c
+++ b/parsers/c.c
@@ -65,7 +65,7 @@ typedef enum eException {
 
 /*  Used to specify type of keyword.
  */
-typedef enum eKeywordId {
+enum eKeywordId {
 	KEYWORD_ALIAS, KEYWORD_ATTRIBUTE, KEYWORD_ABSTRACT,
 	KEYWORD_BOOLEAN, KEYWORD_BYTE, KEYWORD_BAD_STATE, KEYWORD_BAD_TRANS,
 	KEYWORD_BIND, KEYWORD_BIND_VAR, KEYWORD_BIT,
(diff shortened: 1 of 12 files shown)


b4n:
BTW, I don't *actually* think Clang will start to actually optimize this out, because it's likely too much code around depend on it, but still.

Also, another, perhaps better, way to fix this would be to replace the `typedef enum {} keywordId;` with `enum {}; typedef int keywordId; /* to allow KEYWORD_NONE */`

b4n:
Updated to use the different typedef instead.  If we want this, it should be squashed.

masatake:
LGTM. Could you squash the two into one?

techee:
LGTM too.

b4n:
@masatake done

masatake:
Thank you.

Issue #1117, opened 2016-09-07
Segmentation fault while trying to index file


dim-an:
ctags crashes with segmenation fault when trying to index [ptw32_threadStart.c](https://gist.github.com/dim-an/0ed34c1470ea8ee36a967ed2ab3fe23c) file

```
⚡ ./ctags --extra=+qf --fields=+aiS --c++-kinds=+p ~/ptw32_threadStart.c
[1]    665689 segmentation fault  ./ctags --extra=+qf --fields=+aiS --c++-kinds=+p ~/ptw32_threadStart.c
```

techee:
This looks similar to #1111 (no newline at the end of the file). Would you try if https://github.com/universal-ctags/ctags/pull/1113 helps (or simply if adding newline at the end of the file helps)?

dim-an:
neither worked for me :(

b4n:
Valgrind:
```valgrind
$ vg ./ctags --extra=+qf --fields=+aiS --c++-kinds=+p /tmp/ptw32_threadStart.c -o -
==30758== Memcheck, a memory error detector
==30758== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==30758== Using Valgrind-3.12.0.SVN and LibVEX; rerun with -h for copyright info
==30758== Command: ./ctags --extra=+qf --fields=+aiS --c++-kinds=+p /tmp/ptw32_threadStart.c -o -
==30758== 
==30758== Invalid read of size 1
==30758==    at 0x4C2DDD2: strlen (vg_replace_strmem.c:454)
==30758==    by 0x523ADA2: vfprintf (vfprintf.c:1637)
==30758==    by 0x41B603: mio_vprintf (mio.c:804)
==30758==    by 0x41B786: mio_printf (mio.c:870)
==30758==    by 0x411E05: addExtensionFields (output-ctags.c:147)
==30758==    by 0x411E05: writeCtagsEntry (output-ctags.c:200)
==30758==    by 0x403C07: writeTagEntry (entry.c:1096)
==30758==    by 0x405424: uncorkTagFile (entry.c:1149)
==30758==    by 0x413D93: createTagsWithFallback1 (parse.c:2174)
==30758==    by 0x415A1C: createTagsWithFallback (parse.c:2206)
==30758==    by 0x415A1C: parseFile (parse.c:2366)
==30758==    by 0x40C706: createTagsForEntry (main.c:259)
[...]

PR #1118, opened 2016-09-07
CXX: c namespace crash test

b4n:
Test case for crash described in #1117.


masatake:
I'm afraid I injected a bug related to this.

Issue #1120, opened 2016-09-09
cxx: incorrect huge typeref values


techee:
It looks the compiler can get confused and here

https://github.com/universal-ctags/ctags/blob/master/parsers/cxx/cxx_tag.c#L323

it can generate tokens with huge strings. For instance, for

https://github.com/torvalds/linux/blob/master/lib/locking-selftest.c

I get typeref in one tag like

```
[...]

PR #1123, opened 2016-09-12
CXX: Mitigate problem with preprocessor macros when emitting types.

pragmaware:
Attempt to filter out types that contain too many macros or other oddities gathered by mistake.

### DIFF
--- a/parsers/cxx/cxx_parser_function.c
+++ b/parsers/cxx/cxx_parser_function.c
@@ -1090,7 +1090,7 @@ int cxxParserEmitFunctionTags(
 		CXXToken * pTypeName;
 
 		if(pInfo->pTypeStart)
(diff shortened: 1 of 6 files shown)


masatake:
Do you have any test case?

pragmaware:
Heh... it's still not really a fix, so I don't have added one.

This code filters out some obvious cases like the ones referenced in #1120.
( See https://github.com/torvalds/linux/blob/master/lib/locking-selftest.c for an example).

However the real underlying issue is not fixed. The parser will still mistake (few/short) macros for parts of typenames.. Don't know if it's meaningful to add a test. I have just verified that this code does not break the old ones, that is it does not filter out too much.

pragmaware:
A better fix would be to filter out the macros from type names, but that would still work only in files in that the macros are defined. So I don't know if it's worth the effort.

masatake:
Oh, I see.

PR #1124, opened 2016-09-12
CXX: Fix bug 1117

pragmaware:
Refuse to handle :: scopes in C language. Fixes bug #1117


masatake:
LGTM. Thank you.

b4n:
Looks good.  Though, what about altering the tokenizer so it doesn't emit `CXXTokenTypeMultipleColons` in C mode?

PR #1128, opened 2016-09-13
lxcmd: use path arranged by ctags main side as %{input}

### LINKED ISSUE #1122: Pass --tag-relative to coffeetags
Hi, I have a setup for automatically run ctags on changes in my git repos and I use the multiple worktree setup, which means that the files (including the TAGS file) live somewhere else.

For normal tags it works just fine, but for the coffeescript tags it points into oblivion. After a bit of digging, I realized that the --tag-relative flag wasn't propagated to coffeetags (which supports it). It would be pretty handy if it did though.

masatake:
Close #1122.

--tag-relative option passed to ctags command was ignore when
xcmd is used because ctags prints the name of input files (%{input})
returned from the xcmd directly. (ctags doesn't pass --tag-relative option
to the xcmd).

This commit is for reflecting the specification of --tag-relative option
to tags files; instead of using the value returned from xcmd directly,
use the path arranged by ctags main side.

### DIFF
--- a/Tmain/xcmd-tag-relative-option.d/input.dog
+++ b/Tmain/xcmd-tag-relative-option.d/input.dog
@@ -1,0 +1,2 @@
+define: dog
+define: cat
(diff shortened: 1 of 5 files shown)


masatake:
TODO: update expected.tags files related to coffeetags

PR #1130, opened 2016-09-16
CXX: Add test case for pointer-to-array types

b4n:
Welcome to the pretty world of corner-case C syntax :)


b4n:
BTW, as the test shows, this doesn't work yet with the new C++ parser.  The OldC parser did get it right though (well, at least the names themselves):
```console
$ ./ctags --languages=+OldC --language-force=OldC -o- Units/parser-cxx.r/pointer-to-array.b/input.c
bar	Units/parser-cxx.r/pointer-to-array.b/input.c	/^int2ptr_func bar = foo;$/;"	v
baz	Units/parser-cxx.r/pointer-to-array.b/input.c	/^int2ptr (*baz)(void) = foo;$/;"	v
foo	Units/parser-cxx.r/pointer-to-array.b/input.c	/^static int (*foo(void))[2] {$/;"	f	file:
int2ptr	Units/parser-cxx.r/pointer-to-array.b/input.c	/^typedef int (*int2ptr)[2];$/;"	t	file:
int2ptr_func	Units/parser-cxx.r/pointer-to-array.b/input.c	/^typedef int (*(*int2ptr_func)(void))[2];$/;"	t	file:
main	Units/parser-cxx.r/pointer-to-array.b/input.c	/^int main(void) {$/;"	f
```

pragmaware:
Merging, so I can work on it.

b4n:
> Sometims I wonder why OldC works well for these a bit crazy input.

I would imagine it's due to the parser being being old, so having and many fixes, but also it extracting a lot less information.  For example, extracting the typedef name from `typedef int (*(*foo)(int))[2]` is easier if you are merely searching for an identifier that would be the name, instead of understanding the construct to extract name and actual type, etc.

pragmaware:
Heh... the nested parentheses make me go nuts :D
I need to think about something "generic" to make sense of them once for all.

masatake:
> Merging, so I can work on it.

@pragmaware, thank you.

PR #1131, opened 2016-09-17
CXX: Improve complex type parsing

pragmaware:
Improve parsing for complex variable and function types.
Addresses test case added in #1130 and also improves Units/parser-cxx.r/variables-prototypes.cpp.b/.


pragmaware:
Yep, it would be nice to find some "common" parsing routines... some refactoring, as the code is getting pretty complicated in certain places. I'll keep it in mind in the future commits.

PR #1132, opened 2016-09-17
awk: register aliases

masatake:
Though aliases (gawk and mawk) were declared, they were not registered.


masatake:
@dkearns, I missed this when reviewing your patch.

dkearns:
Sorry I must have missed that hunk.

You should have made me write the test. :)

PR #1133, opened 2016-09-17
Units, Awk: use a more realistic program for the simple test

dkearns:


masatake:
Could you add your test case instead of replacing the existing one?
This is not strong option but I think keeping the original one simple is not bad idea.
The simple one is not suitable for testing but helpful for understanding the original intent.

masatake:
I will take over this.

masatake:
I added your test case as 'simple2-awk.d'.
Thank you.

PR #1137, opened 2016-09-20
experimental yaml support based on libyaml

masatake:
yaml parser can pick anchor and alias.
ansibleplaybook parser based on the yaml parser can pick only names of plays.

More work is needed but better than nothing.


masatake:
Added support for recording end line.

parsers/ansibleplaybook.c
```diff
@@ -0,0 +1,233 @@
...
+						  yaml_token_t *token)
+{
+	struct yamlBlockTypeStack *s;
+
+	s = state->type_stack;
+	state->type_stack = s->next;
+
+	s->next = NULL;
+	if (s->associatedCorkIndex != CORK_NIL)
+		{
```
b4n: GNU style? :)

parsers/ansibleplaybook.c
```diff
@@ -0,0 +1,233 @@
...
+	else if (token->type == YAML_STREAM_END_TOKEN)
+		popAllBlockType (data, token);
+}
+
+static void *ansiblePlaybookPrepare(void)
+{
+	static struct ansiblePlaybookState state = {
+		.type_stack = NULL,
+		.play_detection_state = DSTAT_PLAY_NAME_INITIAL,
+	};
```
b4n: Unless I'm mistaken, this won't reset the state when parsing second and next files, won't that be a problem?

parsers/meta-yaml.h
```diff
@@ -0,0 +1,37 @@
...
+#define yaml_token_t void
+#endif
+
+typedef void (* yamlCallback) (yaml_token_t *token, void *data);
+
+struct yamlParserClient {
+	langType lang;
+	yamlCallback callback;
+	void* (* prepareForNewInput) (void);
+	void  (* teardownForTheLastInput) (void*);
```
b4n: I'm not sure those callbacks are well named, at first I'd imagine they are called mostly like the parser's `initialize()` and `finalize()`, as in my head "last input" has to mean "last input file" or something, I didn't expect it to be the start and end of *each* input file.
Maybe `inputStart` and `inputEnd` or something?

Issue #1134, opened 2016-09-18
The suffix + of the option argument of commad line option -I doesn't work.


tefbop:
I was trying to create a tag file for /usr/include/c++/5/future, but the parsing stopped at the macro `_GLIBCXX_VISIBILITY(default)`. I've tried using `-I _GLIBCXX_VISIBILITY+` to ignore the macro, but in vain. Manually removing the function-like macro invocation does work though.

masatake:
Thank you for reporting.
Here is a simplified one.

```
namespace foo _GLIBCXX_VISIBILITY(default)
{
  int x;
}
```

Though I specified `-I  _GLIBCXX_VISIBILITY`, both `foo` and `x` are not captured.

masatake:
In cxxParserParseNextToken of  `parsers/cxx/cxx_parser_tokenizer.c`:
```
	// FIXME: Handle ignore parens!
```

What @tefbop reported is about this FIXME.
That means this bug report should be recorded as a `.b` test case. I will do it.

pragmaware:
Fixed by #1151.

PR #1142, opened 2016-09-22
CXX: Fix bux #1134

pragmaware:
Implement skipping of parenthesis after ignored tokens specified as -I SOMETHING+.
Fixes bug #1134


parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1088,16 +1140,30 @@ boolean cxxParserParseNextToken(void)
...
 			{
 				CXX_DEBUG_PRINT("Ignore token %s",vStringValue(t->pszWord));
-				// FIXME: Handle ignore parens!
+
 				if(szReplacement && *szReplacement)
 				{
+					CXX_DEBUG_PRINT("The token has replacement %s: applying",szReplacement);
 					vStringClear(t->pszWord);
 					vStringCatS(t->pszWord,szReplacement);
+					// FIXME: Do we need to interpret the replacement?
```
b4n: it probably would be great, but I also think it's not likely to be very important.  But yeah if it could understand replacements to at least keywords like `const`/`extern`/`static` and combination of those, it probably would be awesome for when ignore tokens are really useful.

pragmaware: I can easily handle single keywords. Combinations are harder to do since they would require the token to be splitted again with a custom algorithm, then one would need to keep a backlog of tokens yet to be returned which would affect the performance of the existing algo... I don't think it's worth the overhead. I'll add a check for keywords.

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1161,7 +1161,45 @@ boolean cxxParserParseNextToken(void)
...
 
 				if(szReplacement && *szReplacement)
-					return TRUE; // already have a token to return
+				{
+					// Already have a token to return.
+					// Check again for keywords.
+					
+					// There is a bit of code duplication here but I'd rather
+					// avoid a function call and a switch as the keyword check
+					// must be fast (especially the one above).
```
b4n: would it be so slow or problematic, really?  at worse, an inlining hint to the compiler should help.
I'm a bit afraid of the code duplication here because it's already fairly non-trivial, so likely to grow even more complex in the future.

If the part handling `__attribute__` differently make it too annoying/complex, maybe just leave this part out, and simply have a `ccxParserLookupKeyword()` that only does the lookup and special case handling.

masatake: I support @b4n's comment.

pragmaware: There :)

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1165,43 +1169,10 @@ boolean cxxParserParseNextToken(void)
...
-							if(iCXXKeyword == CXXKeyword__ATTRIBUTE__)
-							{
-								// special handling for __attribute__
-								return cxxParserParseNextTokenCondenseAttribute();
-							}
-						}
-					}
-
-					return TRUE;
+					goto check_keyword;
```
b4n: while tempting, this can lead to infinite loop with inappropriate ignore tokens referencing one another, e.g. `-Ifoo=foo`

masatake: Should I add code for avoiding circular referencing to the main part?

pragmaware: I guess it's hard as loops can have an arbitrary number of steps.. I'll better just add a simple check in the cxx parser to disallow recursive ignore tokens.

masatake:
I would like to review this.Please,give me time.

masatake:
LGTM. Sorry for being late.

pragmaware:
My last conflict fixing commit seems to be quite big. See a734947. However the cumulative changes listed in this page seem to be OK. Am I messing up something?

masatake:
I think it is better to do `git rebase master` after updating master.

masatake:
If github interface convinces you, you can do everything locally and the do `git push upstream master:master`. A bit dangerous but sometimes I do so.

`git push --force upstream master:master` is forbidden in github side. So we will not get serious trouble through the direct accessing to `master` branch of upstream.

parsers/cxx/cxx_parser_tokenizer.c
b4n: Yeah resolving loops in the main part would be very tricky, and what do actually do with them?  Better have the user side not re-consider replacements (or if it has to, drop ignore tokens once processed, C-preprocessor-style).

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1079,29 +1136,58 @@ bool cxxParserParseNextToken(void)
...
+						goto check_keyword;
+					}
+	
 					return cxxParserParseNextToken();
 				}
 			}
 		}
 
-		t->bFollowedBySpace = t->bFollowedBySpace | isspace(g_cxx.iChar);
+		t->bFollowedBySpace = isspace(g_cxx.iChar);
```
b4n: How is that related?

pragmaware:
I'm a bit confused with the merging and not sure if it's OK anymore. I think it's safer if I re-do a new clean branch, reapply and recreate the pull request.

parsers/cxx/cxx_parser_tokenizer.c
pragmaware: If I'm not wrong that t->bFollowedBySpace should always read to false on that line.

masatake:
Feel free to do 'git push --force' this PR (as I did frequently :-P).

pragmaware:
Recreated as pull #1151

PR #1143, opened 2016-09-22
Misc: Replace true/false with TRUE/FALSE

pragmaware:
Minor fixes.

### DIFF
--- a/main/routines.c
+++ b/main/routines.c
@@ -392,10 +392,10 @@ extern boolean strToUInt(const char *const str, int base, unsigned int *value)
 	unsigned long ulong_value;
 
 	if(!strToULong(str, base, &ulong_value) || ulong_value > UINT_MAX)
-		return false;
+		return FALSE;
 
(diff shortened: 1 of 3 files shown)


masatake:
LKGTM. Thanks.

codebrainz:
Why use non-standard macros (with no namespace prefix) when `true` and `false` are part of the standard library? Seems like it would be better to do the opposite replacement across the whole code base.

t-b:
What C standard support is required for compiling utags? If C99 than stdbool.h can always be used [1] and mybool.h can be ditched. 

[1]: https://en.wikibooks.org/wiki/C_Programming/C_Reference/stdbool.h

masatake:
You are right. We can use stdbool.h.
I will merge this PR for consistency. Can I ask you to make a PR for 
using stdbool.h then?

t-b:
@masatake I'll come up with a PR.

PR #1144, opened 2016-09-24
Prefer stdbool.h over mybool.h

t-b:
Since 5fa82fbf (make: Force use of GNU 99 C standard, 2015-08-10) we
require a C99 compiler on automake/autoconf platforms.

And since 4da1666e (build: Arrange cpp '-Dbool=int', '-Dtrue=1', and
'-Dfalse=0' flags, 2015-10-02) we either have use stdbool.h or an
equivalent workaround.

Therefore only platforms without a C99 compiler and not using
automake/autoconf still require an equivalent for mybool.h.
Add this equivalent in e_msoft.h for Visual Studio.

In all other places use the default type bool and true or false.


codebrainz:
FWIW, Visual Studio 2013 and newer [has stdbool.h](https://blogs.msdn.microsoft.com/vcblog/2013/07/19/c99-library-support-in-visual-studio-2013/)

dsl/es-lang-c-stdc99.c
```diff
@@ -190,7 +190,7 @@ static EsObjectClass classes[] = {
     .print   = es_boolean_print,
     .atom_p  = 1,
     .obarray = (void*)1,
-    .name    = "boolean",
```
masatake: This is nothing to do with stdboo.h issue.

dsl/qualifier.c
```diff
@@ -94,11 +94,11 @@ struct sCode {
 	{ "<=",      builtin_le,     NULL, CHECK_ARITY, 2 },
 	{ ">=",      builtin_ge,     NULL, CHECK_ARITY, 2 },
 	{ "prefix?", builtin_prefix, NULL, CHECK_ARITY, 2,
-	  .helpstr = "(prefix? TARGET<string> PREFIX<string>) -> <boolean>" },
+	  .helpstr = "(prefix? TARGET<string> PREFIX<string>) -> <bool>" },
```
masatake: This is nothing to do with stdbool.h issue.

dsl/qualifier.c
```diff
@@ -94,11 +94,11 @@ struct sCode {
...
 	{ "suffix?", builtin_suffix, NULL, CHECK_ARITY, 2,
-	  .helpstr = "(suffix? TARGET<string> SUFFIX<string>) -> <boolean>" },
+	  .helpstr = "(suffix? TARGET<string> SUFFIX<string>) -> <bool>" },
```
masatake: This is nothing to do with stdbool.h issue.

dsl/qualifier.c
```diff
@@ -94,11 +94,11 @@ struct sCode {
...
 	{ "substr?", builtin_substr, NULL, CHECK_ARITY, 2,
-	  .helpstr = "(substr? TARGET<string> SUBSTR<string>) -> <boolean>" },
+	  .helpstr = "(substr? TARGET<string> SUBSTR<string>) -> <bool>" },
```
masatake: This is nothing to do with stdbool.h issue.

dsl/qualifier.c
```diff
@@ -109,7 +109,7 @@ struct sCode {
 	  .helpstr = "input file name" },
 	{ "$access",         value_access,         NULL, MEMORABLE, 0UL },
 	{ "$file",           value_file,           NULL, MEMORABLE, 0UL,
-	  .helpstr = "file scope<boolean>" },
+	  .helpstr = "file scope<bool>" },
```
masatake: This is nothing to do with stdbool.h issue.

masatake:
Translated source files still include "TRUE" and "FALSE". 
Could you chnage the translator, `misc/optlib2c`, too?

Thank you.

t-b:
Thanks for the quick review.

I've reverted the changes in `dsl/*`, adapted misc/optlib2c and removed a few more references to mybool from the mingw/msvc build scripts.

t-b:
@codebrainz  Thanks, I've made the check in e_msoft.h honour that.

codebrainz:
I didn't realize [-std=gnu99 was unconditionally forced on](https://github.com/universal-ctags/ctags/commit/42a56708f6972669e354a50ec9684ae274bef6fb#diff-67e997bcfdac55191033d57a16d1408aR455) when I submitted https://github.com/universal-ctags/ctags/commit/25808ebec037938bb9eb3ecca43d61db554e7e80. We should probably take that out unless GCC extension to C and the standard library are desired (if so we should add a check to make sure the flag works/compiler is GCC instead of just forcing it on).

masatake:
@codebrainz, do you have any objection for merging this PR?

t-b:
@codebrainz Are  you talking about C99 vs. gnu99 or do you want  a codebase which is ANSI C89?

masatake:
@codebrainz, thank you.

masatake:
@t-b, thank you.

PR #1148, opened 2016-09-27
Main: Fix handling of -I TOKEN=REPLACEMENT

pragmaware:
Make -I TOKEN=REPLACEMENT actually work.


Units/parser-cxx.r/ignoring-macro-with-parameters.b/args.ctags
```diff
@@ -1,2 +1,2 @@
--I _GLIBCXX_VISIBILITY
+-I _GLIBCXX_VISIBILITY -I _MY_INT=int
```
masatake: As far as I know options should be listed line by line like:
```
-I _GLIBCXX_VISIBILITY
-I _MY_INT=int
```

Comare args2.ctags and args.ctags in following session:
```
[jet@localhost ctags]$ cat args.ctags 
-I X -I Y
[jet@localhost ctags]$ ./ctags --verbose --options=./args.ctags ./main/main.c 
...
Considering option file ./args.ctags: reading...
  Option: -I X -I Y
    ignore token: X
    ignore token: -I
    ignore token: Y
...
[jet@localhost ctags]$ cat args2.ctags 
-I X
-I Y
[jet@localhost ctags]$ ./ctags --verbose --options=./args2.ctags ./main/main.c 
...
Considering option file ./args2.ctags: reading...
  Option: -I X
    ignore token: X
  Option: -I Y
    ignore token: Y
...
```

I wonder why the test is passed....

Units/parser-cxx.r/ignoring-macro-with-parameters.b/args.ctags
```diff
@@ -1,2 +1,3 @@
 -I _GLIBCXX_VISIBILITY
```
b4n: shouldn't this be `-I _GLIBCXX_VISIBILITY+`? (skip also the parentheses)

masatake:
@b4n, it is my mistake.
I would like to add comprehensive cases for testing `-I` option.

Issue #1149, opened 2016-09-28
[C++] some member are not parsed


sim590:
Let the following code snippet
```cpp
struct Ctags {
    int x {};
};
```
Executing ctags, you get
```
Ctags	ctags.cpp	/^struct Ctags {$/;"	s	file:
```
While you should get the same as with the following
```cpp
struct Ctags {
    int x;
};
```
which would be
```
Ctags	ctags.cpp	/^struct Ctags {$/;"	s	file:
x	ctags.cpp	/^    int x;$/;"	m	struct:Ctags	typeref:typename:int	file:
```

Ctags version:
```
Universal Ctags 0.0.0(aef7003), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Sep 27 2016, 18:33:39
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +option-directory, +coproc, +xpath, +json
```

masatake:
Reproduced.

I tried following patch:
```
diff --git a/parsers/cxx/cxx_parser_function.c b/parsers/cxx/cxx_parser_function.c
index 31355b4..9a715fc 100644
--- a/parsers/cxx/cxx_parser_function.c
+++ b/parsers/cxx/cxx_parser_function.c
@@ -478,8 +478,15 @@ bool cxxParserTokenChainLooksLikeConstructorParameterSet(
 	// ending parenthesis.
 
 	if(pChain->iCount < 3)
-		return false; // type var() is NOT valid C++
-
+	  {
+		  if (pChain->iCount == 2 &&
+		      strcmp (vStringValue (cxxTokenChainLast (pChain)->pszWord), "}") == 0 &&
+		      strcmp (vStringValue (cxxTokenChainFirst (pChain)->pszWord), "{") == 0)
+			  return true; // type var {}
+		  else
+			  return false; // type var() is NOT valid C++
+	  }
+	
 	return cxxParserTokenChainLooksLikeFunctionCallParameterSet(pChain);
 }
 
```

And  I got:

```
[jet@localhost ctags]$ cat /tmp/foo.cpp
struct C {
  int x {};
};

[jet@localhost ctags]$ ~/var/ctags/ctags  -o - /tmp/foo.cpp
C	/tmp/foo.cpp	/^struct C {$/;"	s	file:
[...]

PR #1150, opened 2016-09-28
CXX: Improve variable parsing

pragmaware:
Handle C++11
    type var {};
    type var[] { ... };
bracket initializers.

Fixes issue #1149 .

(PS. Commit messages should contain 'type var' instead of 'var type'... but I'm too lazy to redo them :D)

### DIFF
--- a/Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/variable-declarations.cpp.d/expected.tags
@@ -19,14 +19,25 @@ l18	input.cpp	/^	struct Struct1 * l16, l17, l18[10];$/;"	l	function:main	typeref
 l19	input.cpp	/^	Struct1 l19 = {};$/;"	l	function:main	typeref:typename:Struct1	file:
 l20	input.cpp	/^	std::string ** l20[SIZE];$/;"	l	function:main	typeref:typename:std::string ** []	file:
 l21	input.cpp	/^	std::string l21[1 << 2];$/;"	l	function:main	typeref:typename:std::string[]	file:
-l22	input.cpp	/^	std::string * const l22[SIZE][SIZE];$/;"	l	function:main	typeref:typename:std::string * const[][]	file:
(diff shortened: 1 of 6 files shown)


masatake:
LGTM. But could you fix the commt log as you wrote.
`git rebase -i master` will help you.

b4n:
Agreed I'd fix the commit messages, although I might not be so problematic.
But `git rebase -i` will make it a breeze :)

masatake:
Thank you but you should fix the log. 

> (PS. Commit messages should contain 'type var' instead of 'var type'... but I'm too lazy to redo them :D)

I don't understand this. I understand writing good commit log is not easy. However updating the log itself is not difficult. If you have a trouble about git operation, feel free to ask us here.

I will show the steps. Herer I used vim.

0typo

1git-rebase-i

2git-rebase-i

3git-rebase-i

4git-rebase-i

5editor

6editor

7fin

8ok

masatake:
> Agreed I'd fix the commit messages, although I might not be so problematic.

I don't think so. This is a bit problematic. I learn both how C++ parser is written and C++ itself from @pragmaware's code. I don't know C++ well. So from the log I can misunderstand C++ has syntax like
`var type[] {}`. I think I am the only the person.

pragmaware:
I must admit that I'm not such a git master yet and feel afraid of messing up when doing such non-standard tasks. I'll try to learn though.

masatake:
At the first time git looks difficult for me, too. However, the members at this project help me. I'll do the same for you. `git rebsae -i` is really worth to learn. Other my recommendations are `git rebase [--abort|--continue]`, `git branch -m`, `git cherry-pick`, `git push --force`. These are the command I used to handle conflicts.

There is no mailing list. So there was no good place to these question to other developers.
Temporary I  added "git question" label.

b4n:
A tip when starting to learn rebasing: do that on a temporary branch (or backup the original one, same deal).
So, if you really mess things up, you can always just throw your temporary branch away and start again with the original.  It can also help double-checking the modification you did by diffing the rebased branch against the original one and see if it makes sense.

masatake:
Oh, I forget two more. `git stash` and `git ref-log`.

PR #1152, opened 2016-10-01
Some more uses of pool and other optimizations

techee:
I went through the more common languages, profiled them and checked if pool helps there - it seems to help for php and javascript. The numbers in the commit messages are after https://github.com/universal-ctags/ctags/pull/1091 is used.

On the way I noticed some more things that can be done - avoid common allocation of string in Fortran parser, apparently forgotten (or lost during rebase) use of isIdentifierChar() macro for the python parser.

Finally, the go patch doesn't bring any improvement - it just makes copyToken() work like in other parsers and makes it easier to adopt object pool if ever needed in the future.

### DIFF
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -582,10 +582,11 @@ static void makeLabelTag (vString *const label)
 
 static lineType getLineType (void)
 {
-	vString *label = vStringNew ();
+	static vString *label = NULL;
 	int column = 0;
 	lineType type = LTYPE_UNDETERMINED;
 
+	label = vStringNewOrClear (label);
 	do  /* read in first 6 "margin" characters */
 	{
 		int c = getcFromInputFile ();
@@ -647,7 +648,6 @@ static lineType getLineType (void)
 
 	if (vStringLength (label) > 0)
 		makeLabelTag (label);
-	vStringDelete (label);
 	return type;
 }
 
--- a/parsers/go.c
+++ b/parsers/go.c
@@ -136,15 +136,13 @@ static tokenInfo *newToken (void)
 	return token;
 }
 
-static tokenInfo *copyToken (tokenInfo *other)
+static void copyToken (tokenInfo *const dest, const tokenInfo *const other)
 {
-	tokenInfo *const token = xMalloc (1, tokenInfo);
-	token->type = other->type;
-	token->keyword = other->keyword;
-	token->string = vStringNewCopy (other->string);
-	token->lineNumber = other->lineNumber;
-	token->filePosition = other->filePosition;
-	return token;
+	dest->type = other->type;
+	dest->keyword = other->keyword;
+	vStringCopy(dest->string, other->string);
+	dest->lineNumber = other->lineNumber;
+	dest->filePosition = other->filePosition;
 }
 
 static void deleteToken (tokenInfo * const token)
@@ -570,7 +568,9 @@ static void parseFunctionOrMethod (tokenInfo *const token)
 
 	if (isType (token, TOKEN_IDENTIFIER))
 	{
-		tokenInfo *functionToken = copyToken (token);
+		tokenInfo *functionToken = newToken ();
+
+		copyToken (functionToken, token);
 
 		// Start recording signature
 		signature = vStringNew ();
@@ -623,7 +623,8 @@ static void parseStructMembers (tokenInfo *const token, tokenInfo *const parent_
 				if (first)
 				{
 					// could be anonymous field like in 'struct {int}' - we don't know yet
-					memberCandidate = copyToken (token);
+					memberCandidate = newToken ();
+					copyToken (memberCandidate, token);
 					first = false;
 				}
 				else
@@ -698,7 +699,8 @@ static void parseConstTypeVar (tokenInfo *const token, goKind kind)
 			{
 				if (kind == GOTAG_TYPE)
 				{
-					typeToken = copyToken (token);
+					typeToken = newToken ();
+					copyToken (typeToken, token);
 					readToken (token);
 					if (isKeyword (token, KEYWORD_struct))
 						makeTag (typeToken, GOTAG_STRUCT, NULL, GOTAG_UNDEFINED, NULL);
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -32,6 +32,7 @@
 #include "read.h"
 #include "routines.h"
 #include "vstring.h"
+#include "objpool.h"
 
 /*
  *	 MACROS
@@ -41,6 +42,8 @@
 #define isIdentChar(c) \
 	(isalpha (c) || isdigit (c) || (c) == '$' || \
 		(c) == '@' || (c) == '_' || (c) == '#')
+#define newToken() (objPoolGet (TokenPool))
+#define deleteToken(t) (objPoolPut (TokenPool, (t)))
 
 /*
  *	 DATA DECLARATIONS
@@ -122,6 +125,8 @@ static tokenInfo *NextToken;
 
 static langType Lang_js;
 
+static objPool *TokenPool = NULL;
+
 typedef enum {
 	JSTAG_FUNCTION,
 	JSTAG_CLASS,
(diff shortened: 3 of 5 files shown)


parsers/php.c
```diff
@@ -409,15 +412,20 @@ static tokenInfo *newToken (void)
...
+	return token;
+}
+
+static void clearToken (tokenInfo *const token)
+{
 	token->type			= TOKEN_UNDEFINED;
 	token->keyword		= KEYWORD_NONE;
-	token->string		= vStringNew ();
-	token->scope		= vStringNew ();
-	token->lineNumber   = getInputLineNumber ();
```
b4n: line number seems to have got lost in translation

…although, if it didn't break the test suite, maybe it's not so useful?

parsers/fortran.c
```diff
@@ -582,10 +582,11 @@ static void makeLabelTag (vString *const label)
 
 static lineType getLineType (void)
 {
-	vString *label = vStringNew ();
+	static vString *label = NULL;
```
b4n: This leads to a small "leak" (mostly theoretical only though), if we don't want it this could be created in `initialize()` and destroyed in `finalize()`.

parsers/go.c
```diff
@@ -136,15 +136,13 @@ static tokenInfo *newToken (void)
 	return token;
 }
 
-static tokenInfo *copyToken (tokenInfo *other)
+static void copyToken (tokenInfo *const dest, const tokenInfo *const other)
 {
-	tokenInfo *const token = xMalloc (1, tokenInfo);
```
b4n: Why wouldn't it be just as easy to use objpool here inside `copyToken()`?
I don't mean the change is bad, but I don't really see why it's that useful.

techee: OK, yeah, you're right, that would be possible. On the other hand I searched copyToken() and none of the other parsers creates a token there so maybe better to make it consistent with the rest.

parsers/php.c
techee: Oops, that was unintentional. Will fix.

parsers/go.c
b4n: Yes if it's he odd kid doing it like that it might still be interesting to change it indeed.

techee:
> About objpool, I wonder whether it'd be better to "hide" the its use from the parsing code by doing something like
>
> #define newToken() (objPoolGet(TokenPool))
> #define deleteToken(t) (objPoolPut(TokenPool, (t)))

That's actually a really good idea - I think it's more readable indeed and the code looks more like other parsers. Will do this.

parsers/fortran.c
techee: Yeah, but there are quite a few of them - it's basically every "static vString" and there are quite many of them in the code already. On the other hand this isn't a super-important commit, the Fortran parser is already quite fast even without this patch.

techee:
I've repushed the patches with fixes.

parsers/fortran.c
masatake: @b4n, are you talking about "reachable"? 
Consider a memory object assigned to a static pointer variable. As far as NULL (or something other memoryu object) is assigned to the variable, I don't say it is "leaked". Instead "it is still reachable at the end of process.". These two are different things as valgrind reported.

codebrainz: It's a "leak" in the sense that memory is allocated by the application and never released by it. "Reachable" is just a type of leak that still has a pointer pointing to the leaked memory at the end of the process.

masatake: Basically, "Reachable" type of leak is acceptable. Of course only if its size is not so large.

codebrainz: Maybe in CTags specifically, but not in general. Sometimes, like for immutable global data (ex. lookup tables) leaving reachable leaks can speed up program shutdown, but in this case it's just for a small optimization that could probably be achieved better in other ways (ex. stack allocation) and would be quite problematic if CTags and/or the parsers were ever (made to be) used in any kind of multi-threaded scenario.

techee: I just followed the general pattern used in ctags. If you search for "static vString", you'll find lots of them in the code. Yes, it's not good if you want to make it concurrent but there's really a LOT of work if you want to make the code thread-safe (and personally I'd rather go the multi-process way than the multi-threaded way).

How do you want to achieve this using stack allocation?

As I said, I can drop this patch, I don't care much but it doesn't solve anything - code like this is everywhere.

b4n: @masatake yes I meant what Valgrind reports as "still reachable".  And no, depending on the context I don't really see it as a leak, hence the quotes around the word.

So yes, we can totally decide it's fine, but it can be a slight imperfection if/when ctags parsers become a library, because it can then long-live, and a caller could want to limit resource usage and deinit parsers as soon as it can reasonably think it doesn't need them anymore soon (maybe when closing the last file of that filetype in an editor).

Anyway, if we have this pattern in many places already, it doesn't matter in the context of this PR, so if it should be a discussion at all, it probably should happen on its own.

parsers/jscript.c
```diff
@@ -41,6 +42,8 @@
 #define isIdentChar(c) \
 	(isalpha (c) || isdigit (c) || (c) == '$' || \
 		(c) == '@' || (c) == '_' || (c) == '#')
+#define newToken() (objPoolGet(TokenPool))
+#define deleteToken(t) (objPoolPut(TokenPool, (t)))
```
b4n: style: space before the argument list (in the expansion side, obviously).
Feel free to ignore :)

parsers/jscript.c
```diff
@@ -1941,6 +1952,17 @@ static void initialize (const langType language)
 {
 	Assert (ARRAY_SIZE (JsKinds) == JSTAG_COUNT);
 	Lang_js = language;
+
+	TokenPool = objPoolNew (16, (objPoolCreateFunc)newPoolToken,
+		(objPoolDeleteFunc)deletePoolToken, (objPoolClearFunc)clearPoolToken);
```
b4n: Now the functions are specific to the pool, they could have the appropriate signature, and the conversion be done internally.  Doesn't matter too much, but could allow the compiler to check those in case they ever change.

parsers/php.c
```diff
@@ -19,8 +19,11 @@
 #include "entry.h"
 #include "routines.h"
 #include "debug.h"
+#include "objpool.h"
 
 #define isIdentChar(c) (isalnum (c) || (c) == '_' || (c) >= 0x80)
+#define newToken() (objPoolGet(TokenPool))
+#define deleteToken(t) (objPoolPut(TokenPool, (t)))
```
b4n: style (ditto)

parsers/php.c
```diff
@@ -1650,14 +1661,35 @@ static void findZephirTags (void)
 	findTags (true);
 }
 
+static void initializePool (void)
+{
+	if (TokenPool == NULL)
+		TokenPool = objPoolNew (16, (objPoolCreateFunc)newPoolToken,
+			(objPoolDeleteFunc)deletePoolToken, (objPoolClearFunc)clearPoolToken);
```
b4n: ditto

parsers/python.c
```diff
@@ -25,6 +25,10 @@
 #include "xtag.h"
 #include "objpool.h"
 
+#define isIdentifierChar(c) \
+	(isalnum (c) || (c) == '_' || (c) >= 0x80)
+#define newToken() (objPoolGet(TokenPool))
+#define deleteToken(t) (objPoolPut(TokenPool, (t)))
```
b4n: ditto

parsers/python.c
```diff
@@ -1278,16 +1277,16 @@ static void findPythonTags (void)
...
 }
 
 static void initialize (const langType language)
 {
 	Lang_python = language;
 
-	TokenPool = objPoolNew (16, (objPoolCreateFunc)newToken, (objPoolDeleteFunc)deleteToken,
-		(objPoolClearFunc)clearToken);
+	TokenPool = objPoolNew (16, (objPoolCreateFunc)newPoolToken,
+		(objPoolDeleteFunc)deletePoolToken, (objPoolClearFunc)clearPoolToken);
```
b4n: ditto

parsers/jscript.c
techee: I'm not sure I understand.

techee: Oh, you mean to avoid the cast here and do it inside the functions, right?

b4n: I mean that e.g. `newPoolToken()` could be `static void *newPoolToken(void)` (`objPoolCreateFunc`), and then avoid the cast here.  So, if the function signatures expected by `objPoolNew()` ever change and someone forgets to update that, a compiler could spot it.

b4n: yes :)  (cross post!)

parsers/fortran.c
codebrainz: > How do you want to achieve this using stack allocation?

It was just an example, but at a glance just using an automatic character array variable appears like it would work, since the code seems to limit the string to around 6-8 characters or something. Another example would be just recycling small allocations like (AFAIK) glibc's allocator already does.

Anyway, I only commented because a) it is an actual memory leak (albeit intentional), and b) it's a code-smell (IMO). It's certainly not a big deal, especially if it's a widely used pattern already.

masatake:
...We can do more:

Some part of what you have done in the parsers  can be moved to main part.

```
diff --git a/main/parse.h b/main/parse.h
index f61d62c..5c743fd 100644
--- a/main/parse.h
+++ b/main/parse.h
@@ -18,6 +18,7 @@
 #include "dependency.h"
 #include "field.h"
 #include "kind.h"
+#include "objpool.h"
 #include "parsers.h"  /* contains list of parsers */
 #include "strlist.h"
 
@@ -166,6 +167,11 @@ struct sParserDefinition {
 	stringList* currentExtensions; /* current list of extensions */
 	stringList* currentAliases;    /* current list of aliases */
 	unsigned int anonumousIdentiferId; /* managed by anon* functions */
+
+	objPool **tokenPool;
+	objPoolCreateFunc tokenPoolCreateFunc;
+	objPoolDeleteFunc tokenPoolDeleteFunc;
+	objPoolClearFunc  tokenPoolClearFunc;
 };
 
 typedef parserDefinition* (parserDefinitionFunc) (void);
```

When tokenPull`"`Funcs are given, the main part prepare objPool to the place where tokenPool points.
So each parser doesn't have to care about the construction  and destruction of its objPool.

I'm not sure whether `**`` or `*` is better of tokenPool field. It is related to #754.

techee:
@masatake That would be possible but I'm not sure it the use of pool for tokens is so widespread that it would require a special entry in parserDefinition. There are just 4 parsers using it and from what I saw, not many other parsers will need it (at least for now). It's obviously usable only for token-based parsers and even these don't always need the pool - e.g. when they create new tokens rarely or none at all.

(Also for the parsers I chose the size of pool 16 which was enough in these cases but I can imagine cases - e.g. when some function creating tokens is called recursively - where the size of the pool should be larger which would require additional entry to configure its size.)

masatake:
@techee, O.k. Let's merge this. I myself have never written real token based parser. After I get more experiences about writing parser, I would like to think this topic(adding fileds for objPool in parser definition) again.

PR #1155, opened 2016-10-03
docs, style: add dir-locals file for setting the style to linux in GNU emacs

masatake:
Though .editorconfig was introduced, it is not enough for GNU emacs users to make
the coding style consistent with existing code of ctags.

The default style of cc-mode of GNU emacs is "gnu". The "gnu" style adds
extra indentation in block opening:

________if (...)
__________{

The style is not consistent with the existing code.
This commit introduces .dir-locals.el file which forces the style
to "linux". With "linux" style the block opening is handled as follows:

________if (...)
________{

The difficulties is that the basic indent size of "linux" is 8.
It is different from .editorconfig.

So in the updated document, I wrote the way to apply .editorconfig
configuration after installing "linux" style.

### DIFF
--- a/docs/contributions.rst
+++ b/docs/contributions.rst
@@ -123,7 +123,46 @@ functional changes with style fixes makes reviewing harder.
 If possible, don't use file static variables. Find an alternative way
 that uses parameters.
 
-For indentation I'm using the linux style in cc-mode of GNU Emacs.
+
+.. NOT REVIEWED YET
+
+Notes for GNU emacs users
+.........................................................................
+
+If you use GNU emacs, utilize the `.editorconfig` configuration based
+on non-GNU C style. Here non-GNU C style means
+"align a keyword for control flow and `{` of the block start".
+
+GNU style:
+.. code-block:: C
+
+	if (...)
+	    {
+		...
+
+non-GNU style:
+.. code-block:: C
+
+	if (...)
+	{
+		...
+
+For combining the style and `.editorconfig` configuration, put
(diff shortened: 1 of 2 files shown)


masatake:
@dkearns, can I ask you to review the updated document?
This is NOT hurry.

docs/contributions.rst
```diff
@@ -123,7 +123,43 @@ functional changes with style fixes makes reviewing harder.
...
+	(add-hook 'hack-local-variables-hook
+		(lambda () (editorconfig-apply)))
+
+`.dir-locals.el` in ctags source tree applies "linux" style of `cc-mode`.
+Above code snippet applies the `.editorconfig` configuration AFTER
+installing the "linux" style to the current buffer.
+
+I like GNU style, but for keeping consistency in existing code of
+Exuberant-ctags, the origin of Universal-ctags, I introduced the style
+and configuration to my .emacs.  Please, do the same.
```
b4n: While warm and all, I don't think we need personal opinions like that when they don't fit what we expect from contributors ^^

I also do find GNU style fairly nice myself (although a little funny, but good funny), but it's not relevant when contributing to a project that already have a coding style.

Just a thought though, you might want to have docs not too formal.

masatake: Thank you. Updated.

masatake:
For the region newly written, I put marker "NOT REVIEWED YET". @dkearns, I'm happy if you review the region I marked when you find time.

PR #1160, opened 2016-10-06
Meta parsers misc fix suggested in #1137

### LINKED ISSUE #1158: reflect the comments about yaml meta parsrr
I got comments from @b4n about my PR #1137.
There is no way to reopen PR. So I register my task here.

masatake:
Close #1158.

### DIFF
--- a/parsers/ansibleplaybook.c
+++ b/parsers/ansibleplaybook.c
@@ -185,11 +185,18 @@ static void ansiblePlaybook (yaml_token_t *token, void *data)
 
 static void *ansiblePlaybookPrepare(void)
 {
-	static struct ansiblePlaybookState state = {
-		.type_stack = NULL,
-		.play_detection_state = DSTAT_PLAY_NAME_INITIAL,
-	};
+	static struct ansiblePlaybookState state;
+
+	state.play_detection_state = DSTAT_PLAY_NAME_INITIAL;
+
 	return &state;
+}
+
+static void ansiblePlaybookInputEnd(void *data)
+{
+	struct ansiblePlaybookState *state CTAGS_ATTR_UNUSED = data;
+
+	Assert (state->type_stack == NULL);
 }
 
 static void
@@ -202,6 +209,7 @@ findAnsiblePlaybookTags (void)
 static struct yamlParserClient AnsiblePlaybookYamlClient = {
 	.prepareForNewInput = ansiblePlaybookPrepare,
 	.callback = ansiblePlaybook,
+	.inputEnd = ansiblePlaybookInputEnd,
 };
 
 static void initializeAnsiblePlaybookParser (const langType language)
--- a/parsers/ansibleplaybook.c
+++ b/parsers/ansibleplaybook.c
@@ -183,7 +183,7 @@ static void ansiblePlaybook (yaml_token_t *token, void *data)
 		popAllBlockType (data, token);
 }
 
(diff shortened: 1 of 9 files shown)


parsers/meta-m4.h
```diff
@@ -28,8 +28,8 @@ struct m4ParserClient
 	langType lang;
 	void *data;
 
-	char quoteOpen;
-	char quoteClose;
+	const char quoteOpen;
+	const char quoteClose;
```
b4n: why?  how will `changequote()` be handled then?

parsers/yaml.c
```diff
@@ -122,17 +122,17 @@ static void call_callback (void *key CTAGS_ATTR_UNUSED, void *value, void *user_
...
 
 	pushLanguage (client->lang);
 	client->callback (token, client->data);
 	popLanguage ();
 
 	if (token->type == YAML_STREAM_END_TOKEN)
 	{
-		if (client->teardownForTheLastInput)
-			client->teardownForTheLastInput (client->data);
+		if (client-> inputEnd)
```
b4n: extra space

parsers/ansibleplaybook.c
```diff
@@ -185,10 +185,11 @@ static void ansiblePlaybook (yaml_token_t *token, void *data)
 
 static void *ansiblePlaybookPrepare(void)
 {
-	static struct ansiblePlaybookState state = {
-		.type_stack = NULL,
-		.play_detection_state = DSTAT_PLAY_NAME_INITIAL,
-	};
+	static struct ansiblePlaybookState state;
+
+	Assert (state.type_stack = NULL);
```
b4n: is it really a good idea to perform such a check in the init phase?  If it's useful, shouldn't it rather be at the end of the previous run, as IIUC it checks whether the state was properly poped or something?

parsers/meta-m4.h
masatake: `changequote()` is handled in `m4Quote{Open|Close}` in m4.c.
The const'ed fields are for initializing `m4Quote{Open|Close}`.

masatake:
I found 861e49b is wrong. I will fix it.

parsers/meta-m4.h
b4n: Oops, okay sorry.

parsers/ansibleplaybook.c
```diff
@@ -181,13 +181,14 @@ static void ansiblePlaybook (yaml_token_t *token, void *data)
 		popBlockType (state, token);
 	else if (token->type == YAML_STREAM_END_TOKEN)
 		popAllBlockType (data, token);
+
+	Assert (state->type_stack = NULL);
```
b4n: should be `==`, but I guess you noticed

parsers/pythonloggingconfig.c
```diff
@@ -90,7 +90,7 @@ static bool pythonLoggingConfigProbeLanguage (const char *section, const char *k
 
 static struct iniconfParserClient PythonLoggingConfigIniconfClient = {
 	.lang = LANG_IGNORE,
-	.prepareForNewInput = pythonLoggingConfigPrepareForNewInput,
+	.inputStart = pythonLoggingConfiginputStart,
```
b4n: alignment

b4n:
otherwise LGTM

Issue #841, opened 2016-03-13
Provide the way to define parser own options


masatake:
What kind of option format do you want?

My idea:

```
--opt-<LANG>=<OPT>:<VALUE>
```

pragmaware:
Mh... maybe for simplicity the parser specific options could be stored in the global options structure.
We could just define a policy for naming as --<LANG>-<OPTION>:

   --c++-preserve-templates
   --python-something=value
   --c-do-something=value
  etc...

so basically no additional per-parser code would be needed ?

masatake:
The language common prefix is needed two reasons.

* Consider language named "help" as an example. 
   Without the prefix, we have to introduce `--help-something=value`.
    Users and we will be confused. 
    The name of language is obviously wrong but we cannot choose the name of language.

   This is the reason why I introduced `--kind-<LANG>` option.

* I have nerver tried implementing but I guess the common prefix will make implementing bash/zsh completion rules for ctags easier.

How about `-O <LANG>-something=value` or `-O <LANG>:something=value` ?

pragmaware:
Hm.. we already have --<LANG>-kinds.

One point I would like to throw in is that the ctags options are already quite complex and hard to use. It took me quite a while and a lot of man page readings to understand how to use kinds, fields, extras etc.
We should avoid to increase the complexity further more...

masatake:
> Hm.. we already have --<LANG>-kinds
I would like to remove `--<LANG>-kinds`. However, it breaks compatibility.

> One point I would like to throw in is that the ctags options are already quite complex and hard to use. It took me quite a while and a lot of man page readings to understand how to use kinds, fields, extras etc.
We should avoid to increase the complexity further more...

Comparing following two:
```
--c++-preserve-templates
```
```
-O C++:preserve-templates
```
The former will be listed in `--help`.
The later will be not. Just `-O` is listed but `C++:preserve-templates` is not.
`preserve-templates` will be listed in ``--list-parser-option=C++``.

I think the later is a bit simpler than the former in this aspect.

pragmaware:
Hm... don't know.

(in favor of --c++-preserve-templates)
Using more commands to figure out options looks more complex to me.
--c++-preserve-templates=value is quite standard (and you will find it directly in the --help output).
-O C++:preserve-templates=value is not really standard and one will first have to discover it by stumbling upon --list-parser-options and then memorize how to use it by repeatedly looking at --help

(in favor of -O c++:preserve-templates)
[...]

PR #1161, opened 2016-10-08
CXX: Better macro handling

pragmaware:
- Replacement of macros defined with -I is now fully parsed, so it can contain whatever C/C++ code the user likes
- If a replacement of a + macro contains the $0,$1...$9 sequences, then they are replaced with the 0-th, 1-st...9-th parameter of the skipped parenthesis. This can be used to define almost real C/C++ macros.
- If a replacement of a + macro contains the $\* sequence then it's replaced by the whole parameter set extracted from the skipped parenthesis. This allows for handling of simple variadic macros.

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -57,7 +57,7 @@
 #define EXTENSION_SEPARATOR '.'
 #define PATTERN_START '('
 #define PATTERN_STOP  ')'
-#define IGNORE_SEPARATORS   ", \t\n"
+#define IGNORE_SEPARATORS   ",\n"
 
 #ifndef DEFAULT_FILE_FORMAT
 # define DEFAULT_FILE_FORMAT  2
@@ -2194,11 +2194,11 @@ static void saveIgnoreToken(const char * ignoreToken)
 	if(tokenEnd <= tokenBegin)
 		return;
 
-	
 	ignoredTokenInfo * info = (ignoredTokenInfo *)eMalloc(sizeof(ignoredTokenInfo));
 	
 	info->ignoreFollowingParenthesis = ignoreFollowingParenthesis;
 	info->replacement = replacement ? eStrdup(replacement) : NULL;
+	info->replacementLength = replacement ? strlen(replacement) : 0;
 
 	hashTablePutItem(Option.ignore,eStrndup(tokenBegin,tokenEnd - tokenBegin),info);
 
--- a/Units/parser-cxx.r/complex-macros.d/args.ctags
+++ b/Units/parser-cxx.r/complex-macros.d/args.ctags
@@ -1,0 +1,7 @@
+--c++-kinds=*
+--fields-c++=+{properties}
+-I DECLARE_FUNCTION+=$0 $1()
+-I ARGS+=$*
+-I DECLARE_VARS+=int $0a; int $0b;
+-I IMPLEMENT_FUNCTIONS+=void $0a(){}; void $0b(){};
+-I DEPRECATED+=$* __attribute__((deprecated))
--- a/Units/parser-cxx.r/complex-macros.d/expected.tags
+++ b/Units/parser-cxx.r/complex-macros.d/expected.tags
@@ -1,0 +1,14 @@
+DECLARE_FUNCTION	input.cpp	/^#define DECLARE_FUNCTION(/;"	d	file:
+DECLARE_FUNCTION_2	input.cpp	/^#define DECLARE_FUNCTION_2(/;"	d	file:
+DECLARE_VARS	input.cpp	/^#define DECLARE_VARS(/;"	d	file:
+DEPRECATED	input.cpp	/^#define DEPRECATED(/;"	d	file:
+IMPLEMENT_FUNCTIONS	input.cpp	/^#define IMPLEMENT_FUNCTIONS(/;"	d	file:
+f1a	input.cpp	/^IMPLEMENT_FUNCTIONS(f1);$/;"	f	typeref:typename:void
+f1b	input.cpp	/^IMPLEMENT_FUNCTIONS(f1);$/;"	f	typeref:typename:void
+la	input.cpp	/^	DECLARE_VARS(l);$/;"	l	function:main	typeref:typename:int	file:
+lb	input.cpp	/^	DECLARE_VARS(l);$/;"	l	function:main	typeref:typename:int	file:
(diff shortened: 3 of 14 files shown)


main/options.c
```diff
@@ -57,7 +57,7 @@
 #define EXTENSION_SEPARATOR '.'
 #define PATTERN_START '('
 #define PATTERN_STOP  ')'
-#define IGNORE_SEPARATORS   ", \t\n"
```
masatake: Why this one is needed?

main/options.c
```diff
@@ -2194,11 +2194,11 @@ static void saveIgnoreToken(const char * ignoreToken)
 	if(tokenEnd <= tokenBegin)
 		return;
 
-	
 	ignoredTokenInfo * info = (ignoredTokenInfo *)eMalloc(sizeof(ignoredTokenInfo));
 	
 	info->ignoreFollowingParenthesis = ignoreFollowingParenthesis;
 	info->replacement = replacement ? eStrdup(replacement) : NULL;
+	info->replacementLength = replacement ? strlen(replacement) : 0;
```
masatake: Another approach is holding `replacement` in a vString.
vString maintains the length of string.

masatake:
First of all, this is interesting change.

Your commits change the command line interface. So we must go carefully. As you main know I am straggle for releasing.  One of the reason is that I added too many random options.

First of all you must update document. If you are good at English, you can update man/ctags.1.rst.in directly. If not, at least you must update docs/news.rst. In that case we must ask @dkearns reviewing.

About 22c8815 I wonder why /bin/bash is referred in the commit log.

You explain the changes in this pull request head comment only. Don't do so.
Following is the basic rule I would like you to follow:

If a change has well visible to users and/or developers in the other paresrs, write docs/\*.rst first.
Then write commit log well. Finally write the head comment of pull request.

If not, write commit log well.Finally write the head comment of pull request.

If you are busy and the change is so obvious, commit log is enough.

We depend so many things on github. github is really great place to work with, but we should not forget about happens in e-ctags. Primary work log should be in git log, not in PR comments.
If a PR comment is important, it should be copied to docs/\*.rst.

masatake:
I wonder some code of `cxxParserParseNextTokenApplyReplacement` can be moved to lcpp.c.

If you can enhance lcpp.c instead of cxx parser, the other parsers using lccp.c get benefits.

Ideally the value passed from `-I` should be managed in lcpp.c.
Further more I want to move lcpp.c to parsers/cpreproc.c.

masatake:
I'm sorry I made comments only. Basicaly I will do the things by my side; the main part should provide good API to parser developers. However, I don't have much time now.

pragmaware:
Ah!... /bin/bash is probably replaced $0. I used git commit -m "....$0...."

main/options.c
masatake: man page said about -I
```
Otherwise, identifier-list is a list of identifiers (or identifier pairs) to be spe‐
            cially handled, each delimited by a either a comma or by white space (in which case the list should  be
            quoted  to keep the entire list as one command line argument).
```

whitespace is considered as a char for separation. Your new code may break the compatiblity.
Let's avoid it if possible.

My idea is that introducing `--opt-<LANG>=<PARAM>:<VALUE>` facility.
In this case the option will be `--opt-Cpreproc=define:AnythingYouWant`.

masatake:
Could you see #1163?
Before doing interesting things, I think facilitating code is needed in lcpp.c.

masatake:
I cannot sleep when thinking about interesting extension for ctags. Typical symptoms of bad maintainer.

@pragmaware, how do you think about suspending this PR ?
(I assume you agree with me about introducing --opt-<LANG> option.)

Before introducing extension proposed in this PR, I would like to think more about -I option.
Though the option is about lcpp.c, partially it is handled in options.c.
The evaluation is done in cxx side.
I would like to move these things into lcpp.c side as much as possible.

To keep the CLI compatiblity ctags accpets `-I` option. However, internally ctags translate -I option to `--opt-Cpreprocessor=ignore:FOO`. options.c does this translation. The rest of work, registering the FOO to a htable in lcpp.c is done by "ignore" call back function defined in lcpp.c. Only lcpp.c knows the htable.
The interface for cxx declared in lcpp.h will not change.
Of course `--list-opt=<LANG>` must be implemented.

I will implement above things. 
Then it it your turn.
You can add `--opt-Cpreprocessor=define:FOO` or anything you want. You don't touch `-I` option. It is maintained only for compatibility. Really interesting things can be done in `--opt-<LANG>` area.
Of course you can add `--opt-C` or `--opt-C++` if you want.

Do I make sense? 
I choose `--opt-` as prefix because I guess it is command-line-completion-engine friendly.

masatake:
I was supprised that I already wrote the same thing in #841.

pragmaware:
Hm... you suggest to implement the preprocessor as a separate parser to recycle the option handling routines right?

I also had a backward compatiliby concern and I've been thinking to use a dedicated -D instead, to somewhat emulate the gcc's -D. That would be cool: copy and paste the -D options from the gcc commandline and you're done. However this would require a slightly more complex implementation: instead of $0, $1 and so on one would need to use the real macro parameter names. With some effort it could be done.

Also another couple of interesting points:

-  -I (and -D) could be actually used as preprocessing stage for any language. 
- #define lines could actually add macros on-the-fly

By the way, I'm already using this patch in my editor to detect foreach() macros.. it works great.

masatake:
> Hm... you suggest to implement the preprocessor as a separate parser to recycle the option handling routines right?

Yes. 

Wait for awhile. The most of all things are done.

masatake:
I will work on updating these changes.

main/options.c
masatake: > Because macro replacements with parameters generally contain spaces.

How about ','? 

macor replacements with parameters can contain spaces, can't they?

masatake:
-D is already used for debugging purpose. However, it is not written in man page. So we can use for macro definition. I will use -d for specifying debugging level.

masatake:
Let's continue the discussion at #1191.

main/options.c
pragmaware: All macros can contain spaces and , characters.
Maybe we should look into a way of keeping -I backward-compatible and -D supporting full macros.
Though this can be done in steps. In the long term I would also like to parse real macros, not the "fake" ones with $ characters.

PR #1162, opened 2016-10-08
Convert vStringPut macro to inline function (Was: Fix coding issues)

t-b:
Found with make CC=/rest/inst/gcc-6.1.0/usr/local/bin/gcc CFLAGS="-Wall -pedantic".

### DIFF
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -79,18 +79,6 @@ extern vString *vStringNew (void)
 	return string;
 }
 
-#ifndef VSTRING_PUTC_MACRO
-extern void vStringPut (vString *const string, const int c)
-{
-	if (string->length + 1 == string->size)  /*  check for buffer overflow */
-		vStringResize (string, string->size * 2);
-
-	string->buffer [string->length] = c;
-	if (c != '\0')
-		string->buffer [++string->length] = '\0';
-}
-#endif
-
 extern vString *vStringNewCopy (const vString *const string)
 {
 	vString *vs = vStringNew ();
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -25,15 +25,6 @@
 /*
 *   MACROS
 */
-#ifndef DEBUG
-# define VSTRING_PUTC_MACRO 1
-#endif
-#ifdef VSTRING_PUTC_MACRO
-#define vStringPut(s,c) \
-	(void)(((s)->length + 1 == (s)->size ? vStringResize ((s), (s)->size * 2) : 0), \
-	((s)->buffer [(s)->length] = (c)), \
-	((c) == '\0' ? 0 : ((s)->buffer [++(s)->length] = '\0')))
-#endif
 
 #define vStringValue(vs)      ((vs)->buffer)
 #define vStringItem(vs,i)     ((vs)->buffer[i])
@@ -66,9 +57,6 @@ typedef struct sVString {
 extern void vStringResize (vString *const string, const size_t newSize);
 extern vString *vStringNew (void);
 extern void vStringDelete (vString *const string);
-#ifndef VSTRING_PUTC_MACRO
-extern void vStringPut (vString *const string, const int c);
-#endif
 extern void vStringStripNewline (vString *const string);
 extern void vStringStripLeading (vString *const string);
 extern void vStringChop (vString *const string);
@@ -94,5 +82,19 @@ extern char    *vStringDeleteUnwrap (vString *const string);
 
 extern void vStringCatSWithEscaping (vString* b, const char *s);
 extern void vStringCatSWithEscapingAsPattern (vString *output, const char* input);
+
+/*
+*   INLINE FUNCTIONS
+*/
+
+static __inline void vStringPut (vString *const string, const int c)
+{
+	if (string->length + 1 == string->size)  /*  check for buffer overflow */
+		vStringResize (string, string->size * 2);
+
+	string->buffer [string->length] = c;
+	if (c != '\0')
+		string->buffer [++string->length] = '\0';
+}
 
 #endif  /* CTAGS_MAIN_VSTRING_H */


main/vstring.h
```diff
@@ -30,7 +30,7 @@
 #endif
 #ifdef VSTRING_PUTC_MACRO
 #define vStringPut(s,c) \
-	(void)(((s)->length + 1 == (s)->size ? vStringResize ((s), (s)->size * 2) : 0), \
+	(void)(((s)->length + 1 == (s)->size ? vStringResize ((s), (s)->size * 2) : (void) 0 ), \
 	((s)->buffer [(s)->length] = (c)), \
 	((c) == '\0' ? 0 : ((s)->buffer [++(s)->length] = '\0')))
 #endif
```
masatake: Instead how about following code?
```
diff --git a/main/vstring.h b/main/vstring.h
index 203a5c2..2300478 100644
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -29,10 +29,14 @@
 # define VSTRING_PUTC_MACRO 1
 #endif
 #ifdef VSTRING_PUTC_MACRO
-#define vStringPut(s,c) \
-	(void)(((s)->length + 1 == (s)->size ? vStringResize ((s), (s)->size * 2) : 0), \
-	((s)->buffer [(s)->length] = (c)), \
-	((c) == '\0' ? 0 : ((s)->buffer [++(s)->length] = '\0')))
+#define vStringPut(s,c)							\
+	do {										\
+		if ((s)->length + 1 == (s)->size)		\
+			vStringResize ((s), (s)->size * 2);	\
+		(s)->buffer [(s)->length] = (c);		\
+		if ((c) != '\0')						\
+			(s)->buffer [++(s)->length] = '\0';	\
+	} while (0);
 #endif
 
 #define vStringValue(vs)      ((vs)->buffer)
```

I believe C compilers may generate the same code.
How do you think?

t-b: @masatake Yes your suggestion is equivalent.

But if I ought to rewrite the code I would write something like

```cpp
+#define vStringPut(s,c)                            \
+   do {                                        \
+       if ((s)->length + 1 >= (s)->size)        \
+           vStringResize ((s), (s)->size * 2); \
+
+       (s)->buffer [(s)->length] = (c);          \
+      (s)->length++;                                     \
+       (s)->buffer [(s)->length] = '\0';          \
+   } while (0);
```

As that is much more pleasant to read and should execute at the same speed.

codebrainz: You could also use an inline function and it would execute at the same speed and has the advantages of not evaluating arguments multiple times and producing more useful compiler errors when misused.

masatake: @codebrainz, thank you.

@t-b, your macro is better than mine. Could you convert it to an inline functionas @codebrainz wrote. Then add the change as a commit to this PR; your original change and converting the original macro to the inline function should be in separated commits.

t-b: @masatake @codebrainz 
I've removed the macro and added my changes to the function. I've not declared the function as inline, from all I know compilers nowadays always get that decision better by themselves.

main/vstring.c
```diff
@@ -81,12 +81,12 @@ extern vString *vStringNew (void)
 
 extern void vStringPut (vString *const string, const int c)
```
masatake: As far as I know `inline` function should be at `*.h file.
Could you move vStringPut to vstring.h, add inline keyword?
I'm not sure whether `static` should be added or not.

codebrainz: I usually use `static inline` and put definition in the header. There are more details on the differences [here](https://gcc.gnu.org/onlinedocs/gcc/Inline.html). Without putting in the header and no `inline` keyword like above, I think it pretty much guarantees it won't be inlined. Also `extern` as used here is superfluous as it's the default for functions like this, it's akin to putting `auto int x;` for a local variable, AFAIK.

masatake: > @masatake So you think we still need the inline keyword? As I mentioned from my experience the compiler will ignore it in any case.

Are you using compiler other than gcc?

```
$ objdump -d ./ctags | grep '\(callq.*vStringPut\)\|>:'
...
000000000044ec30 <plistFindTagsUnderKey>:
  44ed88:	e8 c3 cf fc ff       	callq  41bd50 <vStringPut>
...
```
`vStringPut` is still `call`ed. It is not inlined. gcc may do so only if a function is `static`. 
 "Linker and Loader"  says optimizing corss-compile-units can be done with linker. GNU ld may  do it if we give more option flags. https://gcc.gnu.org/wiki/LinkTimeOptimization

Don't make a PR for introducing LTO. Instead think about rewriting ctags with javascript or go:-P

masatake: @t-b, Let's use `static inline`.

masatake: @codebrainz, thank you. You know C and these areas well. This is not the first time helping me with your knowledges of the area. Could you help the development of u-ctags explicitly? I sent an invitation.

main/vstring.c
```diff
@@ -81,12 +81,12 @@ extern vString *vStringNew (void)
...
 {
-	if (string->length + 1 == string->size)  /*  check for buffer overflow */
+	if (string->length + 1 >= string->size)  /*  check for buffer overflow */
```
masatake: After thinking this is wrong approach.
As the desing of vString, the condition, `string->length + 1 == string->size`, must be enough.

main/vstring.c
```diff
@@ -81,12 +81,12 @@ extern vString *vStringNew (void)
...
 		vStringResize (string, string->size * 2);
 
 	string->buffer [string->length] = c;
-	if (c != '\0')
-		string->buffer [++string->length] = '\0';
+	string->length++;
```
masatake: Repeatedly putting `\0` extends the string.
This can hogs memory.

main/vstring.c
codebrainz: > As I mentioned from my experience the compiler will ignore it in any case.

With which compiler? I think GCC always tries its best to inline functions marked with `inline`.

masatake:
> @t-b, your macro is better than mine. 

Sorry but it seems that my comment is wrong as far as seeing the result of test.

masatake:
@k-takata, or windows experts, it seems that visual c++ compiler(?) doesn't understand `inline` keyword? I need c99 compiler. What should I do?

main/vstring.c
t-b: Visual Studio seems not to like the `static inline`.

t-b:
@masatake Its not about `inline` it is about `static inline`, see example 1 in https://msdn.microsoft.com/en-us/library/aa299373%28v=vs.60%29.aspx.

masatake:
I merged your change partially for focusing the issue about "inline".

main/vstring.c
codebrainz: @t-b I think until VS 2013 it didn't support C99, so it might be better to use `__inline` which AFAIK works on most compilers (MSVC, GCC, Clang, ICC, etc). Another option is to define some macros perhaps like (untested):

```c
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
# define CTAGS_INLINE static inline
#elif defined(_MSC_VER)
# define CTAGS_INLINE static __inline
#elif defined(__GNUC__) || defined(__clang__)
# define CTAGS_INLINE static __inline__
// #elif ... other compilers/tests here ...
// # define CTAGS_INLINE ...
#else
# define CTAGS_INLINE static
#endif
```

codebrainz: @masatake ok, thanks.

main/vstring.c
codebrainz: It seems like another possibility would be to drop this `if` branch and just call `vStringResize(string, string->size + 1);` unconditionally. It would incur a function call overhead but would make the code simpler and keep the resize policy (double the size when growing) in one place in `vStringResize`.

t-b:
@masatake From my POV this is ready to merge :)

masatake:
I wonder why we have to use __inline instead of inline. As far as I know inline is part of C99.
I must inspect this more but I cannot find time now.

codebrainz:
Because MSVC sucks :)

masatake:
> Because MSVC sucks :)

In such case it is better to define `CTAGS_INLINE`, isn't it?

codebrainz:
> In such case it is better to define CTAGS_INLINE, isn't it?

I think so, though `__inline` is fairly portable (at least MSVC, GCC, Clang, probably others). In one of the comments I put such code to define `CTAGS_INLINE`, but it seems to be gone now.

codebrainz:
Found it: https://github.com/universal-ctags/ctags/pull/1162#discussion_r83127325

masatake:
> Found it: #1162 (comment)

Oh, yes It is what we need.

masatake:
@codebrainz, @t-b, I will merge this. THen I will replace `__inline` with CTAGS_INLINE.

k-takata:
@masatake Sorry, I didn't notice your comment.
VC2013 partly supports C99, but it doesn't support `inline`.
VC2015 supports `inline`.

masatake:
@k-takata, thank you.
If you have time, could you look at  #893?

PR #1163, opened 2016-10-12
Standalone cpreprocessor parser

masatake:
@pragmaware, how do you think about this?

You wan to put more code to lcpp.c.
Before doing much, I think it is better to take efforts for making lcpp.c more extendable.

More of my work about main part if for parser. lcpp.c is not a parser. It is semi-meta-parser.
This commit converts lcpp.c to a parser and meta parser.
That means lcpp.c now gets its own parserDefinition. So we can add more things.

For exapmle per parser options name space `--opt-<LANG>=...` can be attached to lcpp.c.

### DIFF
--- a/main/lcpp.c
+++ b/main/lcpp.c
@@ -84,6 +84,38 @@ typedef struct sCppState {
 	} directive;
 } cppState;
 
+
+typedef enum {
+	CPREPRO_MACRO_KIND_UNDEF_ROLE,
+} cPreProMacroRole;
+
+static roleDesc CPREPROMacroRoles [] = {
+	RoleTemplateUndef,
+};
+
+
+typedef enum {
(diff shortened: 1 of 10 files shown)


masatake:
Currently Options.ignore is in options.c. With parserDefinition, we can move Options.ignore to cpreproc's parserDefinition.

Sorry for short description. I don't have much time now.

main/lcpp.c
```diff
@@ -417,11 +442,17 @@ static void makeIncludeTag (const  char *const name, bool systemHeader)
 	    && isXtagEnabled (XTAG_REFERENCE_TAGS)
 	    && Cpp.headerKind->roles [ role_index ].enabled)
 	{
+		if (Cpp.headerKind == CPreProKinds + CPREPRO_HEADER)
```
masatake: This condtion should be a separated function: CppStandaloneMode.

PR #1169, opened 2016-10-15
Geany sync 2

techee:
These are small things I discovered while working on https://github.com/geany/geany/pull/1263. Nothing terribly interesting except the bugs introduced by me which got fixed: https://github.com/universal-ctags/ctags/commit/42207a4ce6ccbdf91e2472b9d79cf5e1cf7ca6f2 and https://github.com/universal-ctags/ctags/commit/5331fb1b3c76dd6275bdf67c73e44633eb2f14e0 (in the later patch I'm not sure if the copyFile() is needed at all as none of the integration tests has triggered it so maybe the code isn't needed at all).

### DIFF
--- a/main/args.c
+++ b/main/args.c
@@ -120,12 +120,12 @@ static char* nextFileArg (FILE* const fp)
 static char* nextFileLine (FILE* const fp)
 {
 	char* result = NULL;
+	Assert (fp != NULL);
 	if (! feof (fp))
 	{
 		vString* vs = vStringNew ();
 		int c;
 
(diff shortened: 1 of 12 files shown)


masatake:
Looks good to me.

NOTE: It seems that --sort is never covered in Tmain as far as seeing 42207a4. Before introducing big change like introducing mio, I should write test cases for verifying the area getting impact.

techee:
There seem to be no objections and the --sort problem should be fixed so merging now.

Issue #1165, opened 2016-10-13
build and install issue


hucan7:
When I run the autogen.sh, it failed.
The Error information are as following:

```
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force
aclocal:configure.ac:12: warning: macro `AM_SILENT_RULES' not found in library
configure.ac:210: warning: AC_ARG_PROGRAM invoked multiple times
autoreconf: configure.ac: tracing
configure.ac:210: warning: AC_ARG_PROGRAM invoked multiple times
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force
configure.ac:210: warning: AC_ARG_PROGRAM invoked multiple times
configure.ac:12: error: possibly undefined macro: AM_SILENT_RULES
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:277: error: possibly undefined macro: AC_PROG_CC_C99
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

masatake:
Do you install automake?
Could you tell me the version of autoconf?

hucan7:
yes
The version of autoconf is 2.59-12.
Should I update that？

masatake:
Though  `AC_PREREQ([2.59])` is written in configure.ac, autoconf-2.63 may be needed.
Could you try newer version of autoconf?

I will make a PR for updating the line of `AC_PREREQ` in configure.ac.

hucan7:
Yes, I have updated the version of autoconf and automake.
Now the verison is autoconf-2.69, automake-1.15
But there is still error existing as following:

autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /.temp-tool/bin/autoconf --force
configure.ac:120: error: possibly undefined macro: AC_DEFINE
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.

masatake:
Could you run the following command line and show the output?
```
$ /.temp-tool/bin/autoconf --force --verbose
```

The output in my environment is that
```
[yamato@x201]~/var/ctags-github% /usr/bin/autoconf --verbose --force
/usr/bin/autoconf --verbose --force
[...]

PR #1170, opened 2016-10-16
buildsys: check the existence of xmlPreviousElementSibling in libxml2

masatake:
It seems that older libxml2 doesn't have xmlPreviousElementSibling which
is used in PlistXML parser. libxml2 shipped as part of RHEL5 is such old
library. (See #1165).

To avoid build error, configure should check the existence of xmlPreviousElementSibling
in libxml2.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -587,7 +587,9 @@ AM_CONDITIONAL([ENABLE_XCMD], [test "xyes" = "x$enable_xcmd"])
 
 AH_TEMPLATE([HAVE_LIBXML],
 	[Define this value if libxml is available.])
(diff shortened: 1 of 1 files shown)


codebrainz:
A more simple way would be to specify the [version required](https://git.gnome.org/browse/libxml2/tree/doc/symbols.xml#n1738) in `PKG_CHECK_MODULES`, for example: `PKG_CHECK_MODULES([LIBXML], [libxml-2.0 >= 2.7.3], ...)`

masatake:
@codebrainz, thank you. I will use PKG_CHECK_MODULES.

PR #1171, opened 2016-10-17
Use the same line-reading code both for iFileGetLine() and readLineRaw()

techee:
As both of them now use mio_gets() we can factor-out the common line
reading code into a separate function (readLine()) and use it for both
functions. The line-reading code from iFileGetLine() was used as it's
a bit simpler and requires less MIO operations.

The first patch was necessary since the line-reading function calls mio_eof() and as the positions are read in makePatternStringCommon(), it can happen that a pattern string is read for a tag at the end of file first and this sets the EOF flag which causes that subsequent calls of readLineRaw() will to a wrong thing as they think EOF has been reached.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -674,47 +674,58 @@ extern void ungetcToInputFile (int c)
 		File.ungetchBuf[File.ungetchIdx++] = c;
 }
 
-static vString *iFileGetLine (void)
+static void readLine (vString *const vLine, MIO *const mio)
 {
 	char *str;
 	size_t size;
-	bool haveLine;
 
-	File.line = vStringNewOrClear (File.line);
-	str = vStringValue (File.line);
-	size = vStringSize (File.line);
+	vStringClear (vLine);
+
+	str = vStringValue (vLine);
+	size = vStringSize (vLine);
 
 	for (;;)
 	{
 		bool newLine;
 		bool eof;
 
-		mio_gets (File.mio, str, size);
-		vStringSetLength (File.line);
-		haveLine = vStringLength (File.line) > 0;
-		newLine = haveLine && vStringLast (File.line) == '\n';
-		eof = mio_eof (File.mio);
+		if (mio_gets (mio, str, size) == NULL)
+		{
+			if (!mio_eof (mio))
+				error (FATAL | PERROR, "Failure on attempt to read file");
+		}
+		vStringSetLength (vLine);
+		newLine = vStringLength (vLine) > 0 && vStringLast (vLine) == '\n';
+		eof = mio_eof (mio);
 
 		/* Turn line breaks into a canonical form. The three commonly
 		 * used forms of line breaks are: LF (UNIX/Mac OS X), CR-LF (MS-DOS) and
 		 * CR (Mac OS 9). As CR-only EOL isn't haneled by gets() and Mac OS 9
 		 * is dead, we only handle CR-LF EOLs and convert them into LF. */
-		if (newLine && vStringLength (File.line) > 1 &&
-			vStringItem (File.line, vStringLength (File.line) - 2) == '\r')
+		if (newLine && vStringLength (vLine) > 1 &&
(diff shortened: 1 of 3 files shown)


main/read.c
```diff
@@ -674,47 +674,51 @@ extern void ungetcToInputFile (int c)
...
-	char *str;
-	size_t size;
-	bool haveLine;
+	char *str = vStringValue (vLine);
+	size_t size = vStringSize (vLine);
 
-	File.line = vStringNewOrClear (File.line);
-	str = vStringValue (File.line);
-	size = vStringSize (File.line);
+	vStringClear (vLine);
```
b4n: while it's probably actually harmless, I'd rather call `Clear()` *before* fetching the internal pointer (and size), just in case it ever changes to reset the buffer to the default size or something.

BTW, the whole reliance on a `vString` always having a size larger than 0 seems a bit weird, but well, nothing new in this patch.

main/read.c
```diff
@@ -815,65 +819,18 @@ extern const unsigned char *readLineFromInputFile (void)
...
-				reReadLine = true;
-			}
-			else
-			{
-				char* eol;
-				vStringLength(vLine) = mio_tell(mio) - startOfLine;
-				/* canonicalize new line */
-				eol = vStringValue (vLine) + vStringLength (vLine) - 1;
-				if (*eol == '\r')
-					*eol = '\n';
```
b4n: handling of CR line ends is different (but that probably doesn't matter as apparently it doesn't work anyway)

main/read.c
```diff
@@ -815,65 +819,18 @@ extern const unsigned char *readLineFromInputFile (void)
...
-			long startOfLine;
-
-			startOfLine = mio_tell(mio);
-			reReadLine = false;
-			*pLastChar = '\0';
-			result = mio_gets (mio, vStringValue (vLine), (int) vStringSize (vLine));
-			if (result == NULL)
-			{
-				if (! mio_eof (mio))
-					error (FATAL | PERROR, "Failure on attempt to read file");
```
b4n: I'm not totally sure what is the implication of removing this (fatal) check.  It seems fairly fine, but I'm not totally sure -- and checking the implication is a little subtle.

main/read.c
```diff
@@ -815,65 +819,18 @@ extern const unsigned char *readLineFromInputFile (void)
...
-					*eol = '\n';
-				else if (vStringLength (vLine) != 1 && *(eol - 1) == '\r'  &&  *eol == '\n')
-				{
-					*(eol - 1) = '\n';
-					*eol = '\0';
-					--vLine->length;
-				}
-			}
-		} while (reReadLine);
+		readLine (vLine, mio);
```
b4n: `readLineRawWithNoSeek()` could probably be removed now `readLineRaw()` doesn't (uselessly) seek anymore.

Though, `readLineRawWithNoSeek()` has a slightly different behavior from what I can tell: it doesn't put newlines in the buffer, skips consecutive newlines, and can set EOF early.  Not sure if all those differences are really expected given the name is so close, but maybe, not sure.  Also, it takes a `FILE`, not a `MIO`, so maybe it should stay separate.

main/read.c
techee: I can add this check back to the current implementation - a good idea to write an error message if something's wrong.

main/read.c
techee: Yeah, exactly, gets() could never return '\r' at the end of the line (unless at the end of file but there it's better to handle it the same way as for the previous lines).

b4n:
@masatake if the changes @techee mentioned in 4d82b4dc3e08f2b8cda1fce5a181c324366d27ee's message are OK with you, I think it's good.

masatake:
@b4n, thank you very much. I will take a look at 4d82b4d'.

masatake:
LGTM. Sorry to be late to repond

masatake:
Thank you.

PR #1172, opened 2016-10-19
read: Fix theoretical access to freed memory

b4n:
`readLineRawWithNoSeek()` could get the pointer to the internal string
buffer before last modifying it, if reading the last line of the input
and that line has no terminating `\r` or `\n`.  Thus, if the last
modification to the string resulted in an allocation move (e.g. if
`realloc()` returned a different pointer), `readLineRawWithNoSeek()`
would return the pointer to the previous, now invalid, string buffer.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -1022,8 +1022,6 @@ char* readLineRawWithNoSeek (vString* const vline, FILE *const pp)
 				break;
 		}
 
-		result = vStringValue (vline);
-		
 		if (c == '\n' || c == '\r')
 			nlcr = true;
 		else if (nlcr)
@@ -1033,6 +1031,8 @@ char* readLineRawWithNoSeek (vString* const vline, FILE *const pp)
 		}
 		else
 			vStringPut (vline, c);
+
+		result = vStringValue (vline);
 	}
 
 	return result;


masatake:
Excellent. I wonder how you can find this.

b4n:
I just saw this looking at the function when checking whether it couldn't be removed in the context of #1171.

But it could be reproduced artificially by e.g. altering vString so that it:

1. only allocates the exact size it needs
2. use free/malloc instead of realloc to make sure the allocation changes.

Though, the currently existing callers only use `result` to check whether it was `NULL` or not, so it's actually harmless.  But that would change as soon as someone starts using the return value.

With the following patch which only alters internal `vString` buffer sizing, uses the return value of `readLineRawWithNoSeek()`, and makes sure to receive the problematic input, you'll see the issue.  Use `valgrind ./ctags --options=NONE --langdef=dog --langmap=dog:.dog --xcmd-dog=Tmain/xcmd-tag-relative-option.d/xcmd-dog --extra=+f --fields=+n -o - --list-kinds`: with this branch, it's fine, but with master it shows a fair bunch of memory errors.
```diff
diff --git a/Tmain/xcmd-tag-relative-option.d/xcmd-dog b/Tmain/xcmd-tag-relative-option.d/xcmd-dog
index fd8a279..dd0fc20 100755
--- a/Tmain/xcmd-tag-relative-option.d/xcmd-dog
+++ b/Tmain/xcmd-tag-relative-option.d/xcmd-dog
@@ -1,6 +1,6 @@
 #!/bin/sh
 if test "z$1" = "z--list-kinds=dog"; then
-    echo "d  definition"
+    echo -n "d  definition"
 else
     cat <<EOF
 !_TAG_FILE_FORMAT	2	/extended format/
diff --git a/main/lxcmd.c b/main/lxcmd.c
index 1d7d3c7..a560a3e 100644
--- a/main/lxcmd.c
+++ b/main/lxcmd.c
@@ -350,14 +350,13 @@ static bool loadPathKinds  (xcmdPath *const path, const langType language)
 	if (pp)
 	{
 		vString* vline = vStringNew();
+		char* line;
 
-		while (readLineRawWithNoSeek (vline, pp))
+		while ((line = readLineRawWithNoSeek (vline, pp)))
 		{
-			char* line;
 			char  kind_letter;
 
 			vStringStripNewline (vline);
-			line = vStringValue (vline);
 			if (!loadPathKind (path, line, argv))
 				break;
 
diff --git a/main/vstring.c b/main/vstring.c
index 6a7ba96..79e6520 100644
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -23,7 +23,7 @@
 /*
 *   DATA DEFINITIONS
 */
-static const size_t vStringInitialSize = 32;
+static const size_t vStringInitialSize = 1;
 
 /*
 *   FUNCTION DEFINITIONS
@@ -37,13 +37,16 @@ extern void vStringResize (vString *const string, const size_t newSize)
 {
 	size_t size = vStringInitialSize;
 
-	while (size < newSize)
-		size *= 2;
+	if (size < newSize)
+		size = newSize;
 
 	if (size > string->size)
 	{
+		char *tmp = xMalloc(size, char);
+		memcpy(tmp, string->buffer, string->size);
+		eFree(string->buffer);
+		string->buffer = tmp;
 		string->size = size;
-		string->buffer = xRealloc (string->buffer, size, char);
 	}
 }
 
diff --git a/main/vstring.h b/main/vstring.h
index 7fab511..cc9f478 100644
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -90,7 +90,7 @@ extern void vStringCatSWithEscapingAsPattern (vString *output, const char* input
 static __inline void vStringPut (vString *const string, const int c)
 {
 	if (string->length + 1 == string->size)  /*  check for buffer overflow */
-		vStringResize (string, string->size * 2);
+		vStringResize (string, string->length + 2);
 
 	string->buffer [string->length] = c;
 	if (c != '\0')
```

techee:
@b4n IMO it's actually fine to drop this function completely as you suggested in #1171 . I've just done that in the last commit.

masatake:
What I should do this? I thinks this should be merged. If there is no objection, I will merge this.

b4n:
@masatake I think we should rather merge #1171, which obsoletes this PR as @techee mentioned.

masatake:
@b4n, thank you.
It seems that @techee is in busy.
So I would like to merge this. #1171 can make this commit obsolete later.

techee:
@masatake I'm not that busy - if #1171 needs some work, I can do it (I believe I addressed all the review comments).

By the way I didn't really test it myself unless the unit tests already test xcmd (are there any known external applications that use it)?

masatake:
> @masatake I'm not that busy - if #1171 needs some work, I can do it (I believe I addressed all the review comments).

@b4n, could you check the updated commits?

> By the way I didn't really test it myself unless the unit tests already test xcmd (are there any known external applications that use it)?

xcmd's related code are tested in Tmain.

b4n:
@masatake sure.  I'm a little busy right now, but I will look at it later today, tomorrow at worse.

masatake:
@b4n, thank you.

masatake:
#1171 is merged.

PR #1173, opened 2016-10-20
Introduce CTAGS_INLINE macro

masatake:
@codebrainz, could you take over this branch?
This one is based on your comment.
At least I would like you to put your copyright notice.

If you make your new pull request, I will close this one.

BTW, I sent invitation. Did you get?

### DIFF
--- a/main/inline.h
+++ b/main/inline.h
@@ -1,0 +1,26 @@
+/*
+ *
+ *  Copyright (c) 2016, 2017 Matthew Brush
+ *
+ *
+ *   This source code is released for free distribution under the terms of the
+ *   GNU General Public License version 2 or (at your option) any later version.
+ *
+ */
+
+#ifndef CTAGS_MAIN_INLINE_H
+#define CTAGS_MAIN_INLINE_H
+
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+# define CTAGS_INLINE static inline
+#elif defined(_MSC_VER)
+# define CTAGS_INLINE static __inline
+#elif defined(__GNUC__) || defined(__clang__)
+# define CTAGS_INLINE static __inline__
+// #elif ... other compilers/tests here ...
+// # define CTAGS_INLINE ...
+#else
+# define CTAGS_INLINE static
+#endif
+
+#endif /* CTAGS_MAIN_INLINE_H */
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -20,6 +20,7 @@
 
 #include <stdio.h>
 
+#include "inline.h"
 #include "mio.h"
 
 /*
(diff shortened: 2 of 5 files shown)


main/inline.h
```diff
@@ -0,0 +1,18 @@
+/* Taken from a comment at #1162 of github */
```
b4n: Should likely have a normal file header (including the license)

codebrainz:
> @codebrainz, could you take over this branch?

Ok, I might get to it this weekend.

> This one is based on your comment.

Is there an existing header where such general/platform/compat macros go? It seems a little unusual to add a whole separate header for this (but OK, if you like).

> BTW, I sent invitation. Did you get?

Yes, thanks.

masatake:
> Is there an existing header where such general/platform/compat macros go? It seems a little unusual to add a whole separate header for this (but OK, if you like).

You can introduce a unified header file where the macros go after merging inline.h.

masatake:
How do you do?

main/inline.h
masatake: Yes, this should be done by @codebrainz.

masatake:
How do you do?

masatake:
@codebrainz, how do you do about this?

codebrainz:
I tried to update it, but there are conflicts in some kind of Visual Studio files, I don't have such a version of VS to test with.

masatake:
@codebrainz, I make the branch rebase on master. 
Could you look at inline.h?
If, O.k., I would like you to fill the head of the header file like:
```
/*
 *
 *  Copyright (c) 2017, YOUR NAME HERE
 *
 *
 */
```

Opening a pull request for inline-macro branch of masatake/ctags repo, or 
Opening a new pull request for this universal-ctags/ctags repo is o.k.

codebrainz:
@masatake you can use my [Github info](https://github.com/codebrainz), it has my real full name and email address. You can also just put your own info, since you made the actual commit, it's so trivial, I don't mind at all.

masatake:
@codebrainz, thank you. I fiiled the comment block with your real name.

b4n:
Doesn't [U-CTags depend on C99](https://github.com/universal-ctags/ctags/blob/master/configure.ac#L263)?  If so, the macro is kind of irrelevant as there is `inline` in the standard, and actually it will always expand to `static inline` as `__STDC_VERSION__` will be `199901L` or higher for a C99 compiler.

masatake:
@b4n, this is an exception. As far as I can remember, @k-takata told me VisualC++ compiler doesn't understand inline keyword. @k-takata, am I correct?

codebrainz:
@masatake I believe it is true that MSVC doesn't support `inline` in plain C, and is not a C99 compiler. I think more C99 support was added in VisualStudio 2013, but as far as I can tell from a quick search, it still doesn't support `inline`.

@b4n are you sure `inline` is exactly like `static inline`. I might be thinking of C++, but I thought extern and static `inline` had different linkages.

k-takata:
https://github.com/universal-ctags/ctags/pull/1162#issuecomment-254395726
> VC2013 partly supports C99, but it doesn't support inline.
> VC2015 supports inline.

b4n:
@codebrainz my wording was misleading, I didn't mean that `inline` was the same as `static inline`, just that the macro `CTAGS_INLINE` would always expand to the C99 version `static inline` as we request for a C99 compiler.  But if it's still useful for MSVC, sure, keep it.

PR #1174, opened 2016-10-24
main: quote output file name before passing it to system(3) function

masatake:
Following command line doesn't work:

```
  $ ctags -o 'a b' ...
```

because a shell lauched from system(3) deals a whitespace between 'a'
and 'b' as a separator. The output file name is passed to system(3)
to run external sort command.

This commit adds code to put double and single quoets around the output
file name before passing it to system(3).

The issue is reported by Lorenz Hipp lhipp@idealbonn.de in a private mail.

Signed-off-by: Masatake YAMATO yamato@redhat.com


b4n:
I'm not sure if this code is used on Windows, but I don't think it would work there, because AFAIK Windows has different (and weird) quoting rules.

Otherwise, the fix looks fairly fine on *NIX at least, yet I'd rather see a builtin sort implementation used instead of resorting to `system()`.

masatake:
@b4n, thank you.

I wonder what kind of algorithm is used in sort command.
The internal sort loads all contents of tags file on to memory.

I found following comments in sort.c of coreutils.

```
/* Sort the array LINES with NLINES members, using TEMP for temporary space.
   NLINES must be at least 2.
   The input and output arrays are in reverse order, and LINES and
   TEMP point just past the end of their respective arrays.

   Use a recursive divide-and-conquer algorithm, in the style
   suggested by Knuth volume 3 (2nd edition), exercise 5.2.4-23.  Use
   the optimization suggested by exercise 5.2.4-10; this requires room
   for only 1.5*N lines, rather than the usual 2*N lines.  Knuth
   writes that this memory optimization was originally published by
   D. A. Bell, Comp J. 1 (1958), 75.  */
```

masatake:
I will add a stub verison of appendCstringWithQuotes for WIndows.

Issue #815, opened 2016-02-27
Reconsider backslashs in tag name


masatake:
Related to #582, #787.

There is a confliction about usage of backslash.
PHP uses backslash as a separator in an item in a namespace.
u-ctags uses backslash as an escape character.

Basically this is an issue of upper layer tools, not u-ctags.
u-ctags will provide the information about separators used in tags file via pseudo-tags.
The tools should use it.

Anyway I will ask you what we can do.
Implementing an option is one of the approach: --print-name-no-escaping.
Do you want the option?

Anyway I should write about the escaping clearly in docs/format.rst before releasing 1.0.

markwu:
Finally, I found the root cause. That's why I get double backslash in namesapce and class name.  

I like `--print-name-no-escaping` option, It not only solve the tag name confliction issue, and also add back compatbility for current completion tools.

pjkack:
Will the new option also be applied to the input/source-filename in the tag file? I don't want the double backslashes I'm seeing on Windows due to the following code
`
static int writeCtagsEntry (const tagEntryInfo *const tag)
{
	int length = mio_printf (TagFile.fp, "%s\t%s\t",
			      escapeName (tag, FIELD_NAME),
			      escapeName (tag, FIELD_INPUT_FILE));
`

masatake:
@pjkack,  `renderFieldName` in field.c controls it.

It seems tat `--print-name-no-escaping` and `--print-input-no-escaping` are needed.


I may have to add a pseudotag for the option.
Consider a tags is here. A tool may wonder whether it is generated with `--print-name-no-escaping` or not.

I'm thinking about embedding all options and information about environment where ctags is run as pseudo tags.

pjkack:
@masatake Your proposal sounds good to me. Out of curiosity, what's the reason for escaping by default over keeping the exuberant ctags behavior by default?

masatake:
See https://github.com/universal-ctags/ctags/issues/499
and http://docs.ctags.io/en/latest/format.html#proposal-for-extended-vi-tags-file-format

I have to write how the characters are escaped docs/format.rst.

masatake:
The backslash is introduced for recording a name with a tab character.
If --print-name-no-escaping is given and ctags finds a name with a tab character, what ctags should do?
I think just ignoring such names is enough. How do you think?

About pseudo tag, I'm thinking adding `!_TAG_FILE_ESCAPED`.
I will start implementing. However, comments are really welcome.

[...]

PR #1177, opened 2016-11-09
main: introduce the way to switch the backend for renderFieldEscaped function

masatake:
The way to escape meta characters in output may be different in each
output format. However, renderFieldEscaped only considers
(Universal-ctags native) tags format only.

This commit allows switching the backend for renderFieldEscaped function.
Each writer can customize the behavior of renderFieldEscaped.

This is the initial step for solving #815.


masatake:
Updated.

PR #1179, opened 2016-11-10
cxx: fix a typo causing a syntax error when building ctags with defin…

masatake:
…ing CXX_DEBUGGING_ENABLED


masatake:
...we should have a build test with defining CXX_DEBUGGING_ENABLED on travis or somewhere.

PR #1189, opened 2016-11-16
cxx: add a function for decoding CXXTokenType

masatake:
For making CXX_DO_DEBUGGING debug output more readable,
cxxDebugTypeDecode is introduced.

cxxDebugTypeDecode takes enum CXXTokenType typed parameter and returns
its C string representation. It can be used in CXX_DEBUG_PRINT.

misc/gencxxtypedumper.sh generates cxxDebugTypeDecode function from
enum CXXTokenType definition in cxx_token.h.

misc/gencxxtypedumper.sh is not integrated with ctags building scripts
because the script itself requires ctags internally. Generated source
code (cxx_debug_type.c) is committed to our git repository.


pragmaware:
Cool! I've always thought that something like this would be useful.

masatake:
Feel free to fix and modify the script generating the function.

PR #1190, opened 2016-11-17
C: handle attributes attached to array definitions

### LINKED ISSUE #1145: C: attribute after array definition
```
struct fb_info *registered_fb[FB_MAX] __read_mostly;
```

Above code is in linux kernel. `registered_fb` is not captured.
If I delete __read_mostly from the input, it is captured.

For ths input even OldC parser doesn't work expectedly.

masatake:
Close #1145.

	struct fb_info *registered_fb[FB_MAX] __read_mostly;

With this commit, C parser can skip __read_mostly and capture
registered_fb well.

This commit is imperfect. What the commit cannot handle is booked
as .b test case.

Though imperfect, with this commit, the result of parsing linux
kernel code is much better.


masatake:
@pragmaware, how about this one?

#1161 is good. I would like to merge. However, ignoring token after '[]' is also useful.

PR #1191, opened 2016-11-18
CXX Better macros handling (take 2)

masatake:
rebased version of #1161 

Instead of extending -I, we would like to introduce -D option.
However, the specification of arguments for -D is not well defined yet. See my comment in #1161 about `,`.

After defining the spec of -D, we have to do

* update args.ctags of test cases
* update news.rst at least (write about debugging level, too)


masatake:
I will merge c001ed1 first.

pragmaware:
Let's merge this and see where it brings us.

masatake:
I introduced -D option.
@pragmaware, you used `-I foo+=...`. I replaced it with `-D foo=....`


I removed `+` because gcc doesn't do so. Is it o.k.?

I also write about `-D` option in news.rst.

I need this feature in asm parser.

PR #1194, opened 2016-11-20
CPP: Better preprocessor and minor bug fixes

pragmaware:
Support real macros just like gcc does, including token pasting, stringification and variable arguments
Also a small bug fix related to function return type processing.

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -26,6 +26,8 @@
 #include "parse.h"
 #include "xtag.h"
 
+#include "cxx/cxx_debug.h"
+
 /*
 *   MACROS
 */
@@ -582,6 +584,10 @@ static void makeIncludeTag (const  char *const name, bool systemHeader)
 static vString *signature;
 static int directiveDefine (const int c, bool undef)
 {
+	// FIXME: We could possibly handle the macros here!
+	//        However we'd need a separate hash table for macros of the current file
+	//        to avoid breaking the "global" ones.
+
 	int r = CORK_NIL;
 
 	if (cppIsident1 (c))
@@ -1171,6 +1177,8 @@ process:
 
 static void findCppTags (void)
 {
+	CXX_DEBUG_INIT();
+
 	cppInit (0, false, false, false,
 			 NULL, 0, NULL, 0, 0);
 
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
  *  Token ignore processing
  */
 
-static hashTable * ignoreTokenTable;
+static hashTable * defineMacroTable;
 
 /*  Determines whether or not "name" should be ignored, per the ignore list.
  */
-extern const cppIgnoredTokenInfo * cppIsIgnoreToken(const char * name)
+extern const cppMacroInfo * cppFindMacro(const char * name)
 {
-	if(!ignoreTokenTable)
+	if(!defineMacroTable)
 		return NULL;
 
-	return (const cppIgnoredTokenInfo *)hashTableGetItem(ignoreTokenTable,(char *)name);
+	return (const cppMacroInfo *)hashTableGetItem(defineMacroTable,(char *)name);
 }
 
-static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
+extern vString * cppBuildMacroReplacement(
+		const cppMacroInfo * macro,
+		const char ** parameters, /* may be NULL */
+		int parameterCount
+	)
+{
+	if(!macro)
+		return NULL;
+	
+	if(!macro->replacements)
+		return NULL;
+	
+	vString * ret = vStringNew();
+
+	cppMacroReplacementPartInfo * r = macro->replacements;
+
+	while(r)
+	{
+		if(r->parameterIndex < 0)
+		{
+			if(r->constant)
+				vStringCat(ret,r->constant);
+		} else {
+			if(parameters && (r->parameterIndex < parameterCount))
+			{
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringCatS(ret,"\"");
+			
+				vStringCatS(ret,parameters[r->parameterIndex]);
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_VARARGS)
+				{
+					int idx = r->parameterIndex + 1;
+					while(idx < parameterCount)
+					{
+						vStringCatS(ret,",");
+						vStringCatS(ret,parameters[idx]);
+						idx++;
+					}
+				}
+
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringCatS(ret,"\"");
+			}
+		}
+		
+		r = r->next;
+	}
+	
+	return ret;
+}
+
+
+static void saveIgnoreToken(const char * ignoreToken)
 {
 	if(!ignoreToken)
 		return;
 
-	Assert (ignoreTokenTable);
+	Assert (defineMacroTable);
 
 	const char * c = ignoreToken;
 	char cc = *c;
@@ -1210,10 +1270,6 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
 	const char * tokenEnd = NULL;
 	const char * replacement = NULL;
 	bool ignoreFollowingParenthesis = false;
-
-
-	if (strcmp (param_name, "define") == 0)
-		ignoreFollowingParenthesis = true;
 
 	while(cc)
 	{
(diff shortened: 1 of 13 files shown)


pragmaware:
Hm... I can't see where coverage is decreased. We probably need to fix coveralls authentication?

parsers/cxx/cxx_parser_function.c
```diff
@@ -1143,13 +1148,16 @@ int cxxParserEmitFunctionTags(
...
+		(
+			(pInfo->pTypeStart == pInfo->pScopeStart) ||
+			(pInfo->pTypeStart == pInfo->pIdentifierStart)
+		)
 	)
 		pInfo->pTypeStart = pInfo->pIdentifierEnd->pNext;
 	
 	CXX_DEBUG_ASSERT(pInfo->pTypeEnd != pInfo->pIdentifierEnd,"The type should never end at identifier");
-
+
```
masatake: This should not be included.

masatake:
> Hm... I can't see where coverage is decreased. We probably need to fix coveralls authentication?

I don't understand this. Do you want to change the settings or someting of  coveralls.
In that case, tell me, I may have to do something.

Basically, the output of coveralls is just hint for writing test cases. Maybe because of compiler optimisation, the result of coveralls is not intuitive.

pragmaware:
Coveralls says that parsers/c.c decreased coverage but it won't show the source code (because of failed authentication?) so I don't know which new lines it doesn't like.

pragmaware:
How do I build news.rst so I can see the output?

masatake:
> Coveralls says that parsers/c.c decreased coverage but it won't show the source code (because of failed authentication?) so I don't know which new lines it doesn't like.

I'm very sorry. I should know the situation earlier.
I will loot at how to give you access to the lines. Please, wait for awhile.

> How do I build news.rst so I can see the output?

If you are using GNU/Linux, you need to install sphinx doc generator. On Fedora it is called python-sphinx or python3-sphinx.

After installing it, do 'cd docs; make html'.

masatake:
@pragmaware, could you acces to https://coveralls.io/builds/8912094 ?

masatake:
As far as seeing project settings in coveralls, there is no parameter to control to access coverage information.
I guess everyone can see it.

pragmaware:
I can access the build, but when I click on c.c (the file with decreased coverage) it says:

"The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API."

masatake:
@pragmaware, do you login to coveralls? You have a github account, so you can login via the account.
I will check more.

masatake:
Members in this organization, could you see the line by line information in coveralls?
I can. How about you?

masatake:
@pragmaware, could you look at this?
https://github.com/lemurheavy/coveralls-public/issues/199#issuecomment-33511612

I think you hide your the member of u-ctags organization.

parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+	{
+		if(r->parameterIndex < 0)
+		{
+			if(r->constant)
+				vStringCat(ret,r->constant);
+		} else {
+			if(parameters && (r->parameterIndex < parameterCount))
+			{
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringCatS(ret,"\"");
```
b4n: `vStringPut(ret, '"')`

parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+			
+				vStringCatS(ret,parameters[r->parameterIndex]);
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_VARARGS)
+				{
+					int idx = r->parameterIndex + 1;
+					while(idx < parameterCount)
+					{
+						vStringCatS(ret,",");
```
b4n: here too

parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+						vStringCatS(ret,parameters[idx]);
+						idx++;
+					}
+				}
 
-	return (const cppIgnoredTokenInfo *)hashTableGetItem(ignoreTokenTable,(char *)name);
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringCatS(ret,"\"");
```
b4n: and here

parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+		return NULL;
+	
+	if(!macro->replacements)
 		return NULL;
+	
+	vString * ret = vStringNew();
+
+	cppMacroReplacementPartInfo * r = macro->replacements;
+
+	while(r)
```
b4n: I'd use a for loop `for (r = macro->replacements; r; r = r->next)`

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+	while(*c && (isalnum(*c) || (*c == '_')))
+		c++;
+	
+	const char * identifierEnd = c;
+
+#ifdef CXX_DO_DEBUGGING
+	char tmp = *c;
+	*(char *)c = 0;
+	CXX_DEBUG_PRINT("Macro identifier '%s'",identifierBegin);
+	*(char *)c = tmp;
```
b4n: instead of messing with a constant string, I'd suggest
```C
CXX_DEBUG_PRINT("Macro identifier '%.*s'", identifierEnd - IdentifierBegin, identifierBegin`);
```

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+				paramBegin[iParamCount] = c;
+				c++;
+				while(*c && (*c != ',') && (*c != ')') && (!isspacetab(*c)))
+					c++;
+				paramEnd[iParamCount] = c;
+
+#ifdef CXX_DO_DEBUGGING
+				char tmp = *c;
+				*(char *)c = 0;
+				CXX_DEBUG_PRINT("Macro parameter %d '%s'",iParamCount,paramBegin[iParamCount]);
```
b4n: here too

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+				while(*c && (isalnum(*c) || (*c == '_')))
+					c++;
+
+				// check if it is a parameter
+				int tokenLen = c - tokenBegin;
+
+#ifdef CXX_DO_DEBUGGING
+				char tmp = *c;
+				*(char *)c = 0;
+				CXX_DEBUG_PRINT("Check token '%s'",tokenBegin);
```
b4n: ditto

parsers/cpreprocessor.c
masatake: my TODO: misc/src-check should detect this.

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+						{
+							p--;
+							if(!isspacetab(*p))
+							{
+								p++;
+								break;
+							}
+						}
+
+						vStringTruncate(lastReplacement->constant,p-v);
```
b4n: `vStringStripTrailing()`?

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+				}
+				
+				if(i >= iParamCount)
+				{
+					// no parameter found
+					ADD_CONSTANT_REPLACEMENT(tokenBegin,tokenLen);
+				}
+				
+				begin = c;
+				continue;
```
b4n: I don't like this `continue` and the others in this loop, that, unless I'm mistaken, don't give anything over a simple `if() else if() else if() else` construction

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,380 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+	if(info->replacements)
+	{
+		cppMacroReplacementPartInfo ** p = &(info->replacements);
+		while(*p)
+		{
+			if((*p)->constant)
+				vStringDelete((*p)->constant);
+			*p = (*p)->next;
+		}
+	}
```
b4n: why not simply this?
```
for (cppMacroReplacementPartInfo *r = info->replacements; r; r = r->next)
{
    if (r->constant)
        vStringDelete(r->constant);
}
```

`info` is freed anyway, and if it's just to get `info->replacements` `NULL` at the end, it could simply be explicit instead of making "hiding" it behind the pointer indirection -- but again, unnecessary here as the location is freed right after.

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1019,119 +1018,53 @@ static bool cxxParserParseNextTokenSkipIgnoredParenthesis(CXXToken ** ppChain)
...
-					}
-				}
-
-			}
-
-			c++;
-			b = c;
-		} else {
-			c++;
+			aParameters[iParameterCount] = vStringValue(pParam->pszWord);
```
b4n: probably a dumb question from not knowing this new C++ parser enough yet, but is `pszWord` always set, for all tokens?  And is it already "compressed" to work with params like function calls and stuff?

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1261,32 +1194,32 @@ bool cxxParserParseNextToken(void)
...
 				static int iReplacementRecursionCount = 0;
 
-				if(pIgnore->replacement)
+				if(pMacro->replacements)
 				{
 					CXX_DEBUG_PRINT(
-							"The token has replacement %s: applying",
-							pIgnore->replacement
+							"The token has replacements: applying",
+							pMacro->replacements
```
b4n: seems like the format is lost, but not the argument (which seems wrong now)

b4n:
@masatake no I can't see https://coveralls.io/builds/8912245/source?filename=parsers%2Fc.c -- as @pragmaware I see  this:

> ## Source Not Available
>
> The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.

masatake:
@b4n, you also hide your membership. You may have to make it public. Do you know how to do it?

b4n:
@masatake I just changed the setting, and it doesn't help.  But I'm not an "Owner", maybe it doesn't help?
Though, I would think it's rather either the error pointed out by coveralls, or some setting there to allow non-owners.

masatake:
@b4n, as far as I can see, there is no such setting in coveralls side.
I will add you to "Owner" group, and we can see what happens.

masatake:
I converted @b4n to Owner. @b4n, do you see something changed?

b4n:
Okay wait, I wasn't signed in on coverall.  But I can't seem to login there, the OAuth seems to fail trying again and again until GitHub says "Coveralls Pro" made an unusually high number of requests and… voila.

b4n:
…and that Coverall Pro thing wants full r/w access to all repos.  Bad idea.

b4n:
Okay I don't know, even when I'm signed in normally (no pro account), I can see the basic stuff, but when I try and go to the file view it asks me (most of the time, but sometimes it shows me the same message as mentioned above) to login to "Coverall Pro" -- and that doesn't even work, but well, I don't have a pro account anyway.

It worked at some point in the past, I did see some detailed results a while back.  Dunno what's going on, if it's a problem on their part, a setting change somewhere or a new policy, but that makes the thing unusable.

pragmaware:
OK. Let's see if I can do this better.

parsers/cxx/cxx_parser_tokenizer.c
pragmaware: pszWord is always set. In this specific piece of code it's also explicitly "compressed" by the preceding calls so the tokens of the chain contain whole parameters of the macro.

pragmaware:
Closing this pull. Will reopen a new, polished, one.

parsers/cpreprocessor.c
pragmaware: Right... now that I look at it there's even a memory leak there.

parsers/cpreprocessor.c
b4n: Well, I see it the other way around: when I see the next `if`s, I have to wonder what happened before them, I can't simply ignore the block content.  So without the `continue`s, I can have a high level view of the loop in *"it 'switches' over the value of `c`"*, and then I only have to care about how each specific case is handled.

But well, that's probably just a question of taste if you know why you did it, no biggie.

masatake:
(https://coveralls.zendesk.com/hc/en-us/articles/206384305-How-do-I-add-users-to-a-repo-on-Coveralls-)

masatake:
All, I wonder you may need to turn on Universal-ctags/ctags repository in YOUR coveralls account.
Could you try?

PR #1203, opened 2016-11-25
CXX: Better preprocessor macros - take 2

pragmaware:
Support real macros just like gcc does, including token pasting, stringification and variable arguments.

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -26,6 +26,8 @@
 #include "parse.h"
 #include "xtag.h"
 
+#include "cxx/cxx_debug.h"
+
 /*
 *   MACROS
 */
@@ -582,6 +584,10 @@ static void makeIncludeTag (const  char *const name, bool systemHeader)
 static vString *signature;
 static int directiveDefine (const int c, bool undef)
 {
+	// FIXME: We could possibly handle the macros here!
+	//        However we'd need a separate hash table for macros of the current file
+	//        to avoid breaking the "global" ones.
+
 	int r = CORK_NIL;
 
 	if (cppIsident1 (c))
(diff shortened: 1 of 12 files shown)


parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+	{
+		if(r->parameterIndex < 0)
+		{
+			if(r->constant)
+				vStringCat(ret,r->constant);
+		} else {
+			if(parameters && (r->parameterIndex < parameterCount))
+			{
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringPut(ret,'\"');
```
b4n: no need to escape `"` inside `''`

parsers/cpreprocessor.c
```diff
@@ -1184,24 +1192,76 @@ static void findCppTags (void)
...
+					while(idx < parameterCount)
+					{
+						vStringPut(ret,',');
+						vStringCatS(ret,parameters[idx]);
+						idx++;
+					}
+				}
+
+				if(r->flags & CPP_MACRO_REPLACEMENT_FLAG_STRINGIFY)
+					vStringPut(ret,'\"');
```
b4n: ditto

parsers/cpreprocessor.c
```diff
@@ -1244,54 +1300,375 @@ static void saveIgnoreToken(const char * ignoreToken, const char * param_name)
...
+
+		info->hasParameterList = true;
+
+		c++;
+		while(*c)
+		{
+			while(*c && isspacetab(*c))
+				c++;
+
+			if(*c && (*c != ','))
```
b4n: Hum, won't this add one parameter too many, as it'll be entered by the closing `)`?

PR #1204, opened 2016-11-25
Fix handling assignment to a member of an array element

### LINKED ISSUE #1193: js: a bug in capturing anonymous function
```
[jet@localhost ~]$ cat /tmp/foo.js 
x[i].f = function () {
}

y.f = function () {
}
[jet@localhost ~]$ ~/var/ctags/ctags  -o - /tmp/foo.js 
f	/tmp/foo.js	/^y.f = function () {$/;"	f	function:y
x	/tmp/foo.js	/^x[i].f = function () {$/;"	f
```

`x` is captured as a function. Is this expected behaviour?

(See my_global_var4 in review-needed.r/simple.js.t/.)

b4n:
Fixes #1193.

---

I'm not totally sure what should be emitted here, but at least it makes more sense now.


b4n:
(just repushed with a more correct wording in the commit message)

masatake:
LGTM. What a short fix!

b4n:
Yeah, fortunately it was just a small bug :)

PR #1205, opened 2016-11-25
JavaScript: Add support for generators

b4n:
Add support for them `function*` generator syntax, and report such generators as a separate kind.

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*)


masatake:
LGTM.

PR #1209, opened 2016-11-26
Exuberant ctags writer

### LINKED ISSUE #815: Reconsider backslashs in tag name

masatake:
Close #815.

`--output-format=ctags` is removed.
Instead `--output-format=u-ctags` and --output-format=e-ctags` are introduced.
u-ctags is the default output format. With the format ctags can record names that include white spaces.
It uses '\\' for escaping.

e-ctags provides compatible output. It cannot record names that include white spaces; it just thow away them. However, it doesn't use '\\' for escaping.

I wonder we should use a name ctags instead of u-ctags...


masatake:
Update: the last json-writer specific patch is squashed.

masatake:
It seems that converalls is gone.

PR #1215, opened 2016-12-01
buildsys: put "automatically generated" marker as comment

masatake:


autogen.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh
 
 ctags_files=`make -f makefiles/list-translator-input.mak`
-misc/dist-test-cases > makefiles/test-cases.mak && \
+{
+	echo "# Generated by $0"
+	misc/dist-test-cases
```
b4n: shouldn't the comment rather be emitted by `misc/dist-test-cases` itself?

masatake:
@b4n, You are right.

b4n:
Off topic: *misc/dist-test-cases* uses `&&` at every line *end*, but no `\`, that's odd.  Shouldn't this script simply `set -e` and drop the `&&`s?

masatake:
Yes, `-e` can be used.

> but no `\`,
I don't understand it well. What do you mean?
As far as I know `&&` at the end of line is enough for chaining condition statements.
Am I wrong ?

b4n:
> As far as I know `&&` at the end of line is enough for chaining condition statements.
Am I wrong ?

Oh, no you're right, but I didn't know that :)  So I learned something new today :)
So my comment is moot then, the script is fine.

masatake:
I see.

PR #1219, opened 2016-12-01
json-writer: ensure pattern is always populated

### LINKED ISSUE #1218: JSON output format does not contain pattern for some language
Languages like C and Ruby produce proper pattern output:

```
$ ./ctags --output-format=json -o - test.c
{"_type": "tag", "name": "jsonErrorPrinter", "path": "test.c", "pattern": "/^bool jsonErrorPrinter (const errorSelection selection, const char *const format, va_list ap, voi/", "typeref": "bool", "kind": "function"}

$ ./ctags --output-format=json -o - test.rb
{"_type": "tag", "name": "perform", "path": "test.rb", "pattern": "/^      def perform$/", "kind": "method", "scope": "Job", "scopeKind": "class"}
```

But languages like Go and JavaScript will not produce any pattern:

```
$ ./ctags --output-format=json -o - test.go
{"_type": "tag", "name": "Add", "path": "test.go", "pattern": "", "kind": "func"}

$ ./ctags --output-format=json -o - test.js
{"_type": "tag", "name": "onSelection", "path": "test.js", "pattern": "", "kind": "function"}
```

tmm1:
fixes #1218

### DIFF
--- a/Tmain/json-output-format.d/input.go
+++ b/Tmain/json-output-format.d/input.go
@@ -1,0 +1,4 @@
+package main
+
+func main() {
+}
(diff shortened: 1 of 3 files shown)


tmm1:
@masatake Any issue with this PR?

masatake:
I wonder this is an issue of json-writer or not.
Some parsers fill the pattern field. The others  don't. I wonder why. I should know this. Please, wait for awhile.

masatake:
It is related to corkAPI.

masatake:
@tmm1, could you look at #1222 ?

I think this one not an issue of json writer. This issue should be solved in upper layer.
I picked up your test case.

masatake:
Merged via #1222.

PR #1222, opened 2016-12-03
main: revise the code for rendering pattern field

masatake:
The responsibility for rendering pattern field was not clear.
json and ctags writers did their own way.

This commit moves the rendering code to field.c and make the
writers use functions exported from field.c.

### DIFF
--- a/Tmain/json-output-format.d/input.go
+++ b/Tmain/json-output-format.d/input.go
@@ -1,0 +1,4 @@
+package main
+
+func main() {
+}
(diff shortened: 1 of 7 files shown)


tmm1:
👍 very nice cleanup.

masatake:
Thank you. I will merge this after passing the test on CI.

masatake:
Updated. A file tag has not pattern. (As you mentioned in a comment in .c file.)

masatake:
I will squash some of commits.

PR #1224, opened 2016-12-03
main: support multiline regex parser

### LINKED ISSUE #219: Does ctags have multiline regex support?
Hello, 
I have some Java code

``` java
@Subscribe
public void catchEvent(SomeEvent
```

And I want to match it with something like this in my ~/.ctags

```
--langdef=javaspring
--langmap=javaspring:.java
--regex-javaspring=/@Subscribe(\s)*([a-z ]+)\s([a-zA-Z]*)\(([a-zA-Z]*)/\3-\4/
```

This should generate something like "catchEvent-SomeEvent" but it does not. If I remove the newline between @Subscribe and public void it gets it.

``` java
@Subscribe public void catchEvent(SomeEvent
```

``` sh
rm -rf tags && ctags -R . && cat tags|grep Subs
catchEvent-SomeEvent    PathToMyFile.java    /^@Subscribe public void catchEvent(SomeEvent event) {$/;"      r
```

Do you have any idea if I can do this with the current ctag implementation. Am I doing something wrong or ctags does not support multi-line regex now.

Thank you!

masatake:
Close #219

A pattern marked with `_multiline` is applied to whole file contents, not line by line.

Next example is based on an issue #219 posted by @andreicristianpetcu::

    $ cat input.java
    @Subscribe
    public void catchEvent(SomeEvent e)
    {
	return;
    }

    @Subscribe
    public void
	recover(Exception e)
    {
	return;
    }

    $ cat spring.ctags
    --langdef=javaspring
    --langmap=javaspring:.java
    --regex-javaspring=/@Subscribe([[:space:]])*([a-z ]+)[[:space:]]*([a-zA-Z]*)\(([a-zA-Z]*)/\3-\4/s,subscription/{_multiline=3}
    --excmd=mixed
    --fields=+ln

    $ ./ctags -o - --options=./spring.ctags input.java
    Event-SomeEvent	input.java	/^public void catchEvent(SomeEvent e)$/;"	s	line:2	language:javaspring
    recover-Exception	input.java	/^    recover(Exception e)$/;"	s	line:10	language:javaspring

* `{_multiline=N}`

	This tells the patern should be applied to whole file
	contents, not line by line.  `N` is the number of a group in the
	pattern. The specified group is used to record the line number
	and the pattern of tag. In the above example 3 is
	specified. The start position of the group 3 within the whole
	file contents is used.

NOTE: This flag doesn't work well with scope related flags and ``exclusive`` flags.

### DIFF
--- a/Units/regex-multiline-flag.d/args.ctags
+++ b/Units/regex-multiline-flag.d/args.ctags
@@ -1,0 +1,5 @@
+--langdef=javaspring
+--langmap=javaspring:.java
+--regex-javaspring=/@Subscribe([[:space:]])*([a-z ]+)[[:space:]]*([a-zA-Z]*)\(([a-zA-Z]*)/\3-\4/s,subscription/{_multiline=3}
+--excmd=mixed
+--fields=+ln
--- a/Units/regex-multiline-flag.d/expected.tags
+++ b/Units/regex-multiline-flag.d/expected.tags
@@ -1,0 +1,2 @@
+Event-SomeEvent	input.java	/^public void catchEvent(SomeEvent e)$/;"	s	line:2	language:javaspring
+recover-Exception	input.java	/^    recover(Exception e)$/;"	s	line:10	language:javaspring
--- a/Units/regex-multiline-flag.d/input.java
+++ b/Units/regex-multiline-flag.d/input.java
@@ -1,0 +1,13 @@
+@Subscribe
+public void catchEvent(SomeEvent e)
+{
+    return;
+}
+
+
+@Subscribe
+public void
+    recover(Exception e)
+{
+    return;
+}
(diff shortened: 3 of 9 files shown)


masatake:
Added a parser for linker script.

masatake:
Some commits are squashed.

masatake:
The parser for ld-script is removed because it is not good exapmle as a multiline regex parser.

andreicristianpetcu:
Thank you! I'll try it out but I need to figure out how to build it locally :D I use Ubuntu Fedora 24 at home (I will update to 25 this month) and 16.04 at work

masatake:
See http://docs.ctags.io/en/latest/autotools.html

You may not need do "make install". "make" is enough for generating ctags executable.

Issue #1188, opened 2016-11-16
ctags doesn't find subsequent functions when a macro defines an else part for an if


Smjert:
Hello, i have this short example (test.c):

```
int main()
{
        if (something) {
                printf("hello");
#ifdef world
        } else
                printf(" world\n");
#else
        }
#endif
}

int func()
{
}
```

Calling 'ctags test.c' only generate the tag for main() function but not for func().

Ctags Info:

> Universal Ctags 0.0.0(0b1b759), Copyright (C) 2015 Universal Ctags Team
> Universal Ctags is derived from Exuberant Ctags.
> Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
>   Compiled: Nov 14 2016, 17:05:27
>   URL: https://ctags.io/
>   Optional compiled features: +wildcards, +regex, +option-directory, +coproc

b4n:
You example code is invalid if `world` is defined, as it results in unbalanced braces.  Use `} else {` for valid syntax in all branches.

b4n:
…and actually your edited example is invalid in any case, still unbalanced braces.

Smjert:
I'm sorry i did a mess when copying (there were spaces everywhere) and acidentally deleted the #else part.

b4n:
Okay, now it's messy code, but it is valid alright.

@pragmaware @masatake OldC handles it OK (probably because it retries with another preprocessor config on unbalanced braces?)

masatake:
This is the hardest issue in things ctags deals with.
I will never say this is a bug of parser. However, if ctags can handl these input well, people may be happy.

b4n:
Agreed that I don't see it as a high priority thing.  It'd be nice to support, sure, but it's fairly tricky code for no reason.

I believe the OldC parser did a second pass with another option to `lcpp` when it reached EOF with unbalanced braces.  It might even be "simple" to support it the same way.

Smjert:
Thanks for the attention on the issue. On my side i'll fix the braces, i saw that it works... unfortunately is a very old code so it's riddled by things like that ^^'.

pragmaware:
Forcing a second pass in this case is pretty easy: see https://github.com/pragmaware/ctags/commit/2972067276dfd6eefb3eea02784ad6e27a88deae .

[...]

PR #1226, opened 2016-12-04
CPP/CXX: More preprocessor work

pragmaware:
Switch to single preprocessor conditional branch mode when nested {} block level at entry and exit of the first conditional is different. Fixes bug #1188 .

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -51,6 +51,7 @@ typedef struct sConditionalInfo {
 	bool singleBranch;       /* choose only one branch */
 	bool branchChosen;       /* branch already selected */
 	bool ignoring;           /* current ignore state */
+	int enterExternalParserState;          /* the parser state when entering this conditional: used only by cxx */
 } conditionalInfo;
 
 enum eState {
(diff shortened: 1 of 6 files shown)


parsers/cxx/cxx_parser_block.c
```diff
@@ -582,3 +582,19 @@ bool cxxParserParseBlock(bool bExpectClosingBracket)
...
+// or the statement ends. When a characteristic token is found it usually enters
+// a specialized scanning routine (e.g for classes, namespaces, structs...).
+// When the statement ends without finding any characteristic token the chain
+// is passed to an analysis routine which does a second scan pass.
+//
+bool cxxParserParseBlock(bool bExpectClosingBracket)
+{
+	cppSetExternalParserState(cppGetExternalParserState() + 1);
+	bool bRet = cxxParserParseBlockInternal(bExpectClosingBracket);
+	cppSetExternalParserState(cppGetExternalParserState() - 1);
```
masatake: Does Cxx parser uses the value returned from cppGetExternalParserState()?
If not, how about introducing cppPushExternalParserState/cppPopExternalParserState instead of Get/Set?
So you don't have to write +/- 1 for each invocation.

parsers/cpreprocessor.c
```diff
@@ -439,8 +462,37 @@ static bool isIgnoreBranch (void)
...
+	*     yyyy;
+	* #endif
+	*
+	* Follow single branch example: (different block level at enter and exit)
+	*
+	*    if {
+	* #if something
+    *    } else x;
+	* #else
+	*    }
```
masatake: A white space after `}`.
Do git diff. It will be rendered with red color.

masatake:
The other than the white space issue, LGTM.

Issue #1220, opened 2016-12-02
request: CUDA support


sunbubble:
I've been using ctags with langmap c++ for CUDA ( *.cu and *.cuh filetypes), and it works just fine.
Is there a way to have this be default behavior?

$ ctags --langmap=c++:+.cu.cuh

masatake:
@pragmaware, how do you think?

I got the same situation in https://github.com/universal-ctags/ctags/issues/1086 about jsx.
I accepted the patch (without agreement from @b4n, the maintainer of JavaScript parser.)
I'm still not sure this is good or not.

I don't know CUDA. As far as my quick reserach, it has some syntax level extension like `<<<...>>>`.
If a language has its own syntax, ctags should put the name of language to "language:" field.
So I don't want to say "ctags supports CUDA" with just adding " --langmap=c++:+.cu.cuh".

Ideally, CUDA subparser stacked on Cxx parser is implemented.
 keep in mind that your initial motivation of rewriting C and C++ parsers;
The original c.c was too complicated because it supports multiple `{' `}' syntax languages.

Just adding parserDefinition as a stub is one of approach.
```
diff --git a/parsers/cxx/cxx.c b/parsers/cxx/cxx.c
index a3e396d..e83c44a 100644
--- a/parsers/cxx/cxx.c
+++ b/parsers/cxx/cxx.c
@@ -135,3 +135,25 @@ parserDefinition * CppParser (void)
 
 	return def;
 }
+
+parserDefinition * CudaParser (void)
+{
+	static const char * const extensions [] =
+	{
+		"cu",
+		"cuh",
+		NULL
+	};
+	static parserDependency dependencies [] = {
+		{ DEPTYPE_KIND_OWNER, "C" },
+	};
+
+	static selectLanguage selectors[] = { selectByObjectiveCKeywords, NULL };
+
+	parserDefinition* def = parserNew("Cuda");
+
+	/* The same as C++. */
+	def->parser2 = cxxCppParserMain;
+	/* The same as C++. */
+	return def;
+}
```

At least we should require a test case as I did in #1086.

pragmaware:
In theory CUDA is C with some additional keywords and builtin types. The only syntax difference that I'm aware of is the <<<>>> kernel calling syntax which probably can be ignored. 
[...]

PR #1227, opened 2016-12-04
CXX: CUDA support

pragmaware:
Add basic CUDA language support as a new parser.
Requested by #1220


masatake:
When the development becomes stable, please, consider squashing some of commits.
We don't have to record the history of each try and error commit like ee3cc92, for example.
Other than  		4c2b478 can be squashed into one, I think.

pragmaware:
OK. Closing this and reopening with a simpler commit structure.

PR #1230, opened 2016-12-06
Improve path delimiters/separators handling

masatake:
Windows related fixes. Appveyor said o.k.


masatake:
#893 can be merged if this commit is acceptable.

masatake:
Long standing issues about path separators on Windows are not solved.

PR #1234, opened 2016-12-09
[CXX] CUDA support (take 3)

pragmaware:
Add basic CUDA support as a cxx parser sub-language.


masatake:
Thank you for squashing. LGTM. Please, merge this after adding CUDA to news.rst!


Ideally, The commit should be splitted to steps of logical changes.
For example. you changed `enum CXXKeyword` to `CXXKeyword`.

Logical changes should not be squashed.  Non logical changes should be squashed into a commit.
https://github.com/universal-ctags/ctags/pull/1224 is an example.
A set of commits are not the developer's hacking log. Of curser, git can record them. However, I think the power of git is that it can rearrange the commit history. You can rewrite commits your own story.

The reviewing becomes much easier if the commits are logically separated well.

If you have a trouble in using git, please, open an issue.  Thank you.

masatake:
Let's merge!

PR #1237, opened 2016-12-09
SQL: Ignore token after WITH and WITHOUT

### LINKED ISSUE #1235: SQL parser generates wrong results when PostgreSQL's CREATE CAST is used in file
When running ctags on a file containing

    CREATE CAST (type_a AS type_b)
        WITH FUNCTION type_conversion_function(type_a)
        AS IMPLICIT;

    CREATE OR REPLACE VIEW my_view AS
    SELECT 1;

the resulting tags file is

    !_TAG_FILE_FORMAT       2       /extended format; --format=1 will not append ;" to lines/
    !_TAG_FILE_SORTED       1       /0=unsorted, 1=sorted, 2=foldcase/
    !_TAG_OUTPUT_MODE       u-ctags /u-ctags or e-ctags/
    !_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
    !_TAG_PROGRAM_NAME      Universal Ctags /Derived from Exuberant Ctags/
    !_TAG_PROGRAM_URL       https://ctags.io/       /official site/
    !_TAG_PROGRAM_VERSION   0.0.0   /6587204/
    type_conversion_function        ctags_testcase.sql   /^    WITH FUNCTION type_conversion_function(type_a)$/;"        f

This is wrong, because the view my_view is not detected (which works fine if I remove the CREATE CAST).

I'm not sure whether type_conversion_function should appear in the tags file. Is the "f" tag used for SQL function declarations only, or also for functions usages?

masatake:
SQL: Ignore token after WITH and WITHOUT

Close #1235.

SQL parser tries to capture token after FUNCTION keyword.
However, the keyword is used not for defining FUNCTION:

    /* https://www.postgresql.org/docs/9.4/static/sql-createcast.html */

    CREATE CAST (source_type AS target_type)
	WITH FUNCTION function_name (argument_type [, ...])
	[ AS ASSIGNMENT | AS IMPLICIT ]

    CREATE CAST (source_type AS target_type)
	WITHOUT FUNCTION
	[ AS ASSIGNMENT | AS IMPLICIT ]

FUNCTION keyword after WITH keyword and WITHOUT keyword should
not handle specially.

The test case is taken from the issue submitted by @karlb.

### DIFF
--- a/parsers/sql.c
+++ b/parsers/sql.c
@@ -46,7 +46,7 @@
 	/*
 	 * Other databases are less restrictive on the first character of
 	 * an identifier.
-	 * isIdentChar1 is used to identify the first character of an 
+	 * isIdentChar1 is used to identify the first character of an
 	 * identifier, so we are removing some restrictions.
 	 */ \
 	(isalpha (c) || (c) == '@' || (c) == '_' )
@@ -135,7 +135,7 @@ enum eKeywordId {
 	KEYWORD_handler,
 	KEYWORD_comment,
 	KEYWORD_create,
-	KEYWORD_go
+	KEYWORD_go,
 };
 typedef int keywordId; /* to allow KEYWORD_NONE */
 
@@ -328,7 +328,7 @@ static tokenType parseSqlFile (tokenInfo *const token);
 
 static bool isCmdTerm (tokenInfo *const token)
 {
-	DebugStatement ( 
+	DebugStatement (
 			debugPrintf (DEBUG_PARSE
 				, "\n isCmdTerm: token same  tt:%d  tk:%d\n"
 				, token->type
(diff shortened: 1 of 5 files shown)


masatake:
@b4n, could you review 8e83ee0?

parsers/sql.c
```diff
@@ -478,6 +480,21 @@ static void parseIdentifier (vString *const string, const int firstChar)
...
 
+static tokenType parseDollarQuote (vString *const string, const int firstChar)
+{
+	int c = firstChar;
+	do
+	{
+		vStringPut (string, c);
+		c = getcFromInputFile ();
+		if (c == EOF)
+			return TOKEN_EOF;
```
b4n: Probably not very important, but I would rather simply cut off the end and still return the current token, and let the next one return `TOKEN_EOF`, like string does.  But that probably doesn't change anything.

masatake:
> I can give this a shot in a few minutes if you'd like.

Yes. I would like you to take over this pull request.

I'm working on moving `review-needed.r/*sql*.t` to `parser-sql.r`. I would like to merge #1238 first.

masatake:
(I merged #1238.)

PR #1238, opened 2016-12-10
Review sql test cases

masatake:


masatake:
Though there are many test cases for SQL, its coverage is not high enough.

PR #1240, opened 2016-12-12
Asm: fix crashing when encountering extra .endm

masatake:

### DIFF
--- a/parsers/asm.c
+++ b/parsers/asm.c
@@ -208,9 +208,12 @@ static void makeAsmTag (
 													 kind_for_directive);
 				break;
 			case K_PSUEDO_MACRO_END:
-				macro_tag = getEntryInCorkQueue (*lastMacroCorkIndex);
-				macro_tag->extensionFields.endLine = getInputLineNumber ();
-				*lastMacroCorkIndex = CORK_NIL;
+				if (*lastMacroCorkIndex != CORK_NIL)
+				{
+					macro_tag = getEntryInCorkQueue (*lastMacroCorkIndex);
+					macro_tag->extensionFields.endLine = getInputLineNumber ();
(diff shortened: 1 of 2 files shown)


parsers/asm.c
```diff
@@ -208,9 +208,12 @@ static void makeAsmTag (
...
 			case K_PSUEDO_MACRO_END:
-				macro_tag = getEntryInCorkQueue (*lastMacroCorkIndex);
-				macro_tag->extensionFields.endLine = getInputLineNumber ();
-				*lastMacroCorkIndex = CORK_NIL;
+				if (*lastMacroCorkIndex != CORK_NIL)
+				{
+					macro_tag = getEntryInCorkQueue (*lastMacroCorkIndex);
+					macro_tag->extensionFields.endLine = getInputLineNumber ();
+					*lastMacroCorkIndex = CORK_NIL;
+				}
```
b4n: A more solid fix could be
```C
				macro_tag = getEntryInCorkQueue (*lastMacroCorkIndex);
				if (macro_tag)
					macro_tag->extensionFields.endLine = getInputLineNumber ();
				*lastMacroCorkIndex = CORK_NIL;
```
so whether the entry exits or not it won't crash, even if `*lastMacroCorkIndex` references a non-existing one.

But I guess it's not very important as this shouldn't ever happen.

masatake: Yes, however, in that case I got a crash report. As far as my understanding, if lastMacroCorkIndex is non-zero,  macro_tag should have value. If this assumption is not satisfied, there is another bug I injected.

b4n: yeah it definitely should.  I was just suggesting checking `macro_tag` before dereferencing it to be on the safest side, but here it probably doesn't matter

masatake:
Thank you.

PR #1243, opened 2016-12-16
Token info api (and parser for GNU LD script)

masatake:
Though I have worked on Universal-ctags more than two years, I have never written a real parser(crafted parser).

This is the first time to write a crafted parser.
For long time I have wanted to write a parser for GNU linker script, which is important for understanding Linux kernel. Maybe you, @b4n, are not interested in the language. The parser is not the target of reviewing I ask.

For writing the crafted parser, I have studied parsers written by you(php, javascript, and  python). I have found some code common between your parers. I convert it to tokeninfo.[ch]. These files are the target of reviewing. Could you give me comments?


masatake:
(More test cases for ldscript parser are needed. Especially about operators.)

masatake:
Add a missing test case for improving coverage.

masatake:
I will merge this because I would like to use ldscript.c in daily job.
@b4n, feel free to fix the code if you think I made wrong design/decision.

b4n:
@masatake I think I'll have quite a few remarks, and would have done that fairly differently, but I probably won't have time to work on something that size in the upcoming days, so go ahead, and feel free to remember me about this in case I forget.

But I like the goal here in any case :)

masatake:
> But I like the goal here in any case :)

Thank you. I will write more parrses based on then new API to understand what I have done:-P

PR #1253, opened 2016-12-29
etags-writer: truncate too long hints (step 1)

### LINKED ISSUE #1247: etags: severe performance and file-size issue
I'm generating etags for a large Rails app. Prior to using Emacs, I was generating ctags for Vim:
```
$ git ls-files | \
  ctags --tag-relative=yes -L - -f ".git/tags" --languages=-javascript,sql
```
This took about a second and produced a 7MB file. With the `-e` flag, it takes 80 seconds and produce an ***8GB*** file.
```
$ git ls-files | \
  ctags --tag-relative=yes -L - -e -f ".git/tags" --languages=-javascript,sql
```

I initially thought it was a problem with etags in general, but I've since found that the `etags` executable that came with Emacs 25.1 takes 13 seconds and produces a 3MB file, while `exuberant-ctags` takes 2 seconds and produces a 1MB file.

I'm on a 2013 MacBook Air running macOS Sierra 10.12.1

masatake:
Close #1247.

Like javascript and css optimized by tools, there are source files that are made
up of one very long line. Such input files can make TAGS file much larger.

As we did in tags output, this commit truncates hints in TAGS output if they are
longer than the value specified with --pattern-length-limit option.

TODO: json output, xref output.

### DIFF
--- a/Tmain/omit-long-patterns-etags.d/stdout-expected.txt
+++ b/Tmain/omit-long-patterns-etags.d/stdout-expected.txt
@@ -1,0 +1,5 @@
+
+input.sh,349
+                                                                                                func96(func961,0
+                                                                                               func95()func955,110
+                                                                                                 func97func979,219
--- a/Tmain/omit-long-patterns-etags.d/input.sh
+++ b/Tmain/omit-long-patterns-etags.d/input.sh
@@ -1,0 +1,16 @@
+                                                                                                func96()
+{
+
+}
+                                                                                               func95()
+{
+
+}
+                                                                                                 func97()
+{
+
+}
+
+func96
+func95
+func97
--- a/Tmain/omit-long-patterns-etags.d/run.sh
+++ b/Tmain/omit-long-patterns-etags.d/run.sh
@@ -1,0 +1,7 @@
+# Copyright: 2015 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
+# strlen ("func95()") => 8
+${CTAGS} --quiet --options=NONE --pattern-length-limit=`expr 96 + 8` -e -o - ./input.sh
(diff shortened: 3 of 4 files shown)


Tmain/omit-long-patterns-etags.d/stdout-expected.txt
```diff
@@ -0,0 +1,5 @@
+
+input.sh,349
+                                                                                                func96(func961,0
```
dgutov: Are we supposed to see truncation in this example?

masatake: Yes.

masatake:
Comparing with tags output, etags output is still slow. In the current implementation, truncation is done after reading line.
Instead, we can the same effect by limiting the length of reading.
I will work on this item after merging this.

Tmain/omit-long-patterns-etags.d/stdout-expected.txt
dgutov: I don't see it. There are just as many spaces as in the source file (and if truncated, the function name would either be missing or at least incomplete, in the first field).

masatake: Oh, sorry. My code truncates lines from their tail. 
I think it should do from head. fixup commit is pushed. Could you look at the commit again?

b4n: I though truncating the head wasn't an option because etags matching used an anchored search? (`^...`)

b4n: @dgutov I though the goal was to truncate after N characters, and add the name in a separate field, so what @masatake did originally seemed correct.

Of course with the spaces example it won't be a very useful search line, but that's pretty artificial IMO.

dgutov: Yes, sorry. The first version seems to have been the correct one.

I simply missed the truncation because it was by 1 or 2 characters.

PR #1254, opened 2016-12-29
C,CUDA: remove unnecessary CASE_INSENSITIVE_FILENAMES block

masatake:
In case-sensitive platform/environment (in other word #ifndef CASE_INSENSITIVE_FILENAMES),
using C (upper case) as extension for C source code is not popular.
Like C, using CUDA (upper case) as extension for CUDA source code is not popular.

What is popular is using C as extension for C++ source code.

### DIFF
--- a/parsers/cxx/cxx.c
+++ b/parsers/cxx/cxx.c
@@ -77,9 +77,6 @@ parserDefinition * CParser (void)
 	static const char * const extensions [] =
 	{
 		"c",
-#ifndef CASE_INSENSITIVE_FILENAMES
-		"C",
-#endif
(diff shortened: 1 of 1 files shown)


masatake:
```
extern parserDefinition* FortranParser (void)
{
	static const char *const extensions [] = {
		"f", "for", "ftn", "f77", "f90", "f95",
#ifndef CASE_INSENSITIVE_FILENAMES
		"F", "FOR", "FTN", "F77", "F90", "F95",
#endif

```

How about fortran? Is F(upper case) popular exntesion in case-Sensitive platfrom?

PR #1255, opened 2016-12-29
main: implement case insenstive glob matching

masatake:
ctags uses file name (glob) patterns and extensions to choose a
proper parser for given input.

extensions work well in case insenstive environment.
However, patterns don't. This commit fixes this.

GNU implementation provides FNM_CASEFOLD.
I didn't use it to keep portability.

### DIFF
--- a/main/strlist.c
+++ b/main/strlist.c
@@ -216,7 +216,21 @@ static bool fileNameMatched (
 		const vString* const vpattern, const char* const fileName)
 {
 	const char* const pattern = vStringValue (vpattern);
+
+#ifdef CASE_INSENSITIVE_FILENAMES
+	{
+		bool r;
+
+		char* const p = newUpperString (pattern);
+		char* const f = newUpperString (fileName);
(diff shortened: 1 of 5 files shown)


b4n:
LGTM if it works

masatake:
@b4n, thank you fore reviewing.

main/strlist.c
```diff
@@ -216,7 +216,21 @@ static bool fileNameMatched (
...
 {
 	const char* const pattern = vStringValue (vpattern);
+
+#ifdef CASE_INSENSITIVE_FILENAMES
+	{
+		bool r;
+
+		char* const p = newUpperString (pattern);
+		char* const f = newUpperString (fileName);
+		r =  (bool) (fnmatch (p, f, 0) == 0);
```
codebrainz: No big deal, but this `(bool)` cast (and the existing one below) isn't needed. Also the `r` variable could be initialized directly like `bool r = (fnmatch (p, f, 0) == 0);` instead of doing it in two steps.

PR #1258, opened 2017-01-01
Python: Updates to support 3.6

b4n:
Not many code changes were needed, but some, and test cases for newer features.

* [f-strings](https://www.python.org/dev/peps/pep-0498/)
* [variable annotations](https://www.python.org/dev/peps/pep-0526/)
* [async methods](https://www.python.org/dev/peps/pep-0492/)
* [underscore in numeric literals](https://www.python.org/dev/peps/pep-0515/)
* [`@` operator](https://www.python.org/dev/peps/pep-0465/)


masatake:
Thank you. This is very important updates.

Could you add args.ctags to each test case with following contents?
```
--sort=no
```

Ideally this should be done in unit test harness side. However, I didn't because I have to update too many test cases (expected.tags).

b4n:
@masatake sure, done.

masatake:
LGTM.

masatake:
@b4n, Could you merge this?

I would like to capture `@task` in a python script using celery, http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html . For the purpose I would like to hack python.c.

b4n:
@masatake sure, done

masatake:
> @masatake how is this different from normal decorators?

```
[yamato@master]~/var/ctags-github% cat /tmp/myapp2.py 
# celery-3.1.11/examples/app/myapp.py /tmp
from celery import Celery

app = Celery(
    'myapp',
    broker='amqp://guest@localhost//',
    # add result backend here if needed.
    #backend='rpc'
)

@app.task()
def add(x, y):
    return x + y

if __name__ == '__main__':
    app.start()
[yamato@master]~/var/ctags-github% ./ctags --fields=+lK --fields-python='+{decorators}' -o - /tmp/myapp2.py
add	/tmp/myapp2.py	/^def add(x, y):$/;"	function	language:Python	decorators:app.task()
add	/tmp/myapp2.py	/^def add(x, y):$/;"	task	language:PythonCelery
app	/tmp/myapp2.py	/^app = Celery($/;"	variable	language:Python
```

I'm thinking about introducing high level view for a python script using python-celery.
The view is implemented as PythonCelery subparser stacked on Python parser.

In the above exanmple, the decorators field helps user understand the code but `add` is captured as a function in python level.  The subparser captures it again as a task.

masatake:
`PythonCelery` is a just example for my study. I have a chance to read a script using it.
What I would like to expolre is how ctags can handle frameworks and DSLs.

Issue #164, opened 2014-12-03
shell: allow "-" in function name (zsh, bash)


blueyed:
The shell parser does not accept `-` in function names, but this is valid for at least bash and zsh.

I think the shell parser should be made to accept `-`, too.

I do not think it makes sense to have separate/derived parsers (and therefore lang definitions) for those, but maybe the shell parser could try to detect if it's zsh/bash (by looking at the extension and shebang line), and only then apply other rules?

Source ref: https://github.com/fishman/ctags/blob/master/sh.c#L120

PR #1262, opened 2017-01-05
sh: use isBashFunctionChar as default check_char func

### LINKED ISSUE #1261: Bash functions do not need to be prefixed by `function`
I have a bash script with some functions.

    #!/usr/bin/env bash

    a::test() {
        echo "test"
    }

    function a::anotherTest() {
        echo "another test"
    }

Running `ctags` on this file produces the following.

    $ ctags -f - --guess-language-eagerly a
    a::anotherTest	a	/^function a::anotherTest() {$/;"	f
    test	a	/^a::test() {$/;"	f

I would expect the second line to say `a::test` instead of `test`.  I believe this boils down to the parser's `findShTags` function not checking if the file type is `bash` somehow.  I would suggest you change line 201 of `parsers/sh.c` from this

    check_char = isIdentChar;

to something more like this

    if (strncmp(somehowGetCurrentFileType, "bash", 4)) {
        // Not bash
        check_char = isIdentChar;
    } else {
        check_char = isBashFunctionChar;
    }

I don't know how much this would break.  Another option would be to modify the area near line 263 of `parsers/sh.c` to look backwards and change the detected function name if it can be extended backwards.

masatake:
Close #1261.

bash accepts many non isalnum characters as function identifiers even when
not using the `function` keyword.

If `function` keyword is used, ctags has a chance to switch check_char
var to a function specialized to bash. For the case `function` keyword
is not used, this commit change the default value of check_char var to
isBashFunctionChar.

### DIFF
--- a/Units/parser-sh.r/function-identifiers-no-function-keyword-bash.d/expected.tags
+++ b/Units/parser-sh.r/function-identifiers-no-function-keyword-bash.d/expected.tags
@@ -1,0 +1,2 @@
+a::anotherTest	input.bash	/^function a::anotherTest() {$/;"	f
+a::test	input.bash	/^a::test() {$/;"	f
--- a/Units/parser-sh.r/function-identifiers-no-function-keyword-bash.d/input.bash
+++ b/Units/parser-sh.r/function-identifiers-no-function-keyword-bash.d/input.bash
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+#
+# Taken from the original bug report (https://github.com/universal-ctags/ctags/issues/1261)
+#
+
+a::test() {
+    echo "test"
+}
+
+function a::anotherTest() {
+    echo "another test"
+}
--- a/parsers/sh.c
+++ b/parsers/sh.c
@@ -198,18 +198,18 @@ static void findShTags (void)
 				}
 			}
 
-			check_char = isIdentChar;
+			check_char = isBashFunctionChar;
 
 			if (strncmp ((const char*) cp, "function", (size_t) 8) == 0  &&
 				isspace ((int) cp [8]))
 			{
-				check_char = isBashFunctionChar;
 				found_kind = K_FUNCTION;
 				cp += 8;
 			}
 			else if (strncmp ((const char*) cp, "alias", (size_t) 5) == 0  &&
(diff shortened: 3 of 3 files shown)


b4n:
I'll do a proper review later but:

* I'm not sure if this might break parsing some legitimate constructs.  I don't think so, but I'm not sure
* what about specially handling bash only, e.g. with `.bash` extension or `bash` shebang?  I know there are many `sh`-labeled scripts that actually are Bash out there, but as many distributions stop linking */bin/sh* to *bash* I'm guessing it's getting less of a problem.  So, restricting Bash handling to what truly looks like Bash might be sensible.

However, if handling Bash function names everywhere doesn't cause any issue, I guess it's alright and there's no need to discriminate these cases: IMO it's alright to emit wrong tags in case of wrong input, so if it only catches invalid stuff that's fine.

masatake:
@b4n, thank you.

> what about specially handling bash only, e.g. with .bash extension or bash shebang? I know there are many sh-labeled scripts that actually are Bash out there, but as many distributions stop linking /bin/sh to bash I'm guessing it's getting less of a problem. So, restricting Bash handling to what truly looks like Bash might be sensible.

To distinguish a name is bash function or sh function the parser must accept the name from input stream first to store it to vString.  So the parser accepts the characters making bash function anyway. After accepting the parser 
can say whether it is sh function name or bash function name by inspecting each character in the vString instance. This is o.k. but I wonder how the parser represent the difference (sh or bash) in tags file. By introducing 'b,bashFunction' kind?

b4n:
@masatake No, I meant having 2 parsers from the external POV: one "bash" used for `.bash` extension and `bash` shebang, and another "sh" one for the rest.  But maybe it's too much for nothing.

I still can't think of a problem this PR would create, so maybe just go for it and we'll see.

masatake:
> @masatake No, I meant having 2 parsers from the external POV: one "bash" used for .bash extension and bash shebang, and another "sh" one for the rest. But maybe it's too much for nothing.

Oh, I see. I remember your comment in https://github.com/universal-ctags/ctags/issues/164 .
I cannot find enough merit for introducing a new parserDef for bash.

> I still can't think of a problem this PR would create, so maybe just go for it and we'll see.

I see. Thank you for taking time.

PR #1264, opened 2017-01-07
DTD: New parser

### LINKED ISSUE #996: parser for dtd
I need it in my daily work.

masatake:
Close #996.


masatake:
I have worked on this DTD parser instead of the CMake parser for testing and refining tokeninfo.[ch].

tokeinfo.[ch] are very useful. 
After writing ldscript parser and DTD parser, I would like to add `scopeIndex` field `tokenInfo` struct.
A pitfall I found is the way to handle scopeIndex field in tokenUnread function.

tokenUnread copies the value of given tokenInfo instance to an internal queue.
The value of field `scopeIndex` is also copied. Next time calling tokenRead, the scopeIndex field
of a tokenInfo passed to tokenRead is filled with the value stored to the queue.

Both in ldscript and DTD parsers, this behaviour causes troubles.  It seems that it is better that the scopeIndex field of a token should not  be touched in tokenRead/tokenUnread.


I wonder how to refect this knowledge to the API.

masatake:
Some commits are squashed.

PR #1270, opened 2017-01-13
Tokeninfo: add scope index as a filed to tokeninfo struct

masatake:
tokenInfo is really useful. Now even I can write a simple parser!

These commits reflects my experience writing ldscript and dtd parsers.

* add scopeIndex to tokenInfo,
* reset scopeIndex to CORK_NIL in tokenUnread


masatake:
To write more complicated parser, tokenInfo is not enough.

As the next step, I will study @pragmaware's pTokenChain and condensing functions in Cxx parser.
I will extract its essences to main and rewrite ruby parser.

masatake:
I cannot get convince that it is better that a tokenInfo has such context information.

PR #1271, opened 2017-01-14
JavaScript: Add very basic ES6 class support

pragmaware:
Add very basic support for simple ES6 classes.
Also add minimal tracing suport to jscript parser.

### DIFF
--- a/Units/parser-javascript.r/js-es6-class.d/input.js
+++ b/Units/parser-javascript.r/js-es6-class.d/input.js
@@ -1,0 +1,18 @@
+
+class Class1
+{
+  method1(arg1,arg2)
+  {
+  }
+  
+  method2(arg1,arg2)
+  {
+  }
+}
+
+class Class2
+{
+  method3(arg = 10)
+  {
+  }
+}
--- a/Units/parser-javascript.r/js-es6-class.d/input.js
+++ b/Units/parser-javascript.r/js-es6-class.d/input.js
@@ -1,18 +1,18 @@
 
 class Class1
 {
-  method1(arg1,arg2)
-  {
-  }
+	method1(arg1,arg2)
+	{
+	}
   
-  method2(arg1,arg2)
-  {
-  }
+	method2(arg1,arg2)
+	{
+	}
 }
 
 class Class2
 {
-  method3(arg = 10)
-  {
-  }
+	static method3(arg = 10)
+	{
+	}
 }
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -33,6 +33,7 @@
 #include "routines.h"
 #include "vstring.h"
 #include "objpool.h"
+#include "trace.h"
 
 /*
  *	 MACROS
@@ -44,6 +45,52 @@
 		(c) == '@' || (c) == '_' || (c) == '#')
 #define newToken() (objPoolGet (TokenPool))
 #define deleteToken(t) (objPoolPut (TokenPool, (t)))
+
+/*
+ * Debugging
+ *
+ * Uncomment this to enable extensive debugging to stderr in jscript code.
+ * Please note that TRACING_ENABLED should be #defined in main/trace.h
+ * for this to work.
+ *
+ */
+//#define JSCRIPT_DEBUGGING_ENABLED 1
+
+#if defined(DO_TRACING) && defined(JSCRIPT_DEBUGGING_ENABLED)
+	#define JSCRIPT_DO_DEBUGGING
+#endif
+
+#ifdef JSCRIPT_DO_DEBUGGING
+
+#define JSCRIPT_DEBUG_ENTER() TRACE_ENTER()
+#define JSCRIPT_DEBUG_LEAVE() TRACE_LEAVE()
+
+#define JSCRIPT_DEBUG_ENTER_TEXT(_szFormat,...) \
+	TRACE_ENTER_TEXT(_szFormat,## __VA_ARGS__)
+
+#define JSCRIPT_DEBUG_LEAVE_TEXT(_szFormat,...) \
+	TRACE_LEAVE_TEXT(_szFormat,## __VA_ARGS__)
+
+#define JSCRIPT_DEBUG_PRINT(_szFormat,...) \
+	TRACE_PRINT(_szFormat,## __VA_ARGS__)
+
+#define JSCRIPT_DEBUG_ASSERT(_condition,_szFormat,...) \
+	TRACE_ASSERT(_condition,_szFormat,## __VA_ARGS__)
+
+#else //!JSCRIPT_DO_DEBUGGING
+
+#define JSCRIPT_DEBUG_ENTER() do { } while(0)
+#define JSCRIPT_DEBUG_LEAVE() do { } while(0)
+
+#define JSCRIPT_DEBUG_ENTER_TEXT(_szFormat,...) do { } while(0)
+#define JSCRIPT_DEBUG_LEAVE_TEXT(_szFormat,...) do { } while(0)
+
+#define JSCRIPT_DEBUG_PRINT(_szFormat,...) do { } while(0)
+
+#define JSCRIPT_DEBUG_ASSERT(_condition,_szFormat,...) do { } while(0)
+
+#endif //!JSCRIPT_DO_DEBUGGING
+
 
 /*
  *	 DATA DECLARATIONS
@@ -77,7 +124,9 @@ enum eKeywordId {
 	KEYWORD_catch,
 	KEYWORD_finally,
 	KEYWORD_sap,
-	KEYWORD_return
+	KEYWORD_return,
+	KEYWORD_class,
+	KEYWORD_extends
 };
 typedef int keywordId; /* to allow KEYWORD_NONE */
 
@@ -170,7 +219,9 @@ static const keywordTable JsKeywordTable [] = {
 	{ "catch",		KEYWORD_catch				},
 	{ "finally",	KEYWORD_finally				},
 	{ "sap",	    KEYWORD_sap    				},
-	{ "return",		KEYWORD_return				}
+	{ "return",		KEYWORD_return				},
+	{ "class",		KEYWORD_class				},
+	{ "extends",	KEYWORD_extends				}
 };
 
 /*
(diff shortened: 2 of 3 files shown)


parsers/jscript.c
```diff
@@ -126,7 +126,8 @@ enum eKeywordId {
 	KEYWORD_sap,
 	KEYWORD_return,
 	KEYWORD_class,
-	KEYWORD_extends
+	KEYWORD_extends,
+	KEYWORD_static
```
masatake: Let's put  a comma (`,`) for the last element:
```
KEYWORD_static,
```

C language allows this.  When someone wants to add more elements in the future, the last `comma` will make the changes one line smaller. Smaller diff makes previewers happy:-)

Units/parser-javascript.r/js-es6-class.d/input.js
```diff
@@ -1,18 +1,18 @@
...
 }
 
 class Class2
 {
-  method3(arg = 10)
-  {
-  }
+	static method3(arg = 10)
+	{
+	}
```
masatake: About input files of test case, I don't care about whitespaces. 
However, I wonder why these indentation changes are needed.

I take following workflow to avoid injecting this kind of changes:

1. Do git commit
2. See the output of git show and find whether I introduced unwanted change(unnecessary change) in the last commit or not.
3. If I found one, remove it with my editor.
4. Do git add, git commit --amend. (or mark `fixup` in git rebase -i); then Goto 2.
5. Make a pull request

parsers/jscript.c
```diff
@@ -221,7 +222,8 @@ static const keywordTable JsKeywordTable [] = {
 	{ "sap",	    KEYWORD_sap    				},
 	{ "return",		KEYWORD_return				},
 	{ "class",		KEYWORD_class				},
-	{ "extends",	KEYWORD_extends				}
+	{ "extends",	KEYWORD_extends				},
+	{ "static",		KEYWORD_static				}
```
masatake: You can put `,` at the end of `static` element.

parsers/jscript.c
```diff
@@ -2005,6 +2007,9 @@ static bool parseES6Class(tokenInfo *const token,tokenInfo * const parent)
 		if(isType(token,TOKEN_CLOSE_CURLY))
 			break;
 		
+		if(isKeyword(token,KEYWORD_static))
+			readToken(token);
+
```
masatake: I don't know javascript, but I guess. like C language, `static` is important information.
So I wonder we can store it to a field of tag as you did in `property:` in Cxx.
This is just an optional idea.

parsers/jscript.c
```diff
@@ -46,6 +47,52 @@
...
+#define JSCRIPT_DEBUG_LEAVE_TEXT(_szFormat,...) do { } while(0)
+
+#define JSCRIPT_DEBUG_PRINT(_szFormat,...) do { } while(0)
+
+#define JSCRIPT_DEBUG_ASSERT(_condition,_szFormat,...) do { } while(0)
+
+#endif //!JSCRIPT_DO_DEBUGGING
+
+
+/*
```
masatake: Do you think this layer is really needed?
I think using TRACE_xxx directly is o.k.

@b4n, how do you think?

Even if we decide to use the prefix, I have a comment.
This change, adding JSCRIPT_ prefixed macros should be in a separated commit.

parsers/jscript.c
```diff
@@ -77,7 +124,9 @@ enum eKeywordId {
 	KEYWORD_catch,
 	KEYWORD_finally,
 	KEYWORD_sap,
-	KEYWORD_return
+	KEYWORD_return,
+	KEYWORD_class,
+	KEYWORD_extends
```
masatake: You can add a comma char after `KEYWORD_extends`.

parsers/jscript.c
```diff
@@ -170,7 +219,9 @@ static const keywordTable JsKeywordTable [] = {
 	{ "catch",		KEYWORD_catch				},
 	{ "finally",	KEYWORD_finally				},
 	{ "sap",	    KEYWORD_sap    				},
-	{ "return",		KEYWORD_return				}
+	{ "return",		KEYWORD_return				},
+	{ "class",		KEYWORD_class				},
+	{ "extends",	KEYWORD_extends				}
```
masatake: Ditto. You can add the comma char af the last element.

parsers/jscript.c
```diff
@@ -1666,9 +1737,10 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
 				else
 				{
 					if (! ( isType (name, TOKEN_IDENTIFIER)
-						|| isType (name, TOKEN_STRING) ) )
+						|| isType (name, TOKEN_STRING) || isType(name, TOKEN_KEYWORD) ) )
 					{
-                                                /* Unexpected input. Try to reset the parsing. */
+						/* Unexpected input. Try to reset the parsing. */
```
masatake: (Optional comment). This change can be removed from the commit.

parsers/jscript.c
pragmaware: Not sure about the additional layer. On one side it will help if we want to add a lot of tracing everywhere around the code. We will be then able to selectively enable tracing of single subsystems. On the other side it obviously adds a layer of complexity...

parsers/jscript.c
pragmaware: Sure, we could add properties.
I'm still a bit afraid of messing up too much with this parser. Need to understand it better first.

Units/parser-javascript.r/js-es6-class.d/input.js
pragmaware: I think I'll re-make a cleaner pull request with these changes in the next days.

masatake: You don't have to make new one. You can do 'git rebase -i & git push --force'.

parsers/jscript.c
```diff
@@ -1666,9 +1737,10 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
 				else
 				{
 					if (! ( isType (name, TOKEN_IDENTIFIER)
-						|| isType (name, TOKEN_STRING) ) )
+						|| isType (name, TOKEN_STRING) || isType(name, TOKEN_KEYWORD) ) )
```
b4n: style: new condition should be on the next line to follow current style

parsers/jscript.c
```diff
@@ -1877,9 +1949,99 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
...
+	// FIXME: Handle base?
+
+	JSCRIPT_DEBUG_PRINT("Emitting tag for class '%s'",vStringValue(className->string));
+
+	makeJsTag(className,JSTAG_CLASS,NULL);
+
+	if(base)
+		deleteToken(base);
+
+	while(!isType(token,TOKEN_OPEN_CURLY) && !isType(token,TOKEN_EOF))
```
b4n: AFAICT this could reject a few more things to avoid dropping too much on partial/broken input, e.g. `;`

parsers/jscript.c
```diff
@@ -1877,9 +1949,99 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
...
+	JSCRIPT_DEBUG_ENTER();
+
+	tokenInfo * className = newToken ();
+	tokenInfo * base = NULL;
+
+	readToken(className);
+	// FIXME: Check if it's an identifier?
+
+	readToken(token);
+	if(isType(token,TOKEN_KEYWORD) && (token->keyword == KEYWORD_extends))
```
b4n: style: keywords like `if`, `while`, etc. should be followed by a space.  applies all along

parsers/jscript.c
```diff
@@ -1877,9 +1949,99 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
...
+	{
+		readToken(token);
+		
+		if(isType(token,TOKEN_CLOSE_CURLY))
+			break;
+		
+		if(!isType(token,TOKEN_IDENTIFIER))
+			break;
+
+		// assume method
```
b4n: I wonder if function parsing code couldn't be slightly amended to be able to handle this?

parsers/jscript.c
```diff
@@ -1877,9 +1949,99 @@ static bool parseStatement (tokenInfo *const token, tokenInfo *const parent, boo
 	deleteToken (method_body_token);
 	vStringDelete(saveScope);
 
+	JSCRIPT_DEBUG_LEAVE();
+
 	return is_terminated;
 }
 
+static bool parseES6Class(tokenInfo *const token,tokenInfo * const parent)
```
b4n: would be useful (and probably not too hard) to support `var foo = class { … }`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Class_expressions
Especially unnamed ones that is, as they would break parsing here I think

parsers/jscript.c
b4n: @pragmaware hint: it's not very clean or robust to new changes :S

parsers/jscript.c
b4n: What about instead having a symbol a module can define before including *trace.h* that select whether the macros are enabled?  Could even add a prefix or something if we wanted.

I don't know, like
```C
#define DO_MODULE_TRACING JS_DO_DEBUGGING
#include "trace.h"
```
and in *trace.h* do something like
```C
#if defined(DO_TRACING) && (! defined (DO_MODULE_TRACING) || DO_MODULE_TRACING)
…
```
so if a file defined `DO_MODULE_TRACING` before, it conditionally enables `TRACE_…`, but leaves things not defining that as it is currently.

Though, that sounds like maybe simply
```
#if ! JS_DO_DEBUGGING
# undef DO_TRACING
#endinf
#include "trace.h"
```
would be the same.

Anyway, I somewhat agree that granularity would be nice, but also that @masatake is right and this is too verbose and should be made more generic.  As one can't declare a macro inside a macro, direct use of the `TRACE_*` calls in the modules would be needed I guess, so having the granularity in *trace.h*.

---
and agreed on the separate commit, would be nicer

b4n:
Ah and BTW, I try and link some reference in the commit/above the function when implementing new features.  Both because it's good for other too see what you tried to do when they are not too familiar with it, and because well, I'm not too familiar with it :)

b4n:
https://github.com/b4n/fishman-ctags/tree/pragmaware/es6-support (no important changes yet, just style and split)

PR #1273, opened 2017-01-15
JavaScript: Basic ES6 class support

b4n:
Based on #1271.


parsers/jscript.c
```diff
@@ -1307,24 +1381,108 @@ static bool parseMethods (tokenInfo *const token, tokenInfo *const class)
...
+		is_anonymous = false;
+	}
+	else
+	{
+		copyToken (token, className, true);
+		/* We create a fake name so we have a scope for the members */
+		if (! targetName)
+		{
+			char buf[32];
+			snprintf (buf, sizeof buf, "AnonymousClass%u", ++AnonymousID);
```
pragmaware: While working on the C++ parser I've found that it's useful to have "stable" and "non-colliding" anonymous identifiers.
That is, when parsing the same file twice the generated anonymous identifiers should be the same and when
parsing two different files the generated anonymous identifiers should (ideally) be all different. This is useful for tools that parse the source tree in parts (think of an editor that re-parses the current file while it's being edited).
We worked out a generic tool for this purpose. See usage of anonGenerate() at:
https://github.com/universal-ctags/ctags/blob/77886eb5903f4bcc60a075fbb0961b5210293fb3/parsers/cxx/cxx_token.c#L127

Units/parser-javascript.r/js-es6-mixin.d/input.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function Mixin(superclass) {
+  return class extends superclass {
+    hello() {
+      return "hello from mixin";
+    }
+  }
+}
+
```
pragmaware: Wow... the es6 mixins are tricky :)

parsers/jscript.c
b4n: oh, right.  I'll look into that

Units/parser-javascript.r/js-es6-mixin.d/input.js
b4n: Yeah got me a bit of reading to understand it -- especially when the examples also make use of arrow functions, another thing I'm not familiar with.  But the actual concept is fairly simple when you accepted the dynamic nature of it: a function that returns an anonymous class inheriting from a parameter.  It has the strange property of not being an "instance" of the mixin as the `Mixin` function returns a new class each time, but apart from that it's fairly "simple".

Units/parser-javascript.r/js-es6-mixin.d/input.js
```diff
@@ -0,0 +1,28 @@
...
+  foo() { return "foo"; }
+}
+
+class Bar extends Foo {
+  bar() { return "bar"; }
+}
+
+function Mixin(superclass) {
+  return class extends superclass {
+    hello() {
```
masatake: Just a question. `hello` method(?) in the anonymous class is not captured. Is this expected behaviour?

masatake:
@b4n, could you do `git am` following change?

```
commit a267483ac771356460075b0817323bbdf7dafd1b
Date:   Sun Jan 22 02:39:29 2017 +0900

    jscritp: use anonGenerate() to generate class name for anonymous classes
    
    Suggested by @pragmaware.

diff --git a/Units/parser-javascript.r/js-es6-class.d/expected.tags b/Units/parser-javascript.r/js-es6-class.d/expected.tags
index f66cfe0..8d560d5 100644
--- a/Units/parser-javascript.r/js-es6-class.d/expected.tags
+++ b/Units/parser-javascript.r/js-es6-class.d/expected.tags
@@ -1,4 +1,4 @@
-AnonymousClass1        input.js        /^class {$/;"   c
+AnonymousClass152c525b0101     input.js        /^class {$/;"   c
 Class1 input.js        /^class Class1$/;"      c
 Class2 input.js        /^class Class2$/;"      c
 Class3 input.js        /^var Class3 = class {$/;"      c
@@ -11,4 +11,4 @@ method4       input.js        /^      method4(){}$/;" m       class:Class3
 method5        input.js        /^      static method5(){}$/;"  m       class:Class3
 method6        input.js        /^      method6(){}$/;" m       class:Class4
 method7        input.js        /^      static method7(){}$/;"  m       class:Class4
-method8        input.js        /^      method8(n) { return n * n; }$/;"        m       class:AnonymousClass1
+method8        input.js        /^      method8(n) { return n * n; }$/;"        m       class:AnonymousClass152c525b0101
index 7fbcbe6..28631dc 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -174,8 +174,6 @@ typedef struct sTokenInfo {
 static tokenType LastTokenType;
 static tokenInfo *NextToken;
 
-static unsigned int AnonymousID;
-
 static langType Lang_js;
 
 static objPool *TokenPool = NULL;
@@ -1424,11 +1422,7 @@ static bool parseES6Class (tokenInfo *const token, const tokenInfo *const parent
                copyToken (token, className, true);
                /* We create a fake name so we have a scope for the members */
                if (! targetName)
-               {
-                       char buf[32];
-                       snprintf (buf, sizeof buf, "AnonymousClass%u", ++AnonymousID);
-                       vStringCopyS (className->string, buf);
-               }
+                       anonGenerate (className->string, "AnonymousClass", JSTAG_CLASS);
        }
 
        if (! targetName)
@@ -2200,7 +2194,7 @@ static void findJsTags (void)
 {
        tokenInfo *const token = newToken ();
 
-       AnonymousID = 0;
+       anonReset ();
        NextToken = NULL;
        ClassNames = stringListNew ();
        FunctionNames = stringListNew ();
```

masatake:
(Now appveyor works fine. What I did is that just requested building again after cancelling all building at appveyor.  It seems that there is a trouble in appveyor side.)

Units/parser-javascript.r/js-es6-mixin.d/input.js
b4n: Well, the anonymous class here is ignored altogether, because what comes after `return` is skipped.  So… no, ideally it'd be captured I guess, but it's not really a "bug", rather a limitation.  I'll look into it, now we have anonymous classes anyway, it could be mapped to something.

b4n:
> @b4n, could you do git am following change?

Sure, done.  Though, pasting diffs like this is problematic, not sure if it's GitHub's fault or the way you used to copy and paste it, but tabs were expanded to spaces, thus making the patch not apply.  Fixed it though.  Also, `git am` complained it didn't understand the format.  Maybe it's the Git 2.1 I have on this computer though, not sure.

parsers/jscript.c
b4n: Done by @masatake in 763d9df4021fbba030a3711b378bfb3829e10a83

masatake:
> Sure, done. Though, pasting diffs like this is problematic, not sure if it's GitHub's fault or the way you used to copy and paste it, but tabs were expanded to spaces, thus making the patch not apply. Fixed it though. Also, git am complained it didn't understand the format. Maybe it's the Git 2.1 I have on this computer though, not sure.

Oh, sorry. I should make a pull request for your branch.

b4n:
@masatake or you can simply push somewhere and give me a link to the commit itself so I can get the real patch from that, or push and give me the branch name so I can fetch and cherry-pick :)

masatake:
Is there any blocker for merging this?

b4n:
I think I have something left I wanted to add, I'll check tomorrow or Saturday.

masatake:
@b4n, I see. Thank you

Units/parser-javascript.r/js-es6-mixin.d/input.js
b4n: Fixed in 67637d9e96e0089775a3f4d4001cfe1b1772a63d

b4n:
@masatake Here is all I got for the moment, so it's fine for review+merge if approved.

masatake:
@b4n, I will review the changes.

Units/parser-javascript.r/js-es6-class.d/expected.tags
```diff
@@ -4,11 +4,17 @@ Class2	input.js	/^class Class2$/;"	c
...
 method1	input.js	/^	method1(arg1,arg2)$/;"	m	class:Class1
+method10	input.js	/^		method10() {$/;"	m	class:InnerClass2
 method2	input.js	/^	method2(arg1,arg2)$/;"	m	class:Class1
 method3	input.js	/^	static method3(arg = 10)$/;"	m	class:Class2
 method4	input.js	/^	method4(){}$/;"	m	class:Class3
 method5	input.js	/^	static method5(){}$/;"	m	class:Class3
 method6	input.js	/^	method6(){}$/;"	m	class:Class4
 method7	input.js	/^	static method7(){}$/;"	m	class:Class4
 method8	input.js	/^	method8(n) { return n * n; }$/;"	m	class:AnonymousClass152c525b0101
+method9	input.js	/^		method9() {$/;"	m	class:InnerClass1
```
masatake: (This is MY TODO) 
This output about method9
```
method9	input.js	/^		method9() {$/;"	m	class:InnerClass1
```
is better than

```
method9	input.js	/^		method9() {$/;"	m	class:func1.InnerClass1
```

Just specifying just one upper scope object is enough.

I may have to overhaul corkAPI. This may be related to #1280.

Units/parser-javascript.r/js-many-functions.d/expected.tags
```diff
@@ -0,0 +1,16 @@
+AnonymousFunction8797a7c50100	input.js	/^  return function(x) {$/;"	f	function:f1
+C1	input.js	/^var C1 = {$/;"	c
+K1	input.js	/^class K1 {$/;"	c
+a	input.js	/^function a() {$/;"	f
+b	input.js	/^  function b() {$/;"	f	function:a
+c	input.js	/^    function c() {$/;"	f	function:a.b
```
masatake: Following question is to myself but I would like to hear your comment.

For c, a.b is set as scope. None defines what is "scope". However, just one upper level of c is b.
So I think just printing b is enough.

I know there many inconsistency in ctags about this topic. How do you think?

Units/parser-javascript.r/js-es6-class.d/expected.tags
b4n: > Specifying just one upper scope object is enough and is the job of ctags.

No, it's not: full scope is useful for knowing the real hierarchy.  Imagine `A.Foo.b` and `B.Foo.b`, how do you discriminate both if not for their full scope?

What you see here is a bug of the JS parser, but AFAIK it's unrelated to the new changes and I was unable to fix it easily yet.

parsers/jscript.c
```diff
@@ -1410,7 +1412,11 @@ static bool parseES6Class (tokenInfo *const token, const tokenInfo *const parent
 	bool is_anonymous = true;
 	bool is_terminated = true;
 
+	copyToken (className, token, true);
 	readToken (className);
```
b4n: Is it worth it?

Units/parser-javascript.r/js-many-functions.d/expected.tags
b4n: As said above, I think full scope is the Right Thing™ to have.

Units/parser-javascript.r/js-es6-class.d/expected.tags
masatake: > magine A.Foo.b and B.Foo.b, how do you discriminate both if not for their full scope?
@b4n, you are right. I'm wrong.

I have to write this example in somewhere in docs/.
> What you see here is a bug of the JS parser, but AFAIK it's unrelated to the new changes and I was unable to fix it easily yet.

I see.

PR #1274, opened 2017-01-16
vString: Implement vStringStripLeading() in a more efficient way

b4n:
Don't move the whole string several times if there are more than one
byte to strip; instead count the bytes to strip and move accordingly in
one go.

Also, use memmove() that is highly likely to be more efficient but on
very tiny buffers.

---
**DISCLAIMER**: I did **not** do performance tests on this.  I changed it because I saw the implementation and it hurt my eyes.  I'm pretty confident it'll be a lot faster when stripping more than one byte (because the move will happen only once), or when the string is longer than *[insert some small number here]* (because `memmove()` is a *lot* faster at actually moving the bytes in any current architecture, so the question is only whether calling cost outweights the benefits).

### DIFF
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -156,13 +156,14 @@ extern void vStringStripNewline (vString *const string)
  */
 extern void vStringStripLeading (vString *const string)
 {
-	while (isspace ((int) string->buffer [0]) && string->length > 0)
+	size_t n = 0;
+
+	while (n < string->length && isspace ((int) string->buffer [n]))
+		n++;
+	if (n > 0)
 	{
-		size_t i;
-		for (i = 1  ;  i < string->length  ;  ++i)
-			string->buffer [i - 1] = string->buffer [i];
-		--string->length;
-		string->buffer [string->length] = '\0';
+		memmove (string->buffer, string->buffer + n, string->length - n);
+		vStringTruncate (string, string->length - n);
 	}
 }
 


b4n:
@techee I added you as reviewer because you're great at this kind of stuff, but feel free to decline if you don't have time for this.

codebrainz:
[Before](https://gist.github.com/codebrainz/1dc6734d796c0e85ca7a86a884123946#file-before-s) and [after](https://gist.github.com/codebrainz/1dc6734d796c0e85ca7a86a884123946#file-after-s).

b4n:
@codebrainz what does this actually mean in term of performances or otherwise? :)  I see fairly different assembly, but I'm not skilled enough at reading that to get any idea of how it actually impacts runtime :)

masatake:
I would like to know how to read Before and  after. Very interesting.
"Before" executes the first callq always. However, in some case in "after", callq is not executed.
As far as I know callq is expensive, so "after" looks better...

codebrainz:
@b4n I dunno, profile it :) I just posted since I had already generated it to see if it was wildly different since optimizers can often do magic with loops, producing similar code as memmove() does. The main observation I took from it was that it appears to inline the call to `vStringTruncate()` that you added.

@masatake I don't think here the `callq` instruction is expensive here since it's just using as a trick to save the address on the stack and going to the next instruction. It would be expensive if it was doing a full function call (setting up registers and stuff as per calling convention and then jumping into a different function).

techee:
@codebrainz LGTM.

I noticed this code myself too when doing some of the vString changes - the double loop is definitely ugly and one just wonders why anyone wrote it this way. Your code is much nicer.

The reason I didn't do anything about it was that vStringStripLeading() isn't used that much and when used, there will usually be at most 1 character that needs stripping in which case the performance will be about the same. But even though the performance difference will be minimal, your patch is the "right" way to do it.

techee:
Aha, @codebrainz should have been @b4n in my comment above :-).

b4n:
@techee yeah I don't expect much actual performance difference, as as you say the few callers most generally will strip very few bytes (generally 1 I guess).  I just did it because I felt bad having seen this code :)
Thanks for the review :)

@masatake do you know what happens to AppVeyor?  it still hasn't given any results here, yet it's been 2 days.

masatake:
> @masatake do you know what happens to AppVeyor? it still hasn't given any results here, yet it's been 2 days.

Strange. but I have no idea. I restart the jobs abd see what happens.

main/vstring.c
```diff
@@ -156,13 +156,14 @@ extern void vStringStripNewline (vString *const string)
...
+		n++;
+	if (n > 0)
 	{
-		size_t i;
-		for (i = 1  ;  i < string->length  ;  ++i)
-			string->buffer [i - 1] = string->buffer [i];
-		--string->length;
-		string->buffer [string->length] = '\0';
+		memmove (string->buffer, string->buffer + n, string->length - n);
+		vStringTruncate (string, string->length - n);
```
codebrainz: If the string is already zero-terminated, you could replace these two lines with:

```c
memmove(string->buffer, string->buffer + n, (string->length -= n) + 1);
```

To avoid the (proven to be inlined) function call and re-terminating the buffer with zero redundantly. IIRC this is what `g_strchug()` does. Maybe not really worth the hassle though.

codebrainz:
LGTM, even if it was slower it's worth it for the more clear code.

main/vstring.c
b4n: Yeah, but I didn't do that for the sake of the `DebugStatement()` in `vStringTruncate()` that 0s-out the remaining part of the allocation.  Not sure if it's of any use, but I was about to duplicated it and decided calling that should not be too much of a problem -- and you proven the compiler was clever enough :)

codebrainz: Makes sense

techee: Calling vStringTruncate() is necessary because it sets the length field of the vString structure - it has to be kept up-to-date and correspond to strlen(string->buffer) - other vString functions rely on it. One could also set string->length manually of course instead of calling vStringTruncate() but there would still have to be 2 lines of code. 

This is something where vString differs from g_str-stuff. Having length stored helps performance because when constructing tokens and reading input, parsers add the read characters to the end of the current token one by one and strlen() would have to be called all the time.

codebrainz: My single line alternative sets the length and copies one more than it (since the zero term is already in the buffer, it just `memmove`s it), though I'm not sure I'm fond of tucking an assignment side effect into a function call argument like that :) It would be clearer to put it on two lines, but it's mostly pointless since in release build most/all of the overhead of the way in this PR goes away.

techee: @codebrainz Ah, OK, I missed that - probably because I'm not fond of these single-line things :-).

PR #1276, opened 2017-01-20
buildsys: add configure options to control xml, json, and yaml support

radhermit:
This enforces explicit feature usage regardless of the libraries
available on the build system. This avoids automagic dependencies which
can cause issues for source-based distros [1].

[1]: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -586,46 +586,49 @@ AC_CHECK_FUNCS(fork waitpid execv pipe,[enable_xcmd=yes],[enable_xcmd=no])
 AM_CONDITIONAL([ENABLE_XCMD], [test "xyes" = "x$enable_xcmd"])
 
 AC_ARG_ENABLE([xml],
-	[AS_HELP_STRING([--enable-xml],
-		[enable xml support [no]])])
+	[AS_HELP_STRING([--disable-xml],
+		[disable xml support])])
 
 AH_TEMPLATE([HAVE_LIBXML],
 	[Define this value if libxml is available.])
 dnl About the condition of version
 dnl see https://mail.gnome.org/archives/xml/2010-February/msg00008.html
-AS_IF([test "${enable_xml}" = "yes"], [
+AS_IF([test "x$enable_xml" != "xno"], [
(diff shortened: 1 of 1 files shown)


masatake:
If this patch is merged, a user of ctags who does "git clone" from our repository have to specify these options for configure explicitly to use the libraries in ctags?

Do another project accept this kind of patch?

radhermit:
There, now running ./configure keeps the previous behavior so the options don't have to be explicitly specified so the automagic behavior is default. However, if they're specified they're respected.

configure.ac
```diff
@@ -585,30 +585,48 @@ AC_CHECK_FUNCS(scandir)
...
+
 AH_TEMPLATE([HAVE_LIBXML],
 	[Define this value if libxml is available.])
 dnl About the condition of version
 dnl see https://mail.gnome.org/archives/xml/2010-February/msg00008.html
-PKG_CHECK_MODULES(LIBXML, [libxml-2.0 >= 2.7.7],
-		       [have_libxml=yes
-		       AC_DEFINE(HAVE_LIBXML)],
-		       [have_libxml=no])
+AS_IF([test "${enable_xml}" = "yes"], [
```
masatake: Contributors of ctags will link more libraries to ctags in the future.
So could you add comments to configure.ac about "avoiding automagic dependencies"?
So the contributors may do the same as you did in this pull request like:

```
dnl This enforces explicit feature usage regardless of the libraries
dnl available on the build system. This avoids automagic dependencies which
dnl can cause issues for source-based distros [1].
dnl [1]: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies
```

masatake:
> There, now running ./configure keeps the previous behavior so the options don't have to be explicitly specified so the automagic behavior is default. However, if they're specified they're respected.

Thank you !

configure.ac
radhermit: I've added that comment in a new commit.

PR #1277, opened 2017-01-21
Fix sed usage in Tmain/sorted-help-message.d

blueyed:
I was seeing the following test failure when building the
universal-ctags-git AUR package:

```
Testing nolang-modeline-emacs-firstline1 as C               passed
stdout                                                      passed
stderr                                                      Testing nolang-modeline-vim0 as C                           failed (diff: /home/user/.cache/pacaur/universal-ctags-git/src/ctags/Tmain/sorted-help-message.d/stderr-diff.txt)
exit                                                        passed
```

The stderr-actual.txt file:

```
@@ -1,2 +0,0 @@
-sed: character class syntax is [[:space:]], not [:space:]
-sed: character class syntax is [[:space:]], not [:space:]
```

I've tried to reproduce it in my universal-ctags clone/checkout, but
there this test passed?!
It might be related to some special environment variable that is being
set during package builds, which makes `sed` more strict?!

I have `sed (GNU sed) 4.3` installed.

### DIFF
--- a/Tmain/sorted-help-message.d/run.sh
+++ b/Tmain/sorted-help-message.d/run.sh
@@ -27,13 +27,13 @@ extract_short_options()
 extract_long_options()
 {
     sed -n '/Usage:/,$p'  | \
-    sed -n 's/\(^  --[[:alnum:]][<>[:alnum:]_-]*\).*/\1/p'
+    sed -n 's/\(^  --[[:alnum:]][<>[[:alnum:]]_-]*\).*/\1/p'
 }
 
 extract_debug_options()
 {
     sed -n '/Usage:/,$p'  | \
-    sed -n 's/\(^  --_[<>[:alnum:]_-]*\).*/\1/p'
+    sed -n 's/\(^  --_[<>[[:alnum:]]_-]*\).*/\1/p'
 }
 
 


masatake:
```
[jet@localhost ctags]$ cat /tmp/foo.sh
sed --version
echo ---------------------------------
echo ORIGINAL:
sed -n 's/\(^  --[[:alnum:]][<>[:alnum:]_-]*\).*/\1/p' <<EOF
  --abc=N
  an example (1)
  --efg-<LANG>=PAT
  an example (2)
EOF

echo PROPOSED:
sed -n 's/\(^  --[[:alnum:]][<>[[:alnum:]]_-]*\).*/\1/p' <<EOF
  --abc=N
  an example (1)
  --efg-<LANG>=PAT
  an example (2)
EOF
[jet@localhost ctags]$ bash /tmp/foo.sh
sed (GNU sed) 4.2.2
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
and Paolo Bonzini.
GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Be sure to include the word ``sed'' somewhere in the ``Subject:'' field.
---------------------------------
ORIGINAL:
  --abc
  --efg-<LANG>
PROPOSED:
[jet@localhost ctags]$ 
```

masatake:
I don't think your fix is correct:

```
[jet@localhost ycmd]$ echo x  | sed -e 's/[[:alnum:]]/9/' 
9
[jet@localhost ycmd]$ echo x  | sed -e 's/[[[:alnum:]]]/9/' 
x
```

How about using `[a-zA-Z0-9]` instead of `[:alnum:]` ?

blueyed:
Nevermind, for some reason I was missing 0f13d82b when building the package, which is the real fix here.
(Arch's `makepkg` considered an older ref to be `origin/HEAD` when updating the package..)

PR #1279, opened 2017-01-22
tests: use --options=NONE with more tests

blueyed:
Using `--fields=+ln` in `~/.ctags` caused these tests to fail for me.

### DIFF
--- a/Tmain/interactive-mode.d/run.sh
+++ b/Tmain/interactive-mode.d/run.sh
@@ -12,6 +12,8 @@ s()
 	sed -e s/':"'/': "'/g
 }
 
+CTAGS="$CTAGS --options=NONE"
+
 echo identification message on startup
 echo =======================================
 ${CTAGS} --_interactive < /dev/null |s
--- a/Tmain/output-input-field-with-no-escape.d/run.sh
+++ b/Tmain/output-input-field-with-no-escape.d/run.sh
@@ -12,9 +12,9 @@ if ! ( echo '#define foo' > 'a\b.c' ) 2> /dev/null; then
 fi
 
 echo '# u-ctags'
-${CTAGS} --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=u-ctags -o - 'a\b.c'
+${CTAGS} --options=NONE --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=u-ctags -o - 'a\b.c'
 
 echo '# e-ctags'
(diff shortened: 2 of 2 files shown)


blueyed:
btw: I've seen a comment for the AUR package (https://aur.archlinux.org/packages/universal-ctags-git/), where a user reported the following tests to fail for them because of their config:
```
Failed tests
abnormal-output-file-names/stderr-compare
lxcmd-accept-reserved-kind/stderr-compare
lxcmd-backend-not-found/stderr-compare
lxcmd-backend-premature-death/stderr-compare
lxcmd-broken-kind/stderr-compare
lxcmd-disabled-kind/stderr-compare
lxcmd-exit-status-override/stderr-compare
lxcmd-list-kinds-full/stderr-compare
lxcmd-list-kinds/stderr-compare
lxcmd-ptag-kind-desc/stderr-compare
lxcmd-reject-invalid-notAvailableStatus/stderr-compare
lxcmd-reject-reserved-kind/stderr-compare
lxcmd-reject-specifying-in-dot-ctags/stderr-compare
xcmd-dont-run-setgid-executable/stderr-compare
xcmd-dont-run-setuid-executable/stderr-compare
xcmd-tag-relative-option/stderr-compare
 
make: *** [Makefile:6126: tmain] Error 1
``` (http://pastebin.com/4M3VUtdh)

masatake:
LGTM.

blueyed:
Given that there might still be more tests with similar issues, isn't there a better approach to use `--options=NONE` always?!

blueyed:
What about setting `HOME=/dev/null` during (most?) tests?

masatake:
In `make units`,  units script adds `--quiet --options=NONE` to command line.
However,in `make tmain`, the script doesnt' do so becase I thought it, adding the option or not should be decided by the person who writes a test case.

`tmain` was designed for testing the all areas where `units` cannot cover.  So I think I have to gave maximum flexibility to test case developers.

masatake:
Could you tell me more about HOME=/dev/null. I don't understand how useful it is.

b4n:
@masatake so it doesn't picks up user's config I guess.  But it might break more stuff not having `$HOME` pointing to something sensible, not sure.

masatake:
Oh, I see. Thank you. About Tmain, I would like to keep as is. 

"HOME=somewhere-testing-envrionment" can be used to test how ctags handles ~/.ctags well.

masatake:
Anyway @blueyed, I think your change is o.k.

blueyed:
Thanks, merged.

PR #1280, opened 2017-01-23
Rpmspec  n option in package directive

masatake:
This commit makes ctags capture `PACKAGE` in `%package -n PACKAGENAME`.
`-n` option in the line prevented ctags capturing PACKAGE.

### DIFF
--- a/parsers/rpmspec.c
+++ b/parsers/rpmspec.c
@@ -154,7 +154,7 @@ static void found_package_cb (const char *line,
 		vString *name = vStringNew ();
 		tagEntryInfo tag;
 
-		vStringNCopyS (name, line + matches[1].start, matches[1].length);
+		vStringNCopyS (name, line + matches[2].start, matches[2].length);
 		initTagEntry (&tag, vStringValue (name), RpmSpecKinds + K_PACKAGE);
 		tag.extensionFields.scopeIndex = *(int *)userData;
 		makeTagEntry (&tag);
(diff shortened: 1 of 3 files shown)


parsers/rpmspec.c
```diff
@@ -154,7 +154,7 @@ static void found_package_cb (const char *line,
 		vString *name = vStringNew ();
 		tagEntryInfo tag;
 
-		vStringNCopyS (name, line + matches[1].start, matches[1].length);
+		vStringNCopyS (name, line + matches[2].start, matches[2].length);
```
b4n: these changes should likely be in the next commit instead

b4n:
Does this make sense?  For me FQ tags should be `<scope>.<name>`, so to avoid that, one would not set scope altogether.
I'm not knowledgeable about RPM spec files, but this sounds odd to me, likely not the right solution.

parsers/rpmspec.c
masatake: You are correct. Thank you for spotting this. I will do "git push --force" after rearranging the commits.

masatake:
```
name: A
%package B
%package -n C
```

For this spec file rpmbuild command, which generates rpm package, will make following packages:

```
A.rpm
A-B.rpm
C.rpm
```

Without this commit, ctags generated following tags (--extra=+q is given):

```
A
B
A-B  (+q)
```

With this commit, ctags can generate following tags (--extra=+q is given):

```
A
B
A-B (+q)
C
```

The point is that it may be natural for people who knows rpm well that a tag A-C is not included in tags output. The first commit (fc8e0f6) is needed not to emit A-C. 

In other hand A-B is expected by the rpm users.

b4n:
I understand the output you want, but I wonder if forcing not emitting the FQ tag is the way to go.  I mean, what do you get in the `scope` field?  for your `C` you'll have `A`, right?  That doesn't seem right, does it?  If it doesn't, then the fix would be to have the scope not contain `A`, which makes the FQ tag no contain it either AFAICT.

I'm just trying to make sure you're not introducing a new feature to bypass some automatic processing without a very good reason :)

masatake:
@b4n, you may be right. `noAutomaticFQTag` introduces extra complexity.
Recording A-C as an fq tag is not a critical issue.

PR #1281, opened 2017-01-24
added support for robot framework

driechers:
Added support for robot framework keywords and variables.

Keywords in robot can contain variables for example:
```
*** Keywords ***
a keyword ${with variable in} name
    foo
    bar

*** Testcases ***
Test 1
    # The call to the keyword
    a keyword this is a variable name
```
I was unsure of the best solution so I take as much of the keyword as I can so in this case the tag would be "a keyword"
I would appreciate suggestions if there is a better solution.

### DIFF
--- a/Tmain/interactive-mode.d/run.sh
+++ b/Tmain/interactive-mode.d/run.sh
@@ -12,6 +12,8 @@ s()
 	sed -e s/':"'/': "'/g
 }
 
+CTAGS="$CTAGS --options=NONE"
+
 echo identification message on startup
 echo =======================================
 ${CTAGS} --_interactive < /dev/null |s
--- a/Tmain/output-input-field-with-no-escape.d/run.sh
+++ b/Tmain/output-input-field-with-no-escape.d/run.sh
@@ -12,9 +12,9 @@ if ! ( echo '#define foo' > 'a\b.c' ) 2> /dev/null; then
 fi
 
 echo '# u-ctags'
-${CTAGS} --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=u-ctags -o - 'a\b.c'
+${CTAGS} --options=NONE --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=u-ctags -o - 'a\b.c'
 
 echo '# e-ctags'
-${CTAGS} --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=e-ctags -o - 'a\b.c'
+${CTAGS} --options=NONE --pseudo-tags=TAG_OUTPUT_MODE --extra=+p --output-format=e-ctags -o - 'a\b.c'
 
 rm 'a\b.c'


parsers/robot.c
```diff
@@ -0,0 +1,92 @@
+/*
+ * robot.c
+ * parser for robot framewor
+ * Author - Daniel Riechers
+ * License GPL-2
+ */
+
+#include "general.h"
```
masatake: If possible, can I ask you to chnage the license to "GPL2 or laster" like:

```
 /*
*   Copyright (c) 2017, Your name here
*
*

Comment (masatake) [inline parsers/robot.c]: Could you add reference(url?) to "robot" language syntax?

File: parsers/robot.c [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+        char replaceWith = '_';
+        char toReplace = ' ';
+
+        if(strchr(s->buffer, '_'))
+        {
+            replaceWith = ' ';
+            toReplace = '_';
+        }
+
+        for(int i=0; i < s->length; i++)
```
masatake: Could you use vStringLength instead of accessing directly to length field?

parsers/robot.c
```diff
@@ -0,0 +1,92 @@
...
+static void keywords (const char *const line, const regexMatch *const matches,
+                               const unsigned int count, void *data)
+{
+    if (count > 1)
+    {
+        vString *const name = vStringNew ();
+        vStringNCopyS (name, line + matches [1].start, matches [1].length);
+        makeSimpleTag (name, RobotKinds, K_KEYWORD);
+        whitespaceSwap(name);
+        makeSimpleTag (name, RobotKinds, K_KEYWORD);
```
masatake: (@masatake's TODO:  `--extra-<LANG>` option is needed to control the emission of this extra tag.)

parsers/robot.c
```diff
@@ -0,0 +1,92 @@
...
+            if(s->buffer[i] == toReplace)
+                s->buffer[i] = replaceWith;
+}
+
+static void keywords (const char *const line, const regexMatch *const matches,
+                               const unsigned int count, void *data)
+{
+    if (count > 1)
+    {
+        vString *const name = vStringNew ();
```
masatake: `name` must be vStringDelete'ed at the end of this block.

parsers/robot.c
```diff
@@ -0,0 +1,92 @@
...
+        vStringNCopyS (name, line + matches [1].start, matches [1].length);
+        makeSimpleTag (name, RobotKinds, K_VARIABLE);
+        whitespaceSwap(name);
+        makeSimpleTag (name, RobotKinds, K_VARIABLE);
+    }
```
masatake: `name` must be vStringDelete'ed here.

parsers/robot.c
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        for(int i=0; i < s->length; i++)
+            if(s->buffer[i] == toReplace)
+                s->buffer[i] = replaceWith;
+}
+
+static void keywords (const char *const line, const regexMatch *const matches,
+                               const unsigned int count, void *data)
+{
```
masatake: The function name `keywords` is a noun. Could you change it to a verb like `tagKeywords`?

source.mak
```diff
@@ -184,6 +184,7 @@ PARSER_SRCS =				\
 	parsers/pythonloggingconfig.c	\
 	parsers/r.c			\
 	parsers/rexx.c			\
+	parsers/robot.c			\
```
masatake: In addition you may have to update win32/ctags_vs2013.vcxproj  and win32/ctags_vs2013.vcxproj.filters to build ctags on windowns.  I don't know the files well but to find where you have to do update is not difficult.

masatake:
@driechers, I need more examples to answer your question.
A test case (input.robot and expected.tags) may help us.

driechers:
@masatake Thanks for your fast and helpful feedback. I'll get to fixing these issues!

parsers/robot.c
driechers: fixed

source.mak
driechers: fixed

parsers/robot.c
```diff
@@ -42,7 +43,7 @@ static void whitespaceSwap (vString *const s)
             toReplace = '_';
         }
 
-        for(int i=0; i < vStringLength(s); i++)
+        for(i=0; i < vStringLength(s); i++)
```
codebrainz: I don't think it's necessary to avoid C99.

driechers: I thought so at first too, but the mingw build failed and pointed to this loop. So I greped through the code and saw everything else was c89, changed my code and then all the continuous integration checks passed.

codebrainz: Hmm, sounds like a bug in the CI build scripts.

masatake: I'm checking build script.

masatake:
I'm lerning robot framework.

masatake:
> I was unsure of the best solution so I take as much of the keyword as I can so in this case the tag would be >"a keyword"
> I would appreciate suggestions if there is a better solution.

Capturing whole the line may be better: `a keyword ${with variable in} name`.
This behaviour I wrote here is well aligned to what `make` file parser does:

```

[jet@localhost tmp]$ cat Makefile 
MAIN_TARGET=all

$(MAIN_TARGET):
	compiler $(input_files)

[jet@localhost tmp]$ ~/var/ctags/ctags -o - Makefile 
$(MAIN_TARGET)	Makefile	/^$(MAIN_TARGET):$/;"	t
MAIN_TARGET	Makefile	/^MAIN_TARGET=all$/;"	m
```

masatake:
Do you have any reason you don't capture test cases?

parsers/robot.c
masatake: I would like to know whitespaceSwap more.
You call whitespaceSwap then call makeSimpleTag to make a new tag.

My understanding is that you convert the representation of keyword from robot to  python, a language used in writing library level keyword. Am I correct?

driechers:
@masatake In robot test cases are syntactically the same as keywords but unlike keywords they cannot be called. So there wouldn't be a reference to a tag from which a user could jump to a test case.

masatake:
However, I think a user may want to jump to a test case from command line output of robot.
```
==============================================================================
Example test suite
==============================================================================
First test :: Possible test documentation                             | PASS |
------------------------------------------------------------------------------
Second test                                                           | FAIL |
Error message is displayed here
==============================================================================
Example test suite                                                    | FAIL |
2 critical tests, 1 passed, 1 failed
2 tests total, 1 passed, 1 failed
==============================================================================
Output:  /path/to/output.xml
Report:  /path/to/report.html
Log:     /path/to/log.html
```

In the example taken from http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html, a user may want to jump to the definition of "Second test" because it is failed.

driechers:
@masatake I'll start hacking away at tagging testcases.

parsers/robot.c
driechers: White space swap doesn't have to do with python keywords. White space swap is there because in robot spaces are the same as underscores except in the case of delimiting white space which is two or more spaces. Python keywords can in fact be called with spaces or underscores.

`Keyword ABC 123
    [Arguments]      ${arg1}      ${arg2}
    foo   ${arg1}
    bar    ${arg2}
`

The calls to the keyword above are equivalent:
`Keyword ABC 123      this is my arg1      this is my arg2` or `Keyword_ABC_123      this is my arg1      this is my arg2`

masatake:
@driechers, look at parsers/yacc.c. I think there are some hints in it.

Yacc parser changes its state when it reads "%%". You may want to change the state of your parser when it reads "*** Test Case ***" and "*** Keyword ***".

parsers/robot.c
masatake: Oh, I see.  Thank you.

Here after I write about an issue of ctags main part, not your parser. YOU DON"T HAVE TO CHANGE the parser based on this comment.
After working on more than two years, I think ctags should just record names appeared on input streams. Arranged or decorated names should be recorded only when `--extra=[+-]flags` option is given.
At least recording them should be controlled by the option. However, there is no suitable option  flags for controlling recording "white space swapped" names. After merging this parser, I will work on adding `--extra-Robot=+{whiteSpaceSwapped}` option.

masatake:
@driechers, I merged #1282 so you can drop ab81465.
Feel free to rebase your branch and do 'push --force'.

masatake:
BTW, you are using master branch for hacking. This is bad idea. I recommend you to make and use a topic branch for the robot parser.

Units/simple-robot.d/expected.tags
```diff
@@ -4,18 +4,16 @@
 !_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
 !_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
 !_TAG_PROGRAM_URL	https://ctags.io/	/official site/
-!_TAG_PROGRAM_VERSION	0.0.0	/e5be352/
+!_TAG_PROGRAM_VERSION	0.0.0	/f6247b0/
```
masatake: Pseudo tags (tags started from `!_`) should be removed from expected.tags. I should write this earlier. Sorry.

parsers/robot.c
```diff
@@ -48,24 +52,47 @@ static void whitespaceSwap (vString *const s)
                 s->buffer[i] = replaceWith;
 }
 
-static void tagKeywords (const char *const line, const regexMatch *const matches,
+static void changeSection (const char *const line, const regexMatch *const matches,
```
masatake: You can use strncasecmp declared in main/general.h.
With strncasecmp, you can avoid allocating vstring and copy the bytes.

parsers/robot.c
```diff
@@ -16,12 +16,16 @@
...
 
 typedef enum {
+        K_TESTCASE,
         K_KEYWORD,
         K_VARIABLE,
         COUNT_KIND
 } RobotKind;
 
+static RobotKind section = -1;
+
```
masatake: It seems that this variable is initialized only here.
Though it is a bit verbose, this variable should be initialized each time when new input is given to this parser.
  findRobotTags() is the good place where `section` should be initialized.

Units/simple-robot.d/input.robot
```diff
@@ -18,24 +18,23 @@ Test case 1
...
 Keyword 1
+    [Arguments]  ${arg}
     python keyword 1
 
 Keyword 2
+    [Arguments]  ${arg}
     python_keyword_2
 
 Keyword_3
+    [Arguments]  @{arg1}
```
masatake: what happens if `[Arguments]` is placed at the same line of the keyword like:

`Keyword_4 [Arguemtns] ...`

As far as I know this is valid syntax. In this case ` [Arguemtns] ...` should not be in part of a tag name.

driechers:
@masatake I'm closing this pull request and opening a new one the correct way.

PR #1282, opened 2017-01-24
main, buildsys: Specify -std=gnu99 in building on mingw

masatake:
(Ideally this should be c99.)

Inspired by https://github.com/universal-ctags/ctags/pull/1281#discussion_r97571242

See what happens.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -162,9 +162,7 @@ static void rememberMaxLengths (const size_t nameLength, const size_t lineLength
 
 static void addCommonPseudoTags (void)
 {
-	int i;
-
-	for (i = 0; i < PTAG_COUNT; i++)
+	for (int i = 0; i < PTAG_COUNT; i++)
 	{
 		if (isPtagCommonInParsers (i))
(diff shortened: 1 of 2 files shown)


codebrainz:
@masatake any idea why it wants `gnu99` instead of `c99`, and even why it needs either?

On my Linux box, I can compile this without any additional flags to GCC:

```c
#include <stdio.h>
int main() {
	for (int i=0; i < 99; i++)
		printf("i: %d\n", i);
	return 0;
}
```

Maybe it's an old version of GCC when `gnu99` wasn't the default?

k-takata:
On gcc 4.7.0, `gnu90` is the default.
https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Standards.html
I think MinGW's gcc is 4.7 (or maybe earlier).

codebrainz:
@k-takata oh it is old then. On my 16.04 Ubuntu the (i686-w64-)mingw has GCC 5.3.1.

This PR seems reasonable then, to make older Mingw GCC behave like more recent versions.

PR #1283, opened 2017-01-25
Revise json output

### LINKED ISSUE #1212: json writer may not need escaping in ctags level
```
$ (cd Units/parser-php.r/php-full-qualified-tags.d; ../../../ctags --extra=+q -o - --output-format=json ./input.php)
{"_type": "tag", "name": "foo\\bar::__construct",... "scope": "foo\\\\bar", "scopeKind": "class"}
```

To many back slashes are added to scope field.

masatake:
Close #1212.

@tmm1, could you try this change?
"line" field is now represented in an integer.
In addition I extend `--list-fields` option. I belive you may like the introspection feature.

The way to handle FIELDTYPE_INTEGER is ugly; the value is converted from an integer to a string in common layer and then the string is converted to an integer back in json-writer level. Providing the way to accessing directly from writer-json to field data structure is wanted.  The way to handle FIELDTYPE_BOOLEAN is also ugly.  However, these are internal issue. This change provides more stable json output *format* to client tools.


tmm1:
:+1: very nice

masatake:
I have to add a test case that emits a boolean.

PR #1292, opened 2017-01-31
Call anonReset in main part

### LINKED ISSUE #1288: anonReset should be called from main side
anonReset function is exported from parse.c. The original intent is that the function should be called from each parser which uses anonGenerate.

However, developers including I may forget this. This should be done in main side.

masatake:
Close #1288.

### DIFF
--- a/Units/parser-yacc.r/c-anon-ids.d/args.ctags
+++ b/Units/parser-yacc.r/c-anon-ids.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=no
+--extra=+s
+--fields=-t
+
--- a/main/parse.c
+++ b/main/parse.c
@@ -2747,7 +2747,7 @@ extern bool makeKindDescriptionsPseudoTags (const langType language,
 extern void anonReset (void)
 {
 	parserDefinition* lang = LanguageTable [getInputLanguage ()];
-	lang -> anonumousIdentiferId = 0;
+	lang -> anonymousIdentiferId = 0;
 }
 
 static unsigned int anonHash(const unsigned char *str)
@@ -2764,14 +2764,14 @@ static unsigned int anonHash(const unsigned char *str)
 extern void anonGenerate (vString *buffer, const char *prefix, int kind)
 {
 	parserDefinition* lang = LanguageTable [getInputLanguage ()];
-	lang -> anonumousIdentiferId ++;
+	lang -> anonymousIdentiferId ++;
 
 	char szNum[32];
 
 	vStringCopyS(buffer, prefix);
 
 	unsigned int uHash = anonHash((const unsigned char *)getInputFileName());
-	sprintf(szNum,"%08x%02x%02x",uHash,lang -> anonumousIdentiferId, kind);
+	sprintf(szNum,"%08x%02x%02x",uHash,lang -> anonymousIdentiferId, kind);
 	vStringCatS(buffer,szNum);
 }
 
--- a/main/parse.c
+++ b/main/parse.c
@@ -55,8 +55,6 @@ typedef struct {
 	enum specType specType;
 }  parserCandidate;
 
-static ptrArray *parsersUsedInCurrentInput;
-
 /*
  * FUNCTION PROTOTYPES
  */
@@ -66,7 +64,8 @@ static void installKeywordTable (const langType language);
 static void installTagRegexTable (const langType language);
 static void installTagXpathTable (const langType language);
 static void anonResetMaybe (parserDefinition *lang);
-static void clearParsersUsedInCurrentInput (void);
+static void setupAnon (void);
+static void teardownAnon (void);
 
 /*
 *   DATA DEFINITIONS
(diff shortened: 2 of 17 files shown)


masatake:
Added a missing test case.

masatake:
Updated: the original code is too inefficient.

main/parse.h
```diff
@@ -125,7 +125,9 @@ struct sParserDefinition {
 	stringList* currentPatterns;   /* current list of file name patterns */
 	stringList* currentExtensions; /* current list of extensions */
 	stringList* currentAliases;    /* current list of aliases */
-	unsigned int anonumousIdentiferId; /* managed by anon* functions */
+	unsigned int anonymousIdentiferId; /* managed by anon* functions */
+	parserDefinition *parsersAnonReset; /* list of parsers which anonymousIdentiferId
+					       is set for the current input. */
```
b4n: that's not very clear IMO that it's a `next` linked list pointer

b4n: BTW, does it make sense to implement it like that?  It seems a bit overcomplicated to me, couldn't we just have an array indexed on lang?

parsers/yacc.c
```diff
@@ -52,12 +54,20 @@ static  void change_section (const char *line CTAGS_ATTR_UNUSED,
...
-		while (readLineFromInputFile ())
-			;
+		endCharOffset = 0;
+		last = NULL;
+		while ((tmp = readLineFromInputFile ()))
+			last = tmp;
+		if (last)
+			endCharOffset = strlen ((const char *)last);
+		/* If `last' is too long, strlen returns a too large value
+		   for the positive area of `endCharOffset'. */
```
masatake: endCharOffset is passed to `mio_seek` as the second argument:
```
int mio_seek (MIO *mio, long offset, int whence);
```

Though it is long, not int, but mio_seek takes the signed value anyway.
So a cast operation from size_to to long is needed. Do you think doing it here is wrong?

I cannot image `last` has such length, but theoretical it has.

main/parse.h
masatake: You are right. I should use ptrArray.

masatake:
I pushed the implementation using ptrArray.

parsers/yacc.c
b4n: Oh.  OK well no then I guess it's fine, disregard that comment.  And yeah that FILE API isn't the best ever, but well.

Units/parser-yacc.r/c-anon-ids.d/args.ctags
```diff
@@ -0,0 +1,3 @@
+--sort=no
+--extra=+s
+--fields=-t
```
b4n: maybe add a newline at the end

main/ptrarray.c
```diff
@@ -125,6 +125,16 @@ extern bool ptrArrayHasTest (const ptrArray *const current,
 	return result;
 }
 
+static bool ptrEq (const void *ptr, void *userData)
+{
+	return !!(ptr == userData);
```
b4n: `==` yields a value that is perfectly fine as a `bool`, why the `!!`?

parsers/yacc.c
b4n: I guess using `long` would be slightly better, but not very important then

main/parse.c
```diff
@@ -54,6 +55,8 @@ typedef struct {
 	enum specType specType;
 }  parserCandidate;
 
+static ptrArray *parsersUsedInCurrentInput;
```
b4n: as the initial value is used, please initialize explicitly to `NULL` so it's clear what happens.

b4n: Hum, actually it's probably better if the array is created and destroyed explicitly, isn't it?  Right now it would be found as a `still reachable` leak in Valgrind, and IIRC you tried to get rid of those in the parsers, so I guess the same applies to the main part, doesn't it?

parsers/yacc.c
masatake: I will fix this.

main/ptrarray.c
masatake: I'm wrong about this.

Units/parser-yacc.r/c-anon-ids.d/args.ctags
masatake: I wil fix this. Thank you.

main/parse.c
masatake: About setting NULL, static variables are set to zero, aren't they?
This may be about style but in my experience marking static is enough.

> 

Hum, actually it's probably better if the array is created and destroyed explicitly, isn't it? Right now it would be found as a still reachable leak in Valgrind, and IIRC you tried to get rid of those in the parsers, so I guess the same applies to the main part, doesn't it?

I have made too many `still reachable` in main part. Implementing finalizer or them and calling the finalizer  at the end of main() is not small change.

As you wrote I added such finalizer to parsers. However, as far as I can remember, I worked on eliminating  UNreachable leak.

b4n: > This may be about style but in my experience marking static is enough.

yes a static variable has defined initialization rules, and here it's equivalent ton setting it to `NULL`  my remark is purely style, to make clear you expect this default value.

> However, as far as I can remember, I worked on eliminating UNreachable leak.

I remember you removing this kind of static variables holding "cache" value.  But anyway, it's no biggie, and if the main part has many it won't change too much, so it's fine.

b4n:
@masatake oops, I see that me merging #1273 breaks this PR (as the former adds an `anonReset()` call)

PR #1298, opened 2017-02-06
main: use long type to represent byte offsets of area in promise API

masatake:
mio_seed used behind promise API expects long values.


masatake:
> One detail, there's a typo in the commit message, it's mio_seek(), not _seed().

Thank you. Fixed.

Issue #1299, opened 2017-02-07
namespace as structure member name throws C parser off


vladak:
**Test file:**

```
#ifdef	__cplusplus
extern "C" {
#endif

typedef struct foo_s {
	char *foo;
	char *namespace;
} foo_t;

typedef struct bar_s {
	char *bar;
} bar_t;

#ifdef	__cplusplus
}
#endif
```

**ctags version**:

```
$ ./ctags --version
Universal Ctags 0.0.0(1d3dbc8), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Feb  7 2017, 13:35:30
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +option-directory, +coproc, +xpath, +json, +interactive
```

**actual result from universal ctags**

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	/1d3dbc8/
foo	foo.h	/^	char *foo;$/;"	m	struct:foo_s	typeref:typename:char *
foo_s	foo.h	/^typedef struct foo_s {$/;"	s
```

**exuberant ctags**

```
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Apr 11 2012, 01:48:44
  Addresses: <dhiebert@users.sourceforge.net>, http://ctags.sourceforge.net
  Optional compiled features: +wildcards, +regex
```

**actual result from exuberant ctags**

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR	Darren Hiebert	/dhiebert@users.sourceforge.net/
[...]

PR #1302, opened 2017-02-08
[FYI] getting raw string listeral from cpreprocessor

masatake:
Related to #1299.

Information about raw string literal is needed to recognize  "C" in 

```
extern "C"
```
because there can be `extern "C++"`.
This commit adds cppGetFull for the purpose.

This commit doesn't stand alone. This can be included in the future solution for #1299.


masatake:
I made a mistake.

PR #1304, opened 2017-02-11
CXX: Improve handling of C++ keywords when C++ language isn't confirmed

### LINKED ISSUE #1299: namespace as structure member name throws C parser off

pragmaware:
Try to handle C++ keywords as identifiers when they are found in unexpected places and C++ language isn't confirmed yet (as happens in headers).

Fixes #1299 .

### DIFF
--- a/parsers/cxx/cxx_keyword.h
+++ b/parsers/cxx/cxx_keyword.h
@@ -129,12 +129,56 @@ bool cxxKeywordIsConstant(CXXKeyword eKeywordId);
 bool cxxKeywordMayBePartOfTypeName(CXXKeyword eKeywordId);
 bool cxxKeywordIsTypeRefMarker(CXXKeyword eKeywordId);
 bool cxxKeywordExcludeFromTypeNames(CXXKeyword eKeywordId);
+bool cxxKeywordIsCPPSpecific(CXXKeyword eKeywordId);
+
 
 const char * cxxKeywordName(CXXKeyword eKeywordId);
 
 // uLanguage is really CXXLanguage, but we keep it as unsigned int to avoid
 // problems with header inclusions. It works anyway.
 void cxxBuildKeywordHash(const langType eLangType,unsigned int uLanguage);
+
+// Keyword enabled/disabled state management.
+//
+// public, protected, private, class, namespace... keywords are C++ only.
+// However when parsing .h files we don't know if they belong to a C program or C++
+// one and thus for safety we parse them as C++. If our guess is wrong then the parser
+// may become confused and in some cases even bail out.
+//
+// For this reason we enable/disable the processing of certain keyword sets
+// in certain contexts.
+
+
+//
+// "public,protected,private" keywords
+//
+// In header files we disable processing of such keywords until we either figure
+// out that the file really contains C++ or we start parsing a struct/union.
+//
+// This flag is meaningful only when parsing a .h file as C++ since in C
+// public/protected/private are never keywords and we assume that .cpp files
+// have C++ content (so public/protected/private are always keywords).
+//
+// This function returns the previous state of the public/protected/private keywords
+// enabled flag so it can be easily restored.
+bool cxxKeywordEnablePublicProtectedPrivate(bool bEnableIt);
+
+//
+// "namespace"
+//
+// This is not valid inside a class/struct/union
+//
+bool cxxKeywordEnableNamespace(bool bEnableIt);
+
+//
+// "final" keyword
+//
+// This is actually special at C++ level: it's a keyword only within a specific part
+// of a class declaration. In other contexts it's not a keyword.
+void cxxKeywordEnableFinal(bool bEnableIt);
+
+// Is the specific keyword currently disabled?
+bool cxxKeywordIsDisabled(CXXKeyword eKeywordId);
 
 
 #endif //!ctags_cxx_keyword_h_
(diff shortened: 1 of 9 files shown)


parsers/cxx/cxx_parser_typedef.c
```diff
@@ -281,8 +282,25 @@ void cxxParserExtractTypedef(
 
 		if(!t)
 		{
-			CXX_DEBUG_LEAVE_TEXT("Didn't find an identifier: something nasty is going on");
-			return;
+			// Not found yet.
+			// If we're in C++ mode but we haven't confirmed that the language is really C++
```
b4n: I don't see the "confirmed" check in the conditional below, is that OK?

masatake:
```
[jet@localhost ctags]$ grep cxxKeywordEnableNamespace parsers/cxx/*.[ch]
parsers/cxx/cxx_keyword.h:bool cxxKeywordEnableNamespace(bool bEnableIt);
```

Implementation for cxxKeywordEnableNamespace is not found. Am I wrong?

masatake:
Following keyword are used in .h file of Linux kernel, which I have to read daily:
* virutla
* class
* wchar_t
* new
* private
* export
* this
* operator
* namespace
* protected
* delete

How about introducing following more generic function:
```
void cxxKeywordEnable(CXXKeyword eKeyword, bool bEnableIt)
{
       if(bEnableIt)
               g_aCXXKeywordTable[eKeyword].uFlags &= ~CXXKeywordIsDisabled;
       else
               g_aCXXKeywordTable[eKeyword].uFlags |= CXXKeywordIsDisabled;
}
```

b4n:
> * wchar_t

Why is that considered as a keyword by the parser?  OK, it's technically one (why?), but I can't see a good reason why it would be actually useful to distinguish from a custom/library type.  Actually, same applies to all type-keywords, like `bool`, `char`, etc., the only reason I could see would be help recognizing types in some situations, but that can't be relied on as custom types exist.  But maybe I miss something?

parsers/cxx/cxx_keyword.h
```diff
@@ -129,12 +129,56 @@ bool cxxKeywordIsConstant(CXXKeyword eKeywordId);
...
+// This function returns the previous state of the public/protected/private keywords
+// enabled flag so it can be easily restored.
+bool cxxKeywordEnablePublicProtectedPrivate(bool bEnableIt);
+
+//
+// "namespace"
+//
+// This is not valid inside a class/struct/union
+//
+bool cxxKeywordEnableNamespace(bool bEnableIt);
```
b4n: spotted by @masatake: this is unused, and the function isn't defined.

b4n:
@masatake Good catch.  BTW, I can't resist mentioning `git grep` to you in case you don't know about it, it's great ;)

masatake:
> @masatake Good catch. BTW, I can't resist mentioning git grep to you in case you don't know about it, it's great ;)

Thank you. I should use it.

parsers/cxx/cxx_keyword.h
pragmaware: Right.

parsers/cxx/cxx_parser_typedef.c
pragmaware: Ops... great catch!

pragmaware:
@masatake: yes, the initial idea was to be able to enable/disable single keywords or groups, so I've implemented a generic method (instead of the old specific booleans). However for now I've found real usage only fro two groups (public/protected/private and final). Maybe we'll find usage for more in the future.

@b4n: Knowing that a particular token is a (part of) type is useful to the parser at least in a couple of places, as far as I remember. When the information is certain, we should use it.

masatake:
> @masatake: yes, the initial idea was to be able to enable/disable single keywords or groups, so I've implemented a generic method (instead of the old specific booleans). However for now I've found real usage only fro two groups (public/protected/private and final). Maybe we'll find usage for more in the future.

I see. LGTM.

PR #1313, opened 2017-03-02
Cxx: handle struct var with initializer

### LINKED ISSUE #1311: In C, ctags misses local variable defined as an anonymous structure
Hello,

I'm using ctags the first time. Great tool!

At the moment I'm observing that ctags does not list a local variable if it is defined as an anonymous struct. Example source:

```
int
myfunction(int a, int b, int c)
{
        int local_x;
        struct {
                int h;
                char *j;
        } mystruct = { 3, "XXX" };
        int i;

        for (i=0; i<10; i++) {
                local_x = i;
        }
        return local_x;
}
```

When I now run "ctags -u -x --c-kinds=fl testfile.c", I get:

```
myfunction       function      2 testfile.c       myfunction(int a, int b, int c)
local_x          local         4 testfile.c       int local_x;
i                local         9 testfile.c       int i;
```

And with "ctags -u -x --c-kinds=fls testfile.c" the anonymous structure is also listed. But still the variable "mystruct" is missing in the list:

```
myfunction       function      2 testfile.c       myfunction(int a, int b, int c)
local_x          local         4 testfile.c       int local_x;
__anonf81d8a160108 struct        5 testfile.c       struct {
i                local         9 testfile.c       int i;
```

Am I doing something wrong?

Thank you in advance for your help.

Best regards,
Bodo

masatake:
Close #1311.

Though s in

    struct {
      int h;
    } s ;

was tagged well,

s in

    struct {
      int h;
    } s = { 1 };

was not.

This commit makes ctags recognize `=' (CXXTokenTypeAssignment)
as the end marker of struct variable; an expression after the
marker is just skipped.

### DIFF
--- a/Units/parser-c.r/c-struct-var-with-initializer.d/expected.tags
+++ b/Units/parser-c.r/c-struct-var-with-initializer.d/expected.tags
@@ -1,0 +1,10 @@
+__anon61565e570108	input.c	/^struct {$/;"	struct	file:
+a	input.c	/^  int a;$/;"	member	struct:__anon61565e570108	typeref:typename:int	file:
+b	input.c	/^} b = { 1 };$/;"	variable	typeref:struct:__anon61565e570108
+c	input.c	/^int c;$/;"	variable	typeref:typename:int
+d	input.c	/^int d(int e) {$/;"	function	typeref:typename:int
+e	input.c	/^int d(int e) {$/;"	parameter	function:d	typeref:typename:int	file:
+__anon61565e570208	input.c	/^  struct {$/;"	struct	function:d	file:
+f	input.c	/^    int f;$/;"	member	struct:d::__anon61565e570208	typeref:typename:int	file:
+g	input.c	/^  } g = { 1 };$/;"	local	function:d	typeref:struct:d::__anon61565e570208	file:
+h	input.c	/^  int h;$/;"	local	function:d	typeref:typename:int	file:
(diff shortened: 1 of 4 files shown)


Units/parser-c.r/c-struct-var-with-initializer.d/expected.tags
```diff
@@ -0,0 +1,10 @@
+__anon31494def0108	input.c	/^struct {$/;"	struct	file:
+a	input.c	/^  int a;$/;"	member	struct:__anon31494def0108	typeref:typename:int	file:
```
pragmaware: These anonymous identifiers come out different in my tests and the units fails. I wonder why...as the CMDLINE path seems to be relative?

parsers/cxx/cxx_parser.c
```diff
@@ -463,8 +464,24 @@ static bool cxxParserParseEnumStructClassOrUnionFullDeclarationTrailer(
...
+		   if we are at --------^.
+		*/
+		if(cxxTokenTypeIs(g_cxx.pToken,CXXTokenTypeAssignment) &&
+		   (!cxxParserParseUpToOneOf(CXXTokenTypeEOF | CXXTokenTypeSemicolon)))
+		{
+			CXX_DEBUG_LEAVE_TEXT("Failed to parse up to EOF/semicolon");
+			return false;
+		}
+	}
+
```
pragmaware: CXXTokenTypeAssignment should be probably handled at all exit points.
If Typedef has been seen, its an error (unless I'm missing something), but we might just be tolerant and handle it anyway.

masatake: I see.

masatake:
Updated.

BodoStroesser:
I tested the patch. For me it works fine.

Thank you!

PR #1314, opened 2017-03-03
CXX: K&R-style functions with enum/struct/union variables in the signature

pragmaware:
Handle K&R-style functions when they have enum/struct/union variables in the signature.
Also refactor a little bit to simplify handling of K&R functions in general.
Based on masatake's patch "fix handling struct keyword in k&r parameter decls".

### DIFF
--- a/Units/parser-cxx.r/k-and-r.d/args.ctags
+++ b/Units/parser-cxx.r/k-and-r.d/args.ctags
@@ -1,0 +1,2 @@
+--kinds-c=*
+--sort=no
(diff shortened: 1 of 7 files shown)


pragmaware:
It looks that coveralls complains about code paths that handle failure cases. We don't have them as unit tests...

masatake:
I'm running chop test. A CPU of my computer may burn:-P.
I will report the result tomorrow.

masatake:
All tests are passed.

masatake:
LGTM.

PR #1318, opened 2017-03-08
[SELF INCOMPATIBLE] main: rename "subparser" field to "guest"

masatake:
Finally I recognize there are two sets of concepts "guest/host" and "base/sub" in
running multiple parsers on single input file.

"guest/host" parser can be formed with promies API. This is stable. 
I'm working on another one: making API for "base/sub' now.

Before introducing the API for "base/sub", the concept level
separation for guest/host and base/sub is needed; guest/host relation
is different form base/sub relation. 

As the first step for the separation, "subparser" extra that was for
controlling running guest parser is renamed to "guest" in this commits.


masatake:
(Typos in commit log must be fixed.)

PR #1323, opened 2017-03-21
Clarifying where rst2man comes from

sunjay:
Without this, the man page isn't built automatically and it is unclear why unless you happen to see this.


masatake:
Thank you.

PR #1332, opened 2017-03-29
Remove redundant gcc-attr.h vcxproj entry

rnk:
Fixes loading the solution in VS 2015 for me.


masatake:
I don't understand why this change is needed. Could you tell me more about this?

rnk:
Otherwise VS says this:
C:\src\ctags\win32\ctags_vs2013.vcxproj : error  : The item "..\main\gcc-attr.h" already exists under the filter "Header Files".

masatake:
Oh, I see. You wrote "redundant" in the title of this PR. I should read more carufully.
Thank you.

Issue #1329, opened 2017-03-29
C: typedef'ed function pointer type returning struct type doesn't captured well


masatake:
```
struct x {
  int y;
};

typedef struct x (f) (int z);
```

`f` is not captured.

masatake:
```
int (f) (int z);
```
In this case `f` captured wel..

pragmaware:
I think this has been fixed some time ago.

PR #1338, opened 2017-04-06
CXX: Handle 'typedef struct X (Y)()' and similar forms

pragmaware:
Fixes bug #1329


masatake:
I'm suprised that the fix is so short.
LGTM. Thank you.

PR #1345, opened 2017-04-14
C improve c++ scope keywords handling in c

masatake:
C: deal with C++ scope keywowrds as identifiers

In *.h file if [: \t] don't follows after protect|private|public,
deal with the token as an identifier, not a keywowrd.

TODO: The language for each tag entry is still C++. It should be C.

### DIFF
--- a/Units/parser-c.r/cxx-keywords-simple.d/args.ctags
+++ b/Units/parser-c.r/cxx-keywords-simple.d/args.ctags
@@ -1,0 +1,3 @@
+--sort=no
+--kinds-C=+p
+--fields=+K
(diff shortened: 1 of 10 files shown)


codebrainz:
Does it handle these?

```c++
struct Foo {
private/*foo*/:
};
```

```c
struct Foo {
int private       ;
};
```

```c
void foo(void) {
  goto private;
private:
  return;
}
```

pragmaware:
I think this is a good idea.

For the @codebrainz comments:

Case1: comments should be stripped out by the preprocessor, so this should work (in theory).
Case 2: we might require the current token chain to be empty? (might have counterexamples?)
Case 3: we might require the current scope to be != function?

masatake:
Thank you for  the comments.
The change is just a workaround; these things should be done in upper layer.
Anyway I will try to improve my change.

masatake:
All cases suggesetd by @codebrainz are added to Units.

Though I got hits from @pragmaware , I implemented the iead in lower level. Using higher level knowledges in the tokenizer may be not good idea.

PR #1352, opened 2017-04-21
main: don't use a temporary file if possible

### LINKED ISSUE #1350: avoid mktemp for -f- mode
I am currently running ctags inside a user namespace sandbox on Linux for security reasons. This is a little fidgety; one source of problems is that the -f -  (stdout output)  creates a temporary file, which requires correct setup of /tmp/ in the sandbox.

It would be great if that could be avoided. I tried using mio_new_memory() to replace the tempFile call, but somehow it eats all newlines. Alternatively, (preferred) the writes could go directly to the stdout filedescriptor. I can sort the output data myself.

masatake:
Close #1350.

In both -o - and --sort=no is given, ctags should be able to
run without a temorary file.

### DIFF
--- a/Tmain/interactive-resource-management.d/input.c
+++ b/Tmain/interactive-resource-management.d/input.c
@@ -1,0 +1,4 @@
+int
+main(void)
+{
+}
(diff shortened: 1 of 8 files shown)


masatake:
...looks strange. I have to inspect the failures more.

masatake:
Let's see what happens.

PR #1363, opened 2017-04-27
Cxx capture array var in parentheses

### LINKED ISSUE #1359: Problem with parentesized arrays in C
Reduced test case:
```c
int *(foo[]);
```

masatake:
Capture foo in int (foo[2]);

Close #1359.


masatake:
Updated: allow `type (var);` and `type *(var);`.

masatake:
I will rework on this.

masatake:
@pragmaware, could you look at this?

Issue #1361, opened 2017-04-25
Appveyor builds don't package debugging symbols


jtanx:
ctags is crashing. It would be really convenient if the debugging symbols were provided - DWARF or PDB, I don't mind, so I don't have to compile this myself.

jtanx:
So I decided to compile it myself - on mingw it segfaults, but it doesn't produce a meaningful stack trace

```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x74b032f6 in SetFilePointer () from C:\WINDOWS\System32\KernelBase.dll
(gdb) bt
#0  0x74b032f6 in SetFilePointer () from C:\WINDOWS\System32\KernelBase.dll
#1  0x7671b478 in msvcrt!_lseeki64 () from C:\WINDOWS\System32\msvcrt.dll
#2  0x00000168 in ?? ()
#3  0x00000000 in ?? ()
(gdb) q
```

On MSVC2015 it's failing on an assertion with `isspace` complaining that the input isn't in the range of -1 to 255 in `vStringStripTrailing` - from the [docs](https://msdn.microsoft.com/en-us/library/y13z34da.aspx):

> The behavior of isspace and _isspace_l is undefined if c is not EOF or in the range 0 through 0xFF, inclusive

jtanx:
It's a stack overflow parsing this HTML file: 
[History9.zip](https://github.com/universal-ctags/ctags/files/954136/History9.zip)

It's from the ghostscript 9.21 docs

masatake:
Thank you for reporting. 
You report two things (1) build configuration on windows and (2) a bug of html parser.
(2) is important, isn't it?

masatake:
Very strange bug.

masatake:
@techee, can I ask you to look at this issue?

I guess too many recursive unbalanced html tags cause the ctags crashing.

techee:
The problem is that right now the parser doesn't recognize void elements

https://www.w3.org/TR/html51/syntax.html#void-elements

and recurses for each of them while it shouldn't. I was aware of this but it didn't cause any problems when generating tags with normal html files so I ignored it. But it's true that with too many void elements it can really cause a stack overflow.

I'll prepare a fix.

jtanx:
Thanks for fixing the crash, but the original issue wrt. appveyor still stands, so can this please be re-opened?

masatake:
It may be better not to change the configuration.
[...]

PR #1364, opened 2017-04-28
Fix HTML parser crash caused by too many open tags

techee:
See the comments in the individual patches for more info. Each of the patches applied separately fixes the crash reported in #1361 for me.

### DIFF
--- a/parsers/html.c
+++ b/parsers/html.c
@@ -20,6 +20,11 @@
 #include "keyword.h"
 #include "promise.h"
 
+/* The max. number of nested elements - prevents further recursion if the limit
+ * is exceeded and avoids stack overflow for invalid input containing too many
+ * open tags */
+#define MAX_DEPTH 1000
+
 
 typedef enum {
 	K_ANCHOR,
@@ -115,7 +120,7 @@ typedef struct {
 static int Lang_html;
 
 
-static void readTag (tokenInfo *token, vString *text);
+static void readTag (tokenInfo *token, vString *text, int depth);
 
 
 static void readTokenText (tokenInfo *const token, bool collectText)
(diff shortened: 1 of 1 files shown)


parsers/html.c
```diff
@@ -20,6 +20,8 @@
 #include "keyword.h"
 #include "promise.h"
 
+#define MAX_DEPTH 1000
```
masatake: Can I ask you to use a bit more long and descriptive name (or putting comments about the constant)?

techee: Yeah, good point, I added a comment explaining the purpose of the constant.

masatake: AFTER merging this change, can I change this to HTML_MAX_DEPTH ?

Ctags has `--maxdepth` option. To avoid confusion, I would like to add a prefix.

```
[yamato@master]~% u-ctags --help | grep max
  --maxdepth=N
       Specify maximum recursion depth.
```

masatake:
I wonder whether adding a new parameter maxdepth to HTML parser is good or not.

```
$ u-ctags --param-HTML=maxdepth:3000 ...
```
http://docs.ctags.io/en/latest/news.html#parser-own-parameter

parsers/html.c
techee: > Thank you for adding the comment.
> AFTER merging this change, can I change this to HTML_MAX_DEPTH ?

Yeah, sure.

techee:
> I wonder whether adding a new parameter maxdepth to HTML parser is good or not.

I'm not sure if it's not an overkill - I don't think anyone will ever need more than 1000 nested HTML elements and don't think it needs to be configurable. Also 1000 was just "some high number" but I think even 10000 would be fine and still won't lead to stack overflow.

masatake:
LGTM. Please, merge this. Thank you.

PR #1366, opened 2017-04-29
main,PHP: use '<?php' at the first line to detect the language

masatake:


masatake:
... parser guessing should be more extensible.

PR #1367, opened 2017-04-30
Change EditorConfig setting for win32 files

k-takata:
Files under the win32 directory should have CRLF EOL.


k-takata:
`*.vcxproj` and `*.vcxproj.filters` normally doesn't have final newline, so the following might be needed for those files:

```
insert_final_newline = false
```

However, final newlines were added by the commit 5c7fc943de0988e4ddd03caf7e2dd330518ec799. So I don't add the setting this time.

PR #1380, opened 2017-05-13
Cxx: reduce blocks when parsing values for array/union initialization

### LINKED ISSUE #1378: Large C array definition blows up ctags memory use
As of c603db3dcd, the following occurs:

```
[dmitri@ubuntu-dmitri ctags]$ wc -l t.c
10005 t.c
[dmitri@ubuntu-dmitri ctags]$ ctags -fx t.c
ctags: cannot sort tag file : Cannot allocate memory
[dmitri@ubuntu-dmitri ctags]$
```

Stock `ctags` works fine:

```
[dmitri@ubuntu-dmitri ctags]$ /usr/bin/ctags -fx t.c

[dmitri@ubuntu-dmitri ctags]$ cat x
!_TAG_FILE_FORMAT       2       /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED       1       /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Darren Hiebert  /dhiebert@users.sourceforge.net/
!_TAG_PROGRAM_NAME      Exuberant Ctags //
!_TAG_PROGRAM_URL       http://ctags.sourceforge.net    /official site/
!_TAG_PROGRAM_VERSION   5.9~svn20110310 //
IOV     t.c     1;"     d       file:
myarray t.c     /^static struct mystruct myarray[] =$/;"        v       typeref:struct:mystruct file:
[dmitri@ubuntu-dmitri ctags]$
```

I am attaching t.c: 
[t.c.txt](https://github.com/universal-ctags/ctags/files/991679/t.c.txt)

masatake:
Close #1378.

### DIFF
--- a/parsers/cxx/cxx_parser.c
+++ b/parsers/cxx/cxx_parser.c
@@ -66,7 +66,8 @@ void cxxParserNewStatement(void)
 //
 bool cxxParserParseAndCondenseCurrentSubchain(
 		unsigned int uInitialSubchainMarkerTypes,
-		bool bAcceptEOF
+		bool bAcceptEOF,
+		bool bCanReduceInnerElements
 	)
 {
 	CXXTokenChain * pCurrentChain = g_cxx.pTokenChain;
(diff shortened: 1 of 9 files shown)


masatake:
I would like to hear about my change.
Do you think it is good solution for the issue?

If yes, I think we can generalize the code(reducer).

pragmaware:
I think it is a good solution. Improvements are always possible but well... in the meantime it's done.
One direction we could investigate is to avoid placing the tokens in the chain at all (so no memory peaks are present).

PR #1387, opened 2017-05-17
Introduce itcl and tcloo subparsers

masatake:


masatake:
This should be merged after #1393 is merged.

PR #1388, opened 2017-05-17
buildsys: Fix cleaning optlib and parsers/cxx

k-takata:
optlib and parsers/cxx were not cleaned by mk_mingw.mak.


masatake:
LGTM.

PR #1391, opened 2017-05-18
win32: Generate repoinfo.h on VC2013

k-takata:
Do the same thing as `misc/gen-repoinfo`, when `mk_mvc.mak` is used.

### DIFF
--- a/win32/gen-repoinfo.bat
+++ b/win32/gen-repoinfo.bat
@@ -1,4 +1,7 @@
 @echo off
+:: Copyright (C) 2017 Ken Takata
+:: License: GPL-2
+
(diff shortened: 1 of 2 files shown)


win32/gen-repoinfo.bat
```diff
@@ -0,0 +1,22 @@
+@echo off
```
masatake: Could you add your copyright notice as I have done in other files?

masatake:
Thank you!

win32/gen-repoinfo.bat
k-takata: Done.

PR #1393, opened 2017-05-19
win32: Import mkstemp() from MinGW-w64

### LINKED ISSUE #1377: 'ctags: cannot open temporary file : File exists'
On windows:
if run `ctags -n -u --fields=Ks  -f- aaa.cpp` for many times, this error occur.
The file name is different every time.
How can this be fixed?

k-takata:
This imports mkstemp() from MinGW-w64 to fix #1377.
Also updates VC project files and classifies the source files properly.


masatake:
LGTM.

k-takata:
Maybe we should include COPYING.MinGW-w64-runtime.txt in Windows zip package.

k-takata:
Updated.

BTW, I'm thinking if we could include HTML version of the document to zip package. Currently it only includes rst files.
Not sure it is easy.

masatake:
@k-takata, on GNU/Linux what I have to do to generate html files are:

1. cd docs
2. make html

With the above command lines, I will get docs/build/_html.

sphinx is used to generate the html files. So Making the files on appveyor is easy if sphinx is available on the platform. If it is not avaiable, we have to put generated html to our git repo.

ctags man page is also generated from rst file. So this will be critical issue when making a release and shipping zip(?) file for windows in the future.

masatake:
Thank you very much.

k-takata:
Merged.

I confirmed that MSYS2 has sphinx package. So maybe we can use it.

PR #1396, opened 2017-05-22
Win32,TclOO: Fix wrong filter spec

masatake:
Suggested by @k-takata.


k-takata:
I consider sorting the files in `*.vcxproj*` in alphabetical order to avoid this kind of mistake.

PR #1398, opened 2017-05-22
win32: Sort files in VC projects

k-takata:
Sort the file lists in alphabetical order.

See https://github.com/universal-ctags/ctags/pull/1396#issuecomment-302991097.


masatake:
LKGTM. 
BTW, how did you do this? Using VC or any other tool?

k-takata:
An editor and VC.

* .vcxproj
  Just sort with the editor.
* .vcxproj.filter
  Join each 3 lines (start with `<ClCompile` and end with `</ClCompile>`) into a line, sort the lines, then open with VC and save it.

PR #1399, opened 2017-05-22
Fix VC warnings

k-takata:
VC shows the following warnings:

```
main\entry.c(860) : warning C4550: expression evaluates to a function which is missing an argument list
main\entry.c(861) : warning C4550: expression evaluates to a function which is missing an argument list
main\options.c(2522) : warning C4114: same type qualifier used more than once
main\parse.c(918) : warning C4090: 'initializing' : different 'const' qualifiers
main\routines.c(383) : warning C4028: formal parameter 1 different from declaration
main\routines.c(397) : warning C4028: formal parameter 1 different from declaration
main\routines.c(559) : warning C4090: 'return' : different 'const' qualifiers
parsers\cpreprocessor.c(1260) : warning C4028: formal parameter 1 different from declaration
parsers\cxx\cxx_parser_tokenizer.c(1064) : warning C4090: 'function' : different 'const' qualifiers
```

See https://ci.appveyor.com/project/masatake/ctags/build/1.0.2058/job/ceca1toq5bfs857s for full log.
This PR fixes them.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -817,8 +817,8 @@ extern void getTagScopeInformation (tagEntryInfo *const tag,
 
 
 static int   makePatternStringCommon (const tagEntryInfo *const tag,
-				      int putc_func (char , void *),
-				      int puts_func (const char* , void *),
+				      int (* putc_func) (char , void *),
+				      int (* puts_func) (const char* , void *),
 				      void *output)
 {
 	int length = 0;
--- a/main/options.c
+++ b/main/options.c
@@ -2519,7 +2519,7 @@ static void processPatternLengthLimit(const char *const option, const char *cons
 		error (FATAL, "-%s: Invalid pattern length limit", option);
 }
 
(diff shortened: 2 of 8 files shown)


masatake:
gcc has following option:
```
      -Werror
           Make all warnings into errors.
```

I wonder VC has the similar option. If it has, could you turn it on?

masatake:
I would like to review this carefully to learn how I wrote wrong code.
Please, give me time.

k-takata:
> I wonder VC has the similar option.

`/WX` is that.
https://msdn.microsoft.com/en-us/library/thxezb7y.aspx
I will add it later.

Note that we currently use the default warning level (`/W1`). If we change the level higher, we will get more warnings.
E.g. If we use `/W3`, many C4018 warnings (signed/unsigned mismatch) will be shown.
https://msdn.microsoft.com/en-us/library/y92ktdf2.aspx

k-takata:
Added `/WX` option.

PR #1400, opened 2017-05-23
Update docs

k-takata:
Update Windows document and fix typos.

* Add description about PDB=1 option.
* Change a section name from "MSYS" to "MSYS / MSYS2" to make clear that the section also mentions MSYS2.
* Add merits using Autotools on MSYS2.
* Recent Visual Studio uses `*.vcxproj*` for its project filename instead of `*.vcproj*`.
* Fix typos.


masatake:
LGTM. Thank you.

PR #1402, opened 2017-05-24
appveyor: Create html docs

k-takata:
Create html docs on AppVeyor.
I put the html docs in `docs/html/`, but I'm not sure this is the right directory structure.

```
 .
├── COPYING
├── COPYING.MinGW-w64-runtime.txt
├── ctags.exe
├── docs
│   ├── *.rst
│   └── html
│       └── *.html
├── iconv.dll
├── README.md
└── readtags.exe
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
       MSYSTEM: MINGW64
     - compiler: msvc_msys2
       ARCH: x86
-      MSYS2_ARCH: x86
+      MSYS2_ARCH: i686
       MSYS2_DIR: msys64
       MSYSTEM: MINGW32
     - compiler: mingw
--- a/win32/appveyor.bat
+++ b/win32/appveyor.bat
@@ -117,11 +117,20 @@ bash -lc "make check APPVEYOR=1"
 goto :eof
 
 :msvc_msys2_package
+:: Build html docs
+bash -lc "for i in {1..3}; do pacman --noconfirm -Su mingw-w64-%MSYS2_ARCH%-python3-sphinx && break || sleep 15; done"
+bash -lc "cd docs; make html"
+move docs\_build\html docs > nul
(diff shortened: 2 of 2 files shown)


masatake:
> but I'm not sure this is the right directory structure.

It is the right directory if you think it is the right directory:-P.

BTW, does pacman provide libxml2? u-ctags utilizes it if it is available.

k-takata:
> It is the right directory if you think it is the right directory:-P.

Haha, BTW, should we provide both rst and html docs?

> BTW, does pacman provide libxml2?

Yes. It also provides jansson and libyaml.
However they cannot be used from VC.
Switching the official build from VC to MSYS2 is the easiest way to provide json-, xml- and yaml-enabled ctags binary.

masatake:
Providing only the html files is enough.  The rst files are kinds of source file. 
SVG files like http://docs.ctags.io/en/latest/_images/output-tag-stream.svg should be included.

How different they (ctags built with VC and ctags built with MSYS2) are from users view?
I understand that the users may have preferences of build-tool and environment if they built a ctags binary by themselves.
However, I guess users may not have preferences of build-tool and environment if they just use existing ctags binary. Am I wrong? Are there big gap in execution performance?

k-takata:
> Providing only the html files is enough.

Okay, I will update later. It seems that `_static` directory is also needed.

> How different they (ctags built with VC and ctags built with MSYS2) are from users view?

Maybe not much difference. One thing I noticed is filesize.
VC: 500 KB + PDB
MSYS2 unstripped: 4 MB
MSYS2 stripped: 600 KB

k-takata:
Updated.

k-takata:
Can I merge this?

After this, I will create a new PR to change the compiler for creating zip package from VC to MSYS2.

masatake:
> Can I merge this?


Sure. Please, merge this.

PR #1403, opened 2017-05-24
Cxx,refactor: introduce generic function for reducing tokens

masatake:
In commit:

        Cxx: reduce blocks when parsing values for array/union initialization

I introduced a loop to delete tokens to make memory usage during parsing
a large array smaller.

Newly introduced function cxxTokenReduceBackward is generalized version
of the loop. So you can apply to another place in the parser.


masatake:
@pragmaware, how about this one?
I think a bit better than my original code.
I'm not sure "reduce" is good term for the purpose.

Issue #1312, opened 2017-03-02
JSON output is not available on ctags for windows


vnStrawHat:
I have download latest version of Ctags in here https://ci.appveyor.com/project/masatake/ctags but when I run ctags.exe with arg --output-format=json I got this message "ctags.exe: unknown output format name supplied for "output-format=json".

So, someone can help me build ctags for windows with json output support ?

Thanks.
Sorry for my English.

masatake:
The json output requires libjansson. I don't know how to install it to windows and link it to ctags.
So I have to say "patches are welcome.".

k-takata:
Now the Windows binaries support JSON, XML and YAML.

Issue #660, opened 2015-11-06
Package for Windows


cweagans:
For Universal Ctags to be truly universal, we should distribute Windows builds too. We can use Appveyor to do the automated nightly/tag builds. I *think* it supports tags as a build trigger. If so, then we can have it build ctags, package it up in an installer, and upload the installer to the Github Release.

Alternatively, if there's something like Homebrew for Windows, it might be better to do that. Cygwin, maybe? There's some docs on that here: https://cygwin.com/faq.html#faq.programming.packages

I cannot do this one, as I don't have a Windows box to work on.

ffes:
I don't think an installer is needed. The ctags is a tool for developers so a simple .zip with the exe, some basic readme-like documentation where to find all the proper docs and a license file could be enough.

Maybe appveyor can provide such a file, never tried that. ctags (cross)build just fine from Linux, so travis could work as well.
And obviously I can provide such a zip as well, especially for releases. If we want nightly builds, we do need one of those services to help us with that.

Cygwin is a different story. Cygwin has it's own packaging system. I could ask the current maintainer off the ctags package to switch to universal, or suggest to take over the package myself (I am the maintainer of the Cygwin "shutdown" package). For a new package to get accepted it helps a lot if any of the major Linux distributions provide universal-ctags packages.

cweagans:
Ah, great, I didn't know we had cross-compilation set up. We should create Windows builds on Travis when there's a new tag - automation is :+1:

I'd really love to get Universal Ctags into cygwin, so if you're willing to take that on, that would be awesome. We've already got some momentum behind having packages for Linux distros - they're just waiting on us to make a tag. So far, the ones that I know of are Debian, Gentoo, and Arch. I still need to track down people that will package an RPM and get it into the RHEL/Fedora/OpenSuse repos, but I'll make sure to firm that up before our 1.0.0 release. My goal is to have Universal Ctags packaged and immediately installable by pretty much anyone that wants it on release day, so we need to make sure to have Linux, Mac, and Windows packaging squared away before that date.

I'll continue working on that going forward.

masatake:
@ffes and @k-takata, I'm thinking about introducing a code generator(ctagst) written in bash to our build process.
See #614. ctagst converts something.ctags to something.c.

Is it acceptable?
If this introduces a big trouble, I will commit both something.ctags and generated something.c to our git repository when I make a change on something.ctags.

masatake:
[...]

PR #1410, opened 2017-05-27
appveyor: Use MSYS2 for packaging

k-takata:
* Use MSYS2 to create zip packages and stop using VC for it.
* Enable following libraries with MSYS2:
  jansson, libxml2, libyaml.
* Enable static link with MSYS2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,9 @@ If you want to try the latest universal-ctags without building it yourself...
 
 ### Windows
 - Go to https://ci.appveyor.com/project/masatake/ctags/branch/master
-  - Click the ```compiler=msvc_msys2, ARCH=x64, ...``` (or ```compiler=msvc_msys2, ARCH=x86, ...```) build.
+  - Click the ```compiler=msys2, ARCH=x64, ...``` (or ```compiler=msys2, ARCH=x86, ...```) build.
   - View the *Artifacts* tab and download ```ctags-XXXXXX-x64.zip``` (or ```ctags-XXXXXX-x86.zip```). (```XXXXXX``` is a version number or a commit ID.)
   - Add the binary folder to your PATH.
-  - If you need PDB files for debugging, download ```ctags-XXXXXX-x64.pdb.zip``` (or ```ctags-XXXXXX-x86.pdb.zip```).
 
 ### Mac
 See [Homebrew Tap for Universal Ctags](https://github.com/universal-ctags/homebrew-universal-ctags)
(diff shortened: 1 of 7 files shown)


k-takata:
This fails because of #1411. It should be fixed before merging this.

k-takata:
I think static link is suitable for distributing Windows binary for portability.
If we distribute a binary linked with jansson, libxml2 and libyaml, I think we should include license files of the libraries.
I couldn't find the license files under the MSYS2 installation directory, so maybe we should store the copies of them in our repository.

The license files are found at here:
https://github.com/akheron/jansson/blob/master/LICENSE
https://git.gnome.org/browse/libxml2/tree/Copyright
https://bitbucket.org/xi/libyaml/src/default/LICENSE?fileviewer=file-view-default

k-takata:
The directory structure of a zip package is:

```
.
├── ctags.exe
├── docs
│   ├── _images
│   │   └── *.svg
│   ├── _static
│   │   └── *.css, *.js, etc.
│   └── *.html
├── license
│   ├── COPYING
│   ├── COPYING.MinGW-w64-runtime.txt
│   ├── Copyright.libxml2
│   ├── LICENSE.janssen
│   └── LICENSE.libyaml
├── README.md
└── readtags.exe
```

k-takata:
I stopped testing VC binaries in this PR, because it takes long time (2 x ~5 minutes), but stopping it has a risk that we cannot catch bugs only occurs with VC.
So, should we still test VC binaries on AppVeyor?

masatake:
Stopping tesing VC binaries during updating this PR is o.k.
When all updating are done, let's turn the testing on again.

k-takata:
Rebased upon the fix of #1418.
Now both stripped/unstripped binaries are available.
The size of ctags.exe is now:

Stripped: ~ 3 MB
Unstripped: ~ 8 MB
(See: https://ci.appveyor.com/project/masatake/ctags/build/1.0.2098/job/hpf3dq305vulj2se/artifacts)

Test of VC binaries are enabled again, but it is only x64. Test of VC x86 is disabled to reduce the CI time.
Now AppVeyor takes about 20 minutes.

k-takata:
I think it's ready to be merged.

k-takata:
This also fixes #1312.

k-takata:
And  also #660.

k-takata:
I updated the docs and added how to build static link binary using MSYS2.

PR #1412, opened 2017-05-27
Fix typos

k-takata:
Please review carefully.

(Some trailing spaces were removed because of the setting of EditorConfig.)

### DIFF
--- a/parsers/cpreprocessor.h
+++ b/parsers/cpreprocessor.h
@@ -78,7 +78,7 @@ extern int cppGetc (void);
 extern int cppSkipOverCComment (void);
 
 /* notify the external parser state for the purpose of conditional
-   branche choice. The CXX parser stores the block level here. */
+   branch choice. The CXX parser stores the block level here. */
 extern void cppPushExternalParserBlock();
 extern void cppPopExternalParserBlock();
 
--- a/parsers/cxx/cxx_parser.c
+++ b/parsers/cxx/cxx_parser.c
@@ -105,11 +105,11 @@ bool cxxParserParseAndCondenseCurrentSubchain(
 
 //
 // This function parses input until one of the specified tokens appears.
-// The current token is NOT checked agains the specified tokens.
+// The current token is NOT checked against the specified tokens.
 //
 // The algorithm will also build subchains of matching
 // pairs ([...],(...),<...>,{...}): within the subchain analysis
-// of uTokenTypes is completly disabled. Subchains do nest.
+// of uTokenTypes is completely disabled. Subchains do nest.
 //
 // Returns true if it stops before EOF or it stops at EOF and CXXTokenTypeEOF
 // is present in uTokenTypes. Returns false in all the other stop conditions
@@ -221,7 +221,7 @@ bool cxxParserParseAndCondenseSubchainsUpToOneOf(
 			continue; // jump up to avoid checking for mismatched pairs below
 		}
 
-		// Check for mismatched brackets/parenthis
+		// Check for mismatched brackets/parentheses
 		// Note that if we were looking for one of [({ then we would have matched
 		// it at the top of the for
 		if(cxxTokenTypeIsOneOf(g_cxx.pToken,uFinalSubchainMarkerTypes))
@@ -247,10 +247,10 @@ bool cxxParserParseAndCondenseSubchainsUpToOneOf(
 
 //
 // This function parses input until one of the specified tokens appears.
-// The current token is NOT checked agains the specified tokens.
+// The current token is NOT checked against the specified tokens.
 //
 // The algorithm will also build subchains of matching pairs ([...],(...),{...}).
-// Within the subchain analysis of uTokenTypes is completly disabled.
+// Within the subchain analysis of uTokenTypes is completely disabled.
 // Subchains do nest.
 //
 // Please note that this function will skip entire scopes (matching {} pairs)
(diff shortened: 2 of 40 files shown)


parsers/cpreprocessor.h
```diff
@@ -78,7 +78,7 @@ extern int cppGetc (void);
 extern int cppSkipOverCComment (void);
 
 /* notify the external parser state for the purpose of conditional
-   branche choice. The CXX parser stores the block level here. */
+   branch choice. The CXX parser stores the block level here. */
```
k-takata: Or should be `branches`?

parsers/cxx/cxx_parser.c
```diff
@@ -973,7 +973,7 @@ static bool cxxParserParseClassStructOrUnionInternal(
 			(!(uInitialKeywordState & CXXParserKeywordStateSeenTypedef))
 		)
 	{
-		CXX_DEBUG_PRINT("Maybe KnR funciton definition?");
+		CXX_DEBUG_PRINT("Maybe KnR function definition?");
```
k-takata: "KnR" or "K&R"?
It's inconsistent in this file.

parsers/cxx/cxx_token_chain.h
```diff
@@ -190,7 +190,7 @@ vString * cxxTokenChainJoinRange(
 		unsigned int uFlags
 	);
 
-// Treat the tochek chain tc as a comma separated sequence
+// Treat the token chain tc as a comma separated sequence
```
k-takata: Is this fix okay?

k-takata:
Another questions.

https://github.com/universal-ctags/ctags/blob/dda51569a8d0bdf360ce5c15f643e113e6f836be/parsers/make.c#L299
What is this "endef"? Mistake of "endif"?

https://github.com/universal-ctags/ctags/blob/dda51569a8d0bdf360ce5c15f643e113e6f836be/parsers/ocaml.c#L418
Is "consecuting" a mistake of "consecutive"?

parsers/cpreprocessor.h
masatake: I don't know. But `branch` may be o.k.

parsers/cxx/cxx_parser.c
masatake: Maybe `K&R` is better. However, the string KnR is used in a name of cunction.  I think both words are acceptable.

masatake:
I have introduced many of wrong spellilng.

I must build infrastructure for running aspell on

1. comments
2. literal strings
3. commit log
4. subwords in identifiers in C source code
5. rst docs

The infrastructure can be integrated to travis.

masatake:
`endef` is endef:-)

Quoted from the info document of GNU Make:
```
   define IMMEDIATE
       DEFERRED
     endef
```

parsers/cxx/cxx_parser.c
k-takata: I will leave this to @pragmaware.

k-takata:
> endef is endef:-)

Oh, I didn't know that.

> Is "consecuting" a mistake of "consecutive"?

Fixed.

k-takata:
> I must build infrastructure for running aspell on

It would be nice, but I suppose it will detect many false positives and need workarounds for them.
I think it's better to leave this for a task after the first release.

BTW, doesn't Emacs have a plugin for spell checking?
Vim has a built-in spell checker and I used it for this PR.

masatake:
Emacs has a spell checker. However, I sometimes forget to run it before making a PR.
I find I want a unified interface for running a source code tree.

```
$ git checkout devel
$ git spell-check --ignore=Tmain,Units --dictionary=tags,extra.dict --commit-log --string-literal=C,Shell --comments=C,Shell --identifiers=C,Shell master...
commit log...o.k.
string literal...o.k.
comments...o.k.
identifiers...o.k.
...
```

So I can run git spell-check before making a pull request:.
I strongly belive someone on github already develop such software.

> I think it's better to leave this for a task after the first release.
You are right. However, `git spell-check` is very attractive idea for me.

k-takata:
I forgot to check man/ctags.1.rst.in.
Updated.

BTW, are the command line options only described in man/ctags.1.rst.in and not in docs?

masatake:
ctags.1.rst.in is not updated though we should do.
Could you review it if I update it?

k-takata:
> Could you review it if I update it?

Yes.

Then, will an html file generated from ctags.1.rst.in be also available on http://docs.ctags.io/en/latest/ ?
Or, will another rst file about command line options be created in the docs directory?

FYI, I found an interesting project for detecting typo: https://github.com/chakki-works/typot .
A Japanese post is here: http://qiita.com/icoxfog417/items/5edb32b34dfb7886d426 .

masatake:
> Yes.

Thank you. 
 I will update ctags.1.rst.in incrementally. Each time, I will add you "Reviewers".

>Then, will an html file generated from ctags.1.rst.in be also available on http://docs.ctags.io/en/latest/ ?
> Or, will another rst file about command line options be created in the docs directory?

```
$ make man/ctags.1
```
Generated the man page if rst2man is available.

The man page is not at http://docs.ctags.io.

> 
> FYI, I found an interesting project for detecting typo: https://github.com/chakki-works/typot .
> A Japanese post is here: http://qiita.com/icoxfog417/items/5edb32b34dfb7886d426 .

Very interesting. We should use it.
I also try my ideas for spell checking. This will not run on Github or CIs behind it. Mine is just run in local. 


Anyway I'm bit busy these days. So, please wait the initial review  request for awhile.

Issue #1317, opened 2017-03-05
[RFH] Writing regex language extensions for utags


t-b:
[Sorry for misusing the issue tracker, I could not find a mailinglist or similiar]

The C++ parser outputs for the following code

```
class MyClass
{
  static MyFunc()
  {
  }
};
```

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	/a789d196/
MyClass	test.cpp	/^class MyClass$/;"	c	file:
MyFunc	test.cpp	/^  static MyFunc()$/;"	f	class:MyClass	file:
```

I would like to use the feature to include more information in the tag line, `f` vs `f	class:MyClass	file:` in a parser written using the `--langdef`, `--regex-$lang` interface.
For the following code

```
static Function MyFunc()
End
```

I would like to generate an entry like

```
MyFunc	test.ipf	/^static Function MyFunc()$/;"	f specifier:static
```

but I'm currently stuck using

```
$ctags --langdef=igorpro --langmap=igorpro:.ipf --regex-igorpro='/^[[:space:]]*static[[:space:]]+Function[\/A-Z]*[[:space:]]+([A-Z0-9_]+)\([^:]+$/\1/f,function definition/i' test.ipf
```

with

```
MyFunc	test.ipf	/^static Function MyFunc()$/;"	f
```

Any hints appreciated.

masatake:
> [Sorry for misusing the issue tracker, I could not find a mailinglist or similiar]
This is the place for you; there is no mailing list.

Before starting discussion I would like to know  2 things about you.

Q1. whether you understand 3 important concepts of ctags: kind, field and extra.
Q2. scope of your interest: implementing a parser for igorpro or extending ctags 

If your interest is a parser for igorpro, I recommend you to write it in C with regex.
See parsers/rpmspec.c and parsers/robot.c. In addition I recommend you to learn
kind, field, and extra. If you are tolerant about my broken English. I can provide more
[...]

PR #1416, opened 2017-05-27
Add --fielddef

masatake:
`--fielddef` option and `_field` long regex flag allows a user to attach extra information to a tag entry.
I don't have real application of the option and flag. So I don't add their support to optlib2c translator
yet. 

The original code for attaching parser own fields to a tag entry was written by @pragmaware and
used in C and C++ parsers. There was limits in the number of fields attaching. The slots for the attaching is statically defined in ctags source code. In this pull request, the limitation is removed. The slot can be dynamically added on demand.

----

Attaching parser own fields
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

.. NOT REVIEWED YET

Exuberant-ctags allows one of the specified group in a regex pattern can be
used as a part of the name of a tagEntry. Universal-ctags offers using
the other groups in the regex pattern.

A optlib parser can have its own fields. The groups can be used as a
value of the fields of a tagEntry.

Let's think about *Unknown*, an imaginary language.
Here is an source file(``input.unknown``) written in *Unknown*:

    public func foo(n, m);
    protected func bar(n);
    private func baz(n,...);

With `--regex-Unknown=...` Exuberant-ctags can capture `foo`, `bar`, and `baz`
as names. Universal-ctags can attach extra context information to the
names as values for fields. Let's focus on `bar`. `protected` is a
keyword to control how widely the identifier `bar` can be accessed.
`(n)` is the parameter list of `bar`. `protected` and `(n)` are
extra context information of `bar`.

With following optlib file(``unknown.ctags``)), ctags can attach
`protected` to protection field and `(n)` to signature field.

.. code-block:: ctags

    --langdef=unknown
    --kinddef-unknown=f,func,functions
    --map-unknown=+.unknown

    --_fielddef-unknown=protection,access scope
    --_fielddef-unknown=signature,signatures

    --regex-unknown=/^((public|protected|private) +)?func ([^\(]+)\((.*)\)/\3/f/{_field=protection:\1}{_field=signature:(\4)}

    --fields-unknown=+'{protection}{signature}'

For the line `    protected func bar(n);` you will get following tags output::

	bar	input.unknown	/^protected func bar(n);$/;"	f	protection:protected	signature:(n)

Let's see the detail of ``unknown.ctags``.

.. code-block:: ctags

    --_fielddef-unknown=protection,access scope

`--_fielddef-<LANG>=name,description` defines a new field for a parser
specified by `<LANG>`.  Before defining a new field for the parser,
the parser must be defined with `--langdef=<LANG>`. `protection` is
the field name used in tags output. `access scope` is the description
used in the output of ``--list-fields`` and ``--list-fields=Unknown``.

.. code-block:: ctags

    --_fielddef-unknown=signature,signatures

This defines a field named `signature`.

.. code-block:: ctags

    --regex-unknown=/^((public|protected|private) +)?func ([^\(]+)\((.*)\)/\3/f/{_field=protection:\1}{_field=signature:(\4)}

This option requests making a tag for the name that is specified with the group 3 of the
pattern, attaching the group 1 as a value for `protection` field to the tag, and attaching
the group 4 as a value for `signature` field to the tag. You can use the long regex flag
`_field` for attaching fields to a tag with following notation rule::

  {_field=FIELDNAME:GROUP}


`--fields-<LANG>=[+|-]{FIELDNAME}` can be used to enable or disable specified field.

When defining a new parser own field, it is disabled by default. Enable the
field explicitly to use the field. See :ref:`Parser own fields <parser-own-fields>`
about `--fields-<LANG>` option.


masatake:
This new feature may solve #1317.

masatake:
I have to add more test cases to cover error handling.

masatake:
passwd parser can utilize `--field-<LANG>` feature.

TODO

* adding more cases for testing broken command lines.
* extending optlib2c to support `--field-<LANG>`,
* adding passwd parser as an example

Issue #1423, opened 2017-05-30
Incorrect handling of types causes a segfault on armv7l (e.g., Raspberry Pi)


wbkang:
Hello,

I was excited to try out your project on my Raspberry Pi but ctags died with a segfault. After running it through gdb, I found the root cause:
```
// main/kind.h
#define KIND_GHOST_INDEX -1
#define KIND_FILE_INDEX -2
// main/parse.c
extern kindDefinition* getLanguageKind (const langType language, char kindIndex)
{
	kindDefinition* kdef;

	Assert (0 <= language  &&  language < (int) LanguageCount);

	switch (kindIndex)
	{
	case KIND_FILE_INDEX:
		kdef = LanguageTable [language].fileKind;
		break;
	case KIND_GHOST_INDEX:
		kdef = &kindGhost;
		break;
	default:
		kdef = getKind (LanguageTable [language].kindControlBlock, kindIndex);
	}
	return kdef;
}
```

On armv7l, char type is actually unsigned. So doing a switch on an unsigned value `kindIndex` with a negative case value actually does not work. My suggestion is to fix the type of `kindIndex` specifically to `signed char`. All unit tests pass if I make that change as well.

masatake:
Thank you for reporting and fixing this.
I will add #1428 , too.

PR #1424, opened 2017-05-30
Incorrect handling of types causes a segfault on arm

wbkang:
On armv7l, char type is actually unsigned. So doing a switch on an unsigned value kindIndex with a negative case value actually does not work. My suggestion is to fix the type of kindIndex specifically to signed char. All unit tests pass if I make that change as well.

make units is passed.

Created an issue: https://github.com/universal-ctags/ctags/issues/1423

I am contributing the code under the terms of the General Public License version 2 or any later version.

### DIFF
--- a/main/parse.h
+++ b/main/parse.h
@@ -147,8 +147,9 @@ extern parserDefinition* parserNew (const char* name);
 extern bool doesLanguageAllowNullTag (const langType language);
 extern bool doesLanguageRequestAutomaticFQTag (const langType language);
 extern const char *getLanguageName (const langType language);
+/* kindIndex has to be explicitly signed because char is not signed in all platforms. */
 extern kindDefinition* getLanguageKindForLetter (const langType language, char kindLetter);
-extern kindDefinition* getLanguageKind(const langType language, char kindIndex);
+extern kindDefinition* getLanguageKind(const langType language, signed char kindIndex);
 extern int defineLanguageKind (const langType language, kindDefinition *def,
 							   freeKindDefFunc freeKindDef);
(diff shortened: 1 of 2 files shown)


masatake:
> On armv7l, char type is actually unsigned. 
I was very suprised at this but you are right; I found the explanation in the K&R book in my hand.

valgrind is avaiable in your environment?

I would like you to run "make units VG=1". I hope valgrind reports something unwanted cses.

masatake:
> I am contributing the code under the terms of the General Public License version 2 or any later version.

Thank you very much.

main/parse.h
```diff
@@ -148,7 +148,7 @@ extern bool doesLanguageAllowNullTag (const langType language);
 extern bool doesLanguageRequestAutomaticFQTag (const langType language);
 extern const char *getLanguageName (const langType language);
 extern kindDefinition* getLanguageKindForLetter (const langType language, char kindLetter);
-extern kindDefinition* getLanguageKind(const langType language, char kindIndex);
+extern kindDefinition* getLanguageKind(const langType language, signed char kindIndex);
```
masatake: Can I ask you to add comments explaining why "signed" is added here?

wbkang:
All test passed with valgrind on as well (it says `#valgrind-error: 0`). For the future reference, you have to comment out `libarmmem.so` in `/etc/ld.so.preload` (See: https://stackoverflow.com/a/42074785/2710) or else most valgrind tests will fail with error 139.

main/parse.h
wbkang: This is done!

masatake:
Thank you.

Issue #1425, opened 2017-05-30
End line for java


adrianmay:
For some reason I don't get the end:123 field for java files like I do for C/C++ files when I specify --fields=e. 

Is this an oversight or is it hard for some reason?

masatake:
Only some parsers implements `end:` field. Java parser does not.

adrianmay:
I'd be very grateful if it did or if you could help me implement it. It's actually the only reason I'm using the software. My motivation is that I'm trying to find out which functions a source control commit affects, so that I can run statistics about which functions have a historical tendency to change together. I know that git and hg both attempt this, but they both get it wrong.

masatake:
"Of course, a patch is welcome", is the motto of this project.

Java parser is implemented in parsers/c.c.  

As you wrote C/C++ parsers can record end: field. C/C++ parser is newly written by @pragmaware who started rewriting it not to maintain the old C/C++ parser. He though it is hard to maintain the old C/C++ parsers.

Java parser is implemented as a variant of the old C/C++ parser.
To add end: to Java parser, there are two ways. 

1. understanding c.c and changing it
2. as @pragmaware did, writing new Java parser

 Anyway, to add `end:` field, you want to use CorkAPI.

Step for adding `end:` 

0. study javascript language quickly if you don't know.
1. read a javascript parser (jscript.c) and understand how `initTagEntry` and `makeTagEntry` functions are used.
    makeSimpleTag doesn't help you.
2. read http://docs.ctags.io/en/latest/internal.html#cork-api
3. See following patch. I have tried to add end: to JavaScript  parser. The change is not perfect but it will help you to understand how to add end:. 
https://gist.github.com/masatake/b09f530f5b1376b909a62bb54e74e533

4. Do the same existing Java parser or your new Java parser.
5. Run existing test: make units and make units VG=1
6.  Write your own test cases to verify end: field is recorded well. See http://docs.ctags.io/en/latest/units.html#units-test-facility
7. make a pull request

Good luck!

adrianmay:
Well I've been struggling with this but it isn't making much sense to me. The structure of c.c and jscript.c look completely different. In the latter, the function findCmdTerm appears to be the crucial one that skips to the closing brace of the function, but I'm not finding an equivalent in c.c. Also, if there's currently no usage of cork in c.c, why would I need to convert it now? 

[...]

PR #1426, opened 2017-05-30
buildsys: Fix `make distclean`

k-takata:
We use Automake, so we can use `make dist` to create a tarball and `make distcheck` to create and check it.
However, `make distcheck` fails because of some reasons.

* Source files are not properly listed. We also need to care about Windows specific source files.
* Creating a tarball fails because of very long filenames.
  See: https://www.gnu.org/software/automake/manual/html_node/List-of-Automake-options.html
  `tar-v7` supports file names with up to only 99 characters.
  However we have some files which names are longer than that.
  So we should use `tar-ustar` or `tar-pax`.
* `make distclean` leaves some files in build directory. So the following error will be shown:
  ```
  ERROR: files left in build directory after distclean:
  ./Units/conflib-recursive.d/BUNDLES
  ./Units/conflib-recursive.d/optlib/lang.d/def.d/a.ctags
  ...
  ```
  The outputs of unit tests should be cleaned by `make clean` or `make distclean`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,9 +1,13 @@
 include $(top_srcdir)/source.mak
 
 # Test cases are added to EXTRA_DIST in makefiles/test-cases.mak
-EXTRA_DIST   = gnu_regex/regcomp.c gnu_regex/regexec.c \
+EXTRA_DIST   = README.md \
+	       gnu_regex/regcomp.c gnu_regex/regexec.c \
 	       gnu_regex/regex_internal.c gnu_regex/regex_internal.h \
-	       $(WIN32_HEADS) $(WIN32_SRCS)
+	       $(WIN32_HEADS) $(WIN32_SRCS) mk_mingw.mak mk_mvc.mak \
+	       win32/appveyor.bat win32/ctags_vs2013.sln \
+	       win32/ctags_vs2013.vcxproj win32/ctags_vs2013.vcxproj.filters \
+	       win32/gen-repoinfo.bat
 
 bin_PROGRAMS = ctags
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,11 +1,14 @@
 include $(top_srcdir)/source.mak
 
 # Test cases are added to EXTRA_DIST in makefiles/test-cases.mak
-EXTRA_DIST   = README.md \
-	       gnu_regex/regcomp.c gnu_regex/regexec.c \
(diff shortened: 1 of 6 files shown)


k-takata:
Note that this PR only fixes the error of `make distcheck`, however there are still some problems with `make dist`. Documents and some files (especially for VC) are lacking from the tarball.

masatake:
I don't object but I don't think we have to add files under docs directory to the release tarball.
Instead I would like to add files under man directory to the tarball when we update the man page.

k-takata:
Updated.
Dotfiles and Win32 files are added.
Maybe some files under `misc` should be also added.

> I don't think we have to add files under docs directory to the release tarball.

Do you consider the files under `docs` are mainly for the Web page and no need to be installed in a user's PC?
Some software install HTML document under `/usr/share/doc/` (or somewhere else).

masatake:
> Do you consider the files under docs are mainly for the Web page and no need to be installed in a user's PC?
> Some software install HTML document under /usr/share/doc/ (or somewhere else).

The docs directory is a kind of staging area.

* Advertising experimental feature. I would like to get feed back from advanced users before releasing.
* Publishing the latest information (news) about u-ctags development. "latest" is important than its quality
  of contents. As you know I'm bad in English.
* Encouraging potential contributors. As I did in  #1425.

I think average users may need only some part of above.

The man page should 

* explain only stable features and interface,
* be written in good? English, and
* be written for average users, not for hackers.

In other word docs files are stocked materials for writing the real main page. Some parts of the files can be used and the other will be dropped.

I will not object to including docs files. However, at least, I would like you to know my intention.

We can add another document like INSTALL, NEWS, ... We have to think about this. However, for awhile, I would like to think about updating ctags.1 (, ctags-format.1, ctags-incompat.1, and...).

Do I make sense?

k-takata:
Thanks. I got your intention.
I added html files generated from docs/*.rst in Windows package with #1410, but then is it better to add an html file generated from man/ctags.1.rst?

Anyway I will merge this PR for now. `EXTRA_DIST` should be kept maintained.

masatake:
> I added html files generated from docs/*.rst in Windows package with #1410, but then is it better to add an html file generated from man/ctags.1.rst?

Yes. man/ctags.1.rst is must be included. docs/*.rst (or html) are optional.
I'm thinking about splitting man/ctags.1.rst into some.

PR #1430, opened 2017-06-01
[nit]remove compilation warning

mattn:


masatake:
Which compiler do you use?

mattn:
mingw64 mingw-w64-x86_64-gcc 6.3.0-3
```
gcc.exe (Rev3, Built by MSYS2 project) 6.3.0
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

masatake:
I see. Thank you. Please, wait for a while. I'm in busy.

masatake:
Could you show me the original warning message?
I cannot image what kind of warning is reported by the compiler.
As far as my understanding, `f` is initialized before referring.

mattn:
Sure.
```
main/fmt.c: In function 'printTagField':
main/fmt.c:78:12: warning: 'f' may be used uninitialized in this function [-Wmaybe-uninitialized]
   else if (isFieldEnabled (f->ftype))
            ^~~~~~~~~~~~~~~~~~~~~~~~~
```

masatake:
This is never "NIT".

I'm using "gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)" however I cannot regenerate(?) the message you saw.

(
I got following warnings instead
```
main/fmt.c: In function ‘printTagField’:
main/fmt.c:69:27: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for (findex = 0; findex < tag->usedParserFields; findex++)
                           ^
main/fmt.c:76:14: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (findex == tag->usedParserFields)
              ^~
```
with CFLAGS='-Wall -Wextra'
)

In your chage you assigned NULL to `f`. However, I think your change doen't fix the issue behind the warning message. Though your gcc didn't report explicitly, I guess it warned my wrong programming about integer operations.

Instead of your patch, could you try following patch?
```
diff --git a/main/fmt.c b/main/fmt.c
index 121fe42..7e2bedd 100644
--- a/main/fmt.c
+++ b/main/fmt.c
@@ -63,17 +63,20 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 		str = renderFieldEscaped (WRITER_XREF, ftype, tag, NO_PARSER_FIELD, NULL);
 	else
 	{
-		unsigned int findex;
+		int findex, used = tag->usedParserFields;
 		const tagField *f;
 
-		for (findex = 0; findex < tag->usedParserFields; findex++)
+		if (used < 0)
+			used = 0;
+
+		for (findex = 0; findex < used; findex++)
 		{
 			f = getParserField(tag, findex);
 			if (isParserFieldCompatibleWithFtype (f, ftype))
 				break;
 		}
 
-		if (findex == tag->usedParserFields)
+		if (findex == used)
 			str = "";
 		else if (isFieldEnabled (f->ftype))
 			/* TODO: Don't use WRITER_XREF directly */

```

 
If you get a trouble about applying the patch, please tell me so. I will make a gist or a branch.

mattn:
> In your chage you assigned NULL to f. However, I think your change doen't fix the issue behind the warning message. Though your gcc didn't report explicitly, I guess it warned my wrong programming about integer operations.

Well, on my environment, I got following message with -Wextra, without my patch.

```
[CC] main/fmt.o
main/fmt.c: In function 'printLiteral':
main/fmt.c:37:72: warning: unused parameter 'tag' [-Wunused-parameter]
 static int printLiteral (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag CTAGS_ATTR_UNUSED)
                                                                        ^~~
main/fmt.c: In function 'printTagField':
main/fmt.c:69:27: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for (findex = 0; findex < tag->usedParserFields; findex++)
                           ^
main/fmt.c:76:14: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (findex == tag->usedParserFields)
              ^~
main/fmt.c:78:12: warning: 'f' may be used uninitialized in this function [-Wmaybe-uninitialized]
   else if (isFieldEnabled (f->ftype))
            ^~~~~~~~~~~~~~~~~~~~~~~~~
```

And with -Wextra and my patch both.

```
[CC] main/fmt.o
main/fmt.c: In function 'printLiteral':
main/fmt.c:37:72: warning: unused parameter 'tag' [-Wunused-parameter]
 static int printLiteral (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag CTAGS_ATTR_UNUSED)
                                                                        ^~~
main/fmt.c: In function 'printTagField':
main/fmt.c:69:27: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for (findex = 0; findex < tag->usedParserFields; findex++)
                           ^
main/fmt.c:76:14: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (findex == tag->usedParserFields)
              ^~
```

So, my change fixes one of warning messages, and my change doesn't hide another warning. Anyway, ctags should fixes all of compilation warnings possibly.

mattn:
FYI, this is all compilation warnings on my environment.

https://gist.github.com/644231dc00bc009711aac7418bf0905a

masatake:
Could you try the latest head?
I changed add "unsigned" to the type of "usedParserFields" field.
As the result, I expect `f` is initialized anytime before the control accesses the field.

mattn:
Do you think that ctags must crashed when usedParserFields is 0. Then, I prefer to add Assert (f != NULL);

masatake:
> Do you think that ctags must crash when usedParserFields is 0. Then, I prefer to add Assert (f != NULL);

if `tag->usedParserFields` is 0, findex == `tag->usedParserFields` becomes true. There for `f` is not accessed.
Am I wrong?

mattn:
I don't make sure. As far as I can see, getParserField possibly return NULL.

masatake:
How about the original compiler warning? Is it gone?
If the warning is still reported, I missed something...

b4n:
I didn't check further, but the fix proposed in the PR itself doesn't make sense: yes, `f` becomes explicitly initialized no matter the branch, but then it's still dereferencing a NULL pointer, which means a crash.

And as @masatake said, `f` can only not be initialized if the loop didn't run once, which means `tag->usedParserFields <= 0`.  If `tag->usedParserFields` is unsigned, this means the guard `findex == tag->usedParserFields` is safe.

If the compiler doesn't agree, IMO it's just that it can't make out all the relations between the branches and shows a false positive.  However, I'd myself rewrite that code as something like the following that seems more obvious to me (if I'm getting *all* the logic at that late time):
```C
		unsigned int findex;

		str = "";
		for (findex = 0; findex < tag->usedParserFields; findex++)
		{
			const tagField *f = getParserField (tag, findex);
			if (isParserFieldCompatibleWithFtype (f, ftype))
			{
				if (isFieldEnabled (f->ftype))
				{
					/* TODO: Don't use WRITER_XREF directly */
					str = renderFieldEscaped (WRITER_XREF, tag->parserFields [findex].ftype,
								  tag, findex, NULL);
				}
				break;
		}
```
as `str` is set to `""` below if it's still `NULL`, it should be equivalent, and be more obvious that the access to `f` only happens if `isParserFieldCompatibleWithFtype (f, ftype)` succeeded.

PR #1433, opened 2017-06-01
appveyor: Include man page in zip package

k-takata:
Include `man/ctags.1.rst` in zip package. See: https://github.com/universal-ctags/ctags/pull/1426#issuecomment-305165075

### DIFF
--- a/win32/appveyor.bat
+++ b/win32/appveyor.bat
@@ -148,7 +148,7 @@ bash -lc "make check APPVEYOR=1"
 goto :eof
 
 :msys2_package
-md package
+md package\man
 :: Build html docs
 bash -lc "cd docs; make html"
 move docs\_build\html package\docs > nul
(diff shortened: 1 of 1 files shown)


k-takata:
`man/ctag.1.html` is not included this time.
When I run `rst2html` locally, some part is not marked up correctly.
Here is a screen shot:

ctags man

You can see that `--append[=yes|no]` and `--exclude=[pattern]` are wrongly marked up.
Is there any setting for fixing this?

k-takata:
Another option might be using man2html.

```
$ groff -Tutf8 -Dutf8 -mandoc man/ctags.1 | man2html > man/ctags.1.html
```

However, I feel the quality is not good. It uses `<pre>` and a monospace font to mark up the whole page.

masatake:
About the broken rst2html output, I guess it is something to do with square brackets in `--append[=yes|no]`. Maybe this notation violates the notation of option list in rst. I will inspect it and fix ctags.1.rst or rst processor itself. For awhile let's use rst2html.

Thank you.

k-takata:
> For awhile let's use rst2html.

Maybe we should add a makefile rule for creating ctags.1.html from ctags.1.rst.
After that I will include the html file in zip package.

PR #1446, opened 2017-06-08
appveyor: Additional fixes for #1445

k-takata:
* It is not recommended to update only specified packages. (ArchWiki says so.)
  Use `pacman -Suy` as usual.
* Use `checkupdates` to check if there are any updates.
* Remove unused toolchain before updating so that reduce the time for updating.
  (Unfortunately, it takes about 3 ~ 5 minutes even with this workaround.)
* Use default setting when updating Cygwin packages.
* Stop building msys2 binary before testing MSVC binary. This reduces about 30 sec.
* Use 64-bit Cygwin instead of 32-bit version. 64-bit version is about 1 minute faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Build Status](https://travis-ci.org/universal-ctags/ctags.svg?branch=master)](https://travis-ci.org/universal-ctags/ctags)
 [![Coverity Scan Build Status](https://scan.coverity.com/projects/4355/badge.svg)](https://scan.coverity.com/projects/4355)
 [![Coverage Status](https://coveralls.io/repos/universal-ctags/ctags/badge.svg?branch=master&service=github)](https://coveralls.io/github/universal-ctags/ctags?branch=master)
-[![Build status](https://ci.appveyor.com/api/projects/status/1c4wwswe8yd99la2/branch/master?svg=true)](https://ci.appveyor.com/project/masatake/ctags/branch/master)
+[![Build status](https://ci.appveyor.com/api/projects/status/6hk2p5lv6jsrd9o7/branch/master?svg=true)](https://ci.appveyor.com/project/universalctags/ctags/branch/master)
 [![RTD build status](https://readthedocs.org/projects/ctags/badge)](http://docs.ctags.io)
 
 universal-ctags has the objective of continuing the development from
@@ -18,7 +18,7 @@ together.
 If you want to try the latest universal-ctags without building it yourself...
 
 ### Windows
-- Go to https://ci.appveyor.com/project/masatake/ctags/history
+- Go to https://ci.appveyor.com/project/universalctags/ctags/history
   - Select one of the builds named ```Daily build: YYYY-MM-DD```.
   - Click the ```compiler=msys2, ARCH=x64, ...``` (or ```compiler=msys2, ARCH=x86, ...```) job.
   - View the *Artifacts* tab and download ```ctags-XXXXXX-x64.zip``` (or ```ctags-XXXXXX-x86.zip```). (```XXXXXX``` is a version number or a commit ID.)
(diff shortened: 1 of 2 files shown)


k-takata:
Tests on AppVeyor take 30 minutes after this change. Very long...

k-takata:
I reduced the build time on AppVeyor, but still take nearly 30 minutes.
Maybe there are two options to reduce the build time:

* Build both 64-bit and 32-bit MSYS2 builds in one job.
  This will reduce overhead of updating MSYS2 system, but the build script becomes complex.
* Stop creating a zip package for each build and stop updating MSYS2 system.
  Use AppVeyor's scheduled build to create Windows zip packages instead. It would be enough to create a zip package once a day.

k-takata:
> Maybe there are two options to reduce the build time:

I think that the second option is the best. I have updated the build script.

* For normal builds:
  * Don't create zip packages.
  * Stop updating msys2 packages.
  * Test only on 32-bit msys2, 64-bit msvc and 64-bit cygwin.
    Skip testing on 64-bit msys2.
* For scheduled daily builds or tagged builds:
  * Create zip packages and test on both 32- and 64-bit msys2.

Normal builds now take less than 15 minutes.
https://ci.appveyor.com/project/masatake/ctags/build/1.0.2177
Scheduled builds take about 20 minutes.
https://ci.appveyor.com/project/k-takata/ctags/build/1.0.204

To make this work, @masatake you need to change the setting of AppVeyor.
See: https://www.appveyor.com/docs/build-configuration/#scheduled-builds
Open "Settings" -> "General" -> "Build schedule", then set to `0 0 * * *` if you want to run the daily builds on 0:00 UTC.
What do you think?

masatake:
@k-takata, currently we use AppVeyor on my account. How do you think use your account instead?
So you can do everything you want, and is good for us. Anyway I will take a look about the schedule.

k-takata:
As far as I know, I don't have enough github permissions to run AppVeyor on my account.
Another option would be that you invite me as a collaborator on AppVeyor. I haven't use AppVeyor's team feature, though.

masatake:
> As far as I know, I don't have enough github permissions to run AppVeyor on my account.

Now I add you to "Code/CI-Admin". As far as I know this one attach you to the highest(?) permission to you.
If you still have a trouble, I will look into AppVeyor's team feature.

k-takata:
Now I'm setting up a new AppVeyor account for this project.
Cf. http://help.appveyor.com/discussions/questions/1154-appveyor-account-for-github-organizations

The project URL is:
https://ci.appveyor.com/project/universalctags/ctags
(Actually I wanted to use `universal-ctags` but it became `universalctags`. Not sure why.)

k-takata:
@masatake 
I added you as a collaborator of universalctags AppVeyor account.
And now, two webhooks are installed on this github project. So I disabled your AppVeyor webhook.

k-takata:
Oops. I mistakenly disabled my new webhook instead of masatake's one. Fixed.

k-takata:
@masatake 
I think I should add my name in `docs/developers.rst` as a maintainer of appveyor setting.
The document says "Keep the list in alphabetical order.", but it is not kept so...

masatake:
> I think I should add my name in docs/developers.rst as a maintainer of appveyor setting.
The document says "Keep the list in alphabetical order.", but it is not kept so...

Can I ask you to fix the order  when adding your name to the list?

k-takata:
I setup scheduled builds to run at 0:00 UTC.
The first daily build will be executed about 7 hours later from now. Let's see what happen.

> Can I ask you to fix the order when adding your name to the list?

okay.

k-takata:
https://ci.appveyor.com/project/universalctags/ctags/build/1.0.7
The first daily build has been built fine.

PR #1447, opened 2017-06-08
parser: Use inttypes.h to fix printf warning

k-takata:
On Windows, the following warning is shown:
https://ci.appveyor.com/project/k-takata/ctags/build/1.0.176/job/i3gcx6qr2wieq3tl#L500
> ```
> parsers/yaml.c:148:37: warning: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'size_t {aka long long unsigned int}' [-Wformat=]
>   verbose("yaml token:%s<%d>@Line:%lu\n", tokenTypeName[token.type], token.type,
>                                     ^
> ```

Include `inttypes.h` and use `PRIuPTR` to fix the warning.
Note that `long` is 32-bit and `size_t` is 64-bit on 64-bit Windows.

### DIFF
--- a/parsers/yaml.c
+++ b/parsers/yaml.c
@@ -10,6 +10,8 @@
 
 #include "general.h"  /* must always come first */
 
+#include <inttypes.h>
+
 #include "debug.h"
 #include "htable.h"
 #include "options.h"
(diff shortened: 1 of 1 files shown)


masatake:
LGTM. Thak you.

Can I ask you to add the description of this thread(?) to the commit log:

```
On Windows, the following warning is shown:
https://ci.appveyor.com/project/k-takata/ctags/build/1.0.176/job/i3gcx6qr2wieq3tl#L500

    parsers/yaml.c:148:37: warning: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'size_t {aka long long unsigned int}' [-Wformat=]
      verbose("yaml token:%s<%d>@Line:%lu\n", tokenTypeName[token.type], token.type,
                                        ^

Include inttypes.h and use PRIuPTR to fix the warning.
Note that long is 32-bit and size_t is 64-bit on 64-bit Windows.
```

What you wrote here is an important property of our project. If we can record it to git log, it is portable and someone including I can read it again in the future.

k-takata:
Updated.

Issue #1442, opened 2017-06-07
RFC: removing XCMD feature


masatake:
I will remove XCMD from the planed first stable release(6.0).

I like the idea but it makes difficult to keep the consistency of ctags internal.
I guess only a few people using the feature. 

Universal-ctags manages kinds more strictly than Exuberant-ctags.
Many of tags generators(xcmds) for specific languages models Exuberant-ctags.
The current Universal-ctags cannot fill the gap in kinds managment.

One of the way to go is that Universal-ctags defines better protocol between
Universal-ctags and xcmds that fill the gaps. However, to define such
protocol, I need much more experiences.

#705 (Process Multiple files with xcmd ) porints the critical weakness of
XCMD feature.

If you are using the feature and you like it, plesae add comments here. Based on the comments, I will revise the feature in 7.0 or later.  

The thumb  down reaction implies "I want it".

TODO: I have to notify this to CoffeeTags people.

masatake:
@kopischke, thank you for the feed-back. Could you tell me how you use xcmd?

kopischke:
@masatake for ripper-tags integration (supported as of [this PR](https://github.com/tmm1/ripper-tags/pull/42)). ctags’ stock Ruby tagging is laughably bad (Ruby’s fault for being a metaprogramming maze, not ctags).

masatake:
@kopischke, I see. Thank you.

Issue #705, opened 2015-12-04
Process Multiple files with xcmd


bangroot:
Some external tag generators may be a bit more heavy-weight, but could lower the burden if they were allowed to process multiple files at once. For instance, I'm currently writing an external tags generator for universal-ctags to get better groovy support. Given that groovy already has AST support, I'm just implementing via a simple groovy script (see [here](https://github.com/bangroot/ctags-castle/blob/master/home/.ctags.d/drivers/groovytags) if you care).

My thought would be that the xcmd configuration could support something like `--xcmdb-groovy=groovytags` to represent that this xcmd supports batch processing (handle multiple file inputs). This way, existing xcmd implementations don't break.

masatake:
This was really important pointing out. Thank you.
However, I removed xcmd feature from u-ctags.

PR #1448, opened 2017-06-10
[SELF INCOMPATIBLE] xcmd: remove xcmd feature

masatake:
I'm very sad that I have to make this pull request.

I will remove XCMD from the planed first stable release(6.0).

I like the idea but it makes difficult to keep the consistency of ctags internal.
I guess only a few people using the feature.

Universal-ctags manages kinds more strictly than Exuberant-ctags.
Many of tags generators(xcmds) for specific languages models Exuberant-ctags.
The current Universal-ctags cannot fill the gap in kinds managment.

One of the way to go is that Universal-ctags defines better protocol between
Universal-ctags and xcmds that fill the gaps. However, to define such
protocol, I need much more experiences.

#705 (Process Multiple files with xcmd ) porints the critical weakness of
XCMD feature.

Please, give us feed-back via https://github.com/universal-ctags/ctags/issues/1442


win32/ctags_vs2013.vcxproj.filters
```diff
@@ -621,4 +615,4 @@
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
masatake: I should not do this.

masatake:
Bye, xcmd.

k-takata:
`libexec/drivers/coffeetags` still exists. Shouldn't it removed?

masatake:
Yes, this should be removed. Can I ask you to open an issue about it?
I'm afraid that I cannot remember this.

PR #1449, opened 2017-06-10
Html improve script handling

masatake:
If  '<' operator of JavaScript is used in a script area, html parser
cannot tell correct information about the area to JavaScript guest
parser because html parser parsers inside of the area as if the area
is written in html; html parser recognized '<' operator as a start
of tag.

```
    <script>
      var i;
      for (i = 0; i < 2; i++) {
         console.log ("hello");
      }
    </script>
```

with this change, html parser skips to </script> just after finding <script>.


masatake:
For testing:

```
[jet@localhost]~/var/ctags% cat a.html
<script>
      var i;
      for (i = 0; i < 2; i++) {
         console.log ("hello");
      }
</script>
[jet@localhost]~/var/ctags% ./ctags --extras=+g -o - a.html
i	a.html	/^      var i;$/;"	v
```

masatake:
It seems that Geany people are in busy.
@techee, I will merge this. Feel open an issue if you have a trouble or objection.

PR #1457, opened 2017-06-13
main: put promises array to the default trashbox

masatake:


masatake:
trashbox is bad name. I should call it memcontext or memctx.

PR #1459, opened 2017-06-14
Force init

masatake:
Introduce `--_force-initializing`. option.
u-ctags initializes parsers in lazy way. However, it makes difficult to test initializing stage of parsers.
This new option will help us conducting such tests.


masatake:
I should add a small test.

PR #1462, opened 2017-06-15
main: skip utf8 BOM at the head of input files

### LINKED ISSUE #1422: Set --input-encoding internally if BOM is found (Was: Handle the UTF-8 BOM)
The current Unicode Standard permits the use of the UTF-8 BOM (without recoomending or requiring it), and it is quite widely used in practice (especially on Windows). Any decent modern source code editor can handle it transparently, and I think ctags should do likewise.

If ctags doesn't skip the UTF-8 BOM when reading a source code file, it may fail to find all the valid tags. For example, the IniConf parser will fail to find the first section if it is placed at the beginning of the file (because the first line will not start with a "[").

masatake:
Partially close #1422


docs/news.rst
```diff
@@ -1627,6 +1627,18 @@ output. It is changed to ``-d``. This change is not critical because
...
+Skipping utf-8 BOM
+---------------------------------------------------------------------
+
+The three bytes sequence('\xEF\xBB\xBF') at the head of an input
+file is skipped when parsing.
+
+TODO:
+
+* Do the same in guessing and selecting parser stage.
+* Refect the BON detection to encoding option
```
k-takata: BON -> BOM

main/read.c
```diff
@@ -592,6 +593,36 @@ extern MIO *getMio (const char *const fileName, const char *const openMode,
...
+/* Return true if utf8 BOM is found */
+static bool checkUTF8BOM (MIO *mio, bool skipIfFound)
+{
+	bool r = false;
+	if ((0xEF == mio_getc (mio))
+		&& (0xBB == mio_getc (mio))
+		&& (0xBF == mio_getc (mio)))
+		r = true;
+
+	if (r && skipIfFound)
```
k-takata: Is this condition correct? Shouldn't be `! (r && skipIfFound)`?

masatake: You are correct!
Terribly bad. I wonder why the test case cannot detect this mistake.

...

rewindInputFile is called via resetInputFile.
resetInputFile is called from createTagsForFile.
This means we can remove duplicated code in rewindInputFile and openInputFile from openInputFile...

PR #1466, opened 2017-06-18
colprint: Dyanmically layouting --list output

masatake:
I introduced many --list options.
The column layout of them were statically defined in source code.
The lines were not sorted.

This pull request introduces colprint API.
It can calculate the layout dynamically.
It provides the way to sort the lines.

The all --list- command I introduced are rewritten base on the API.
--_list-roles are named to --list-roles.
--list-langdef-flags is introduced.


masatake:
It seems that LANG affects the result of tests...

PR #1468, opened 2017-06-19
Self spell checking

masatake:
This introduce two targets: `make cspell` and `make dicts`.

The spelling of all commits from master to the current branch are checked.
These are very experimental but this is needed to write man pages that are for releasing 6.0.

`make cspell` can be integrated to `make check`. However, it is not mature enough. I will use this privately for awhile.


configure.ac
```diff
@@ -648,6 +648,18 @@ AS_IF([test "x$enable_yaml" != "xno"], [
 ])
 AM_CONDITIONAL(HAVE_LIBYAML, test "x$have_libyaml" = xyes)
 
+AH_TEMPLATE([HAVE_ASPELL],
+	[Define this value if aspell is available.])
+AC_ARG_ENABLE([aspell],
+	[AS_HELP_STRING([--disable-aspell],
+		[enable aspell support])])
```
k-takata: enable -> disable

masatake: ... sigh. Thank you. Spell checking is not perfect for foolish mistakes. 

I will make a fixup commit.

masatake:
:-O `+2.6%` 

Writing a test case just for improving coverage is not good idea. (But I did so in this PR.).

masatake:
O.K. I will merge this after autosquashing. So I can go back to writing the man page.

masatake:
The coverage goes down again.???

PR #1485, opened 2017-06-25
Improve list output

masatake:


masatake:
XFMT field in `--list-fields` should removed. I will revise this when writing the option in ctags.1.

PR #1490, opened 2017-06-27
Suppress compiler warnings

masatake:


parsers/html.c
```diff
@@ -346,8 +346,8 @@ static bool skipScriptContent (tokenInfo *token, long *line, long *lineOffset)
 	bool found_start = false;
 	bool found_script = false;
 
-	long line_tmp[2];
-	long lineOffset_tmp[2];
+	unsigned long line_tmp[2];
+	unsigned long lineOffset_tmp[2];
```
codebrainz: Out of curiosity, what warning does this suppress?

masatake: Oh, this is mistake. This doesn't suppress the original warning. The original warnings are:
```
parsers/html.c:439:6: warning: ‘lineOffset_tmp[1]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
      makePromise ("JavaScript", startLineNumber, startLineOffset,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          endLineNumber, endLineOffset, startSourceLineNumber);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/html.c:439:6: warning: ‘line_tmp[1]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

I will remove this commit and add a fix for this warning.

k-takata:
MinGW reports another warning:
https://ci.appveyor.com/project/universalctags/ctags/build/1.0.120/job/e861xd8h2ri4f41p#L355

```
  CC       parsers/ctags-html.o
parsers/html.c: In function 'readTag':
parsers/html.c:439:6: warning: 'lineOffset_tmp[1]' may be used uninitialized in this function [-Wmaybe-uninitialized]
      makePromise ("JavaScript", startLineNumber, startLineOffset,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          endLineNumber, endLineOffset, startSourceLineNumber);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/html.c:439:6: warning: 'line_tmp[1]' may be used uninitialized in this function [-Wmaybe-uninitialized]
```

I think this is false positive, but could you look into this?

masatake:
I will withdraw  	ba1ce89. I will verify the control and data flow of skipScriptContent.

Thank you, guys. I should take more time.

codebrainz:
It's legit, line_tmp[1] is only assigned in a conditional/if but is used unconditionally. The fix would be to initialize it like `long line_tmp[2] = {0,0};`

masatake:
`return type != TOKEN_EOF;` at the end of function seems meaninglessly complicated.
`return found_script;` may be better.

masatake:
Just cleanup. The original warning is still there.

masatake:
skipScriptContent:

`*line` is used called only when skipScriptContent() returns true.
That means found_script is true.
If true is set to found_script, line_tmp[1] is set to *line.
This assignment is done only when found_start is true.
If true is set to found_start, line_tmp[0] is set to line_tmp[1].
The return value of getInputLineNumber () is stored to line_tmp[0] at
the head of loop; this is done always.

Am I wrong?

masatake:
Oh, this is about lineOffset, not about line.

masatake:
I think changes proposed here is o.k.
The warning we talked yesterday is separately tracked in #1496.

PR #1494, opened 2017-06-29
Add mremap to seccomp whitelist.

hanwen:


masatake:
Thank you.

masatake:
... I wonder when mremap is called. In malloc?

hanwen:
realloc I think.

masatake:
> realloc I think.

Oh, I see. I have never think about the implementation of realloc.

PR #1495, opened 2017-06-29
Multi pass parsing over multi source files

### LINKED ISSUE #80: Add dual-pass capability

masatake:
Close #80.

@vhda, how do you think about these changes? See the commits whose logs are started with ` 	[TEMPORARY] `.
I will update docs/internal.rst after getting your approval.

In my version, SystemVerilog emits the most of all tags in the first (-1) pass. However, you can pass some of them two the next mm pass via "barrel". In the seond (-2) pass, tags in the barrel are stored to the keyword table.

For testing I choose typedef kind tags. In the -2 pass, SystemVerilog parser can recognize the next token of typedef'ed type. If such token is recognized in class context, the token can be tagged as "member" kind.
I didn't choose the kind name "variable" though 'v' was chosen. Because a member of class is not a variable. 

This will not work well in interactive mode. I have to research the area more.

```
   $ cat a.sv
    class test;
       t_user t_user_memb;
    endclass
    $ cat b.sv
    typedef int t_user;
    $ ./ctags -o -  a.sv b.sv
    t_user	b.sv	/^typedef int t_user;$/;"	T
    t_user_memb	a.sv	/^   t_user t_user_memb;$/;"	v	class:test
    test	a.sv	/^class test;$/;"	C
```

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -62,7 +62,8 @@ typedef enum {
 	K_PROGRAM,
 	K_PROTOTYPE,
 	K_PROPERTY,
-	K_TYPEDEF
+	K_TYPEDEF,
+	K_MEMBER,
 } verilogKind;
 
 typedef struct {
@@ -123,7 +124,8 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'P', "program",   "programs" },
  { false,'Q', "prototype", "prototypes" },
  { true, 'R', "property",  "properties" },
- { true, 'T', "typedef",   "type declarations" }
+ { true, 'T', "typedef",   "type declarations" },
+ { true, 'v', "member",    "member elements" },
 };
 
 static const keywordAssoc KeywordTable [] = {
@@ -1097,7 +1099,20 @@ static void findTag (tokenInfo *const token)
 	}
 	else if (token->kind == K_TYPEDEF)
 	{
-		processTypedef (token);
+		if (strcmp (vStringValue (token->name), "typedef") == 0)
+			processTypedef (token);
+		else if (currentContext->kind == K_CLASS)
+		{
+			int c;
+			c = skipWhite (vGetc ());
+			if (isIdentifierCharacter (c))
+			{
+				readIdentifier (token, c);
+				token->kind = K_MEMBER;
+				createTagCommon (token, true);
+				/* TODO: typeref field can be filled. */
+			}
+		}
 	}
 	else if (token->kind == K_CLASS)
 	{
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -24,6 +24,7 @@
 #include "debug.h"
 #include "cpreprocessor.h"
 #include "keyword.h"
+#include "mm.h"
 #include "options.h"
 #include "parse.h"
 #include "read.h"
@@ -194,6 +195,8 @@ static const keywordAssoc KeywordTable [] = {
 
 static tokenInfo *currentContext = NULL;
 
+static int systemVerilogPassCount;
+
 /*
  *   FUNCTION DEFINITIONS
  */
@@ -336,6 +339,24 @@ static void buildKeywordHash (const langType language, unsigned int idx)
 		const keywordAssoc *p = &KeywordTable [i];
 		if (p->isValid [idx])
 			addKeyword (p->keyword, language, (int) p->kind);
+	}
+}
+
+static void extendSystemVerilogKeywords (langType language, Barrel *barrel)
+{
+	unsigned int count = countEntryInBarrel (barrel);
+	for (unsigned int i = 0; i < count; i++)
+	{
+		tagEntryInfo *e = getEntryInBarrel (barrel, i);
+
+		if (e->langType == language
+			/* TODO: Ugly */
+			&& ((e->kind - SystemVerilogKinds) == K_TYPEDEF))
+		{
+			addKeyword (eStrdup (e->name), language, K_TYPEDEF); /* MEMLEAK */
+			verbose ("SystemVerilog extends keyword table: %s\n",
+					 e->name);
+		}
 	}
 }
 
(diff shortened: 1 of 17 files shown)


masatake:
After thinking I found barrel is not needed here. Adding typedefed typename to the keyword table in the -1 pass will be enough...

masatake:
Other interesting application of mm is "unknown" kind of python.
"unknown" kind objects tagged in -1st pass may be solved in -2nd pass...

masatake:
What I wrote is something like linker. So I can borrow many concepts and ideas from linker.

vhda:
I have the intention of dedicating part of my weekend to Universal Ctags.
I'll reply ASAP :)

masatake:
I updated the documents.

docs/internal.rst
```diff
@@ -420,3 +420,110 @@ An example can be found in DTS parser:
 
 Setting `requestAutomaticFQTag` to `TRUE` implies setting
 `useCork` to `TRUE`.
+
+Multi passes parsing over multi input files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The main part of Universal-ctags provides APIs for Multi passes
+parsing over Multi source files (MM). The main part applies
```
vhda: Don't understand the meaning of "MM". First "M" is for multi, but what is the second for?
Typically you would do something like: "Multi source Files (MF)". That is, capitalize the words used in the abbreviation.

main/keyword.h
```diff
@@ -25,7 +25,13 @@
 /*
 *   FUNCTION PROTOTYPES
 */
+
+/* `string' should be allocated statically. */
 extern void addKeyword (const char *const string, langType language, int value);
+
+/* addKeywordStrdup does strdup `string'.
+   Duplicated string is freed in  freeKeywordTable() */
```
masatake: Thank you. I will make a fixup! commit.

parsers/verilog.c
```diff
@@ -122,7 +124,8 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'P', "program",   "programs" },
  { false,'Q', "prototype", "prototypes" },
  { true, 'R', "property",  "properties" },
- { true, 'T', "typedef",   "type declarations" }
+ { true, 'T', "typedef",   "type declarations" },
+ { true, 'v', "member",    "member elements" },
```
masatake: Designing kinds is the most important task in ctags development.
I don't know Verilog. So if you are o.k., I'm o.k.
@RadekRR, this is the chance to reflect your idea to the software you are using:-).

masatake:
@vhda, I feel gaps between what you want and what I wrote.

Could you look at

Please, look at the document explaining MM. 
https://github.com/universal-ctags/ctags/pull/1495/commits/6ea782371d7dc78b884a7882dd4dedd81a03f63c

https://github.com/masatake/ctags/blob/6ea782371d7dc78b884a7882dd4dedd81a03f63c/docs/barrel.svg

docs/internal.rst
masatake: MM means "M"ulti pass parsing over "M"ulti source files.
Do I make sense?

parsers/verilog.c
```diff
@@ -1151,6 +1211,18 @@ static void findVerilogTags (void)
...
+	verbose ("SystemVerilog enters pass %d\n", passCount);
+	systemVerilogPassCount = passCount;
+	findVerilogTags ();
+
+	if (passCount < 0)
+		return RESCAN_NONE;
+	else
+		return RESCAN_MM;
+}
+
```
masatake: > Why make the parser aware of passes?

The main part doesn't know which source file(s) should be rescanned.

docs/internal.rst
vhda: Sorry, missed the first M... ups!

masatake:
I read again the original discussion.

My understanding of what you want.

pass 1. just making a hashtable and fill it with type names (or something names). You want to make no tag entries in this stage.

pass2. making tag entries and emiting them to tags file.

These flow can be implemented on MM. What you should is just ignoring Barrel. You can create SystemVerilog own type name table with functions declared in hashtable.h. SystemVerilog parser
must return RESCAN_MM.

However, you should emit  some of tag entries in pass1 if possible. Because, newly introduced `--interactive` mode may expect per-source file processing. It means in the mode ctags runs only pass 1(-1). So the user of `--interactive` mode may recognize SystemVerilog parser does nothing.

wvandamm:
Hi @masatake ,

Could I please inquire what the status is of this pull request? The functionality proposed seems highly interesting in any case!

Also a related question: the C++ parser appears to already be able to identify definitions for variables or objects of a type created with typedef or class. If I understand the discussion well that's something that this pull request was intending to add as well. For the SystemVerilog parser the same functionality does not appear to work yet though. So am I correct in assuming that the C++ parser has solved this problem in the parser itself? Or has some framework been added that other parsers could make use of to add similar functionality?

Thanks!

Wim

masatake:
> Could I please inquire what the status is of this pull request? The functionality proposed seems highly interesting in any case!

There is no progress other than changes I proposed here.

C++ parser may use heuristics. It is not perfect. e.g. about handling template variables.
See files under Units/parser-cxx.r/*.b.

Before implementing Multi-passes/Mult-files parsing, I have to improve the infrastructure for Multi-passes/Singl-files parsing. See #2115.

If you want mm seriously, and want to implement, I will explain what kind of issues are.

PR #1503, opened 2017-07-03
Fix indent

k-takata:
This fixes only important indentation mistakes.
(Many minor mistakes still remain.)


masatake:
LGTM. Though you are the owner(?) of this PR, but I merged this because the indentation changes cause conflicts frequently. Generally the owner should merge but this is special case.

PR #1504, opened 2017-07-06
Misc changes mainly about --list-* options

masatake:
These changing have been mage during updating man pages.


masatake:
circleci comes back!

PR #1507, opened 2017-07-10
Verilog/enum

vhda:
Some improvements to the unit tests I had laying around plus the completion of the half baked work I also had for enumerators and typedefs. The support for these two types is still not perfect, but it's an improvement :)

Will now start working on what we've been discussing lately.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -55,12 +55,14 @@ typedef enum {
 	K_ASSERTION,
 	K_CLASS,
 	K_COVERGROUP,
+	K_ENUM,
 	K_INTERFACE,
 	K_MODPORT,
 	K_PACKAGE,
 	K_PROGRAM,
 	K_PROTOTYPE,
 	K_PROPERTY,
+	K_STRUCT,
 	K_TYPEDEF
 } verilogKind;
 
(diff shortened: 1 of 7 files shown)


masatake:
```
 { true, 'c', "constant",  "constants (define, parameter, specparam)" },
```

The description for kinds must be updated: adding "enum" to it.

parsers/verilog.c
```diff
@@ -116,12 +118,14 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'A', "assert",    "assertions" },
  { true, 'C', "class",     "classes" },
  { true, 'V', "covergroup","covergroups" },
+ { true, 'E', "enum",      "enumerators" },
```
vhda: `enum` kind is already defined here, but for variables to type `enum`.
The possible values that that variable can take were introduced as constants.

vhda:
But I really should include "enum values" in that list.
I'll do so as soon as you finish your review, to avoid multiple iterations.

masatake:
Other than the I already commented, it looks good to me.

BTW, u-ctags allows a parser to have own fields.

Consider the following input:
```
enum bit [1:0] {
    ...
} enum_complex;
```

Potentially u-ctags emits following tags:

```
enum_complex	input.sv	/^    } enum_complex;$/;"	E	class:test enumType:bit
```

or

```
enum_complex	input.sv	/^    } enum_complex;$/;"	E	class:test enumType:bit[1:0]
```

(I'm not sure which one is better.)

If you get interested, search `attachParserField` in parsers/maven2.c.

vhda:
I decided against `attachParserField`, for now. Will look into this feature later.

masatake:
Thank you.

PR #1512, opened 2017-07-24
Don't depend on git being installed.  Switch to portable find

derekschrock:


masatake:
I have many test cases that are not added to git repository.
Your patch adds them to  makefiles/test-cases.mak. None may want them.

derekschrock:
If I understand your comment you're saying you have other tests in your local tree that you don't want to be included?  If so would it be best for you to have these else where?  How can anyone take into account your local tree vs what is in the remote repo.

masatake:
> If I understand your comment you're saying you have other tests in your local tree that you don't want to be included? 

Yes.

>  If so would it be best for you to have these else where? 

I don't think so.

> How can anyone take into account your local tree vs what is in the remote repo.

These conflicts are reported by git each time when git pull or git checkout.

I don't understand why you need this kind of portability.
I think you are using git for getting source file of u-ctags. So you may have git.
Could you tell me the background of this patch?

derekschrock:
On Sun, Jul 23, 2017 at 11:28:23PM EDT, Masatake YAMATO wrote:
> These conflicts are reported by git each time when git pull or git checkout.
> 
> I don't understand why you need this kind of portability.
> I think you are using git for getting source file of u-ctags. So you may have git.
> Could you tell me the background of this patch?

In the process of porting universal-ctags to a FreeBSD port this was one
issue I ran in to.  It's possible this can be ignored since this is used
to generate distfiles and not critical for building the software?

It's also possible to include git in as a build dependency in the port 
however it seems like overkill when you can get the same data via 
portable POSIX utilities.  This was the reason for the patch.

Do you feel it's safe to use ignore makefiles/test-cases.mak, by either 
commenting it out in Makefile.am or making it an empty file, in the
FreeBSD port?

masatake:
git is needed when you get source code from git.
I guess the environment(E1) where getting the source code and environemnt(E2) where running make are different in your case.

In such case you should do run "make dist"  in E1. You will get a tar.gz file. Then copy the tar.gz file to E2.
In the E2 you don't need git to build a binary.

If we release a tar ball, you don't need such step. However, I'm working for releasing but the official tar ball is not available yet.

> In the process of porting universal-ctags to a FreeBSD port this was one issue I ran in to. It's possible this can be ignored since this is used to generate distfiles and not critical for building the software?

I think so but I have nerver tested such condition.

> It's also possible to include git in as a build dependency in the port however it seems like overkill when you can get the same data via portable POSIX utilities. This was the reason for the patch.

Such dependency is needed because you try to build ctags from code taken from git repository.
If this project releases official tar.gz, you don't need such dependency.  

> Do you feel it's safe to use ignore makefiles/test-cases.mak, by either commenting it out in Makefile.am or making it an empty file, in the FreeBSD port?

I feel it is safe.

derekschrock:
On Mon, Jul 24, 2017 at 02:13:01AM EDT, Masatake YAMATO wrote:
> Such dependency is needed because you try to build ctags from code taken from git repository.
> If this project releases official tar.gz, you don't need such dependency.  

No, you can download tarballs of a given commit via github from a direct
URL.  FreeBSD's port infrastructure doesn't use git to download github
(or gitlab) repos.  They're direct URLs so normal checksum validation
can still occur.  Versioning can be a little odd sometimes with
non-tagged releases (rolling release) but once you wrap your head around
the idea it's pretty sane.

For example:
https://codeload.github.com/universal-ctags/ctags/tar.gz/5567535e

Can be used to download universal-ctags/ctags commit 5567535e.  So git
is not required inside a basic FreeBSD build environment to pull in
upstream files for github sources.  Just an http[s] client and checksum
validation.

You can find more information here:
https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/book.html#makefile-master_sites-github-description

> I feel it is safe.

Ok I think this can be closed then we can work around autogen.sh with
FreeBSD port's builtin autoreconf support.

masatake:
I see what happens on your side. How about following patch?

```
diff --git a/misc/dist-test-cases b/misc/dist-test-cases
index f09e9cf..d520a42 100755
--- a/misc/dist-test-cases
+++ b/misc/dist-test-cases
@@ -8,7 +8,10 @@
 echo "# -*- makefile -*-"
 echo "# Generated by $0" &&
 echo "EXTRA_DIST += misc/units \\" &&
-git ls-files | grep 'Units\|Tmain' | sed -e 's/$/\\/' -e 's/^/    /' &&
-echo '$(NULL)' &&
-
+if type git > /dev/null 2>&1; then
+    git ls-files | grep 'Units\|Tmain' | sed -e 's/$/\\/' -e 's/^/    /' &&
+	echo '$(NULL)'
+else
+    echo '$(NO_GIT_EMPTY)'
+fi &&
 exit $?
```

derekschrock:
On Mon, Jul 24, 2017 at 02:54:55AM EDT, Masatake YAMATO wrote:
> I see what happens on your side. How about following patch?

Yes, this seems to be a better way to handle this issue.

PR #1515, opened 2017-07-26
build-sys: fix wrong usage of AC_CHECK_FUNCS used in checking mblen

masatake:
The original code defines `HAVE_MBLEN=1` when mblen is not found.
It should be defined only when the function is found.

AC_CHECK_FUNCS defines the macro automatically when it is defiend
explicitly in the parameter list. So just specifying the name of
function is enough.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -707,7 +707,7 @@ fi
 # Output files
 # ------------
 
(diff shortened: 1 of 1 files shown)


masatake:
This implies the C code surrounded by #ifdef HAVE_MBLEN - #endif is not tested

masatake:
Does anyone have interest to write a test case for mblen?

PR #1521, opened 2017-08-04
circleci: run db_recover in "Install build tools" step

masatake:
Building at circleci fails randomly with following message:

    BDB1539 Build signature doesn't match environment
    failed loading RPMDB

As far as reading [1] and [2], running db_recover may
fix the failure.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=555315
[2] https://bugzilla.redhat.com/show_bug.cgi?id=553998

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,6 +9,9 @@ jobs:
        - run:
            name: Install build tools
            command: |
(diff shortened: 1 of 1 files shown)


masatake:
This change doesn't help us.

PR #1530, opened 2017-08-19
JavaScript: don't make EOF be a part of template string

### LINKED ISSUE #1493: fuzz test generates "(null)" json output
2017/06/29 01:10:09 post "{\"command\":\"generate-tags\",\"filename\":\"Units/afl-fuzz.r/github-issue-539-js.d/input.js\",\"size\":6}\n"
2017/06/29 01:10:09 let`	

2017/06/29 01:10:09 read (null)
2017/06/29 01:10:09 ignoring /home/hanwen/bin/universal-ctags error: unmarshal((null)): invalid character '(' looking for beginning of value

masatake:
EOF char was mistakenly appended to a token of template string.

Close #1493.


parsers/jscript.c
```diff
@@ -511,11 +511,15 @@ static void parseTemplateString (vString *const string)
 		c = getcFromInputFile ();
 		if (c == '`')
 			break;
-		vStringPut (string, c);
+
+		if (c != EOF)
```
b4n: should rather be in the previous `if` clause (even though it kind of makes the loop condition useless, but still):
```c
		if (c == '`' || c == EOF)
			break;
```

masatake: Thank you. Updated.

PR #1541, opened 2017-08-31
CXX: Improve namespace parsing

### LINKED ISSUE #1539: universal ctags not parsing c++ headers
I have been in the habit of running exuberant ctags with the following command:

`ctags -Re --language-force=c++ -h=".h.H.hh.hpp.hxx.h++.inc.def." --c++-kinds=+l --file-scope=no --tag-relative=no -f TAGS /usr/include/c++/7.1.0`

This creates a 2.1 MB TAGS file that I can load up in emacs and look up tags for classes in the standard libraries (e.g. vector).

But running the same command with universal ctags creates a file of size 469k, and that lacks vector, for example.

I'm not sure if this is due to differences in the parsers used, or perhaps my command line options are outdated now?

pragmaware:
Be more tolerant when parsing namespaces.
Fixes parsing of stl headers which contain a macro after the namespace name.
Fixes #1539 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 *.od
 *.pdf
 *.ssexwsp
+*.ssexwsp.user
 *.TMP
 *.tmp
 *~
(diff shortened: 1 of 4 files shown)


masatake:
LGTM.

PR #1542, opened 2017-08-31
byte oriented multiple state regex meta parser and pappetManifest parser

masatake:
See 
https://github.com/universal-ctags/ctags/pull/1542/commits/3a11ff56e20bad163413ccd31d9b5783ab14d7a0


masatake:
@bitterjug,

Your parser cannot ignore `greeting2` of
```
greeting =
    "Hello World!"

{-
greeting2 =
    "Don't capture me"
-}
```

This is the limitation of line oriented regex parser.
A parser defined with --_mtable-regex-<LANG> and --_tabledef-<LANG> can overcome this limitation.
It has stack for tracking the state of parsing.

masatake:
None stops me introducing this crazy feature.
I will not write this in manual in version 6.0.

bitterjug:
@masatake I followed your notes on how this works. It's great. A general purpose FSM parser with regular expression matching is very exciting, and takes  ctags optlib to another level of power. I don't think anyone should stop you adding the feature. You should add it, and we should experiment with what we can achieve with it. 

Having said all that,  I don't think optlib, as you currently have it, is the right way to define these. I want to suggest that this feature is so very different and more sophisticated than what optlib has been used for that it warrants a more friendly syntax than gathering command line arguments together into a file. 

Optlib is ugly. Taking valid command line arguments and putting them into a file effectively defines a new file format for defining optlib parsers. And it looks terrible. We end up with a line-oriented syntax for what ends up as a kind of key-value config where the values can be somewhat complex with  multiple fields. Defining a regexp FSM parser for any non-trivial language will require a bunch of config. Nobody is *really* going to do that with command line options to `ctags`; there would be too many, so anyone using this will *always* be writing a config file. I suggest you use a more friendly config file format. It doesn't matter if it's not compatible with optlib: it's really a completely new feature. 

I think if would be easier to write these in something like YAML. What do you think?

PR #1548, opened 2017-09-07
Fix things reported by cppcheck

masatake:


main/colprint.c
```diff
@@ -273,7 +273,7 @@ void colprintLineAppendColumnInt  (struct colprintLine *line, unsigned int colum
 {
 	char buf[12];
 
-	snprintf(buf, 12, "%d", column);
+	snprintf(buf, 12, "%u", column);
```
codebrainz: The standard doesn't specify whether `char` is signed or unsigned, so cppcheck should also warn for the new version.

[C99 N1256 draft](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf) 6.2.5/15:

> The implementation shall define char to have the same range, representation, and behavior as either signed char or unsigned char.

masatake: ???

Here `column` is unsigned int, not char.

codebrainz: Oops, I just quickly scanned it in the 15 minute window after waking up and before heading off to work, must've seen the `char` above and got confused, sorry for the noise. I should make sure I'm fully awake up before reviewing commits :) BTW, there's a typo in the commit message, if it matters.

masatake: @codebrains, thank you.

PR #1553, opened 2017-09-12
Update Makefile.am

vpereira:
add SECCOMP_CFLAGS to the Makefile.am file

IMPORTANT:

Do we have to check if SECCOMP is enabled? We have checks if libyaml and libxml are enabled but not for SECCOMP.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -64,6 +64,7 @@ ctags_CFLAGS  += $(LIBXML_CFLAGS)
 ctags_CFLAGS  += $(JANSSON_CFLAGS)
 ctags_CFLAGS  += $(LIBYAML_CFLAGS)
 ctags_CFLAGS  += $(ASPELL_CFLAGS)
+ctags_CFLAGS  += $(SECCOMP_CFLAGS)
 
 ctags_LDADD  =
 ctags_LDADD += $(LIBXML_LIBS)


masatake:
> IMPORTANT:
> 
> Do we have to check if SECCOMP is enabled? We have checks if libyaml and libxml are enabled but not for SECCOMP.

??? Are you talking about the availability of seccomp in build-environment?
If yes, it is already checked in configure.ac.

Or are you talking about the advertising the feature availablity in ctags itself?
If yes, it is already done with --list-features option:

```
[yamato@master]/tmp% u-ctags --list-features
u-ctags --list-features
#NAME             DESCRIPTION
aspell            linked with code for spell checking (internal use)
debug             TO BE WRITTEN
interactive       accepts source code from stdin
json              supports json format output
multibyte         TO BE WRITTEN
option-directory  TO BE WRITTEN
regex             can use regular expression based pattern matching
sandbox           linked with code for system call level sandbox
wildcards         can use glob matching
xpath             linked with library for parsing xml input
yaml              linked with library for parsing yaml input
```

sandbox is for seccomp.

Anyway I will merge this PR. Thank you.

vpereira:
I was talking to have something

  if HAVE_SECCOMP
   add_seccomp_header
  add_seccomp_lib 

because you can pass to configure ```--disable-seccomp``` and it may or may not explode if we try unconditionally to add seccomp headers and lib

masatake:
Oh, I see you are talking about conditions in Makefile.am.

Adding source files and header files are guarded with `if` condition like:
```
if HAVE_LIBYAML
PARSER_SRCS += $(YAML_SRCS)
PARSER_HEADS += $(YAML_HEADS)
endif
```

You may wonder why no 'if HAVE_SECCOMP'.

This is o.k. seccomp.o can be linked always because secomp.c itself has big ifdef block.
seccomp.c provides a stub if no libseccomp is available.
See seccomp.c.

PR #1556, opened 2017-09-20
Remove non-posix make flag --no-print-directory from autogen.sh

mqudsi:
This fixes the remaining BSD build issues after the recent bmake fixes
were merged into master. Discussed in #1554. Closes #1554.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -5,7 +5,7 @@ set -xe
 type autoreconf || exit 1
 type pkg-config || exit 1
 
-ctags_files=`make -f makefiles/list-translator-input.mak --no-print-directory`
+ctags_files=`make -f makefiles/list-translator-input.mak`
 misc/dist-test-cases > makefiles/test-cases.mak && \
     if autoreconf -vfi; then
 	if type perl > /dev/null; then


mqudsi:
Just for reference, as seen in the GNU Make man page:

```
       -w, --print-directory
            Print a message containing the working directory before and after
            other processing.  This may be useful for tracking down errors
            from complicated nests of recursive make commands.

       --no-print-directory
            Turn off -w, even if it was turned on implicitly.
```

I'm not sure what implicitly turns on `-w`. Perhaps `-d` which prints debugging info (which isn't specified). We should be OK without `--no-print-directory`.

masatake:
LGTM.

codebrainz:
> I'm not sure what implicitly turns on -w.

It's explained somewhat in [the online docs](https://www.gnu.org/software/make/manual/html_node/Options-Summary.html) after the part you quoted:

> Print a message containing the working directory both before and after executing the makefile. This may be useful for tracking down errors from complicated nests of recursive make commands. See Recursive Use of make. __(In practice, you rarely need to specify this option since ‘make’ does it for you; see The [‘--print-directory’](https://www.gnu.org/software/make/manual/html_node/_002dw-Option.html#g_t_002dw-Option) Option.)__

So I guess because of the `-f` it's considered a submake, or maybe because of the specified make file including other make files. Either way, it's definitively the default behaviour I've observed in my own Make files.

masatake:
Thank you.

PR #1557, opened 2017-09-21
asm: Add .equiv define

markferry:
Simple change to recognize .equiv assembler directive.
  (http://ftp.gnu.org/old-gnu/Manuals/gas/html_chapter/as_7.html#SEC87)


masatake:
Thank you.

1. could you add the reference "ftp://ftp.gnu.org/old-gnu/Manuals/gas/html_chapter/as_7.html#SEC87" 
not only to the comment of this pull request but also the commit log. It is helpful to understand the change with git log and git blame.

2. could you add a test case to Units/parser-asm.r/ ?
    See http://docs.ctags.io/en/latest/units.html to know how to write a test case.

masatake:
Please, forget my lsat comment. I will make a pull request to your branch.

masatake:
I guess you know well about gas.
How do you think about .eqv directive?
```

7.26 `.eqv SYMBOL, EXPRESSION'
==============================

The `.eqv' directive is like `.equiv', but no attempt is made to
evaluate the expression or any part of it immediately.  Instead each
time the resulting symbol is used in an expression, a snapshot of its
current value is taken.
```

markferry:
Thanks @masatake. I hadn't noticed `.eqv`. Looks like it's appropriate to incorporate it. I'll also add test cases and better commit messages. Updated patches to follow...

masatake:
O.k. I change my mind again. I will wait for your updating.
Feel free to ask me any question about writing a test case.

masatake:
Thank you very much. I squashed three commits into one. (I made a typo in the commit log. Sorry.)

PR #1565, opened 2017-10-03
go: improve scope fields

masatake:
Package and receiver are considered when building scope fields.

Let's see an example.

For the input:

```
package main
import (
	"fmt"
)

type Point struct { X, Y int }

func (p *Point) Render() {
	fmt.Printf("<%d,%d>\n", p.X, p.Y)
}

func main() {
	p := &Point{X: 5, Y: 12}
	p.Render()
}
```

Without the changes:
```
main	foo.go	/^package main$/;"	kind:package	line:1	language:Go
Point	foo.go	/^type Point struct { X, Y int }$/;"	kind:struct	line:6	language:Go
X	foo.go	/^type Point struct { X, Y int }$/;"	kind:member	line:6	language:Go	scope:struct:Point
Y	foo.go	/^type Point struct { X, Y int }$/;"	kind:member	line:6	language:Go	scope:struct:Point
Render	foo.go	/^func (p *Point) Render() {$/;"	kind:func	line:8	language:Go	signature:()
main	foo.go	/^func main() {$/;"	kind:func	line:12	language:Go	signature:()
foo.go	foo.go	1;"	kind:file	line:1	language:Go	extras:inputFile	end:15
```

With the changes:
```
main	foo.go	/^package main$/;"	kind:package	line:1	language:Go
Point	foo.go	/^type Point struct { X, Y int }$/;"	kind:struct	line:6	language:Go	scope:package:main
X	foo.go	/^type Point struct { X, Y int }$/;"	kind:member	line:6	language:Go	scope:struct:main.Point
Y	foo.go	/^type Point struct { X, Y int }$/;"	kind:member	line:6	language:Go	scope:struct:main.Point
Point	foo.go	/^func (p *Point) Render() {$/;"	kind:unknown	line:8	language:Go	scope:package:main	role:receiverType	extras:reference
Render	foo.go	/^func (p *Point) Render() {$/;"	kind:func	line:8	language:Go	scope:unknown:main.Point	signature:()
main	foo.go	/^func main() {$/;"	kind:func	line:12	language:Go	scope:package:main	signature:()
foo.go	foo.go	1;"	kind:file	line:1	language:Go	extras:inputFile	end:15
```


parsers/go.c
```diff
@@ -517,6 +533,7 @@ static int makeTag (tokenInfo *const token, const goKind kind,
 		e.extensionFields.signature = argList;
 
 	e.extensionFields.scopeIndex = scope;
+
```
masatake: This line can be removed.

masatake:
Could you wait to review these changes awhile?
I found that code picking up a receiver is not robust enough.

masatake:
Updated. Please, review the changes.

parsers/go.c
```diff
@@ -692,6 +672,7 @@ static void parseConstTypeVar (tokenInfo *const token, goKind kind)
 	do
 	{
 		tokenInfo *typeToken = NULL;
+		scope = scope_original;
```
techee: I think re-setting the scope variable here every loop and setting it to a different value below makes the code a bit confusing. I would suggest keeping the contents of "scope" without change and introduce new variable such as "member_scope" whose value is set in makeTag() below and pass this one to parseStructMembers().

parsers/go.c
```diff
@@ -576,6 +639,8 @@ static void parseFunctionOrMethod (tokenInfo *const token, int scope)
 		if (isType (token, TOKEN_OPEN_CURLY))
 			skipToMatched (token);
 	}
+	else if (receiver_type_token)
```
techee: "else" can be removed here and then you can drop the deleteToken() above and reuse the below one for that case too.

parsers/go.c
```diff
@@ -531,19 +547,60 @@ static int parsePackage (tokenInfo *const token)
 		return CORK_NIL;
 }
 
+static tokenInfo * parseReceiver (tokenInfo *const token)
```
techee: I think this function can be simplified a bit and made more readable. I'll prepare a patch - it's easier than writing what I think should be changed.

techee: The patch is here

https://github.com/techee/ctags/commits/scope_improvement

Feel free to squash it with your commit if you like it.

techee:
LGTM apart from the minor comments above.

masatake:
@techee, thank you very much. Especially about rewriting parserReceiver(). I rewrote it three times before opening this pull request. So I am surprised that one can improve it. Writing simply is very difficult to me.
All comments you gave me are reflected in the new commits.

techee:
@masatake Looks good.

No problem reviewing the patches - it's the advantage of the review that the reviewer sees the patches with fresh eyes and doesn't see the history of the patch. The author of the patch is mentally fixed on certain solution and doesn't see the alternatives. Happened to me many times when @b4n reviewed my code :-).

masatake:
I will squash the changes into two and merge them.

PR #1567, opened 2017-10-05
main: trim whitespace prefix from tag name

### LINKED ISSUE #1141: whiespace prefixed tags and orders in tags file
```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.html 
cat /tmp/foo.html 
<h1> ABC</h1>

[yamato@x201]~/var/ctags-github% ./ctags /tmp/foo.html 
./ctags /tmp/foo.html 
[yamato@x201]~/var/ctags-github% cat tags 
cat tags 
 ABC    /tmp/foo.html   /^<h1> ABC<\/h1>$/;"    h
!_TAG_FILE_FORMAT   2   /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED   1   /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
!_TAG_PROGRAM_NAME  Universal Ctags /Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL   https://ctags.io/   /official site/
!_TAG_PROGRAM_VERSION   0.0.0   /25bad34/
```

Look at the whiespace before `ABC`.

If a tag name is started with a whitespace, it is printed in tags file before pseudo tags.
I guess this behavior make client tools get into crazy.

Could you give me comment and ideas what we can do?

(Please change the tile of this issue if you can find better one.)

masatake:
Close #1141

Original issue:

    [yamato@x201]~/var/ctags-github% cat /tmp/foo.html
    cat /tmp/foo.html
    <h1> ABC</h1>

    [yamato@x201]~/var/ctags-github% ./ctags /tmp/foo.html
    ./ctags /tmp/foo.html
    [yamato@x201]~/var/ctags-github% cat tags
    cat tags
     ABC    /tmp/foo.html   /^<h1> ABC<\/h1>$/;"    h
    !_TAG_FILE_FORMAT   2   /extended format; --format=1 will not append ;" to lines/
    !_TAG_FILE_SORTED   1   /0=unsorted, 1=sorted, 2=foldcase/
    !_TAG_PROGRAM_AUTHOR    Universal Ctags Team    //
    !_TAG_PROGRAM_NAME  Universal Ctags /Derived from Exuberant Ctags/
    !_TAG_PROGRAM_URL   https://ctags.io/   /official site/
    !_TAG_PROGRAM_VERSION   0.0.0   /25bad34/

    Look at the whiespace before ABC.

    If a tag name is started with a whitespace, it is printed in tags file before pseudo tags.
    I guess this behavior make client tools get into crazy.

This patch trims lower characters than '!' in Ascii order at the head
of tag name.


main/entry.c
```diff
@@ -1280,8 +1280,37 @@ static void makeTagEntriesForSubwords (tagEntryInfo *const subtag)
...
+		else
+			enter_body = true;
+
+		if (prefix_length == 0)
+			return NULL;
+
+		total_length++;
+	}
+	return eStrndup (name + prefix_length, total_length - prefix_length);
+}
```
codebrainz: I'm not sure if I misunderstood this function, but it could possibly be implemented more simply like this:

```c
static char *trimPrefixedWhitespaces(const char *name)
{
  const char *start = name;
  for (; isspace(*start); start++)
    ;
  return eStrndup(start, strlen(start));
}
```

It would be surprising to me that the ctags code doesn't already have this kind of function though.

masatake: I should do as you wrote.

At the first time, I used memmove not to allocate an extra memory.
So my trimPrefixedWhitespaces is a bit complicated.

Thank you for the suggestion.

masatake:
ctags has a function for trimming whitespaces at the head of string. However, it is vString based.

codebrainz:
@masatake yeah I noticed that one. I wonder if there are other places in the code doing this inline or with local helper functions? IME, this commonly happens when generic utility functions are added locally into specific source file.

masatake:
nextStringArg in main/args.c is very similar.
O.k. I will introduce a utility function for factoring.

Thank you.

codebrainz:
Nice catch!

I rather like the implementation and API of `g_strchomp` and `g_strchug` [provided by Glib](https://git.gnome.org/browse/glib/tree/glib/gstrfuncs.c). Having the caller pass in a mutable string seems to make it simpler to use, and leaves the memory allocation policy to the caller.

masatake:
I'm thinking about very small one:

```
CTAGS_INLINE const char* eStrskip (const char *str, int (* should_ignore) (int))
{
	const char *tmp;

	for (tmp = str; should_ignore(*tmp); tmp++)
		;
	return tmp;
}
```

codebrainz:
It might be awkward to use it, for example I believe the predicates from `ctype.h` might be allowed to be `inline` or macros, so you couldn't take their address (I could be wrong).

What about something like:

```c
char *eStrskip(char *str, const char *delimiters, int frontOrBack)
{
  if (!delimiters)
    delimiters = " \t\r\n\v\f";
  if (frontOrBack) { // front
    const char *start = str;
    for (; strchr(delimiters, *start) != NULL; start++) ;
    memmove(str, start, strlen(start));
  } else { // back
    size_t len = strlen(str);
    while (len--) {
      if (strchr(delimiters, str[len]) != NULL)
        str[len] = '\0';
      else
        break;
    }
  }
  return str;
}
```

masatake:
I don't want to update the passed string because the passed string can be modified with 'const'.
We will two functions:
* eStrskip (no updating), and
* eStrchop (updating).

masatake:
There are not a few area we can use the new functions.

```
$ grep --color -A 2 -nH -e 'for.\+isspace' *.c
args.c:33:	for (start = *next  ;  isspace ((int) *start)  ;  ++start)
args.c-34-		;
args.c-35-	if (*start == '\0')
--
args.c:42:		for (end = start ;  *end != '\0'  &&  ! isspace ((int) *end)  ;  ++end)
args.c-43-			;
args.c-44-		length = end - start;
--
entry.c:1287:	for (start = name; isspace(*start); start++)
entry.c-1288-		;
entry.c-1289-
--
parse.c:483:		for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-484-			;  /* no-op */
parse.c:485:		for ( ;  *p != '\0'  &&  ! isspace ((int) *p)  ;  ++p)
parse.c-486-			vStringPut (interpreter, (int) *p);
parse.c-487-	} while (strcmp (vStringValue (interpreter), "env") == 0);
--
parse.c:530:	for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-531-		;  /* no-op */
parse.c-532-
--
parse.c:537:		for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-538-			;  /* no-op */
parse.c-539-		for ( ;  *p != '\0'  &&  isLanguageNameChar ((int) *p)  ;  ++p)
--
parse.c:553:		for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-554-			;  /* no-op */
parse.c-555-		if (strncmp(p, "-*-", strlen("-*-")) != 0)
--
parse.c:599:			for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-600-				;  /* no-op */
parse.c-601-			for ( ;  *p != '\0'  &&  isLanguageNameChar ((int) *p)  ;  ++p)
--
parse.c:704:				for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-705-					;  /* no-op */
parse.c-706-				filetype = determineVimFileType(p);
```

codebrainz:
A version for immutable strings could be like this (untested):

```c
void eStripString(const char *str, const char *delims, const char **start,
                  const char **end, bool fromStart, bool fromEnd) {
  if (!delims)
    delims = " \t\r\n\v\f";
  if (fromStart) {
    const char *p = str;
    for (; strchr(delims, *p); p++)
      ;
    *start = p;
  } else {
    *start = str;
  }
  if (fromEnd) {
    size_t len = strlen(*start);
    const char *p = *start + len;
    while (len--) {
      if (strchr(delims, *p))
        p--;
      else
        break;
    }
    *end = p;
  } else {
    *end = *start + strlen(*start);
  }
}
```

masatake:
... I would like to merge this. Introducing common cstring utilities will be done later.

As far as seeing the output for grep

```
parse.c:599:			for ( ;  isspace ((int) *p)  ;  ++p)
parse.c-600-				;  /* no-op */
```

is the most common pattern.

So

```
CTAGS_INLINE const char* eStrskipSpaces (const char *str)
{
	const char *tmp;

	for (tmp = str; isspace(*tmp); tmp++)
		;
	return tmp;
}
```

is not so bad.
Of course, eStripString may be applicable another areas.

codebrainz:
Sounds fine to me.

The only reason I mentioned general and/or stripping tail of string is because of what you mentioned in "nextStringArg in main/args.c" uses it as well as I think vString could use it as well.

Issue #1566, opened 2017-10-04
[RFC] Which language do you want us to support?


masatake:
This is /dev/null (The 2nd meaning of https://en.wiktionary.org/wiki//dev/null) where requests go. We never promise to write the parser you want. I will help you if you want to write a parser only if you will provide the result under 'GPLv2 or later.'  I want to merge the result to our code base.

I would like you to write the document that explains the target language specification.
Much sample input files or information for large code base may help developers.

You can use thumb-up emoji.

For summarizing purpose, I will remove your comment.

For me(@masatake), the most of the languages I will use daily C, Python, Shell, Make, Automake, Autoconf, Asm, RPMspec, and Ld-script are available, and their parsers work well though there  are
many points I have to improve.

Member in the organization, feel free updating this comment.

If you find an implementation including .ctags file for a language, you can report here as I did about asn1 parser. We can import the implementation if the author agrees distributing it in GPL v2 or later.
For importing, we need volunteers who helps us write test cases.

Don't report about issue about a client tool like vim here. ctags just generates tag files. It doesn't provide tree structured UI panel (or something.) If you need a help from ctags side to improve your client tool, open an issue.

My queue:
- [x] merging parsers submitted from Geany,
- [x] merging @b4n's new (mostly rewritten) Cobol parser 
- [ ] merging (and reworking) asn1 parser
- [ ] improving javascript parser with the newly introduced symbol table
- [ ] implementing unoidl parser or thrift parser in peg for study
- [ ] rewriting java parser after extracting the token chain feature from Cxx parser as common rontines
- [x] looking for a volunteer who rewrites SystemVerilog parser

masatake:
Haskell

masatake:
Improving clojure

masatake:
Scala

masatake:
Vala

masatake:
Improving Ruby

masatake:
lex & flex

masatake:
scss

gruz:
I was forwarded from here https://github.com/geany/geany/issues/1329#issuecomment-334461067
So I'd be happy to have full JavaScript support.

masatake:
@gruz, thank you for contacting.
As I wrote in the first comment, this issue(#1566) is /dev/null. 

So could you open a new issue?
In the original issue opened at geany, a screenshot is used to explain the issue.
So I have to write OCF software first to extract the input given to ctags. It takes a long time.
[...]

PR #1568, opened 2017-10-05
Imporve/fix some scopes reported by the JavaScript parser

b4n:


masatake:
For newly added test cases, could you set --sort=no option?
Just put "--sort=no" to args.ctags file.

(In the future, I will put --sort=no to misc/units script.)

b4n:
@masatake OK, but for JS I don't know if it makes much sense as the parser emits some tags out of order.

b4n:
@masatake actually, what's the point in not sorting?  I agree it's easier to check when the parser is emitting things in the order they appear in the file, but otherwise I kind of feel like the tests would relying too much on an implementation detail that never mattered.  Namely, I don't have a really sensible way of "not sorting" the new test I added and that doesn't pass yet: which order should I put?  None is really relevant.
Again, I'm not sure why you don't want sorting, but maybe we should have a mean to sort by line numbers?

masatake:
@b4n,  how about adding `--fields=+n` (or `--fields=+{line}`) instead of `--sort`?

masatake:
I will review them in the week end. Thank you. BTW, is this change something to do with
https://github.com/universal-ctags/ctags/issues/1566#issuecomment-334462281 ?

b4n:
@masatake thanks!  I just squashed the fixups and added `--fields=+n` to the failing test case as you suggested.
And no, it's unrelated to that issue, yet related to Geany.  I was in the process of updating Geany's parser to current U-CTags for https://github.com/geany/geany/issues/934 and realized one of the test cases was giving bad results, so I went ahead and spend a day fixing it -- which itself lead to finding other small issues and trying to fix them, and trying to increase the coverage.

Units/parser-javascript.r/js-scope-resolution.b/expected.tags
```diff
@@ -0,0 +1,10 @@
+Cls	input.js	/^  var Cls = function() {$/;"	c	line:16	function:doesnt_extend
+Cls	input.js	/^var Cls = function() {$/;"	c	line:1
+assert	input.js	/^  function assert(expr, message) {$/;"	f	line:37	function:main
+doesnt_extend	input.js	/^function doesnt_extend() {$/;"	f	line:14
+dummy1	input.js	/^    this.dummy1 = function() {}$/;"	m	line:18	class:doesnt_extend.Cls
+dummy2	input.js	/^  Cls.prototype.dummy2 = function() {$/;"	m	line:20	class:doesnt_extend.Cls
+extend	input.js	/^function extend() {$/;"	f	line:7
+extended	input.js	/^  Cls.prototype.extended = function() {$/;"	m	line:9	class:Cls
```
masatake: Very interesting.

"Cls" in the scope field is obviously useful information about "extended".
However, "extend" in which "extended" is assigned is also useful information.

None defines the meaning of "scope" field. I have just recognized that there are two type of scopes: 
source code aspect scope and runtime scope.  lexical scope and dynamic scope are more suitable terms?

For reading and understanding source code, source code aspect scope information is also useful. The 
information can be regenerated with line: and end: fiedls but recording it in ctags is not bad idea.

Some parsers may records runtime scope and the other parsers may records source code aspect scope as the value for scope field. It may be too late to redefine the meaning of scope field. However, I would like to explore this area in the future. 

@b4n, you have developed an IDE. I would like to hear your comments.

One of my idea is introducing new scope named sourceScope that is filled only when if source code aspect scope and runtime scope about a tag is not matched.

Units/parser-javascript.r/js-crlf.d/expected.tags
```diff
@@ -0,0 +1,12 @@
+v1	input.js	/^var v1, v2$/;"	v
+v2	input.js	/^var v1, v2$/;"	v
+v3	input.js	/^var v3 = 1$/;"	v
+v4	input.js	/^,   v4 = 2$/;"	v
+v5	input.js	/^,   v5 = 3$/;"	v
+f1_c1_m1	input.js	/^    f1_c1_m1:$/;"	m	class:f1.f1_c1
+f1_c1_m2	input.js	/^    f1_c1_m2$/;"	m	class:f1.f1_c1
+f1_c1	input.js	/^  var f1_c1 = {$/;"	c	class:f1
```
masatake: It seems that f1 is a function.

masatake:
Have you ever consider to use corkAPI for representing scopes in jscritp.c?

For filling a scope field, a parser have to track both scope name and scope kind.
tokenInfo of jscript.c has only a structure field for scope name. 
Adding a field of scope kind is an approach.
Using corkAPI is another approach.

In my experince, corkAPI works well in this case.

I can take over the rest of job for introducing corkAPI if you are o.k. In such case, I will merge this pull request first.

masatake:
@b4n, after rebasing the lastet master branch where CTAGS_INLINE macro is introduced, could you consider chery-picking following changes?

https://github.com/masatake/ctags/commit/690c456fd9f160fe2ac6248a8b912dde5cd82f63
https://github.com/masatake/ctags/commit/d4a27f0461e11605a68ff937e6b23c33c7ca4f07
https://github.com/masatake/ctags/commit/cacb33fc3e428bba37c24f2e505b294e67d6689c
https://github.com/masatake/ctags/commit/524dd1bd298d7385b8631cb866bfc976b52fbce2

All of them are in found at https://github.com/masatake/ctags/commits/jscript-make-signature-field-sophisticated

Units/parser-javascript.r/js-commas-and-missing-semicolons.d/expected.tags
```diff
@@ -0,0 +1,11 @@
+A	input.js	/^var A = {}, b = 2;$/;"	v
+b	input.js	/^var A = {}, b = 2;$/;"	v
+c	input.js	/^var c = 3, d = 4;$/;"	v
+d	input.js	/^var c = 3, d = 4;$/;"	v
+mem1	input.js	/^A.mem1 = function() {return 42;}, A.mem2 = function() {return 43;}$/;"	f	function:A
```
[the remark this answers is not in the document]
b4n: The problem is that in JavaScript there's not really anything but objects, a class, a function, no real differences.  Actually you can see a fun thing in Firefox's JS console:
```
>> class foo {}
undefined
>> print(foo)
"function foo() {
    
}"
```

Anyway, my point is that the effective high-level kind is very dynamic.  In this example, A should be a class merely because it gets methods added, but it really becomes a class.  So ideally A should be reported as a class, but when it's first seen and the tag emitted it looks like a naked object, hence the "variable" kind.

Units/parser-javascript.r/js-crlf.d/expected.tags
b4n: Indeed. As mentioned somewhere it's a known limitation of the JS parser, with a `FIXME` in the code.

b4n:
@masatake I though about using that new cork API, but it would require quite some changes, especially as many tags are held back (to find the right kind).  Maybe though it would be quite handy if we can change the kind of a tag after having prepared it.
Anyway, I'm not against using it at all, I just felt it would be a fairly larger undertaking I didn't want to start right away, but I'd be fine reviewing the changes, or trying it at some point.

b4n:
> I think introducing "unknown" kind is better.

Hum… maybe, I'm not sure.  OK, it doesn't report something wrong and a client *could* special-case it, but OTOH the parser is generally right in the assumption, only a few cases are incorrect, so I'm not sure.

masatake:
I found we have js-scope-resolution.b test case.
I'm sorry to be late to know that. 

LGTM. Thank you.

masatake:
@b4h, should I merge this?

b4n:
@masatake You can, yes.
I am in the process of importing your mentioned changes, but wanted to try and simplify a few things, but didn't get there yet.  It's however mostly unrelated to this PR, so I think merging would be fine.
And I have another set of changes for the JS parser (for supporting Unicode identifiers and character escapes) in the works, so it's not like this one would be the last one anyway ☺️

masatake:
Could you look at#1519? It will change the last argument of initTagEntry. It required a pointer to kindDefintion struct. With the change of #1519, it requires an integer instead. This change may have small impact on the code you are working.

Issue #862, opened 2016-04-02
Adding varType TagEntryInfo field from Geany


techee:
I'd like to start syncing Geany's ctags with universal-ctags and there are a few things we'd need in universal-ctags in order to have all the functionality we need. One of the things we need is a

const char *varType;

field in sTagEntryInfo. As the name suggests, it contains a string corresponding to a variable type or a return type of a function. For instance for

```
char *foo;

int bar() {
    return 0;
}
```

it contains "char *" for the variable and "int" for the function. We need this info to be able to perform scope-based autocompletion and also the return value of functions for tooltips. This value is non-NULL only for statically typed languages - the currently supported ones in Geany are the languages from c.c and then Pascal, Rust, Go (c.c implementation in Geany diverges quite a lot and I haven't checked how hard it would be to port it to universal-ctags - the other languages should be easy to port).

Is it OK to add a new field to ctags with this information? I'd prepare a patch doing this. I've also noticed https://github.com/universal-ctags/ctags/pull/857 - one possibility would be to use custom extension field for this info (though the type info should be available for all statically typed languages so quite a big portion so one field defined for all of them might be better).

cc @b4n

masatake:
I'm working hard to generalize field manipulation based on the idea proposed in #857 by @pragmaware.
Could you wait for awhile?

techee:
@masatake Sure, no problem. From the Geany perspective it's just important that we have some kind of access to the field definitions to map them to Geany's internal types. I did this in the last commit of

https://github.com/geany/geany/pull/957

by accessing LanguageTable. If you are interested what we need from ctags you can have a look at tm_ctags_wrappers.h/c in the pull request.

masatake:
@techee, did you see `typeref:typename` introduced by @pragmaware ?
```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.c
cat /tmp/foo.c
char *foo;

int bar() {
    return 0;
}
[yamato@x201]~/var/ctags-github% ./ctags -o - /tmp/foo.c
./ctags -o - /tmp/foo.c
bar	/tmp/foo.c	/^int bar() {$/;"	f	typeref:typename:int
foo	/tmp/foo.c	/^char *foo;$/;"	v	typeref:typename:char *
```

This is not enough?
@pragmaware solved two things.  

1. the way to use typeRef[2] field of extensionFields of tagEntryInfo.
  As @b4n speicified before, ctags main part expects typeRef [0] and typeRef[1] are filled.
[...]

PR #1569, opened 2017-10-05
Golang add anonymous member kind

masatake:
See each commit logs.


masatake:
@techee, can I ask you to review again?

parsers/go.c
```diff
@@ -678,16 +680,33 @@ static void parseStructMembers (tokenInfo *const token, const int scope)
 				}
 				readToken (token);
 			}
-			if (!isType (token, TOKEN_COMMA))
+
+			if (isType (token, TOKEN_DOT)  && (!first))
```
techee: I think it would be nicer to move parsing of the type outside of this loop and keep this loop only for parsing variable names. I think the way you do it now won't cause any trouble for valid input but it seems to be possible to mix variable parsing and type parsing the way it is.

Eventually it would be nice to have some "varType" (https://github.com/universal-ctags/ctags/issues/862) and report the type for all the member variables. That would mean the whole type would have to be always collected and assigned to the variable tags.

parsers/go.c
```diff
@@ -685,6 +693,15 @@ static void parseStructMembers (tokenInfo *const token, const int scope)
...
 				// Throw away the package component.
+				if (!typeForAnonMember)
+					typeForAnonMember = vStringNewCopy (memberCandidate->string);
+				else
+				{
+					vStringPut (typeForAnonMember, '.');
+					vStringCat (typeForAnonMember, memberCandidate->string);
+				}
+
+
```
techee: Like in my previous comment, I would move this part outside of this loop and make another one just for type parsing. Then it would be good to think about how variables are named - if type parsing is detected, memberCandidate->string should be copied to something like "type" variable so it's clearer what is happening - using memberCandidate below for typeForAnonMember creation doesn't make things very obvious.

masatake:
Updated. I agree with you. The code becomes much easier to read. Thank you.
Collecting types for (normal) members has not been done yet. It is next step.
I decide to add `*' to the type of anonymous member.
```
[yamato@master]~/gps/hp-rta-_/meetings% cat /tmp/goo.go
cat /tmp/goo.go
type X struct {
	*T
}
[yamato@master]~/gps/hp-rta-_/meetings% u-ctags -o - /tmp/goo.go 
u-ctags -o - /tmp/goo.go 
T	/tmp/goo.go	/^	*T$/;"	M	struct:X	typeref:typename:*T
X	/tmp/goo.go	/^type X struct {$/;"	s
```

masatake:
@techee, can I merge this?

PR #1578, opened 2017-10-12
Fixes invoking autogen.sh from a makefile again

sdwolf:
This is the same problem that 637f3d8da5abc60a26b03d7cb1dfbbf171522e7f fixed, but with a cross platform solution.

It was my mistake for using a GNU specific flag.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -3,7 +3,7 @@
 type autoreconf || exit 1
 type pkg-config || exit 1
 
-ctags_files=`make -f makefiles/list-translator-input.mak`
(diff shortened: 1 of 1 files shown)


masatake:
Can I ask you to put the initial comment of this pull reuqest
```
This is the same problem that 637f3d8 fixed, but with a cross platform solution.

It was my mistake for using a GNU specific flag.
```
to your commit log? 

Thank you.

sdwolf:
Hello @masatake, I've updated the commit message as you requested.

masatake:
Thank you!

PR #1585, opened 2017-10-25
Tcl fix commnet handling

masatake:
`#` at the middle of line in tcl script is not a comment starter.
But ctags recognized it as a comment starter.
The changes in this pull request fixes the wrong recognition about comment starter.

### DIFF
--- a/parsers/tcl.c
+++ b/parsers/tcl.c
@@ -311,18 +311,24 @@ static void readToken (tokenInfo *const token, void *data CTAGS_ATTR_UNUSED)
 	pstate->lastTokenType = token->type;
 }
 
+static bool tokenIsEOL (tokenInfo *const token)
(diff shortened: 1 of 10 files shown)


parsers/tcl.c
```diff
@@ -217,9 +238,12 @@ static void readToken (tokenInfo *const token, void *data CTAGS_ATTR_UNUSED)
...
-				c = getcFromInputFile ();
-			while (c != EOF && c != '\r' && c != '\n');
+			if (bol)
+			{
+				do
+					c = getcFromInputFile ();
+				while (c != EOF && c != '\r' && c != '\n');
+			}
 			goto getNextChar;
 		}
```
masatake: I have to think about the case that escaped is true.

PR #1586, opened 2017-10-28
sh: run guest parser on heredoc area

masatake:
This change extends sh parser to run a guest parser on a heredoc area.

In the following 4 examples, sh parser   makes 4 promises for running php parser:
```
php <<EOF1
...
EOF1

env php <<EOF2
...
EOF2

someThingCommand > foo.php <<EOF3
...
EOF3

someThingCommand  <<EOF4 > foo.php
...
EOF4
```

`;`, command line separator is not handled well yet.


masatake:
I will merge  f546df6 first.

PR #1590, opened 2017-10-31
Cxx: solve confliction between digraph and template angle char

### LINKED ISSUE #1588: ctags missed generating some symbols
source file: https://github.com/scylladb/seastar/edit/master/core/reactor.hh

some missed symbols are as follows:
line 397:   class thread_pool {
line 442:   class reactor_backend {  
line 525:   inline open_flags operator|(open_flags a, open_flags b) {  
line 1081:  inline reactor& engine() {

cmd: ctags -I __THROW -I __nonnull --c-kinds=+pxd --extras=+fgqr --fields=+CEKRSZaeilmnprx ./reactor.hh

$ ctags --version
Universal Ctags 0.0.0, Copyright (C) 2015 Universal Ctags Team
  Compiled: Oct 31 2017, 20:25:39
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +multibyte, +option-directory

masatake:
CPreProcessor converted digraph "<:" to '['.
However, it causes C++ cannot parser "<::namespace>" well.
If ':' comes next to '<:', CPreProcessor should not do the
conversion.

Close #1588.


masatake:
I recognized this is not perfect.
`[::namespace::a_constant]` may be valid expression.
This can be written with digraph like `<:::namespace::a_constant:>`.

masatake:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3237.html
```
if the next three characters are <:: and the subsequent character is neither : nor >, the < is treated as a preprocessor token by itself (and not as the first character of the alternative token <:);
```

masatake:
I will merge bea7699 and 95ea591 first.

PR #1591, opened 2017-11-01
go: Make struct parsing a little easier to read

techee:
The code right now isn't very easy to read. This patch tries to improve
that by being more explicit what is being parsed - memberCandidate
isn't reused for cases which have nothing to do with being member
candidate and type parsing is performed the same way both for
anonymous and non-anonymous members (right now tag is generated only
for the anonymous case).

### DIFF
--- a/Units/parser-go.r/go-anonmember.d/expected.tags
+++ b/Units/parser-go.r/go-anonmember.d/expected.tags
@@ -17,3 +17,13 @@ lang3	input.go	/^	lang3, lang4 *P.Q.LANG$/;"	m	struct:X
 lang4	input.go	/^	lang3, lang4 *P.Q.LANG$/;"	m	struct:X
 T11	input.go	/^	P.Q.T11$/;"	M	struct:X	typeref:typename:P.Q.T11
 T12	input.go	/^	*P.Q.T12$/;"	M	struct:X	typeref:typename:*P.Q.T12
+Y	input.go	/^type Y struct {$/;"	s
+byte	input.go	/^	byte$/;"	M	struct:Y	typeref:typename:byte
+b	input.go	/^	b byte$/;"	m	struct:Y
+ss	input.go	/^	ss []string$/;"	m	struct:Y
+bss	input.go	/^	bss [][]byte$/;"	m	struct:Y
+f	input.go	/^	f func() error$/;"	m	struct:Y
+ifaces	input.go	/^	ifaces []interface{}$/;"	m	struct:Y
+tags	input.go	/^	tags  map[string]struct{}$/;"	m	struct:Y
+notify	input.go	/^	notify chan<- struct{}$/;"	m	struct:Y
+p	input.go	/^	p *struct {$/;"	m	struct:Y
--- a/Units/parser-go.r/go-anonmember.d/input.go
+++ b/Units/parser-go.r/go-anonmember.d/input.go
@@ -14,3 +14,19 @@ type X struct {
 	P.Q.T11
 	*P.Q.T12
 }
+
+type Y struct {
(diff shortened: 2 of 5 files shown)


techee:
@masatake Sorry for the delay - I have been too busy in the last 2 weeks and didn't have time to review your patch in #1569 properly. This pull request tries to improve the readability of the code a bit and also make the code more ready if someone implements type reporting for struct members.

masatake:
> @masatake Sorry for the delay
NO PROBLEM.  Feel free to block merging by updating the title(subject?) of pull request I opened like

```
[BLOCK MERGING Nov  9 by @techee] go: Make struct parsing a little easier to read #1591
```

if you will review the change. (Of course, just putting a comment for blocking is also o.k.).

I merged #1569 because my colleague who need to read  kubernetes written in Go wants it.

Thank you for new pull request. I will review it.

techee:
@masatake There was no need to block merging and I wasn't sure when I would have time to review it so it was just fine that you merged your patch.

masatake:
`ss` is recorded in `typeref` for `ss`.
I guess the way to detecting `anon` doesn't work expectedly.

```
[yamato@master]~/var/ctags-github% cat input.go
cat input.go
package foo

type Foo struct {
	byte
	b byte
	ss []string
	bss [][]byte
	f func() error
	ifaces []interface{}
	tags  map[string]struct{}
	notify chan<- struct{}
	p *struct {
		x int
		y int
		z int
	}
}

func main () {
}
[yamato@master]~/var/ctags-github% go build input.go
go build input.go
[yamato@master]~/var/ctags-github% ./ctags --sort=no -o - input.go
./ctags --sort=no -o - input.go
foo	input.go	/^package foo$/;"	p
Foo	input.go	/^type Foo struct {$/;"	s	package:foo
byte	input.go	/^	byte$/;"	M	struct:foo.Foo	typeref:typename:byte
b	input.go	/^	b byte$/;"	m	struct:foo.Foo
ss	input.go	/^	ss []string$/;"	M	struct:foo.Foo	typeref:typename:ss
bss	input.go	/^	bss [][]byte$/;"	M	struct:foo.Foo	typeref:typename:bss
f	input.go	/^	f func() error$/;"	M	struct:foo.Foo	typeref:typename:f
ifaces	input.go	/^	ifaces []interface{}$/;"	M	struct:foo.Foo	typeref:typename:ifaces
tags	input.go	/^	tags  map[string]struct{}$/;"	M	struct:foo.Foo	typeref:typename:tags
notify	input.go	/^	notify chan<- struct{}$/;"	M	struct:foo.Foo	typeref:typename:notify
p	input.go	/^	p *struct {$/;"	m	struct:foo.Foo
main	input.go	/^func main () {$/;"	f	package:foo
```

techee:
@masatake Thanks for looking at this - yes, the code was wrong and I basically had to rewrite it in the fixed version I've just pushed. I've also added your example to tests so structs with more complex types are covered.

techee:
> As the result, we can expect "make units" gets successful at any commits in our git history.

The unit test change isn't related to the go.c change - the unit test passes both before the go.c change and after it too. I just added the example you have posted and made the change before I started modifying go.c to be sure it still passes after I perform the modifications.

I can of course change the order but it's unnecessary IMO as "make units" passes for all the three commits.

masatake:
@techee, you are correct. I misunderstood the changes.

masatake:
I have not merged this because I wanted to push changes on your changes. In my changes, I would like to add code filling typeref fields. I found intArray like ptrArray is needed to store corkIds. However, I found I don't have enough time for hacking ctags in this year.

PR #1592, opened 2017-11-01
JavaScript: Add support for Unicode escape sequences

b4n:
Handle both kind of Unicode escape sequences inside strings and identifiers:

`"\u" Hex4Digits`: Hex4Digits is a 4 digits hexadecimal value representing a UTF-16 word (e.g. either a code point or a high or low surrogate).

`"\u{" HexDigits "}"`: HexDigits is any number of hexadecimal digits representing a Unicode code point (so must have an actual value less than or equal to 0x10FFFF).

Note: the previous code handling escape sequences outside string literals didn't seem to follow anything from the ECMAScript standard, so was simply removed.

See http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf §11.8.4 "String Literals" for the definition of "UnicodeEscapeSequence".


b4n:
Okay, OSX and MSVC should be all fixed now

masatake:
I would like to review this after you merge the other pull requests you submitted.

masatake:
I will review wihin this week.

masatake:
Sorry, still I cannot find time for reviewing.  Please, wait for awhile.

masatake:
MY TODO: 
It is nice that a parser can specify the sizeof ungetchBuf of inputFile when opening an input file.
This change specifies it 8 bytes statically.
openInputFile may have ability to do it. However, The consideration for nestedInputStream may be needed. Dynamic ungetchBuf may be needed to write a parser for Jinja2.

masatake:
MY TODO:

The name of feature "multibyte" is really bad name. We should call it "iconv" or something.

masatake:
I'm sorry to be late to review and merge this pull request.
Thank you.

PR #1595, opened 2017-11-07
Units,puppetManifest: add README about .b test cases

### LINKED ISSUE #1539: universal ctags not parsing c++ headers

masatake:
Suggested by @b4n.
Close #1539.

### DIFF
--- a/Units/parser-puppetManifest.r/puppet-tag.b/README
+++ b/Units/parser-puppetManifest.r/puppet-tag.b/README
@@ -1,0 +1,3 @@
+This test case is a placeholder.
+I have not designed how ctags deals with tags of puppetManifest yet.
+So I put only the input file here.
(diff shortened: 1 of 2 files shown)


masatake:
Thank you.

PR #1598, opened 2017-11-10
AutoIt parser

Skif-off:
Hello!

I like Geany, I like AutoIt, I added [syntax highlighting](https://github.com/geany/geany/pull/1017) and I want to ensure the work of autocomplete and Symbols tab (a list of symbols in code, for AutoIt is "functions" only). Geany uses Ctags and now I here: have I to do something else?

### DIFF
--- a/parsers/autoit.c
+++ b/parsers/autoit.c
@@ -29,13 +29,11 @@
 */
 typedef enum {
 	K_FUNCTION,
-	K_INCLUDE,
 	K_REGION
 } AutoItKind;
 
 static kindDefinition AutoItKinds [] = {
 	{ true, 'f', "func", "functions" },
-	{ true, 'i', "include", "include files" },
 	{ true, 'r', "region", "regions" }
 };
 


parsers/autoit.c
```diff
@@ -0,0 +1,77 @@
...
+				++cp;
+			while (isalnum ((int) *cp) || *cp == '_')
+			{
+				vStringPut (name, (int) *cp);
+				++cp;
+			}
+			while (isspace ((int) *cp))
+				++cp;
+			if (*cp == '(')
+				makeSimpleTag (name, K_FUNCTION);
```
masatake: This parser is enough simple. So we can do the same thing with command line based regex parser:

```
[jet@localhost]~/var/ctags% cat /tmp/foo.au3 
; Taken from https://www.autoitscript.com/autoit3/docs/intro/lang_functions.htm

#include <Constants.au3>

Local $iNumber = 10
Local $iDoubled = 0

For $i = 1 To 10
    $iDoubled = MyDouble($iNumber)
    MsgBox($MB_OK, "", $iNumber & " doubled is " & $iDoubled)
    $iNumber = $iDoubled
Next
Exit

Func MyDouble($iValue)
    $iValue = $iValue * 2
    Return $iValue
EndFunc   ;==>MyDouble

func MyDouble0($iValue)
    $iValue = $iValue * 2
    Return $iValue
EndFunc   ;==>MyDouble

FUNC MyDouble1($iValue)
    $iValue = $iValue * 2
    Return $iValue
[jet@localhost]~/var/ctags% cat /tmp/autoit.ctags 
--langdef=AutoIt
--map-AutoIt=+.au3
--kinddef-AutoIt=f,funct,functions
--regex-AutoIt=/^func[[:space:]]{1,}([a-z0-9_]{1,}) *\(/\1/f/{icase}
[jet@localhost]~/var/ctags% ./ctags -o - --options=/tmp/autoit.ctags /tmp/foo.au3  
MyDouble	/tmp/foo.au3	/^Func MyDouble($iValue)$/;"	f
MyDouble0	/tmp/foo.au3	/^func MyDouble0($iValue)$/;"	f
MyDouble1	/tmp/foo.au3	/^FUNC MyDouble1($iValue)$/;"	f
```

Skif-off:
@masatake,
>This parser is enough simple. So we can do the same thing with command line based regex parser

```/parsers/awk.c``` too :) There are two reasons:
1. I or someone else can add something else (```#include```, maybe ```#OnAutoItStartRegister``` & ```#OnAutoItExitRegister```, in the future). Or special tag ```#Region - #EndRegion``` [(but for navigation only)](https://github.com/geany/geany/issues/967#issuecomment-341622338).
2. Geany uses Ctags and without it couldn't works autocomplete and Symbol (it's bad for IDE) and I must first offer Ctags project a parser. 

>At least we need a test case.

I added [input.au3 and expected.tags](https://github.com/Skif-off/ctags/tree/ctags-autoit/Units/parser-autoit.r/simple-au3.d), is it enough? I copied-pasted your example: is it bad idea (license and all that)?

>Do Darren Hiebert and Elias Pschernig write this?

Sorry, fixed. How can I not specify my name? Simple сode is works and I want to give it to society without conditions :)

>Could you add URL to language reference of the target language if it exists?
https://www.autoitscript.com/autoit3/docs/, is this the page?

Ok, done.

> you can write this in regex parser

If I right understand, Geany not uses Ctags regex parsers. I don't know. Is it necessary?

---
I have problem: I've never seen this, but suddenly it occurred to me to try
```autoit
Example()

    	Func Example()
    MsgBox(4096, "Example", "Test indents: OK")
EndFunc
```
I've never seen this, but what do I do now? I should rewrite the findAutoItTags function, isn't it?

masatake:
Just adding
```
	while (isspace ((int) *line))
			++line;
```

 after reading line is enough.

Adding such unexpected input to the test case is also important.

masatake:
How do you think other kinds?
I have found Global and Local are interesting.
I don't mean supporting them in your paser is must. Just talking about extending the parser.

Skif-off:
@masatake, thanks, works fine!
I use ```strlen ((const char *) line) >= 8```, is there a faster way?

Skif-off:
@masatake,
>How do you think other kinds?
>I have found Global and Local are interesting.

Sorry, I have never used Ctags and I don't know how to usually use it (goals and objectives). I think, ```Local``` is difficult: we use it in functions too - we will get all in one heap, but significance is different.
Maybe ```Global Const```?
```autoit
Global Const $GUI_CHECKED = 1
Global Const $WS_OVERLAPPEDWINDOW = 0x00CF0000
Global Const $WM_GETMINMAXINFO = 0x0024
Global Const $GUI_SS_DEFAULT_LABEL = 0
Global Const $SS_NOPREFIX = 0x0080
Global Const $LVS_NOCOLUMNHEADER = 0x4000
Global Const $LVS_NOSORTHEADER = 0x8000
Global Const $LVS_SHOWSELALWAYS = 0x0008
Global Const $LVS_EX_FULLROWSELECT = 0x00000020
Global Const $LVS_EX_GRIDLINES = 0x00000001
Global Const $LVM_FIRST = 0x1000
Global Const $LVM_SETCOLUMNWIDTH = ($LVM_FIRST + 30)
Global Const $ES_LEFT = 0
Global Const $ES_PASSWORD = 32
Global Const $ES_AUTOHSCROLL = 128
Global Const $WM_COMMAND = 0x0111
Global Const $WM_SYSCOMMAND = 0x0112
Global Const $SC_CLOSE = 0xF060
Global Const $GUI_RUNDEFMSG = 'GUI_RUNDEFMSG'
```
Usually we write list of them at the beginning of the file. ```#include```, like header files.

I googled and I found [jEdit4AutoIt](http://thorsten-willert.de/Themen/jEdit/jEdit4AutoIt/index.php): author uses Exuberant Ctags (ctags-5.8 (09 Jul 2009)) and ```ctags.cnf```:
```
--langdef=AutoItScript
--langmap=AutoItScript:.au3.AU3
--regex-AutoItScript=/^[ \t]*func[ \t]+(.*?)[\(]/\1/Functions/i
--regex-AutoItScript=/^[ \t]*global[ \t]+(.*?)[ \t]*/\1/Global Variables/i
--regex-AutoItScript=/^[ \t]*local[ \t]+(.*?)[ \t]*/\1/Local Variables/i
--regex-AutoItScript=/^[ \t]*(local|global)?[ \t]+enum[ \t]+(.*)/\1 \2/Enum/i
--regex-AutoItScript=/^#include(.*?)(<|")(.*?)("|>)/\3/Include Files/i
--regex-AutoItScript=/^#region[ \t]*(.*?)/\1/Regions/i
--regex-AutoItScript=/^[ \t]*(Opt|AutoItSetOption)[ \t]*[\(](.*),(.*)[\)]/\2 \3/Options/i
--regex-AutoItScript=/^(.*?);#__DEBUG__/\1/Debug/i
```
We can use ```#Region``` with indents too (see [example, in function "Example()"](https://www.autoitscript.com/autoit3/docs/functions/UDPSend.htm)), but it's bad for global navigation by file.
I like ```#Region```, for example

I'll think about it tag.

masatake:
```
strlen ((const char *) line) >= 8
```

I think we don't need this now. I will merge you change after I squashing the commits into one or two changes.


 AFTER MERGING, and when I find a time I will rewrite the parser token oriented way. (The current implementation is line oriented.)

masatake:
Oh,

```
strlen ((const char *) line) >= 8
```

You introduce this already. This is o.k.

parsers/autoit.c
```diff
@@ -0,0 +1,83 @@
+/*
+*   Copyright (c) 2017, S
```
masatake: I expect put your real name here.

Skif-off: Fix.

Skif-off:
I added ```#include``` & ```#region``` and some optimizations.

masatake:
Though there are many exceptions, basically, ctags captures definitions.
In 
```
#include <foo>
```
, foo is not defined. Therefore foo should not be captured as a definition tag.
Instead, ctags should capture it as a reference tag. I wrote the idea in 
http://docs.ctags.io/en/latest/news.html?highlight=reference#reference-tags .

We can use `cpreprocessor.[ch]` for capturing foo. However, it is expected that a character-oriented parser, not line oriented parser, uses it. 

@Skif-off , is it o.k. to remove `#include` related code from your change when merging?
I would like to rework on the code handling `#include` later when I find time.

Skif-off:
@masatake 
> is it o.k. to remove ```#include``` related code from your change when merging?
I would like to rework the code handling ```#include``` later when I find the time.

Ok. I added ```#include``` like jEdit4AutoIt but was not sure :)

masatake:
@Skif-off, could you look at #1599 ?
I squashed all your changes into one commit in my pull request.
I added one more commit not to make an empty tag.

Skif-off:
@masatake 
Yes, I've seen, thank you! 
Can I delete [my fork](https://github.com/Skif-off/ctags)?

masatake:
@Skif-off, yes you can delete your branch.

masatake:
I merged #1599. Thank you for the contribution.

In the future we can rewrite this based on tokenInfo.h and cpreprocss.c.
So we can get more tags and fill end: field.

PR #1599, opened 2017-11-20
AutoIt parser based on @Skif-off work

masatake:
Cloes #1598.

This is basedon #1598. I added two more commits.
@Skif-Off, could you take a look the two commits?
If you are o.k., I will squash the all commits including your and mine into one.


Skif-off:
@masatake 
I'm not against :)
Is this will not large affect the speed? It seemed to me that most parsers don't check it himself.

masatake:
> Is this will not large affect the speed? It seemed to me that most parsers don't check it himself.

Making an empty tag is bad idea.

PR #1601, opened 2017-11-20
Rust: Include trait in impl tags

ithinuel:
When multiple implementations are made for the same type, it can be rapidly difficult to distinguish them.

This PR helps by adding which trait is implemented in the tag as "Trait for Type" instead of simply "Type" .

### DIFF
--- a/Units/parser-rust.r/rust-test_input.d/expected.tags
+++ b/Units/parser-rust.r/rust-test_input.d/expected.tags
@@ -3,21 +3,21 @@ Animal	input.rs	/^enum Animal {$/;"	g
 B	input.rs	/^pub struct B$/;"	s
 Bar	input.rs	/^struct Bar(isize);$/;"	s
 Baz	input.rs	/^struct Baz(isize);$/;"	s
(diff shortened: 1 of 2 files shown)


masatake:
I think using "inherits:" field may be better.

```
B	foo.rs	/^impl A for B {$/;"	kind:implementation	      inherits:A
```

This is on the analogy of Java:
```
interface A {
}
class B implements A {
}
```

For the above input ctags emits:
```
A	foo.java	/^interface A {$/;"	kind:interface	line:1	language:Java	end:2
B	foo.java	/^class B implements A {$/;"	kind:class	line:4	language:Java	inherits:A	end:5
```

masatake:
I implemented my idea. Could you look at #1603 ?

I understand how useful using '... for ...' in scope field which is implemented in your pull request.
However, it is not consistent enough with the other area of ctags.

Using full qualified extra tag(--extras=+q) is one of idea.
It will work

ithinuel:
A better implementation of this PR was proposed in #1604 based on @masatake  work in #1603 .

PR #1602, opened 2017-11-21
Rust: use vString API

masatake:
Rust parser directly accesses vString data structure.
Using vStringAPI increases the number of typing. However, it helps us to read the source code of the parser.

### DIFF
--- a/parsers/rust.c
+++ b/parsers/rust.c
@@ -438,7 +438,7 @@ static void addTag (vString* ident, const char* arg_list, int kind, unsigned lon
 	if (kind == K_NONE || ! rustKinds[kind].enabled)
 		return;
 	tagEntryInfo tag;
-	initTagEntry(&tag, ident->buffer, kind);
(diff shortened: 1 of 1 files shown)


parsers/rust.c
```diff
@@ -97,8 +97,7 @@ static void parseBlock (lexerState *lexer, bool delim, int kind, vString *scope)
 /* Resets the scope string to the old length */
 static void resetScope (vString *scope, size_t old_len)
 {
-	scope->length = old_len;
-	scope->buffer[old_len] = '\0';
+	vStringTruncate (scope, old_len);
```
b4n: The function is only used once, and this body is now only 1 call with the same args (e.g. basically an alias for `vStringTruncate`), it could probably be simply removed.  But if you wanna keep the naming for clarity, it also sounds reasonable.

masatake: I would like to keep this because it may help code reader understand the intent of code.

PR #1603, opened 2017-11-21
Rust: fill inherits and end fields

masatake:
(This is based on pull request #1601 submitted by @ithinuel.)

For the input

    impl A for B {

Rust parser emits:

     B	foo.rs	/^impl A for B {$/;"	kind:implementation	inherits:A

This on on the analogy of what Java parser does:

    interface A {
    }
    class B implements A {
    }

For the above input ctags emits for class B:

    B	foo.java	/^class B implements A {$/;"	inherits:A

### DIFF
--- a/Units/parser-rust.r/rust-simple.d/args.ctags
+++ b/Units/parser-rust.r/rust-simple.d/args.ctags
@@ -1,1 +1,1 @@
---fields=afikmsS
+--fields=afikmsSe
--- a/Units/parser-rust.r/rust-simple.d/expected.tags
+++ b/Units/parser-rust.r/rust-simple.d/expected.tags
@@ -1,4 +1,4 @@
-Circle	input.rs	/^impl Circle {$/;"	c
+Circle	input.rs	/^impl Circle {$/;"	c	end:17
 Circle	input.rs	/^struct Circle {$/;"	s
 area	input.rs	/^    fn area(&self) -> f64 {$/;"	P	implementation:Circle	signature:(&self) -> f64
 main	input.rs	/^fn main() {$/;"	f	signature:()
--- a/Units/parser-rust.r/rust-test_input.d/args.ctags
+++ b/Units/parser-rust.r/rust-test_input.d/args.ctags
@@ -1,1 +1,1 @@
---fields=afikmsS
+--fields=afikmsSe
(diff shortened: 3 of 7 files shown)


ithinuel:
Should'nt the trait name also include it's parameters so it distinguish different implement of the same method for the same type/trait combination ?

masatake:
> Should'nt the trait name also include it's parameters so it distinguish different implement of the same method for the same type/trait combination ?

Could you give me an example? I don't know well about Rust (and bad in English).

Is adding trait: field to method kind enough?

```
impl HasArea for Circle {
    fn area(&self) -> f64 {
        std::f64::consts::PI * (self.radius * self.radius)
    }
}

Circle	foo.rs	/^impl HasArea for Circle {$/;"	kind:implementation	inherits:HasArea
area	foo.rs	/^    fn area(&self) -> f64 {$/;"	kind:method	scope:implementation:Circle	trait:HasArea
```

I'm not sure what is "parameters" here.

Do you mean `<i64>` in the following code?
```
impl ConvertTo<i64> for i32 {
    fn convert(&self) -> i64 { *self as i64 }
}
```

ithinuel:
Yes indeed,
typically, for [vec.rs](https://github.com/rust-lang/rust/blob/master/src/liballoc/vec.rs) from rust's sources, with or without `--extrax+q`, `vim` for example always report   
`Vec : implementation` where it would be much more relevant to have something to distinguish the implementations without the need to look at the implemented methods (when the elements are folded for example).

I fetched your branch and "fixed" the qualified name on my fork : ithinuel/ctags@0e554965

Now it really looks exactly as I expect :D

I just don't know how to proceed to update the source branch of my PR (or should I closse this one and create a new one ?).

masatake:
Could you push new changes to your branch where you used to open #1601 with `--force` option?

.... However, I found you used `master` for this work, improving Rust parser.
As far as my understanding about git, using master for such hacking is a bad idea.
`master` is for syncing upstream repository. I recommend you to make a topic branch, as I did in this PR;
I made `masatake:rust-record-trait-for-impl-to-inherits-field` branch.

If you will continue to use your `master` for this work, do `push --force` your changes to `master` where 
 you used to open #1601.
If you will make and use a new branch, open a new pull request and close #1601. Please refer #1601 and this #1603 form the new pull request.  So people having interest in this area can read this fruitful discussion:-)

masatake:
Now Rust is introduced to Linux. I cannot ignore this issue.

PR #1605, opened 2017-11-29
WIP: Update readtags to support escape sequences and fix & improve its testing

b4n:
Following what I started in #1580, this PR aims at updating *readtags* to properly support the current format, as well as fixing & improving its automated test cases (`roundtrip`).


b4n:
TODO: fix OSX and Windows (either the test facility or the program, not 100% sure what is the problem for Windows)

b4n:
OK, apparently OSX works now.  Windows' left, up for grabs if anybody got an idea/has a good way to perform tests (I don't have a Windows machine and waiting for AppVeyor when shooting in the dark is fairly frustrating)

b4n:
OK apparently there's expansions performed on the arguments passed to readtags, so that it actually gets different values than the script passes.  I'm now trying to work around this by escaping those, but there might be better solutions I don't know about.
Any ideas, e.g. @k-takata?

b4n:
OK I don't know how to avoid the issue yet, e.g. `::` leads to *readtags* receiving `;`, leading `/`es to `C:/…`, etc. I tried escaping with `\` and `^`, doesn't seem to help.

b4n:
OK i improved the situation with the last commit after finding https://github.com/msys2/msys2/wiki/Porting#user-content-filesystem-namespaces, but it's still not perfect: there's encoding issues, and some exclude patters break legitimate stuff (like `*` and `/`).  All in all, there's still 14 tests failing on MSYS2.

I don't really know what I can do, but add the ability for *readtags* to read the tag name from a file instead of the command line.  It's not very nice because I can't really see a real use for it, but it would solve the whole problem at once I guess.

masatake:
(I will work on this item in the new year day.)

masatake:
I will take time for this important one.

masatake:
@b4n, I'm reading your patches.
I will devide your patches into two groups; one group(A) gathers the part of patches handling escape sequence, and the other group(B) gathers the part of patches general purpose, nothing about escape sequence. I will merge B into master branch first.

misc/roundtrip
```diff
@@ -112,8 +114,10 @@ OLD_IFS="$IFS"
 IFS='
 '
 set -x
-for tags in $(find "$UNITS" -name expected.tags); do
-    for name in $(sed -e 's/^\([^	]*\)	.*/\1/' "$tags"); do
+tagfiles="$(find "$UNITS" -name expected.tags)"
```
masatake: The first and the last double quote characters are needed.

Does the following code do the same?
```
tagfiles=$(find "$UNITS" -name expected.tags)
```

misc/roundtrip
```diff
@@ -112,8 +114,10 @@ OLD_IFS="$IFS"
...
+for tags in $tagfiles; do
+    tagnames="$(sed -e 's/^\([^	]*\)	.*/\1/' "$tags")"
```
masatake: The same as the last my comment.

masatake:
I will merge following changes to master branch.
(The commit ids are changed because I make your branch be rebased on master.)

```
52c638a8 circleci: Install find for roundtrip
c079b0cc roundtirp: Improve message for failing tests
dc0bce22 readtags: Build with coverage support if requested
c7817168 roundtrip: Cleanup and fix so it actually does check something
85358ece roundtrip: Fix running on non-C locales
16260d46 testing: Define $(SILENT)
85525454 testing: Don't unconditionally silent test rules
                 roundtrip: Exit on command failure
```

misc/roundtrip
b4n: yes, doesn't matter

misc/roundtrip
```diff
@@ -53,7 +53,7 @@ expandEscapeSequences()
 :again
     s/\\/__BACKSLASH__/
     T out
-    s/__BACKSLASH__\\/__LASTBACKSLASH__/;	t again
+    s/__BACKSLASH__\\/__LITBACKSLASH__/;	t again
```
masatake: Could you tell me why this change is needed?

masatake:
@b4n, my understanding is that many changes are about \n and \t.
If we use GNU sed on MacOSX, can we make changes simpler?

misc/roundtrip
b4n: It is not needed per se, but I changed the name for clarity as it's not only used for the last backslash but for all literal backslashes

b4n:
> @b4n, my understanding is that many changes are about \n and \t.
> If we use GNU sed on MacOSX, can we make changes simpler?

Not really, having fixups here is just that I'm too used to GNU sed so I don't think about its extensions in supporting escape sequences; but ultimately the changes are the same just with a slightly different syntax (using the literal value rather than the GNU sed escape sequence).

b4n:
> My idea is that the code introduced by @b4n's patch should be activated for handling u-ctags v2 and v3 formats output only.

As said, I think that my changes are backward compatible.  And I think it is not only with the v2 format, but also the e-ctags format: the only thing that would risk being a problem is if e-ctags could output unescaped `\`es, in which case the updated code would erroneously try and decode it.  However, IIRC e-ctags doesn't do so, in which case we should be safe.

masatake:
Thank you very much. I merged all your changes.

Issue #1606, opened 2017-12-07
Undefined symbols for architecture x86_64 on macOS 10.13


lightingghost:
macOS 10.13
Apple LLVM version 9.0.0 (clang-900.0.38)

The commands run are
```
 ./autogen.sh
 ./configure --prefix=/opt/local
 make
```

The error messages are
```
Undefined symbols for architecture x86_64:
  "_iconv", referenced from:
      _convertString in ctags-mbcs.o
  "_iconv_close", referenced from:
      _closeConverter in ctags-mbcs.o
  "_iconv_open", referenced from:
      _openConverter in ctags-mbcs.o
```

masatake:
Could you show me config.log and generatred Makefile ?

lightingghost:
@masatake Thanks for helping

I have attached the config.log, config.h, and Makefile. I have renamed them to *.txt as GitHub do not allow me to upload .h files.


[config.h.txt](https://github.com/universal-ctags/ctags/files/1537703/config.h.txt)
[config.log](https://github.com/universal-ctags/ctags/files/1537699/config.log)
[Makefile.txt](https://github.com/universal-ctags/ctags/files/1537702/Makefile.txt)

masatake:
I found a reason.
Wait for awhile.

masatake:
I'm sorry. I misuderstood. I cannot find why you got the error.

Could you show me the output of "make V=1" ? I would like to know whether '-liconv' is passed to the compiler. As far as I read Makefile.txt, `-liconv` is passed to the compiler via LIBS macro.

lightingghost:
@masatake Here is the output of `make V=1`

```
./misc/gen-repoinfo main/repoinfo.h
/Library/Developer/CommandLineTools/usr/bin/make  all-am
./misc/gen-repoinfo main/repoinfo.h
[...]

PR #1610, opened 2017-12-10
Address the build process with macports on macOS

lightingghost:


masatake:
@lightingghost, thank you very much.

@cweagans, can I ask you to review this?
This is related to #1606.

cweagans:
These instructions will work, but they do not produce a universal-ctags binary built against the macports-installed iconv. If it were up to me, we wouldn't support macports _at all_ (Homebrew is where the majority of Mac users are moving these days), but if we're going to, then I guess this is an okay solution.

b4n:
If the _libiconv versions are literal equivalents to the _iconv ones we expect, what about suggesting `CPPFLAGS=-D_iconv=_libiconv -D_iconv_open=_libiconv_open -D_iconv_close=_libiconv_close` instead?

cweagans:
^ sounds like a better path forward. I would recommend changing the universal ctags port to do that automatically. Having macports not rename those symbols would be the ideal case, though.

masatake:
See #1615.

masatake:
@lightingghost, ping.

lightingghost:
@masatake I am sorry I missed your previous comments. I have just tested your branch for fixing the problem, it works for me.

masatake:
@lightingghost, thank you for testing. My branch is merged.

PR #1611, opened 2017-12-10
Convert slash to backslash in exclude option

masatake:
A test case is needed.

### DIFF
--- a/Tmain/ptag-kind-desc.d/run.sh
+++ b/Tmain/ptag-kind-desc.d/run.sh
@@ -3,7 +3,7 @@
 
 CTAGS=$1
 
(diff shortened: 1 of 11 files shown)


masatake:
I have no time to write a test case. I will write a test case for this item in the next March.
Anyone, who has knowledge about win32, can add a test case and merge this.

masatake:
The change is not enough. If an user uses a slash as path separator for specifying input files, the change doesn't work well because the change converts slash to backslash in patterns of --exclude.
A stage for normalization is needed when applying the patterns.

masatake:
Crazy...

PR #1619, opened 2017-12-25
yacc.c: updated syntax

jannick0:
updated yacc syntax to bison 3.0.2 and fixing compile bug for MINGW32


masatake:

Could you add a test case for your change?
See http://docs.ctags.io/en/latest/units.html and Units/parser-yacc.d.

masatake:
Could you drop the change for MINGW32?
I think we should take a different apraoch. I'm happy if you open a new issue for the original bug behind the chnage. It is about seccomp on Linux.

codebrainz:
Does anyone else find it ironic that this YACC parser does not use YACC? Just sayin' :)

masatake:
@jannick0, thank you very much for the many contributions these days. However, I would like to merge this change first. Could you consider resubmit this change(writing a test case) before making more pull request?

I don't know you motivation(something about security assurance?), however, I really want people who is interested in improving u-ctags to know the test harnesses.

masatake:
@codebrainz, I will add code to the build script for utilizing  bison/flex after making a initial release.

jannick0:
@masatake I am currently working on that. I was starting off having a little look-head issue, then I was getting deeper into the new multiline feature (which caused some of the other issues I recently posted). I am inclined to dismiss the patch above and probably use a multiline approach using (PCRE complaint regular) expressions. I hope I can pull something including tests shortly if all that will work out well.

As to flex/bison - I am currently thinking that this would be a little bit of a stretch, since we do not want to parse the full flex and bison grammar (for which the grammars are already implemented, of course, - including there actions relevant for flex/bison). However for the tagging purposes here hand-written automatons or multiline regular expressions should probably do I strongly. But again, this depends on if the fully fletched grasp of the flex/bison AST is needed here.

masatake:
BTW, yacc (and rpmspec) parsers have scars; callback functins associated with regular expressions call
 readLineFromInputFile. This is wrong code that I wrote. I must fix this.

jannick0:
Continued in #1640 ...

PR #1621, opened 2017-12-26
[SELF INCOMPATIBLE] main: introduce version 3 format and remove --out…

masatake:
…put-format=u-ctags|e-ctags option

To record meta characters like tab in a name, we have designed new file format.

The file format itself was already introduced. TAG_OUTPUT_MODE pseudo tag was used
as marker telling which format is used in a tag file: Exuberant-ctags(e-ctags)
compatible format or Universal-ctags(u-ctags) file format. u-ctags was the default
format. A user could choose the format with --output-format option.

However, we recognize that the hack using TAG_OUTPUT_MODE pseudo doesn't work well
with readtags command. A data structure of readtags doesn't have a field for
recording TAG_OUTPUT_MODE. It means readtags cannot switch the file format version
dynamically. The pseudo tags can be used in readtags for the purpose, recognizing
the file format is only TAG_FILE_FORMAT.

This commit introduces following SELF INCOMPATIBLE changes:

1. --output-format=u-ctags and --output-format=e-ctags are removed.
   They are unified to --output-format=ctags.

2. TAG_OUTPUT_MODE pseudo tag is removed.
   Instead TAG_FILE_FORMAT=3 is introduced.
   TAG_FILE_FORMAT=2 is compatible with e-ctags.
   TAG_FILE_FORMAT=3 is default. A user can choose one of the formats
   with --format=3 or --format=2.

The new version 3 may break many client tools.
If you are just user of such client tools, use --format=2 option to generate
tags file with compatible file format.
If you are a developer of such client tools, please, consider to support
--format=3.

The rest of work are:

1. more Tmain test cases,
2. more precise documentation about the format 3.
3. updating readtags to support format 3.

About readtags, the most of all work is done by @b4n.
I must merge his changes(#1605).

### DIFF
--- a/Tmain/json-output-format.d/stdout-expected.txt
+++ b/Tmain/json-output-format.d/stdout-expected.txt
@@ -16,7 +16,6 @@
 {"_type": "tag", "name": "main", "path": "input.go", "pattern": "/^package main$/", "language": "Go", "line": 1, "kind": "package"}
 # json --languages=+man --fields=* --extras=*-{subword}
 {"_type": "ptag", "name": "JSON_OUTPUT_VERSION", "path": "0.0", "pattern": "in development"}
-{"_type": "ptag", "name": "TAG_FILE_FORMAT", "path": "2", "pattern": "extended format; --format=1 will not append ;\" to lines"}
 {"_type": "ptag", "name": "TAG_FILE_SORTED", "path": "1", "pattern": "0=unsorted, 1=sorted, 2=foldcase"}
 {"_type": "ptag", "name": "TAG_PROGRAM_AUTHOR", "path": "Universal Ctags Team", "pattern": ""}
 {"_type": "ptag", "name": "TAG_PROGRAM_NAME", "path": "Universal Ctags", "pattern": "Derived from Exuberant Ctags"}
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -912,6 +912,23 @@ Changes to the tags file format
 ---------------------------------------------------------------------
 
 
+
+version 3 format for handling meta characters in tag entry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. REVIEWED
+
+To allow recording names including tab characters and other control
+characters in a tag file, version 3 of the file format extends version
+2 with extra escape sequences.
+
+You can use ``--format=2`` to keep compatibility with the output of
+Exuberant-ctags; however, this version of the format cannot record
+names which include tab characters.
+
+See also :ref:`Compatible output and weakness <compat-output>`.
+
+
 Truncating the pattern for long input lines
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 To prevent generating overly large tags files, a pattern field is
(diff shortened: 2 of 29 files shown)


docs/news.rst
```diff
@@ -911,6 +911,20 @@ Subparsers can be listed with ``--list-subparser``:
 Changes to the tags file format
 ---------------------------------------------------------------------
 
+.. NOT REVIEWED YET
+
+version 3 format for handling meta characters in tag entry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To make allow to record a name including a tab character to a tag file,
+the tags file format is extended, and we call the new file format version 3.
```
b4n: *“To allow recording names including tab characters and other control characters in a tag file, version 3 of the file format extends version 2 with extra escape sequences.”*

docs/news.rst
```diff
@@ -911,6 +911,20 @@ Subparsers can be listed with ``--list-subparser``:
...
+
+You can use ``--format=2`` to keep the compatibility between the output of
+Exuberant-ctgs. However, the old file format cannot record a name including
+a tab character.
```
b4n: *“You can use ``--format=2`` to keep compatibility with the output of Exuberant-CTags. However, this version of the format cannot record a names including tab characters.”*

main/options.c
```diff
@@ -267,8 +267,10 @@ static optionDescription LongOptionDescription [] = {
  {0,"  --format=level"},
 #if DEFAULT_FILE_FORMAT == 1
  {0,"       Force output of specified tag file format [1]."},
-#else
+#elif DEFAULT_FILE_FORMAT == 2
  {0,"       Force output of specified tag file format [2]."},
+#else
+ {0,"       Force output of specified tag file format [3]."},
```
b4n: why not simply use this?
```C
 {0,"       Force output of specified tag file format [" #DEFAULT_FILE_FORMAT "]."},
```

masatake:
@b4n, thank you very much. I will reflect your comment to the changes.

I would like to introduce format=3 atomically. 
So I will use a format-3 branch for developing. When the development of the branch is stable, I will merge it into our master branch. I will merge your readtags work to the branch.

Tmain/json-output-format.d/stdout-expected.txt
```diff
@@ -16,7 +16,7 @@
 {"_type": "tag", "name": "main", "path": "input.go", "pattern": "/^package main$/", "language": "Go", "line": 1, "kind": "package"}
 # json --languages=+man --fields=* --extras=*-{subword}
 {"_type": "ptag", "name": "JSON_OUTPUT_VERSION", "path": "0.0", "pattern": "in development"}
-{"_type": "ptag", "name": "TAG_FILE_FORMAT", "path": "2", "pattern": "extended format; --format=1 will not append ;\" to lines"}
```
masatake: I wonder TAG_FILE_FORMAT in json output is needed or not. Maybe it is not needed.

Tmain/json-output-format.d/stdout-expected.txt
```diff
@@ -16,7 +16,7 @@
...
+{"_type": "ptag", "name": "TAG_FILE_FORMAT", "path": "3", "pattern": "extended format with meta character escaping"}
 {"_type": "ptag", "name": "TAG_FILE_SORTED", "path": "1", "pattern": "0=unsorted, 1=sorted, 2=foldcase"}
```
masatake: Nothing tod owith output format but I wonder soriting json output makes sense or not.

docs/news.rst
codebrainz: *“You can use `--format=2` to keep compatibility with the output of Exuberant-CTags; however, this version of the format cannot record names which include tab characters.”*

Tmain/json-output-format.d/stdout-expected.txt
b4n: It's probably interesting to have a versioned format for the JSON output too (e.g. if one day wanting to change the layout), but it probably doesn't make sense to use the same one as for the regular format, as it's totally different and changes to one wouldn't affect the other.

masatake: Yes.
We have JSON_OUTPUT_VERSION!
So TAG_FILE_FORMAT doesn't make sense. I will remove it in another pull request.

PR #1627, opened 2017-12-27
CTagsSelfTest: call getppid only when HAVE_SECCOMP is defined

### LINKED ISSUE #1625: [DEBUG / MinGW] compilation error since getppid() not existing
A CPP guard excluding this case branch for MinGW https://github.com/universal-ctags/ctags/blob/d1616530d659ba603454bde3506f22e7eb9eae7a/main/parse.c#L3962-L3966 like so
```c
#if DEBUG && ! __MINGW32__
```
would avoid a compilation error, since `getppid()` is not existing for `MinGW` IIRC. Some other OSes supposedly to be dealt with as well.

masatake:
Close #1625.
Inspired by @jannick0.

CTagsSelfTest calls getppid for testing the sandbox feature.
Calling getppid dones't make sense if ctags is built at
a platform where the sandbox feature is not available.

### DIFF
--- a/main/parse.c
+++ b/main/parse.c
@@ -3843,7 +3843,7 @@ extern void addLanguageTagMultiTableRegex(const langType language,
  * A parser for CTagsSelfTest (CTST)
  */
 #define SELF_TEST_PARSER "CTagsSelfTest"
-#ifdef DEBUG
+#if defined(DEBUG) && defined(HAVE_SECCOMP)
 extern void getppid(void);
 #endif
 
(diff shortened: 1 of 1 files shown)


jannick0:
That looks good to me. 

A tiny remark which is certainly up to you how you handle that, while I must admit that the package code is nicely readable: I would rather say
```c
#if DEBUG && HAVE_SECCOMP
```
instead of
```c
#if defined(DEBUG) && defined(HAVE_SECCOMP)
```
which is what we want, since the latter would be sadly true if `DEBUG=0` and `HAVE_SECCOMP=0` were set due to some nasty configuration. The preprocessor interprets any unknown token as 0 by default, which makes our life a lot easier here.

masatake:
Thank you.

I took mistake.

```
#if DEBUG && defined(HAVE_SECCOMP)
```
should be

```
#if defined(DEBUG) && defined(HAVE_SECCOMP)
```

defiend is used all the other areas, so I will use here, too.

PR #1631, opened 2017-12-30
[main/read.c | multiline]: fix SIGSEGV

jannick0:
Technical fix. Logic in `iFileGetLine()` for multiline to be reviewed.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -877,7 +877,7 @@ static vString *iFileGetLine (void)
 			parseLineDirective (vStringValue (File.line) + 1);
 		matchLanguageRegex (getInputLanguage (), File.line);
 
-		if (use_multiline)
+		if (use_multiline && File.allLines)
 			vStringCat (File.allLines, File.line);
 
 		return File.line;


main/read.c
```diff
@@ -877,7 +877,7 @@ static vString *iFileGetLine (void)
 			parseLineDirective (vStringValue (File.line) + 1);
 		matchLanguageRegex (getInputLanguage (), File.line);
 
-		if (use_multiline)
+		if (use_multiline && File.allLines)
```
b4n: is this a real issue you're fixing, or just a theoretical one?  I don't see how this is useful as on https://github.com/universal-ctags/ctags/pull/1631/files#diff-0a4427393487ed8d5f467beb7f72034fR863 `File.allLines` is created unconditionally, and AFAIK the allocation is checked and the program aborts if it fails, so `File.allLines` *has been* created.

masatake:
As @b4n wrote, use_multiline implies Fill.alllines != NULL.
Do you really get SIGSEGV? If yes, I would like to know the condiation.

masatake:
#1640 triggers SEGV. However, it is not a bug of main part. Wrong design of yacc parser triggers SEGV.

k-takata:
The OP doesn't reply to the question, so closing. If this is a real issue, please reopen.

Issue #1632, opened 2017-12-30
[main/read.c] introduce variable lang to avoid multiple calls


jannick0:
... to avoid multiple calls of `getInputLanguage ()` in central function `iFileGetLine`  I would like to suggest. I am happy to file a pull request.

https://github.com/universal-ctags/ctags/blob/4f9f6f83de929385c448e59d40e5b62588b6bbcb/main/read.c#L855-L896

masatake:
What kind of suggestion? I would like to see it in a pull request because My C language skill is better than my English skill. Thank you.

PR #1634, opened 2017-12-30
main/read.c: reduce calls of getInputLanguage

jannick0:
As per #1632.


masatake:
LGTM. Just a question. I wonder why you think this change is needed.
For optimization or for readability?

jannick0:
Thanks. It is about readability and to make sure that the language is not changed by any other function on the fly. Optimzation is not really my cup of tea.

masatake:
Thank you.

Issue #1623, opened 2017-12-26
[main/fmt.c] uninitialized variable 'f'


jannick0:
https://github.com/universal-ctags/ctags/blob/d1616530d659ba603454bde3506f22e7eb9eae7a/main/fmt.c#L78

masatake:
Does your compiler warn this?
If yes, I would like to know which compiler you are using.


Look at the for-loop ahead of the line.
f is initialized in the body of loop when tag->usedParserFields != 0.

If tag->usedParserFields == 0, there is no chance for initialzing f. However, in that case
we can expect findex == tag->usedParserFields. In this case f is not referred.

jannick0:
The compiler does not grasp the semantic logic unfortunately. 

Here a suggestion which essentially implements what you are saying above and would pacify the compiler. Moreover, the case that `getParserField` returns `NULL` (which is not excluded) would be covered then, too.
```diff
@@ -64,7 +64,7 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 	else
 	{
 		unsigned int findex;
-		const tagField *f;
+		const tagField *f = NULL;
 
 		for (findex = 0; findex < tag->usedParserFields; findex++)
 		{
@@ -73,7 +73,7 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 				break;
 		}
 
-		if (findex == tag->usedParserFields)
+		if ( !f)
 			str = "";
 		else if (isFieldEnabled (f->ftype))
 			/* TODO: Don't use WRITER_XREF directly */
```

jannick0:
@masatake Cf. [travis'](https://travis-ci.org/universal-ctags/ctags/jobs/323065084) output with the complaints of mingw's gcc.

b4n:
I quite like the code in https://github.com/universal-ctags/ctags/issues/1623#issuecomment-354044391

masatake:
@b4n, thank you.

@jannick0, could you make a pull request based on your comment?

PR #1635, opened 2017-12-30
fmt.c: zero-initialize variable

jannick0:
As per #1623

### DIFF
--- a/main/fmt.c
+++ b/main/fmt.c
@@ -64,7 +64,7 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 	else
 	{
 		unsigned int findex;
-		const tagField *f;
+		const tagField *f = NULL;
 
 		for (findex = 0; findex < tag->usedParserFields; findex++)
 		{
@@ -73,7 +73,7 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 				break;
 		}
 
-		if (findex == tag->usedParserFields)
+		if (!f)
 			str = "";
 		else if (isFieldEnabled (f->ftype))
 			/* TODO: Don't use WRITER_XREF directly */


main/fmt.c
```diff
@@ -73,7 +73,7 @@ static int printTagField (fmtSpec* fspec, MIO* fp, const tagEntryInfo * tag)
 				break;
 		}
 
-		if (findex == tag->usedParserFields)
+		if (!f)
```
b4n: shoot, sorry, this actually isn't equivalent: here `f` won't be `NULL` unless `tag->usedParserFields == 0`; but before it also caught the case when no field passed the `isParserFieldCompatibleWithFtype (f, ftype)` test.

So this should be fixed, maybe like:
```C
 		for (findex = 0; findex < tag->usedParserFields; findex++)
 		{
            tagField *fitem = getParserField(tag, findex);
 			if (isParserFieldCompatibleWithFtype (fitem, ftype))
            {
 				f = fitem;
                break;
            }
 		}
```
or
```C
 		for (findex = 0; findex < tag->usedParserFields; findex++)
 		{
            f = getParserField(tag, findex);
 			if (isParserFieldCompatibleWithFtype (f, ftype))
                break;
            else
 				f = NULL;
 		}
```

b4n: @masatake why on [line 80](https://github.com/universal-ctags/ctags/pull/1635/files#diff-be3bbc112b04c745703ab241b76456d4R80) does it access `tag->parserFields [findex].ftype` rather than `f->ftype`?  wouldn't it be the same, and not depend on the value of `findex` nor on the internals of `tagEntryInfo` when there is `getParserField()` already?

anyway @jannick0 beware to keep `findex` the same so long as it's still used outside the loop -- I was about to suggest something like 

```C
 		for (findex = 0; ! f && findex < tag->usedParserFields; findex++)
 		{
            tagField *fitem = getParserField(tag, findex);
 			if (isParserFieldCompatibleWithFtype (fitem, ftype))
 				f = fitem;
 		}
```
above, but that changes the value of `findex` outside the loop, which is **_not_** acceptable here (yet).

masatake: @b4n, I find it is my mistake.

`tag->parserFields [findex].ftype` should be `f->type`.
I forgot replacing it in 4e9e52c2d6903ac4704c84358d40b56b28c55d10.

b4n: OK so following @masatake's comment, I'd do something like this:
```C
		for (unsigned int findex = 0; ! field && findex < tag->usedParserFields; findex++)
		{
			tagField *f = getParserField(tag, findex);
			if (isParserFieldCompatibleWithFtype (f, ftype))
 				field = f;
 		}
```
(replacing `f` with `field` for clarity in the function body, and stopping use of `findex` outside the loop body)

masatake: @b4n and @jannick0, I would like to fix what @b4n spotted separately from the change proposed in this pull request.

jannick0: Agree - then you guys take care of the right if condition and the other fix I would presume.

PR #1636, opened 2017-12-30
optlib.rst: fix short flag of 'extend'

jannick0:


masatake:
Good catch. Thank you.

masatake:
By the way, "docs(web):" is the prefix for the commit title(subject?) for these files.

PR #1637, opened 2017-12-30
main/read.c: add line numbers to debug read

jannick0:
Debugging should not yield to excessive line numbers, so `%04d` should be OK here.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -969,7 +969,7 @@ extern const unsigned char *readLineFromInputFile (void)
 	{
 		result = (const unsigned char*) vStringValue (line);
 		vStringStripNewline (line);
-		DebugStatement ( debugPrintf (DEBUG_READ, "%s\n", result); )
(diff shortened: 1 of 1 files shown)


masatake:
Could you add a example session of output to your commit log?

You may think my request is a bit strange. The debug infrastructure of u-ctags is broken. When I started working on this project, I didn't understand u-ctags has such facility.  So I modified many areas without 
considering the existing debu infrastructure. I would like to revise the debug infrastructure, and I 
would like to integrate it with code for tracing written by @pragmaware originally for C++ parser.

However, I cannot find time for it. So, simply saying, I have not understand the debug infra yet.

This is the reason why I want you to write the usage of this change to your commit log.

Anyway, an example input, output and/or tags file always help us.

masatake:
I didn't get the response for my question. Therefore I will clsoe this.

PR #1640, opened 2017-12-31
YACC parser: syntax update

jannick0:
Close #1619.
This PR replaces #1619 (since I am struggling with github a bit to kind of overwrite a prior PR). 

Innocently it uses multiline regexps which can do pretty much of the tagging job here.

More to come soon hopefully, but certainly in the next year:
- sample tests incl. false positives
- more enhancements of the syntax: definition of `%token`s, of `%code{ ...}` environment

**Have a Great and Happy New Year everybody!**

### DIFF
--- a/parsers/yacc.c
+++ b/parsers/yacc.c
@@ -26,10 +26,11 @@ static bool not_in_section2 = true;
 static bool not_in_section3 = true;
 static bool in_union;
 static tagRegexTable yaccTagRegexTable [] = {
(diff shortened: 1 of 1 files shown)


masatake:
(You can use `git push --force ...` for overwriting existing branch.)

masatake:
SEGV occurs becase, as I wrote, yacc parser calls readLineFromInputFile  from a regex callback function.

masatake:
See #1649.

PR #1641, opened 2018-01-03
Improve roundtrip test harness  generic part of b4n s work

masatake:
Generic part of #1605.

### DIFF
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -18,6 +18,10 @@ LANGUAGES=
 CATEGORIES=
 UNITS=
 
+SILENT = $(SILENT_@AM_V@)
+SILENT_ = $(SILENT_@AM_DEFAULT_V@)
+SILENT_0 = @
+
 V_RUN = $(V_RUN_@AM_V@)
 V_RUN_ = $(V_RUN_@AM_DEFAULT_V@)
 V_RUN_0 = @echo "  RUN      $@";
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -18,13 +18,17 @@ LANGUAGES=
 CATEGORIES=
 UNITS=
 
+V_RUN = $(V_RUN_@AM_V@)
+V_RUN_ = $(V_RUN_@AM_DEFAULT_V@)
+V_RUN_0 = @echo "  RUN      $@";
+
 #
 # FUZZ Target
 #
 # SHELL must be dash or bash.
 #
 fuzz: $(CTAGS_TEST)
-	@ \
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -40,7 +44,7 @@ fuzz: $(CTAGS_TEST)
 # NOISE Target
 #
 noise: $(CTAGS_TEST)
-	@ \
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -56,7 +60,7 @@ noise: $(CTAGS_TEST)
 # CHOP Target
 #
 chop: $(CTAGS_TEST)
-	@ \
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -68,7 +72,7 @@ chop: $(CTAGS_TEST)
 		--with-timeout=$(TIMEOUT)"; \
 	$(SHELL) $${c} $(srcdir)/Units
 slap: $(CTAGS_TEST)
-	@ \
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -84,7 +88,7 @@ slap: $(CTAGS_TEST)
 # UNITS Target
 #
 units: $(CTAGS_TEST)
-	@ \
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -115,8 +119,7 @@ clean-units:
 # Test main part, not parsers
 #
 tmain: $(CTAGS_TEST)
-	@ \
-	\
+	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
 		VALGRIND=--with-valgrind;	\
@@ -144,21 +147,17 @@ clean-tmain:
 # Test installation
 #
 tinst:
-	@ \
-	\
+	$(V_RUN) \
 	builddir=$$(pwd); \
 	rm -rf $$builddir/$(TINST_ROOT); \
-	\
 	$(SHELL) $(srcdir)/misc/tinst $(srcdir) $$builddir/$(TINST_ROOT)
 
 #
 # Test readtags
 #
 roundtrip: $(READ_TEST)
-	@ \
-	\
+	$(V_RUN) \
 	builddir=$$(pwd); \
-	\
 	$(SHELL) $(srcdir)/misc/roundtrip $(READ_TEST) $${builddir}/Units
 
 #
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -159,10 +159,14 @@ tinst:
 #
 # Test readtags
 #
+if USE_READCMD
 roundtrip: $(READ_TEST)
 	$(V_RUN) \
 	builddir=$$(pwd); \
 	$(SHELL) $(srcdir)/misc/roundtrip $(READ_TEST) $${builddir}/Units
+else
+roundtrip:
+endif
 
 #
 # Checking code in ctags own rules
--- a/misc/roundtrip
+++ b/misc/roundtrip
@@ -15,6 +15,8 @@
 #   You should have received a copy of the GNU General Public License
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+set -e
+
 # Avoid trouble with weird bytes on non-C locales
 export LC_ALL=C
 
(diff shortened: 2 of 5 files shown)


misc/roundtrip
```diff
@@ -40,16 +38,23 @@ if ! [ -x "${UNITS}" ]; then
...
+# disable path expansion, because we don't need it and it's applied on the
+# result of the commands used to loop on, and we don't want that.
+set -f
+# remove space from IFS as it's valid in tag names
+OLD_IFS="$IFS"
+IFS='
+'
+
 for tags in $(find "$UNITS" -name expected.tags); do
     for t in $(sed -e 's/^\([^\t]*\)\t.*/\1/' "$tags"); do
```
b4n: Use of `\t` breaks OSX support, see https://github.com/universal-ctags/ctags/pull/1605/commits/378c53e653d0db08310a8422a3ed7c2243ed9614

makefiles/testing.mak
```diff
@@ -1,7 +1,7 @@
 # -*- makefile -*-
 .PHONY: check units fuzz noise tmain tinst clean-units clean-tmain clean-gcov run-gcov codecheck cppcheck dicts cspell
 
-check: tmain units
+check: tmain units roundtrip
```
b4n: That's the tricky part: it's useful to have it so roundtrip actually checks something everywhere, but it doesn't work on Windows.  So maybe leave it out until Windows support is fixed?

masatake: I will also attack this crazy issue. If I cannot find a solution, too, I will do as you write and I will wait for @k-takata's help.

misc/roundtrip
masatake: Thank you. I merged 378c53e partially.

makefiles/testing.mak
masatake: I have read your efforts for making roundtrip work on MSYS2. Thank you very much.
I know well how it is painful.

masatake:
It is something to do with MSYS2_ARG_CONV_EXCL, but I think we should not take time for this any more.

b4n:
@masatake I was about to suggest indeed only running `roundtrip` on non-MSYS2.  I think it's the best solution for now.
I'm not sure it's better, but I would have then had `check` not run `roundtrip`, and explicitly run `roundtrip` in CircleCI and Travis' scripts (instead of *not* running it in AppVeyor).  The pros and cons I see is that while it's very nice to have as many tests as possible run on as many machines as possible (so having it on `check` is good), it's very bad for `check` to report false positives as it might prevent legitimate builds to erroneously fail.

So I would suggest to simply remove the `roundtrip` dependency from `check`, and call `roundtrip` from CircleCI and Travis.

masatake:
@b4n, I see. Thank you.

makefiles/testing.mak
```diff
@@ -159,10 +159,16 @@ tinst:
...
 # Test readtags
 #
+if USE_READCMD
 roundtrip: $(READ_TEST)
 	$(V_RUN) \
 	builddir=$$(pwd); \
 	$(SHELL) $(srcdir)/misc/roundtrip $(READ_TEST) $${builddir}/Units
+else
+# apparently Automake doesn't like adding an extra dependency in a conditional,
+# so add it unconditionally and simply have it not do anything if it's disabled.
```
b4n: you could remove this comment now as it doesn't make much sense anymore

makefiles/testing.mak
```diff
@@ -144,22 +151,24 @@ clean-tmain:
...
 roundtrip: $(READ_TEST)
-	@ \
-	\
+	$(V_RUN) \
 	builddir=$$(pwd); \
-	\
 	$(SHELL) $(srcdir)/misc/roundtrip $(READ_TEST) $${builddir}/Units
+else
+# apparently Automake doesn't like adding an extra dependency in a conditional,
+# so add it unconditionally and simply have it not do anything if it's disabled.
```
b4n: Not sure why my comment didn't make it here, but anyway: this comment is kind of outdated and could go

masatake:
`[PATCH 25/30] roundtrip: Don't hide error output from readtags` should go here, too.
Maybe it is introduced for debugging roundtrip on Msys2. However, it will be useful on the other platform.

masatake:
The rest is:
```
0001-readtags-Add-support-for-escape-sequences-in-tag-nam.patch
0005-roundtrip-Try-and-match-unescaped-names-too.patch
0008-roundtrip-Try-and-fix-macos-compatibility.patch
0011-roundtrip-Fix-unescaping.patch
0012-roundtrip-Fix-handling-of-escaped-newlines.patch
0013-readtags-Don-t-expect-escaped-search-string.patch
0015-fixup-roundtrip-Fix-unescaping.patch
0016-roundtrip-Fix-handling-of-trailing-newlines.patch
0017-fixup-readtags-Add-support-for-escape-sequences-in-t.patch
0020-roundtrip-Fix-MacOS-compatibility.patch
0021-roundtrip-Fix-some-escaping-cases.patch
0023-roundtrip-Further-OSX-fixing-hopefully.patch
0024-roundtrip-Add-a-test-case-for-all-escape-sequences.patch
0027-fixup-roundtrip-Further-OSX-fixing-hopefully.patch
```

The fixup changes will be squashed by my side.
My task is that making readtags version sensitive. My understanding is that @b4n's change is for version 3 format. The old code should be used for version 1 and 2. The new @b4n's code should be used for version 3. Such conditions and branch must be added to readtags.

Through the task, I may have enough chance to learn version 3. I will write it to tags(5) new man page.
I will also write ctags-incompatiblities(7) man page. Afte that we will release 5.9.

b4n:
> My task is that making readtags version sensitive. My understanding is that @b4n's change is for version 3 format. The old code should be used for version 1 and 2. The new @b4n's code should be used for version 3. Such conditions and branch must be added to readtags.

My changes should be fully backward compatible with v2: I just introduced new escape sequences which were just not valid format before.
And FWIW, my changes also make readtags handle escape sequences from v2, which was missing.

All in all, I think that with my changes readtags is fully capable of reading all 3 versions of the format without any problem, as AFAIK they are all backward-compatible (e.g. v1 is a subset of v2 which is a subset of v3).

masatake:
Surprisingly it works well both v3 and v2, as you wrote. I understand the trick. I thought such change cannot be done without referring the version ptag.

```
[yamato@master]~/var/ctags-github% cat /tmp/input.php 
cat /tmp/input.php 
Tests for namespaces (unbraced version)

Expected output is

namespaces:
	Bar\Baz
	Foo

classes:
	B [Bar\Baz]
	C [Foo]

functions:
	__construct [Foo\C]
	__construct [Bar\Baz\B]
	a [Foo]
	a [Bar\Baz]
	b [Foo]
	c [Bar\Baz\B]
	d [Foo\C]

<?php

/* namespace "Foo" */
namespace Foo;

function a() {
	return true;
}

function b() {
	return false;
}

class C {
	function __construct() {
		// ...
	}

	function d() {
		return 42;
	}
}

/* namespace "Bar\Baz" */
namespace Bar\Baz;

function a() {
	return true;
}

class B {
	function __construct() {
		// ...
	}

	function c() {
		return 42;
	}
}
[yamato@master]~/var/ctags-github% for v in 2 3; do ./ctags --format=$v -o $v.tags /tmp/input.php ;done
for v in 2 3; do ./ctags --format=$v -o $v.tags /tmp/input.php ;done
[yamato@master]~/var/ctags-github% ./readtags -t 2.tags 'Bar\Baz'
./readtags -t 2.tags 'Bar\Baz'
Bar\Baz	/tmp/input.php	/^namespace Bar\\Baz;$/
[yamato@master]~/var/ctags-github% ./readtags -t 3.tags 'Bar\Baz'
./readtags -t 3.tags 'Bar\Baz'
Bar\Baz	/tmp/input.php	/^namespace Bar\\Baz;$/
```

For the both tags files, readtags can find entries for the same key, `Bar\Baz`.

I found some storange behaviours when I run readtags with -e option.

```
 [yamato@master]~/var/ctags-github% ./readtags -e -t 2.tags 'Bar\Baz'
./readtags -e -t 2.tags 'Bar\Baz'
Bar\Baz	/tmp/input.php	/^namespace Bar\\Baz;$/;"	kind:n
[yamato@master]~/var/ctags-github% ./readtags -e -t 3.tags 'Bar\Baz'
./readtags -e -t 3.tags 'Bar\Baz'
Bar\Baz	/tmp/input.php	/^namespace Bar\\Baz;$/;"	kind:nn
[yamato@master]~/var/ctags-github% ./readtags -e -t 3.tags '__construct'
./readtags -e -t 3.tags '__construct'
__construct	/tmp/input.php	/^	function __construct() {$/;"	kind:f	class:Bar\\Baz\\B
__construct	/tmp/input.php	/^	function __construct() {$/;"	kind:f	class:Foo\\C
[yamato@master]~/var/ctags-github% ./readtags -e -t 2.tags '__construct'
./readtags -e -t 2.tags '__construct'
__construct	/tmp/input.php	/^	function __construct() {$/;"	kind:f	class:Bar\Baz\B
__construct	/tmp/input.php	/^	function __construct() {$/;"	kind:f	class:Foo\C
```

See kind fields. In v2, just `n` is printed. However, for v3, `nn` is printed.
About scope information (started from class:), backslash is used twice in v3 output.

b4n:
Which branch are you using?  I don't have a ready branch with v3 format and readtags changes, should I merge these manually to try?  Anyway, sounds like a bug somewhere indeed; but as my own changes didn't follow any format option I can't really tell just now.

> About scope information (started from class:), backslash is used twice in v3 output.

Well… the thing is that no format version really tell anything about the encoding of the extension fields, only the name part is really mentioned. Anyway I would think it should follow the same escaping rules as the name itself, seems the only reasonable solution to me (no matter the version).

masatake:
Oh, sorry. The branch I'm refering is `format-3` branch. After checkout the branch, I applies three patches you wrote:

```
313fdf87379f52fdc65c07bf077d7ff66dc869e0 ... 0001-readtags-Add-support-for-escape-sequences-in-tag-nam.patch
08673db11511dbd8b4e2186e476f9bc89db9269d ... 0013-readtags-Don-t-expect-escaped-search-string.patch
29fdc7078bba80c3a62a24ac97deec3cb2d3fa73 ... 0017-fixup-readtags-Add-support-for-escape-sequences-in-t.patch
```

> Well… the thing is that no format version really tell anything about the encoding of the extension fields, only the name part is really mentioned. Anyway I would think it should follow the same escaping rules as the name itself, seems the only reasonable solution to me (no matter the version).

In the version 2 format, a field value cannot contain a tab character because the character is
used as a column separator.
With the escaping rule of version 3, a field value can contain a tab character. As you know well,
a backslash is introduced to include a tab character.

> (no matter the version)
I think we cannot introduce such escaping rule to the version 2 format because it breaks compatibility.
This is the reason why we introduce version 3 format.

However, parsers don't know about the format. A parser can make a tag which contains a tab char in its name or in its one of field values. If such tag is passed to `writer-ctags.c` and using version 2 is specified by user, u-ctags just skips printing such tag.

Anyway I have to read your change. That's the first  step.

PR #1643, opened 2018-01-03
[CXX] Fix c++14 digit separator

### LINKED ISSUE #1642: Wrong parsing when number of digit separators is odd [C++14]
If the file containing such a code:
```
int X = 100'000;
int Y = 100000;
```
ctags generates only tag for X variable:
```
X	a.cpp	/^int X = 100'000;$/;"	v	typeref:typename:int
```

However, in the following case:
```
int X = 100'00'0;
int Y = 100000;
``` 
when the number of digit separators in the first number is even, ctags correctly generates both tags:
```
X	a.cpp	/^int X = 10'000'0;$/;"	v	typeref:typename:int
Y	a.cpp	/^int Y = 100000;$/;"	v	typeref:typename:int
```

pragmaware:
Fix parsing of C++14 digit separators.
Fixes #1642.


masatake:
How about using isxdigit()?

masatake:
Could you add a newline before "else" like:
```
}
else if (...)
{
```


Other than this, it looks good for me.

PR #1645, opened 2018-01-04
Adding debug option to readtags

masatake:
For helping us debugging readtags on AppVeyor.

### DIFF
--- a/read/readtags-cmd.c
+++ b/read/readtags-cmd.c
@@ -17,6 +17,7 @@ static int extensionFields;
 static int SortOverride;
 static sortType SortMethod;
 static int allowPrintLineNumber;
+static int debugMode;
 #ifdef QUALIFIER
 #include "dsl/qualifier.h"
 static QCode *Qualifier;
@@ -143,6 +144,7 @@ static const char *const Usage =
 #endif
 	"[-s[0|1]] [-t file] [-] [name(s)]\n\n"
 	"Options:\n"
+	"    -d           Turn on debugging output.\n"
 	"    -e           Include extension fields in output.\n"
 	"    -h           Print this help message.\n"
 	"    -i           Perform case-insensitive matching.\n"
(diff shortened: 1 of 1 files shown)


read/readtags-cmd.c
```diff
@@ -17,6 +17,7 @@ static int extensionFields;
 static int SortOverride;
 static sortType SortMethod;
 static int allowPrintLineNumber;
+static int debugMode;
```
b4n: what about an explicit initialization?  I feel like it's a bit too magical for the casual reader not completely familiar with implicit initialization details in C.

masatake: How about yourself?

I think casual readers are not important to us. 
You are one of the most active developers.  (Focusing on active developers is my policy running this project. ) So if you feel uncomfortable with the code,
we must fix it. However, when introducing explicit initialization, it should be introduced to all the non-local variables.

Currently, I don't initialize explicitly because (1) implicit initializing was rather popular in Exuberant-ctags, and (2) I myself am familiar with implicit initializing.  I myself can adapt
to code using explicit initialization.

If you feel uncomfortable with implicit initialization, please make a big patch.

Anyway, I will merge this. The aim of this pull request is for merging your readtags code.
I'm trying to merge your code in incrementally way.

b4n: Well, I know the rules so I have no problem here, but I'm also used to notice and "fix" these because I know most people aren't familiar with it and leads to them having a harder time following the code.  But whichever you prefer is fine by me, although I like being explicit myself.

masatake: ... I found explicit initialization used widely in glib.

```
[jet@localhost]/srv/sources/sources/g/glib2/2.22.5-6.el6/pre-build/glib-2.22.5% find -name '*.c' | xargs grep '^static' | grep ';' | grep -v '(' | grep -v '='  | wc -l
110
[jet@localhost]/srv/sources/sources/g/glib2/2.22.5-6.el6/pre-build/glib-2.22.5% find -name '*.c' | xargs grep '^static' | grep ';' | grep -v '(' | grep -e '= *\(0\|NULL\)' | wc -l
232
```

```
[jet@localhost]~/var/linux% find -name '*.c' | xargs grep '^static' | grep ';' | grep -v '(' | grep -e '= *\(0\|NULL\)' | wc -l
1079
[jet@localhost]~/var/linux% find -name '*.c' | xargs grep '^static' | grep ';' | grep -v '(' | grep -v '=' | wc -l             
16664
```

How do you do in geany? Do you have any coding rule in the project?

Issue #1653, opened 2018-01-09
Possible NULL dereference in storeAdaTags


SiIky:
https://github.com/universal-ctags/ctags/blob/4dbeb0be3a2e2f710ca8468badeba7611b79f906/parsers/ada.c#L2051-L2171
First possible NULL derefence at line 2088.
https://github.com/universal-ctags/ctags/blob/4dbeb0be3a2e2f710ca8468badeba7611b79f906/parsers/ada.c#L2088
`storeAdaTags()` is called only twice and `token` is not NULL in neither of those, so I think the if should either be removed or have the rest of the function inside it.

SiIky:
I didn't make a pull request because I didn't know which you preferred (to remove the if or not)
No, I used `cppcheck` :p I found another bug today so I'll create a new issue for it later and elaborate on `cppcheck`

masatake:
I see.

SiIky:
Fixed in 047870360a5e6995cd52b79343fdcc7327d19c4e

PR #1654, opened 2018-01-10
Ada,refactor: delete redundant condition

masatake:
Suggested by @Silky in #1653.

The condtion `token != NULL` in storeAdaTags is redundant
because its callers checks the same condtion before calling
it.

To avoid potential programming mistake, I put Assert instead.


silky:
infact, suggested by @Silky not @silky, :)

PR #1656, opened 2018-01-10
Fix NULL check

### LINKED ISSUE #1655: Wrong NULL check
https://github.com/universal-ctags/ctags/blob/796684b6bb67dc9cc738f619baec89f9ece863fa/parsers/cxx/cxx_token_chain.c#L291-L300
Returns early if `t` is non NULL, otherwise, since `vStringCatS()` doesn't touch `t` and `cxxTokenAppendToString()` doesn't check for NULL `t` is dereferenced.

As mentioned in my other issue, I used `cppcheck` to find these.
```sh
find . -name '*.[ch]' |\
 cppcheck --std=c99 -v --language=c -f --enable=warning --file-list=- 2&>cppcheck.log
```
Don't know where all the code and tests are, but thought the parsers directory was a good place to start :) The message cppcheck gave for this one was:
```
[parsers/cxx/cxx_token_chain.c:294] -> [parsers/cxx/cxx_token_chain.c:304]: (warning) Either the condition 't' is redundant or there is possible null pointer dereference: t.
```

SiIky:
Fix #1655

### DIFF
--- a/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/README
+++ b/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/README
@@ -1,0 +1,6 @@
+S::T::E::alpha and S::T::E::beta are not included intentionally
+Taken from the comments of cxx_tag.c::
+
+	// If the scope kind is enumeration then we need to remove the
+	// last scope part. This is what old ctags did.
+
--- a/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/args.ctags
+++ b/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--extra=+q
--- a/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/expected.tags
+++ b/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/expected.tags
@@ -1,0 +1,12 @@
+S	input.cc	/^namespace S {$/;"	n	file:
+T	input.cc	/^  struct T {$/;"	s	namespace:S	file:
+S::T	input.cc	/^  struct T {$/;"	s	namespace:S	file:
+E	input.cc	/^    enum E {$/;"	g	struct:S::T	file:
+S::T::E	input.cc	/^    enum E {$/;"	g	struct:S::T	file:
+alpha	input.cc	/^      alpha, beta,$/;"	e	enum:S::T::E	file:
+S::T::alpha	input.cc	/^      alpha, beta,$/;"	e	enum:S::T::E	file:
+beta	input.cc	/^      alpha, beta,$/;"	e	enum:S::T::E	file:
+S::T::beta	input.cc	/^      alpha, beta,$/;"	e	enum:S::T::E	file:
+elt	input.cc	/^    } elt;$/;"	m	struct:S::T	typeref:enum:S::T::E	file:
+S::T::elt	input.cc	/^    } elt;$/;"	m	struct:S::T	typeref:enum:S::T::E	file:
+s	input.cc	/^struct S::T s = { .elt = S::T::E::alpha };$/;"	v	typeref:struct:S::T
--- a/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/input.cc
+++ b/Units/parser-cxx.r/enum-in-a-struct--with-q-extra.d/input.cc
@@ -1,0 +1,9 @@
+namespace S {
+  struct T {
+    enum E {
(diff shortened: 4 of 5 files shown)


masatake:
```
[jet@localhost]~/var/ctags% cat /tmp/foo.cc 
cat /tmp/foo.cc 
namespace S {
  struct T {
    enum E {
      TOKYO, KYOTO,
    } elt;
  };
}

struct S::T s = { .elt = S::T::E::TOKYO };

[jet@localhost]~/var/ctags% diff -uN WITHOUT-PATCH.tags WITH-PATCH.tags 
diff -uN WITHOUT-PATCH.tags WITH-PATCH.tags 
--- WITHOUT-PATCH.tags	2018-01-11 11:54:21.299130364 +0900
+++ WITH-PATCH.tags	2018-01-11 11:54:33.683074701 +0900
@@ -11,9 +11,9 @@
 E	/tmp/foo.cc	/^    enum E {$/;"	g	struct:S::T	file:
 S::T::E	/tmp/foo.cc	/^    enum E {$/;"	g	struct:S::T	file:
 TOKYO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
-S::TOKYO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
+S::T::TOKYO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
 KYOTO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
-S::KYOTO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
+S::T::KYOTO	/tmp/foo.cc	/^      TOKYO, KYOTO,$/;"	e	enum:S::T::E	file:
 elt	/tmp/foo.cc	/^    } elt;$/;"	m	struct:S::T	typeref:enum:S::T::E	file:
 S::T::elt	/tmp/foo.cc	/^    } elt;$/;"	m	struct:S::T	typeref:enum:S::T::E	file:
 s	/tmp/foo.cc	/^struct S::T s = { .elt = S::T::E::TOKYO };$/;"	v	typeref:struct:S::T
```

The result is obviously improved.
`S::T::E::TOKYO` is not included as a tag but it seems that it is intended.

```
	if(eScopeType == CXXScopeTypeEnum)
	{
		// If the scope kind is enumeration then we need to remove the
		// last scope part. This is what old ctags did.
		if(cxxScopeGetSize() < 2)
			return -1; // toplevel enum
```

Actually, with your patch, u-ctags generates the same tags file as e-ctags generates.

I will a pull request to your branch. So could you push it here?

(
BTW, I think you should not use master branch on your side.
As far as I know, the master branch should be used for synchronizing upstream.
If you modify your master branch with a hacking activity, you cannot do another hack.
)

masatake:
See https://github.com/SiIky/ctags/pull/1.

masatake:
(Thank you for pushing my change.)

SiIky:
Merged
>I think you should not use master branch on your side

Sorry about that, next time I'll do it right. (This was my first PR)

masatake:
Let's wait for the response from the maintainer of C/C++ parsers.

> (This was my first PR)

I'm very proud of u-ctags project attracting a new hacker.
Though the required knowledge for developing u-ctags is only C and standard C library, I think developing u-ctags is very ... attractive. I cannot find a good word representing how I feel. 

I hope that you enjoy hacking  u-ctags more. So I would like to add more comments here.
I would like you to add a prefix to the subject of commit log like:

```
C++: fix NULL check
```
Here `C++` is the prefix. If a patch is about a parser, use the name of parser as prefix.
Use `main` for the change for the common part of code. Please, see the other commit logs about more examples.

Knowing how to write a test case is important in this project. Please look at my patch which I submitted to your branch.  See also http://docs.ctags.io/en/latest/units.html .

Thank you.

pragmaware:
Yep, looks good!
It's interesting that this bug didn't pop out earlier... it's there since the beginning of the parser.

SiIky:
(Didn't have time to reply yesterday)

>I'm very proud of u-ctags project attracting a new hacker.

I don't know if I'll stay for long or contribute much to the project, but I'd like to at least try.

>Though the required knowledge for developing u-ctags is only C and
>standard C library

Not sure about those requirements, it seems to me there's a lot to learn before I can even grasp the ctags big picture and contribute something more valuable... (this PR shows exactly that, a C bug, didn't make tests or anything)

>I would like you to add a prefix to the subject of commit

Done, was that what you wanted? Tell me if there's anything missing

masatake:
Thank you.

PR #1661, opened 2018-01-15
Sh,Units: skip heredoc line if no command follows env command

masatake:
Here document detector in shell parser expects a command
is specified after env commande like:

    env perl <<EOF
    ....
    EOF

In this case, "perl" is the command.
If such command is not specified, ctags crashes.
This change fixes this.

'make noise' the input causing the crash.


masatake:
@b4n, thank you. I have still not find time to review your patch. I'm very sorry.

PR #1665, opened 2018-01-20
CXX: Ignore scopes when the current language is not C++

### LINKED ISSUE #1664: C: "::" in struct definition makes ctags crash
```
[jet@localhost]~/var/ctags% cat /tmp/input7.c    
struct a::b{ 
[jet@localhost]~/var/ctags% ./ctags /tmp/input7.c
zsh: segmentation fault (core dumped)  ./ctags /tmp/input7.c
```

pragmaware:
Scopes are supported only by C++. In other languages we silently
skip them.

Fixes #1664.

### DIFF
--- a/Units/parser-c.r/broken-input-cxx-operator.d/input.c
+++ b/Units/parser-c.r/broken-input-cxx-operator.d/input.c
@@ -1,0 +1,1 @@
+struct a::b{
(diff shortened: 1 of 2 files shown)


masatake:
I made a pull request adding the original test input. Could you merge it?

masatake:
LGTM. Thank you.

PR #1667, opened 2018-01-21
CXX: Improve friend declarations

pragmaware:
C++ friend declarations are messy. Sometimes they act as a forward declaration and sometimes they do not. Their scope is not easy to find out and certainly is not the one we were emitting until now. It is better to ignore them.

There is one special exception: the friend function definition form for which the scope is clear (it's the enclosing namespace). We can handle that.


pragmaware:
I've been testing this since a week. Merging so I can do some work on top of it.

PR #1669, opened 2018-01-26
docs: note that jansson is required for interactive mode

vrischmann:
I just spent around 10 minutes wondering why `ctags --_interactive` wouldn't work and figured out that jansson is required only by looking at the source code.

Hopefully this change will make it a little clearer if someone is confused as I was.


masatake:
Thank you!

PR #1670, opened 2018-01-28
CXX: Improve qt header parsing

pragmaware:
Cleanup some corner cases encountered while parsing the Qt headers.


masatake:
LGTM.

PR #1673, opened 2018-01-31
CXX: Fix #1671

### LINKED ISSUE #1671: ctags doesn't handle some C++ templates correctly
I have this small template in my code:

```
template <long x, typename enabled=void>
struct tabs { const static long value = x; };

// specialize tabs
template <long x>
struct tabs<x,typename enable_if_c<(x<0)>::type> { const static long value = -x; };
```

This is legal C++, however, when ctags encounters this it stops parsing. All code in the source file that appears after the `tabs` declaration is invisible to ctags.  For instance, if it were followed by 

```
void my_function();
```

ctags would not output any mention of `my_function` in the resulting tags file.  The general problem is the `<(x<0)>` in the specialization.  In particular, if you take the inner `<` out ctags works fine.  This kind of code pattern is pretty important because any time you are doing template metaprogramming with integers it's not super unusual to want to compare two integers that happen to be template arguments and have the result of that drive a template specialization. Any idea how easy this is to fix?

pragmaware:
Better handling for class/struct/union template specialization nastiness.
Fixes #1671


davisking:
Sweet, that seems to fix the issue.

PR #1683, opened 2018-02-11
readtags: Add support for escape sequences in tag names

masatake:
(Just squashing the original three changes.)


masatake:
I think I took a mistake during applying the patches.

masatake:
I withdraw this.

PR #1697, opened 2018-02-25
fix undefined behavior in calls to <ctype.h> functions

rillig:
These functions must be called with an "unsigned char" argument,
otherwise the behavior is undefined.

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -460,7 +460,7 @@ static bool readDirective (int c, char *const name, unsigned int maxLength)
 		if (i > 0)
 		{
 			c = cppGetcFromUngetBufferOrFile ();
-			if (c == EOF  ||  ! isalpha (c))
+			if (c == EOF  ||  ! is_alpha (c))
 			{
 				cppUngetc (c);
 				break;
@@ -1070,9 +1070,9 @@ static int skipToEndOfChar (void)
 		}
 		else if (Cpp.hasSingleQuoteLiteralNumbers)
 		{
-			if (count == 1  &&  strchr ("DHOB", toupper (c)) != NULL)
+			if (count == 1  &&  strchr ("DHOB", to_upper (c)) != NULL)
 				veraBase = c;
-			else if (veraBase != '\0'  &&  ! isalnum (c))
+			else if (veraBase != '\0'  &&  ! is_alnum (c))
 			{
 				cppUngetc (c);
 				break;
@@ -1522,7 +1522,7 @@ static void saveMacro(const char * macro)
 		return;
 	}
 
-	if(!(isalpha(*c) || (*c == '_')))
+	if(!(is_alpha(*c) || (*c == '_')))
 	{
 		CXX_DEBUG_LEAVE_TEXT("Macro does not start with an alphanumeric character");
 		return; // must be a sequence of letters and digits
@@ -1530,7 +1530,7 @@ static void saveMacro(const char * macro)
 
 	const char * identifierBegin = c;
 
-	while(*c && (isalnum(*c) || (*c == '_')))
+	while(*c && (is_alnum(*c) || (*c == '_')))
 		c++;
 
 	const char * identifierEnd = c;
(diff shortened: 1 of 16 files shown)


masatake:
The result may be the same but int is better than unsigned char, isn't it?

rillig:
The conversion from `unsigned char` to `int` happens automatically and without changing the value. Up to 2015, [NetBSD recommended](https://nxr.netbsd.org/history/src/lib/libc/gen/ctype.3) first casting to `unsigned char` and then to `int`, but later that recommendation was withdrawn. Therefore I consider it enough to just cast to `unsigned char`.

codebrainz:
It's not very nice to put generically named, unprefixed macros into headers, IMO.

Since (AFAICT) these functions are not what the parsers really want anyway, as they depend on locale, what about replacing them with actual (inline) functions which do the desired character classification themselves? For example for ASCII `islower`:

```c
static inline bool ctags_ascii_islower(int c) {
  return (c >= 'a' && c <= 'z');
}
```

rillig:
Sure, that looks nice.

I just wanted to get rid of the GCC compiler warning `array subscript has type 'char'". And this was the minimum change necessary.

rillig:
> So the change proposed here is plain wrong for EOF.

The change here only affects code that iterates over strings, and in these cases `str[i]` can never be `EOF`.

There are actually two possible cases where the `ctype.h` functions are called:

1. With the result of `fgetc` or similar functions. Then the value is either EOF or in the range [0, UCHAR_MAX]. Everything is fine here.

2. With a character from a string. In this case, the character is in the range [CHAR_MIN, CHAR_MAX], and half of that range _may be_ negative. For this range, the cast is necessary and correct. Passing the character itself is valid from the point of the function call and type conversions, but the requirement inside the function call says that only very specific values are allowed.

All in all, my suggested commit only prevents the undefined behavior. The really proper fix would be to add Unicode support for all parsers, which would allow extended characters in identifiers. But that's another story.

parsers/cpreprocessor.c
```diff
@@ -1070,9 +1070,9 @@ static int skipToEndOfChar (void)
 		}
 		else if (Cpp.hasSingleQuoteLiteralNumbers)
 		{
-			if (count == 1  &&  strchr ("DHOB", toupper (c)) != NULL)
+			if (count == 1  &&  strchr ("DHOB", to_upper (c)) != NULL)
```
b4n: here `c` comes from the file, not from a string

b4n:
@rillig oh, indeed, good point, sorry. Yeah in this case it makes sense indeed, and is needed on most systems (as `char` is most generally signed).

This said, @codebrainz is right: it's unlikely any of these functions do what is expected, as most of the time the parser will expect ASCII only. Which will be OK most of the time, but not always. We probably should stop using those, and add helper for the common space/digit/alphanumeric checks.
Also, the macros introduced here are not very clear on when they should be used, and have an easy to overlook name.

> The really proper fix would be to add Unicode support for all parsers, which would allow extended characters in identifiers. But that's another story.

Yes that would be great.  But it's subtle, as it means we pretty much gotta get the right encoding or else we're in trouble… and it would not solve more than if we were using int/wchar_t/etc. right now with plain 8 bits values in it.

rillig:
>here `c` comes from the file, not from a string

Oops, that's a mistake in my side. I had edited all files with computer warnings and then replaced _all_ instances of the calls. I should have fixed only those which were warned about.

krytarowski:
This needs to be casted to (unsigned char) otherwise we are fiddling with Undefined Behavior and compiler warnings. EOF is checked with other functions like `feof(3)`. This is an ugly part of C, not NetBSD.

This behavior is enforced by the C standard, POSIX and the NetBSD implementation.

masatake:
People, how do you think the idea @codebrainz wrote? I like it.
Not using `_` in the common code between parsers and main part.
So the name should be ctagsIsLower or isLower.

When we introduce more explicit interface beween the parsers and the main part, we should use `ctags` 
as prefix for the components of the interface. However, we will not do it before releasing 6.0. So `isLower` is acceptable.

Do you think adding `ascii` in their names is better?

rillig:
I'm no longer interested in fixing the possible undefined behavior.

PR #1701, opened 2018-03-06
Fix an off-by-one error in the tag stack traversing

LemonBoy:
Since the whole `stack` array isn't completely cleared by the
`initStack` routine ctags would sometimes segfault after reading stale
or invalid data off the topmost item.

Writing a test is slightly complicated by the fact you have to "poison" the stack first.


masatake:
Thank you. LGTM.

masatake:
Thank you.

PR #1703, opened 2018-03-07
docs: don't put "the" before "both"

masatake:
It seems that it is wrong English.


masatake:
@codebrainz, thank you very much.

PR #1707, opened 2018-03-09
 [SELF INCOMPATIBLE] main: reserver '{' in kind description as a meta character

masatake:
These are some small fixes.

The exception is the last one.

I'm thinking about adding flags to `--kinddef-<LANG>` option. For the case, I would like to reserver the char `{`.


masatake:
I have to do the same to `--fielddef-<LANG>` and `--extradef-<LANG>`.

PR #1708, opened 2018-03-10
[SELF INCOMPATIBLE]: rename the feature "multibyte" to "iconv"

masatake:
Using the name "multibyte" is exaggeration.


b4n:
I don't really see why the renaming makes sense, for me it's pretty much the same but "iconv" is technical while "multibyte" is a little less.
But I don't really care about the name here either, so why not.

masatake:
The name "iconv" just tells the library is linked to ctags. It doesn't tell how it is used.

PR #1709, opened 2018-03-11
Refine the concept of role

masatake:
Though the change is large, this PR does only two things essentially:
* Rename "role" field to "roles",
* Assign "def" to "roles" field if a tag is definition tag.


masatake:
A test case for multipe roles is needed.

PR #1712, opened 2018-03-16
Defining role from cmdline and capturing reference tags with regex parser

masatake:
This change set introduces three things:

* --_roledef-<LANG>=<kindletter>.<rolename>,<roleDescription> option,
* _role flag for --regex-<LANG> option, and
* _refonly flag for --kinddef-<LANG> option

They allow to a user to define a role and capture reference tags.
The web document is written in progress. 
The usage is illustrated in "Units: add a case for testing the combination of --_roledef-<LANG> op…".

With this change set, we can improve elm parser (See #1260).
We can also improve puppet parser as I wrote in https://github.com/rodjek/vim-puppet/pull/84#issuecomment-370303183 .

I will not explain them in the man page till releasing 6.0.

### DIFF
--- a/Tmain/broken-extradef.d/run.sh
+++ b/Tmain/broken-extradef.d/run.sh
@@ -1,0 +1,60 @@
+# Copyright: 2018 Masatake YAMATO
+# License: GPL-2
+
+CTAGS="$1 --quiet --options=NONE"
+
+{
+{
+echo '# echo unknown lang'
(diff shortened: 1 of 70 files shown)


masatake:
TODO:
* update optlib2c to support `--_roledef-<LANG>`,
* capture imported moudle in elm,
* capture included class in puppetManifest.

masatake:
I will update puppetManifest.ctags in next patch set.

PR #1727, opened 2018-04-03
Markdown parser

masatake:
See  #1566.


b4n:
I'm not sure it's actually better, but did you consider Geany's markdown parser?

masatake:
Yes, I considered.
I should talk to you first. I put a comment about what I thought in #1566.
Geany's one doesn't write the version of GPL explicitly. I wondred what I should do.

In addition, it doesn't track the level of sections. Geany's markdown parser has only ond kind.
It is different from rst.c. rst.c supports chapter, section, subsection, and subsubsection.
I think doing the same for the other documentation language as rst.c (rest.c in ctags).

PR #1730, opened 2018-04-06
Add installation instructions for Debian

xuhdev:


masatake:
At the section "The latest build and package" of README, we wrote about macos and windows.
About macos, as far as I understand correctly, the building is done by client sides.
The timing for updating is controlled by the users of the client sides.
About windows, as far as I understand correctly, the building is done by server side. Updating is done enough frequently.

About Debian, I'm not sure how frequently the package is updated. It is done by the package maintainer.
(This is a good thing. So I don't have to think about the detail of the distribution.)

If a bug is reported, we (people at universal-ctags) can take an action for the reporter: fixing a code and updating a binary package (or updating a script generating a binary package).
About Debian package, I wonder we can do the same. I can fix a bug, but we cannot ship a new package directly.

If we write about the Debian package on our web page, I think we should take a responsibility to it (as we do for Windows and MacOSX).
However, I think we should not take such responsibility. So I don't want to merge your change.

masatake:
I'm still introducing self-incompatible changes frequetly. So I would like to use the latest version.

xuhdev:
@masatake I understand your concerns. In this case, can you maintain a package at [OpenSUSE build service](https://build.opensuse.org/)? You only need to build the package once and it automatically makes it available across many Linux distributions. In addition, I believe it is also possible to automatically build nightly versions.

masatake:
Sorry, but I don't want to maintain more things.

You can maintain the package at the service as part of the member of Universal-ctags organization as @k-takata doing for Windows, and as @KazuakiM doning for MacOSX.

xuhdev:
@masatake I see. I can try to build it, and let you know when it's ready.

PR #1732, opened 2018-04-09
Fix Ruby parsing for block assigns

tambeta:
Assigning blocks' return values to variables is a common idiom in Ruby. Fix the parser to take this into account.

I'm not sure if matching against regexen would be acceptable in ctags, but this would certainly make things easier. The proposed implementation doesn't use regexps.

I originally submitted this fix to Geany, which includes ctags as part of its source tree - Geany devs recommended posting it to upstream first. The issue is [explained over there](https://github.com/geany/geany/issues/1744).


b4n:
Could you provide test cases for this?

tambeta:
OK, for now I added a test case for determining if the correct tags are extracted if blocks are assigned to variables.

masatake:
@tambeta, Thank you.

Could you paste here an example input that uses `||=` and/or  `+=` if you don't have time?
I will convert it to .b test case.  
http://docs.ctags.io/en/latest/units.html#gathering-test-cases-for-known-bugs

Units/parser-ruby.r/ruby-block-assign.d/expected.tags
```diff
@@ -0,0 +1,6 @@
+Bar	input.rb	/^c = class Bar$/;"	c
+Foo	input.rb	/^m = module Foo$/;"	m
+method_a	input.rb	/^  x = def method_a$/;"	f	class:Bar
+method_b	input.rb	/^  def method_b$/;"	f	class:Bar
+method_c	input.rb	/^  def method_c$/;"	f	class:Bar
+method_d	input.rb	/^  def method_d$/;"	f	class:Bar
```
b4n: Shouldn't `m`, `c` and `x` be reported as something (variables I guess)?

I'm not sure where this syntax is used, but apparently it does something (store the result of the last statement? not sure from experimenting…), so I guess they should ideally be reported, shouldn't they?

tambeta: They could be reported as variables, but I believe this hasn't been the case thus far. Anyway, it isn't in the scope of this PR, as it just aims to fix the annoying bug where class members are reported as free-floating methods if block assignment is used.

tambeta:
@masatake, I took the time to read about the protocol for adding known bug tests and the latest commit adds a failing test where various "shortcut" assignment operators are used (`+=` etc).

tambeta:
Giving it some thought, it wasn't a big job to extend parsing to allow "shortcut" assignment operators. The latest commit introduces this.

parsers/ruby.c
```diff
@@ -117,31 +112,104 @@ static bool canMatch (const unsigned char** s, const char* literal,
...
 	return c == 0 || isspace (c);
 }
 
+/*
+ * Advance 's' while the passed predicate is true. Returns true if
+ * advanced by at least one position.
+ */
+static bool advanceWhile (const unsigned char** s, bool (*predicate) (int))
+{
+	const int s_length = strlen ((const char *)*s);
```
masatake: `size_t` is better.

parsers/ruby.c
```diff
@@ -117,31 +112,104 @@ static bool canMatch (const unsigned char** s, const char* literal,
...
+	unsigned char this_char;
+	int i = 0;
+
+	for (i = 0; i < s_length; ++i)
```
masatake: This function scans the input twice: once by strlen and once by this `for`.
How about replacing them with following `while`?

```
while (**s != `\0`)
```

parsers/ruby.c
```diff
@@ -167,7 +235,7 @@ static bool parseRubyOperator (vString* name, const unsigned char** cp)
 	{
 	    if (canMatch (cp, RUBY_OPERATORS[i], notOperatorChar))
 	    {
-	        vStringCatS (name, RUBY_OPERATORS[i]);
+			vStringCatS (name, RUBY_OPERATORS[i]);
```
masatake: Could you this line as the orignal?

masatake:
I made some minor comments but it looks good to me.

tambeta:
Noted and implemented.

masatake:
Thank you.

masatake:
Could you look at #1734. To make the history of commits simple, I rearranged your changes.

tambeta:
OK, let's take the discussion over there.

masatake:
I merged your work in #1734. Thank you.

tambeta:
Great, thanks.

PR #1733, opened 2018-04-10
SQL: support "create table FOO as select" format 

### LINKED ISSUE #1658: Support for "CREATE TABLE AS" SQL syntax
Running ctags on the following file will lead to correct tags for the table `foo`, but table `bar` is missed completely. I assume this is caused by the `CREATE TABLE AS` syntax.

```sql
CREATE TABLE foo(
    col1 text
);

CREATE TABLE bar AS
SELECT 1 AS col1;
```

Ideally, both the table and the column would be picked up by ctags, but at least the table name should be included in the tags file.

masatake:
Close #1658.


Units/parser-sql.r/sql-create-table-as.d/input.sql
```diff
@@ -0,0 +1,9 @@
+--
+-- Taken from #1658 submitted by @karlb
+--
+CREATE TABLE foo(
+    col1 text
+);
+
+CREATE TABLE bar AS
+SELECT 1 AS col2;
```
b4n: there probably should be a note somewhere that this is *not* the expected output, but just here until the parser supports the other tests.  The thing is that it can be confusing when working on the parser to have a test starting to fail, although you actually just *fixed* (or improved) something.

masatake: I see. I will add README (or extend the parser to capture `col2`).

karlb:
Thanks for working on this! Here are some more interesting test cases:

```sql
-- has columns foo, bar
CREATE TABLE table_with_subselect AS
SELECT foo,
    2 AS bar
FROM (
        SELECT 1 AS foo
    ) subsel;

-- has columns foo, bar, b_bar
CREATE TABLE table_with_join AS
SELECT foo,
    a.bar,
    b.bar AS b_bar
FROM table_with_subselect a
    JOIN table_with_subselect b USING (foo);

-- has columns col1, col3, col4
CREATE TABLE table_with_weird_expressions AS
SELECT 'string AS not_a_colname' AS col1,
    'unknown colname; depends on db',
    'string that goes over '
    'two lines' AS col3,
    1 + (SELECT 1 AS inner_colname) AS col4
FROM (SELECT 1 AS foo) AS not_a_column;

-- has columns foo, bar
CREATE TABLE view_with_subselect AS
SELECT foo,
    2 AS bar
FROM (
        SELECT 1 AS foo
    ) subsel;
```

I've also included on view, because I think the same parsing should be applied in both cases. If that's out of the scope of this change, just ignore it.

masatake:
Thank you.

masatake:
My new patch doesn't work expectedly for the input:
```
CREATE TABLE tmp_table0 AS
SELECT tmp5 + tmp6 FROM tmp_table1;
```

In this example, capturing just tmp_table0 is expected in my understanding.
However, tmp6 is captured, too. I have to learn the syntax of column expression.

karlb:
> Are these o.k.?

Yes, looks great!

I originally intended to write `CREATE VIEW view_with_subselect` instead of `CREATE TABLE view_with_subselect`. Maybe you want to change that in order to check that the same column recognition works for views, too. If you don't the table name is misleading.

masatake:
@karlb, I will  not deal columns in a view in this pull request.

So I rewrote your test case a bit:
```SQL
diff --git a/Units/parser-sql.r/sql-create-table-as.d/input.sql b/Units/parser-sql.r/sql-create-table-as.d/input.sql
index 07fe8183..0cf62b24 100644
--- a/Units/parser-sql.r/sql-create-table-as.d/input.sql
+++ b/Units/parser-sql.r/sql-create-table-as.d/input.sql
@@ -39,7 +39,7 @@ SELECT 'string AS not_a_colname' AS col1,
 FROM (SELECT 1 AS foo) AS not_a_column;
 
 -- has columns foo, bar
-CREATE TABLE view_with_subselect AS
+CREATE TABLE table_with_subselect2 AS
 SELECT foo,
     2 AS bar
 FROM (

```

karlb:
I've been using these changes for the last weeks and things work great for me so far. Thanks!

PR #1734, opened 2018-04-11
Ruby: assign blocks (originally submitted by @tambeta)

masatake:
Simplyfy the commits in #1732.


tambeta:
For some reason, the directory `Units/parser-ruby.r/ruby-block-assign-shortcut.d/` has appeared. During development, I used `Units/parser-ruby.r/ruby-block-assign-shortcut.b/` for known failing tests. A separate test case for shortcut assignment operators is no longer needed, since I implemented the extra logic and added to the existing test case. So this directory should be removed altogether.

masatake:
@tambeta, thank you. It seems that the last commit should be removed.
I removed 6f68d0709dafdcfa19b0da7ee3c27edcff7ae0f5 from this pull request.

PR #1747, opened 2018-04-30
C++: handle function-try-block

### LINKED ISSUE #1746: C++ failed to parse try-catch body block
void foo() try {
	  }catch(...){
	  }

	  void bar() {
	  }

It finds `bar`, but not `foo`

masatake:
Close #1746.

The original bug report opened by @ChoppinBlockParty:

    void foo() try {
    }catch(...){
    }

"foo" is not captured well.

This commit introduces code ignoring "try" token if the last token is
a pair of "(" and ")".

NOTE: This change doesn't handle well the catch block.  For an
example, a local variable inside the catch block should have "foo" as
"scope". The current implementation cannot recognize it.


masatake:
@pragmaware, thank you.

PR #1749, opened 2018-05-01
CXX: Improve parsing of function-try-blocks and try/catch in general

pragmaware:
Improved solution for #1746.

-  catch() is part of the function-try-block:
-- the variables declared inside now have proper scope
-- the function ends after the last catch()
- variables declared in the catch() parenthesis are extracted
- function-try-blocks are reported as properties

### DIFF
--- a/Units/parser-cxx.r/function_try_block.d/args.ctags
+++ b/Units/parser-cxx.r/function_try_block.d/args.ctags
@@ -1,4 +1,4 @@
---c++-kinds=+pflz
+--kinds-c++=+pflz
 --fields=+SsKe
 --fields-c++=+{captures}+{properties}
 --sort=no
--- a/Units/parser-cxx.r/function_try_block.d/args.ctags
+++ b/Units/parser-cxx.r/function_try_block.d/args.ctags
@@ -1,1 +1,4 @@
+--c++-kinds=+pflz
+--fields=+SsKe
+--fields-c++=+{captures}+{properties}
 --sort=no
--- a/parsers/cxx/cxx_parser.c
+++ b/parsers/cxx/cxx_parser.c
@@ -404,12 +404,7 @@ bool cxxParserParseToEndOfQualifedName(void)
 	return true;
 }
 
-
-//
-// Attach the current position of input file as "end" field of
-// the specified tag in the cork queue
-//
-void cxxParserMarkEndLineForTagInCorkQueue(int iCorkQueueIndex)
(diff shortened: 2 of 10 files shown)


parsers/cxx/cxx_parser.c
```diff
@@ -1617,76 +1622,114 @@ bool cxxParserParseAccessSpecifier(void)
...
-			cxxTokenChainDestroyFirst(pChain);
-			// Fake the final semicolon
-			CXXToken * t = cxxTokenChainLast(pChain);
-			t->eType = CXXTokenTypeSemicolon;
-			vStringClear(t->pszWord);
-			vStringPut(t->pszWord,';');
-
-			// and extract variable declarations if possible
-			cxxParserExtractVariableDeclarations(pChain,0);
+					);
```
masatake: Visual Studio makes a warning on this line (https://ci.appveyor.com/project/universalctags/ctags/build/1.0.763/job/wim07v3kr02a88r6). How about using trigraph like:
```
bOkToExtractVariables = cxxTokenChainNextTokenOfType(
						pAndOrStar,
 						CXXTokenTypeAssignment | CXXTokenTypeSemicolon
					)? true: false;
```

or ...
```
} else if ( cxxTokenChainNextTokenOfType(
					pAndOrStar,
 						CXXTokenTypeAssignment | CXXTokenTypeSemicolon)) {
            bOkToExtractVariables = true;
}

File: parsers/cxx/cxx_tag.c [hunk 2]
```diff
@@ -286,6 +286,9 @@ vString * cxxTagSetProperties(unsigned int uProperties)
 		vStringCatS(pszProperties,_szProperty); \
 	} while(0)
 
+	// FIXME: The property names might be too verbose. Maybe it could be a good idea
+	//        to a switch to shorten them to 1-2 letter codes?
+
```
masatake: I think basically using long names is not bad.
Making shortere can be done in a post processor.

Units/parser-cxx.r/function_try_block.d/args.ctags
```diff
@@ -1 +1,4 @@
+--c++-kinds=+pflz
```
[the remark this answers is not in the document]
pragmaware: done

parsers/cxx/cxx_parser.c
pragmaware: done

parsers/cxx/cxx_tag.c
pragmaware: OK

masatake:
Thank you.

PR #1756, opened 2018-05-21
Booking bugs recently reported

masatake:
In addition, remove test cases already reviewed.


masatake:
It seems that a broken tag entry in a tags file triggers a bug of readtags.c.
I have to open a new issue eventually. Here is a temporary memorandum.

tags file:
```
greet	input.java	/^		public void greet (String word) {$;"	m	line:9	method:input.hello.anonXXX
```
The last slash at the end of pattern doesn't exist.

How to run:
```
[yamato@master]~/var/ctags-github% ./readtags  -t tags  - greet
zsh: segmentation fault (core dumped)  ./readtags -t tags - greet
```
How to fix the bug:

```
diff --git a/read/readtags.c b/read/readtags.c
index 81ce43ce..8c43b0ea 100644
--- a/read/readtags.c
+++ b/read/readtags.c
@@ -455,6 +452,7 @@ static void parseTagLine (tagFile *file, tagEntry *const entry)
 				if (p == NULL)
 				{
 					/* invalid pattern */
+					return;
 				}
 				else
 					++p;
@@ -470,6 +468,7 @@ static void parseTagLine (tagFile *file, tagEntry *const entry)
 			else
 			{
 				/* invalid pattern */
+				return;
 			}
 			fieldsPresent = (strncmp (p, ";\"", 2) == 0);
 			*p = '\0';
```

More study is needed.

masatake:
The newly found bug is fixed in https://github.com/universal-ctags/ctags/pull/1757.

PR #1759, opened 2018-05-26
CXX: Fix #1750 (without breaking other unit tests)

### LINKED ISSUE #1750: c++ parsing error
(If you will report an issue about the result of parsing,
fill the following items. If you get the input file somewhere,
write the URL or something.)

*****

The name of the parser:
c++
ctags -o tags bad.h
The content of input file:
```c++
#define int0 int
template <int,
         class _Comp0=less<int>,
         class _Comp1=less<pair<int, int> > >
class Test : public set<int> {
  typedef int xxx;
};
#define int1 int
```

```
int0	bad.h	/^#define int0 /;"	d
```
The tags output you expect:
```
Test	bad.h	/^class Test : public set<int> {$/;"	c
int0	bad.h	1;"	d
int1	bad.h	8;"	d
xxx	bad.h	/^  typedef int xxx;$/;"	t	class:Test
```
 This is what Exuberant Ctags generates

pragmaware:
Rather complicated, but fixes #1750 and does not break existing unit tests.
One can still construct tests that fool the parser. A definitive fix is not feasible without global knowledge
about the program being compiled.


masatake:
Could you the two commits into one?
`git rebase -i master` may help you.

Could you rename the .b test case to .d? "make units" says at the end of output:
```
...
#FIXED:                                 3
	parser-cxx.r/template-nested-triangle-brackets
...
```

pragmaware:
Done. Check it out.

pragmaware:
Do you still need to review this or we can merge?

masatake:
I'm sorry to be late. I have had a troulbe in my health.

I have two requests you to change the commits.

1. Could you remove the lines only whitespaces?
git diff may tell where the lines are.

whitespace-only-lines

2. Could you add a ".b" test case that ctags cannot handle well even with your new code?

ctags can handle your example well:
```
template<char i, bool B = i < 10> void f1(void) { }
```
I would like you to add paranoid but compilable ones because I would like to try my algorithm.

masatake:
LGTM. Thank you very much.

PR #1766, opened 2018-06-18
Ruby block assign: handle variables with sigils

tambeta:
When assigning blocks in Ruby, handle assignment to variables prefixed by `@`, `@@` or `$`.

This is an improvement to #1734.


masatake:
Thank you.

PR #1769, opened 2018-06-23
main,Cxx: allow to enable/disable tracing feature dynamically

masatake:
To enable the tracing feature in Cxx parser required three steps
at build-time.

1. Run ./configure --enable-debugging
2. Define CXX_DEBUGGING_ENABLED in parsers/cxx/cxx_debug.h
3. make clean && make

To disable the tracing, undoing 2, and 3 were needed.
It is difficult to use tracing-enabled ctags in daily.

This change allows enabling/disabling tracing feature at run-time.
To enable the tracing feature in Cxx parser required 2 steps at build-time.

1. Run ./configure --enable-debugging
2. make clean && make

Then you can use --_trace=<LANG>[,<LANG>]* option to enable the tracing feature
in a language (if the language uses the trace API).

    [jet@localhost]~/var/ctags% ./ctags --list-features | grep debug
    debug             TO BE WRITTEN
    [jet@localhost]~/var/ctags% ./ctags main/main.c                 
    [jet@localhost]~/var/ctags% ./ctags --_trace=C main/main.c 2>&1 | head
    [cxxCParserInitialize][at 0] Parser initialize for language C
    [>> cxxCParserMain][at 0] 
	[cxxKeywordEnablePublicProtectedPrivate][at 0] Disabling public/protected/private keywords
	[cppInitCommon][at 0] cppInit: brace format is 0
	[>> cxxParserParseBlockInternal][at 0] 
	    [directiveHash][at 27] Found #elif or #else: ignore is 0
	    [directiveHash][at 39] Found #elif or #else: ignore is 0
	    [cxxParserParseBlockInternal][at 92] Token 'static' of type 0x04
	    [cxxParserParseBlockInternal][at 92] Token 'struct' of type 0x04
	    [cxxKeywordEnablePublicProtectedPrivate][at 92] Enabling public/protected/private keywords


masatake:
@pragmaware, could you look at this one?
This change may have an impact on your workflow.
I introduced `--_trace=<LANG>[,<LANG>]*` option.

pragmaware:
The idea looks good to me.
I mean as long as a compilation with debug disabled keeps tracing disabled so there is no performance impact on release builds.

masatake:
@pragmaware, how do you think using `__func__` instead of `__PRETTY_FUNCTION__`?
`__func__` is part of c99.

pragmaware:
Ow... it would probably be fine too. Anything that prints the current function name is OK.

masatake:
I should write about `--_trace` to the web documents.

Issue #1772, opened 2018-06-26
C: --_xformat ignoring the struct key word while extracting function information the header


sourabhjains:
I am extracting function information in the following format.
**function_name:return_type:signature**

I have used following command to do that
**ctags --language-force=C -x --_xformat="%N:%{typeref}:%{signature}"  --kinds-C=p foo.h**

When I formatting the ouput using --_xformat with typeref tag. typeref doesn't consider struct keyword while formatting the ouput.

for example: foo.h
#include <stdio.h>
int first (int a, int b);
struct mystructure second (int a);
long long third (long a);

The ouput of above command will be:
first:int:(int a,int b)
second:mystructure:(int a)
third:long long:(long a)

Here, in second function --_xformat misssed the struct keyword.

masatake:
Thank you for reporting.
I will work on this item.

masatake:
See #1773.
With the change proposed in #1773, the command line makes following output:
```
first:typename:int:(int a,int b)
second:struct:mystructure:(int a)
third:typename:long long:(long a)
```

This output is not enough for your requirement. However, I don't think I should make ctags specialized to your requirement more. sed may help you as usual.

It is possible to introduce typerefPrefix and typerefName fields. See the commit log of the change if you are interested in introducing new fields.

masatake:
I merged #1773. The change may solve your issue partially. As I wrote you need a post processor like sed for your purpose.

Issue #1510, opened 2017-07-19
Getting a full qualified symbol name for typeref:typename?


FlKo:
For [typeref:typename:xxx](https://github.com/universal-ctags/ctags/blob/master/docs/parser-cxx.rst#typeref):

Is it possible to get a **full-qualified** symbol name here (e.g. **typeref:typename:System::TDateTime** instead of **typeref:typename:TDateTime** ?

masatake:
Thank you for contacting. Basically what you want is not available. However, I would like to understand the requirement for future development.

I don't know well abot C++ and I am bad in English. So an example is very important. Wll understandable short example drives me as you want:-)

Input:
```C++
namespace System {
  // http://docs.embarcadero.com/products/rad_studio/radstudio2007/RS2007_helpupdates/HUpdate4/JA/html/delphivclwin32/System__TDateTime.html
  class TDateTime : public TDateTimeBase {
  };
  TDateTime foo (void);
}

System::TDateTime bar (void);
```

The output of the current implementation:
```
% u-ctags --sort=no --kinds-C++=+p --fields='+{typeref}' -o - /tmp/sysdate.h
System	/tmp/sysdate.h	/^namespace System {$/;"	n
TDateTime	/tmp/sysdate.h	/^  class TDateTime : public TDateTimeBase {$/;"	c	namespace:System
bar	/tmp/sysdate.h	/^System::TDateTime bar (void);$/;"	p	typeref:typename:System::TDateTime
foo	/tmp/sysdate.h	/^  TDateTime foo (void);$/;"	p	namespace:System	typeref:typename:TDateTime
```

I expect what you want is:
```
% u-ctags --sort=no --kinds-C++=+p --fields='+{typeref}' -o - /tmp/sysdate.h
System	/tmp/sysdate.h	/^namespace System {$/;"	n
TDateTime	/tmp/sysdate.h	/^  class TDateTime : public TDateTimeBase {$/;"	c	namespace:System
bar	/tmp/sysdate.h	/^System::TDateTime bar (void);$/;"	p	typeref:typename:System::TDateTime
foo	/tmp/sysdate.h	/^  TDateTime foo (void);$/;"	p	namespace:System	typeref:typename:System::TDateTime
```

Do I understand what you want correctly?

FlKo:
[...]

Issue #1082, opened 2016-08-20
Incompatible behavior: "typeref" field of C


htfy96:
### Source`test.c`

``` c
static int f(int x) { return 42; }
static bool wakeup_spi = false;
```
### Universal Ctags result

``` tags
f   test.c  /^static int f(int x) { return 42; }$/;"    f   typeref:typename:int    file:
wakeup_spi  /tmp/test-spi.c /^  static bool wakeup_spi = false;$/;" v   typeref:typename:bool   file:
```
#### Version

```
Universal Ctags 0.0.0(c984455), Copyright (C) 2015 Universal Ctags Team
```
### Exuberant Ctags result

``` tags
f   test.c  /^static int f(int x) { return 42; }$/;"    f   file:
wakeup_spi  /tmp/test-spi.c /^  static bool wakeup_spi = false;$/;" v   file:
```
## Expected behavior

According to man page of Universal Ctags and Exuberant Ctags:

> t   Type and name of a variable or typedef as "typeref:" field [enabled]

It seems that both the two forks don't conform to man page:
-  Exuberant Ctags only prints this field for definition of variable of struct type
-  Universal Ctags even prints it for functions

masatake:
Related to https://github.com/universal-ctags/ctags/issues/862 .
Maybe we should decide the meaning of fields and update the man page.

PR #1773, opened 2018-06-26
[SELF-INCOMPATIBLE] xformat-writer,json-writer: make typeref: field in xref and json output formats as same as that in tags

masatake:
typeref: field has sub-fields "typeref prefix" and "typeref name" in tags format.
":" combines the two sub-fields and makes the field value.

e.g.

	typename:int
	struct:S

In xref and json output formats, only "typeref name"
sub field was printed when printing typeref: field.

This change fixes this inconsistent.
This is self-incompatible change.

If you want, "typerefPrefix" and "typerefName" fields can be introduced.
Open an issue in that case.

This change is needed to solve #1772.


masatake:
As far as reading "Typeref" section of docs/parser-cxx.rst, I should provide typerefPrefix and typerefName fields.

I have to read again the discussion in  #882, #1510, #1082, and #862.

masatake:
#1510 is out of this scope.

masatake:
I will solve #882 and #862 in another pull request.

Issue #1777, opened 2018-07-01
reStructuredText: handle targets


blueyed:
It would be nice to parse targets in rst files, e.g.:

```
.. target-name:
```

Spec: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#hyperlink-targets

I guess this would not be hard to implement, but also not trivial given the current logic therein: https://github.com/universal-ctags/ctags/blob/35c89c8fbb4c7dd6a008acc31207a0a2a61868e8/parsers/rst.c#L223.

I think there should be a new kind ("T", since "t" is taken already), enabled by default?

masatake:
See #1779.
Very important suggestion for languages for documentation.
A good size of request for a change:-)

Thank you.

masatake:
Implemented in #1779.

PR #1779, opened 2018-07-02
ReStructuredText: add target kind

masatake:
A target appears as following:

  .. _something-target:

or

  .. _ `something target`:

TODO: Do the same on the other languages for documentation
like TeX.

Suggested by @blueyed in #1777.


masatake:
@b4n, you may have interest on this item because this can be ported to geany.

blueyed:
I think it should be more strict maybe:

> Hyperlink targets may be named or anonymous. Named hyperlink targets consist of an explicit markup start (".. "), an underscore, the reference name (no trailing underscore), a colon, whitespace, and a link block:

Sphinx only generates a target for the last one, whereas this PR creates tags for all of them:

```rst
.. _       	     `section target1`   :
.. _       	     `section target2`:
.. _`section target3`  :
.. _`section target4`:
```

masatake:
Updated.

Escaping with backslash is handled, too. 
```
.. _abc\:efg:
```

In this case, ctags records this as:
```
abc\\:efg
```
in tags file.

blueyed:
Awesome, thanks!

masatake:
@b4n, thank you.

PR #1789, opened 2018-07-09
Misc man page fixes

### LINKED ISSUE #1082: Incompatible behavior: "typeref" field of C

masatake:
Close #1082.


main/options.c
```diff
@@ -248,12 +248,12 @@ static optionDescription LongOptionDescription [] = {
  {1,"      Include extra tag entries for selected information (flags: \"Ffq.\") [F]."},
  {1,"  --extras-<LANG|*>=[+|-]flags"},
  {1,"      Include <LANG> own extra tag entries for selected information"},
- {1,"      (flags: --list-extras=<LANG>)."},
+ {1,"      (flags: inquire the output of --list-extras=<LANG> option)."},
```
codebrainz: If I understand correctly[0], I think it would sound more natural to replace "inquire" with "see" or "refer to". Same with below on L256.

[0]: That it means: to see what you can use for the `flags`, use `--list-extras` and it will print them for you.

masatake: Thank you!

masatake:
Refected the comment from @codebrainz, and updated the commit.

PR #1791, opened 2018-07-11
Add support for Matlab classes

andebjor:
Recognize the definition of a class, and allow for functions not to be
defined in the first column but to be indented by white space.


masatake:
Thank you. BTW, could you help us more to improve matlab parser?
Test cases under Units/review-needed.r were came from Exubertant-ctags.
The cases are not reviewed; pairs of an input file and expected.tags for the input are not verified by ctags developers. 

Could you look at matlab_backtracking.m.t/{input.m,expected.tags} and matlab_test.m.t/{input.m,expected.tags} and give me feedback ?

If you say they are o.k., I will move the cases to Unix/parser-matlab.r, a new directory.

andebjor:
See https://github.com/universal-ctags/ctags/pull/1792

PR #1792, opened 2018-07-11
Clean matlab testcases

andebjor:
Re-work the synthetic test case `matlab_test.m.t` a bit to make it more
realistic and stress the parser a bit more.

In addition, collect all Matlab unit tests in a new category 
`parser-matlab.r`.

I'm not completely sure if the test `Units/noext-tg-matlab.d` should be
included in `parser-matlab.r` or not, as the point (if I assume correct)
is that the language should be guessed in that case. It is included now.


masatake:
I made small requests to revise but the most of all parts are good.

andebjor:
Makes sense, I hope I got it right now.

masatake:
Thank you. I will merge this after appveyor says o.k.

Your two pull requests show you have gift to read broken English:-)
I have a special offer for more hacking the parser.

1. instead of improving matlab.c, you can write matlab.ctags.
    See optlib/*.ctags files, especially optlib/elm.ctags.
    These files are translated into C source file and linked to ctags.
2. implementing scope field
    See man/ctags-optlib.7.rst.in.
    See scope= in "FLAGS FOR ``--regex-<LANG>`` OPTION" section.
    You can fill scope: field wit your matlab parser.
3. You can find more crazy things in http://docs.ctags.io/en/latest/optlib.html?highlight=scope .

I will show some examples next.

masatake:
Here is an example of the more complicated parser.
Ideally "super:" parser-own field should be "inherits:" built-in field. 
I used "super:" to avoid the current limitation of ctags.
If you are interested in developing MatLab.ctags, this example can be a start point.

```
[jet@localhost]~/var/ctags% cat /tmp/input.m 
classdef input < handle
    methods(Static)
        function classfunc3;
	% Silly function without any arguments
	    A = magic(4);
	    R = randn(3,4,5);
	
	    disp('A:');
	    disp(A);
	
	    disp('R:');
	    disp(R);
        end
     end
 end
[jet@localhost]~/var/ctags% cat /tmp/MatlabX.ctags 
--langdef=MatLabX
--map-MatLabX=+.m
--kinddef-MatLabX=c,class,classes
--kinddef-MatLabX=f,func,functions
--kinddef-MatLabX=v,var,variables
--_fielddef-MatLabX=super,super class

--regex-MatLabX=/^[ \t]*classdef[ \t]*([a-zA-Z][a-zA-Z0-9]*)[ \t]*<[ \t]*([a-zA-Z][a-zA-Z0-9]*)/\1/c/{scope=push}{_field=super:\2}
--regex-MatLabX=/^[ \t]*function[ \t]*([a-zA-Z][a-zA-Z0-9]*).*/\1/f/{scope=push}
--regex-MatLabX=/^[ \t]*([a-zA-Z][a-zA-Z0-9]*)[ \t]*=.*/\1/v/{scope=ref}

--regex-MatLabX=/^[ \t]*methods.*//{scope=push}{placeholder}
--regex-MatLabX=/^[ \t]*end//{scope=pop}{placeholder}

# To avoid bugs in ctags itself, these items must be last
--fields-MatLabX=+{super}
--map-MatLab=-.m
[jet@localhost]~/var/ctags%  u-ctags --options=/tmp/MatlabX.ctags --fields=+eKsp  -o - /tmp/input.m 
A	/tmp/input.m	/^	    A = magic(4);$/;"	var	func:input.classfunc3
R	/tmp/input.m	/^	    R = randn(3,4,5);$/;"	var	func:input.classfunc3
classfunc3	/tmp/input.m	/^        function classfunc3;$/;"	func	class:input	end:13
input	/tmp/input.m	/^classdef input < handle$/;"	class	end:14	super:handle
[jet@localhost]~/var/ctags% 
```

andebjor:
It looks very nice!

However, I think I need to stop here, at least for now. With the addition of recognizing classes I now have what I need for jumping around Matlab code in Vim, which is my use case.

One note about scope, and the keyword `end` though: This keyword does not only mark the end of functions and `method` blocks, but also `for`-loops, `if`-statements, and so on. I would thus imagine that it would be rather tricky to make the following case work as expected:

```matlab
classdef input < handle
    methods(Static)
        function classfunc3;
            for n=1:10
                if n%2 == 0
                    A = magic(4);
                    R = randn(3,4,5);
	
                    disp('A:');
                    disp(A);
	
                    disp('R:');
                    disp(R);
                end
            end
        end
    end
end
```

What would be nice though to support is class properties:

```matlab
classdef input < handle
    properties (Access = private)
        Data
    end
    methods
        function classfunc3(p)
            disp(p.Data);
        end
    end
end
```

But in practice properties are often given an initial value, which let them be captured as `variable` in the current implementation:
```matlab
classdef input < handle
    properties (Access = private)
        Data = [];
    end
end
```

But as I said, don't hold your breath for new PRs from me, right now :(

PR #1807, opened 2018-07-30
Avoid truncating a pattern in the middle of an UTF-8 character

### LINKED ISSUE #1275: Unicode parse incorrect!
See the result tags file.
screenshot_20161226_094605

The simple example header file.
[a.zip](https://github.com/universal-ctags/ctags/files/716057/a.zip)

### LINKED ISSUE #1805: Universal ctags inserts invalid `utf-8` characters for certain files
(
Thank you for contacting us.

If you are reporting an issue with the parsing output, please fill
the following template.  As your custom CTags configuration can
affect results, please always use `--options=NONE` as the first
option when running `ctags`.

Otherwise, delete the template and write your issue from scratch.
Examples may help developers understanding your issue better.

Use GitHub web interface and markdown notation.
Using mail results broken text rendering that makes
the developers go crazy.
)

*****

The name of the parser:

```
$ ctags -R
```

I dont have any special configurations in `.ctags` or anywhere else. This is a fresh VM that this test was run on. 

The content of input file: https://github.com/pallets/jinja/blob/master/jinja2/_identifier.py

`Universal-ctags` inserts invalid `utf-8` characters under certain circumstances.

The tags output you expect:

Expected tag output with all valide `utf-8` characters.

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(3522685), Copyright (C) 2015 Universal Ctags Team
Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: July 27 1018, 23:16:36
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath
```

How do you get ctags binary:

(
The ctags binary is built on `ubuntu-16.04` VM with no modifications other than installing necessary libraries such as `automate`, `autoreconf` for compiling `ctags` and necessary libraries to compile `vim` based on https://github.com/Valloric/YouCompleteMe/wiki/Building-Vim-from-source#a-for-a-debian-like-linux-distribution-like-ubuntu-type 
)

@lilydjwg pointed out to me that `ctags` was inserting invalid `utf-8` characters even though the file being used to generate the tags have all valid `utf-8` characters here: 
https://github.com/vim/vim/issues/3213#issuecomment-406961075

The compiled version of `ctags` works great in general.

Recently found out, that it turns out that `ctags` has a bug due to which the
old `Execuberant ctags` installed by `sudo apt-get install ctags` on Ubuntu
16.04 doesn't insert any invalid `utf-8` characters, but if I compile
`Universal-ctags` from source and isntall it based on instructions here:
https://github.com/universal-ctags/ctags/blob/master/docs/autotools.rst, it
will insert invalid `utf-8` characters. Heres the evidence:

With `exuberant-ctags` installed using just `sudo apt-get install ctags`:

With `Universal-ctags` compiled from source (latest commit) as of this post,
compiled with instructions from here:
https://github.com/universal-ctags/ctags/blob/master/docs/autotools.rst:

This causes a lot of problems in vim, because if invalid `utf-8` characters are
passed to `vim.eval`, `vim.eval` breaks and this leads to no tags returned at
all. Currently, there is only one way of transfering data contained in a `viml`
variable to the `python-name` space, using `vim.eval`. So, any other plugin in
vim or else where will have similar problems as well.  @ludovicchabant for
example had to post-process his tags file to stop such problems:
https://ludovic.chabant.com/devblog/2017/02/25/aaa-gamedev-with-vim/

Also he had to change `ctrl-py-matcher` to catch this issue.
https://github.com/ludovicchabant/ctrlp-py-matcher/blob/2f6947480203b734b069e5d9f69ba440db6b4698/autoload/pymatcher.py#L22

There are multiple other files I have seen which have similar problems, but I
have just provided one here to narrow down the problem.

My guess is this is a bug, and I don't expect that `ctags` would do this by
design. Can this be rectified, as this used to work fine in Exuberant Ctags
upon which Universal-ctags is based?

Ref: https://github.com/vim/vim/issues/3213#issuecomment-408727629

b4n:
Fixes #1275, fixes #1805.


Tmain/omit-long-patterns-etags.d/stdout-expected.txt
```diff
@@ -1,5 +1,5 @@
 
-input.sh,349
-                                                                                                func96(func961,0
+input.sh,351
+                                                                                                func96()func961,0
                                                                                                func95()func955,110
-                                                                                                 func97func979,219
+                                                                                                 func97(func979,219
```
b4n: @masatake do we wanna change this test case to balance the previous off-by-one-ness?  I'm not sure, because now it properly checks the behavior when the line is the exactly as long, short and longer than the truncation, which seems a good thing to do, but you might remember if you had another goal here maybe.

masatake:
@b4n, thank you very much. I will take a look at this weekend.

Tmain/omit-long-patterns-etags.d/stdout-expected.txt
masatake: It seems that there a bug has been existed since Exuberant-crags about newline handling.
`line` returned from `readLineFromBypassAnyway` can end with or without a newline char.
Both can occurs. Though the code trims the line with `line[len - 1] = '\0''`.

To combine your fix, I think the diff should be:
```
diff --git a/main/writer-etags.c b/main/writer-etags.c
index 762c8378..dbd6470f 100644
--- a/main/writer-etags.c
+++ b/main/writer-etags.c
@@ -98,18 +98,18 @@ static int writeEtagsEntry (tagWriter *writer,
                long seekValue;
                char *const line =
                                readLineFromBypassAnyway (etags->vLine, tag, &seekValue);
-               if (line == NULL)
+               if (line == NULL || line[0] == '\0')
                        return 0;
 
                len = strlen (line);
 
                if (tag->truncateLineAfterTag)
                        truncateTagLineAfterTag (line, tag->name, true);
-               else
-                       line [len - 1] = '\0';
+               else if (line [len - 1] == '\n')
+                       line [--len] = '\0';
 
                if (Option.patternLengthLimit < len)
-                       line [Option.patternLengthLimit - 1] = '\0';
+                       line [Option.patternLengthLimit] = '\0';
 
                length = mio_printf (mio, "%s\177%s\001%lu,%ld\n", line,
                                tag->name, tag->lineNumber, seekValue);
```

masatake:
@b4n, thank you for working on this topic.
Can I ask you to add explanation about the length adjustment for utf-8 input to  `Truncating the pattern for long input lines` section in docs/news.rst?

b4n:
@masatake Thanks for the review!  I made the changes, tell me if it was what you were expecting.

PR #1810, opened 2018-08-02
Fypp parser

p-vitt:
I created a (rudimentary) parser for [Fypp](https://github.com/aradi/fypp), a Fortran preprocessor. Comments are welcome.

I created the parser based on the Fortran parser. The structure and some of the code was taken from there. Should I add some comment about that in the copyright notice?

Right now the parser is able to create tags for macro definitions, which seems to be the most important feature for a preprocessor.

### DIFF
--- a/Units/parser-fypp.r/fypp-macro.d/expected.tags
+++ b/Units/parser-fypp.r/fypp-macro.d/expected.tags
@@ -1,0 +1,1 @@
+assertTrue	input.fy	/^#:def assertTrue(/;"	d
--- a/Units/parser-fypp.r/fypp-macro.d/input.fy
+++ b/Units/parser-fypp.r/fypp-macro.d/input.fy
@@ -1,0 +1,6 @@
+#:def assertTrue(cond)
+if (.not. ${cond}$) then
+  print *, "Assert failed"
+  error stop
+end if
+#:enddef assertTrue
--- a/main/parsers.h
+++ b/main/parsers.h
@@ -66,6 +66,7 @@
 	FalconParser, \
 	FlexParser, \
 	FortranParser, \
+	FyppParser, \
 	GdbinitParser, \
 	GoParser, \
 	HtmlParser, \
--- a/parsers/fypp.c
+++ b/parsers/fypp.c
@@ -1,0 +1,504 @@
+	/*
+ *   Copyright (c) 2018, Peter Vitt
+ *
+ *   This source code is released for free distribution under the terms of the
+ *   GNU General Public License version 2 or (at your option) any later version.
+ *
+ *   This module contains functions to generate tags for Fypp, a Python powered
+ *   Fortran preprocessor.
+ *   See: https://github.com/aradi/fypp
+ */
+
+/*
+ *   INCLUDE FILES
+ */
+#include "general.h"  /* must always come first */
+
+#include <string.h>
+
+#include "keyword.h"
+#include "parse.h"
+#include "read.h"
+#include "routines.h"
+#include "xtag.h"
+
+/*
+ *   MACROS
+ */
+#define isident(c)              (isalnum(c) || (c) == '_')
+#define isBlank(c)              (bool) (c == ' ' || c == '\t')
+#define isType(token,t)         (bool) ((token)->type == (t))
+#define isKeyword(token,k)      (bool) ((token)->keyword == (k))
+#define isSecondaryKeyword(token,k)  (bool) ((token)->secondary == NULL ? \
+	false : (token)->secondary->keyword == (k))
+
+/*
+ *   DATA DECLARATIONS
+ */
+enum eKeywordID {
+	KEYWORD_def,
+	KEYWORD_enddef,
+	KEYWORD_for,
+	KEYWORD_if,
+	KEYWORD_else,
+	KEYWORD_elif,
+	KEYWORD_endif,
+	KEYWORD_del,
+	KEYWORD_set,
+	KEYWORD_mute,
+	KEYWORD_stop,
+	KEYWORD_assert,
+	KEYWORD_include,
+	KEYWORD_getvar,
+	KEYWORD_setvar,
+	KEYWORD_delvar,
+	KEYWORD_globalvar,
+	KEYWORD_defined,
+};
+typedef int keywordId; /* to allow KEYWORD_NONE */
+
+typedef enum eTokenType {
+	TOKEN_UNDEFINED,
+	TOKEN_EOF,
+	TOKEN_CONTROL,
+	TOKEN_DIRECTCALL,
+	TOKEN_EVAL,
+	TOKEN_KEYWORD,
+	TOKEN_IDENTIFIER,
+	TOKEN_INLINESTART,
+	TOKEN_INLINEEND,
+	TOKEN_LINE,
+	TOKEN_LINECONTINUE,
+	TOKEN_NUMERIC,
+	TOKEN_PAREN_OPEN,
+	TOKEN_PAREN_CLOSE,
+	TOKEN_COMMA,
+	TOKEN_STATEMENT_END,
+} tokenType;
+
+typedef enum eTagType {
+	TAG_UNDEFINED = -1,
+	TAG_MACRO_DEFINITION,
+	TAG_CONDITION,
+	TAG_VAR_DEFINITION,
+	TAG_VAR_REMOVAL,
+	TAG_ITERATOR,
+	TAG_INCLUDE,
+	TAG_COMMENT,
+	TAG_MUTE,
+	TAG_STOP,
+	TAG_ASSERT,
+	TAG_COUNT /* must be last */
+} tagType;
+
+typedef struct sTokenInfo {
+	tokenType type;
+	keywordId keyword;
+	tagType tag;
+	vString* string;
+	vString* parentType;
+	vString* signature;
+	bool isMethod;
+	struct sTokenInfo* secondary; /* TODO: make use of this */
+	unsigned long lineNumber;
+	MIOPos filePosition;
+} tokenInfo;
+
+static langType Lang_fypp;
+static int Ungetc;
+
+static kindDefinition FyppKinds [] = {
+	{true,  'd', "def",        "definition"},
+	{false, 'f', "for",        "for loop"},
+	{false, 'i', "if",         "condition"},
+	{false, 'D', "del",        "remove variable"},
+	{false, 'S', "set",        "set variable"},
+	{false, 'e', "eval",       "evaluate"},
+	{false, 'c', "call",       "call"},
+	{false, 'C', "directcall", "direct call"},
+	{false, 'g', "global",     "global"},
+	{false, 'I', "include",    "include"},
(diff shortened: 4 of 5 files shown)


masatake:
A test case is failed but it is nothing to do with your new parser. I'm working on the bug now.

masatake:
Thank you for considering contributing to u-ctags.


We can do what you do in your parser with `.ctags` file like:

```
[yamato@master]/tmp% cat fypp.ctags
--langdef=fypp
--map-fypp=+.fy
--kinddef-fypp=d,definition,definitions
--regex-fypp=/^[ \t]*#:def[ \t]+([a-zA-Z][a-zA-Z0-9_]*).*/\1/d/
[yamato@master]/tmp% cat input.fy 
#:def assertTrue(cond)
if (.not. ${cond}$) then
  print *, "Assert failed"
  error stop
end if
[yamato@master]/tmp% u-ctags --options=./fypp.ctags -o - input.fy 
assertTrue	input.fy	/^#:def assertTrue(cond)$/;"	d
```

Could you tell me why you write your parsre in C?
I guess you may have future plan extending the current version, and for the plan, you may need a real programming language like C.

p-vitt:
> Could you tell me why you write your parsre in C?
I started with looking at the fortran parser, as Fypp is targeted to preprocess Fortran. The Fortran parser is written in C, so I started to experiment with this parser, modified it, extracted some parts, and voila, there was the Fypp parser.

In between I had a look at the optlib stuff, but to be honest, I'm not a big fan of regex magic. I can't get my head around it, and as there was a pretty decent Fortran parser written in C, it wasn't a big step to use C for a Fortran preprocessor parser.

I'm not sure whether I'm going to extent the parser. At the moment, tagging macro definitions fulfills my needs. But we just started to make use of Fypp, so I can't estimate for how long this holds.

masatake:
I see.

I have 3 comments.
About copyright notice, though I'm not a lawyer, I think you should put the name of the original notice and append your name after it if you take the codes from the original parser like:
https://github.com/strace/strace/blob/master/unwind-libdw.c

You define many kinds:

```
static kindDefinition FyppKinds [] = {
	{true,  'd', "def",        "definition"},
	{false, 'f', "for",        "for loop"},
	{false, 'i', "if",         "condition"},
	{false, 'D', "del",        "remove variable"},
	{false, 'S', "set",        "set variable"},
	{false, 'e', "eval",       "evaluate"},
	{false, 'c', "call",       "call"},
	{false, 'C', "directcall", "direct call"},
	{false, 'g', "global",     "global"},
	{false, 'I', "include",    "include"},
	{false, 'm', "mute",       "mute"},
	{false, 's', "stop",       "stop"},
	{false, 'a', "assert",     "assert"},
};
```
The primary mission of ctags is making tags for named definitions though there are some exceptions.
Kinds are for categorizing the definitions in a language.
About FyppKinds, other than "def" don't make sense.
Taking a glance at README.md of Fypp, Only "def" kind is enough for tagging, isn't it?

Currently there are three ways to write a parser: token oriented, line oriented, and optlib.
The token oriented one is the most powerful and flexible. However, it is not easy to write.
Your token oriented parser tracks only one token ("def" keyword). As the result your parser
emits "Hello" for the following input:

```
[yamato@master]/tmp/ctags% cat ../foo.fy
#:if DEBUG > 0
program hello
  print *, 'def Hello World!'
end program hello
#:endif
[yamato@master]/tmp/ctags% ./ctags -o - ../foo.fy 
Hello	../foo.fy	/^  print *, 'def Hello /;"	d
```

My fypp.ctags isn't caught the trick.

Universal-ctags maintainers have to maintain a code. So if there are two parsers for a language and the qualities are the same, I would like to choose smaller and simpler one. For the above input,  fypp.ctags
shows the better result and its code is very short. I cannot find the reason to choose the token oriented parser.

A line oriented parser will be acceptable. parsers/r.c is an example.
I think a line oriented parser is suitable for the syntax of fypp.

If you want to parse incomplete Fortran code in fypp file with fypp parser, there is no choice;
we should choose the token oriented approach. However, do you want to go the crazy way?

p-vitt:
About the kinds: I forgot to strip this down. I created this list when I didn't know yet what a kind is used for. Atm I would only have d, correct. But I could also think of s (set variable) as a possible jump target.

About the rest: I'm fine with your regex parser. Also the possible extension of `#:set var` would be quite easy with a regex, I assume. And for the case that I (or someone) want to extend the parser and actually need the token approach, I can still resuse the branch in my fork.

masatake:
fypp is original designed as a preprocessor for Fortran input.
So I think fypp programmers expect ctags captures both fypp macro and Fortran language objects from fypp inputs.
I am extending my fypp parser to parse fortran. It works though there are still some bugs about memory management and stream handing.
```
[yamato@master]~/var/ctags-github/Units/parser-fypp.r/run-guest.d% cat input.fy                     
#! This test case is derived from #1810, a pull request submitted by @p-vitt
#! and parser-fortran.r/fortran-interface.d/input.f90.
#! Callable needs only string argument
#:def debug_code(code)
  #:if DEBUG > 0
    $:code
  #:endif
#:enddef debug_code

#! Pass code block as first positional argument
#:call debug_code
  if (size(array) > 100) then
    print *, "DEBUG: spuriously large array"x
  end if
#:endcall debug_code

program helloworld
print *, "Hello, world."
end program helloworld

module test_interface
#! Callable needs also non-string argument types
#:def repeat_code(code, repeat)
  #:for ind in range(repeat)
    $:code
  #:endfor
#:enddef repeat_code
  type atype
  end type atype
  ! operator overload
  interface operator(+)
    ! subprogram prototype
    type(atype) function add(a, b)
      import atype
      type(atype), intent(in) :: a, b
    end function add
  end interface operator(+)
  ! wrap subprogram prototypes
  interface ! anonymous interface
    subroutine suba()
    end subroutine suba
    subroutine subb()
    end subroutine subb
  end interface
  ! define generic subprograms
#! fypp preprocessor comments  here, and
  interface get
#! there
    ! subprogram name list
    module procedure get_1d
    module procedure get_2d
  end interface get
contains
  ! definition of subprograms
  subroutine get_1d(a)
    real a(:)
  end subroutine get_1d
#! Pass code block as positional argument and 3 as keyword argument "repeat"
#:call repeat_code(repeat=3)
this will be repeated 3 times
#:endcall repeat_code
  subroutine get_2d(a)
    real a(:, :)
  end subroutine get_2d
end module test_interface


[yamato@master]~/var/ctags-github/Units/parser-fypp.r/run-guest.d% u-ctags --fields=+l --extras=+g -o - input.fy
+	input.fy	/^  interface operator(+)/;"	i	language:Fortran	module:test_interface
atype	input.fy	/^  type atype$/;"	t	language:Fortran	module:test_interface
debug_code	input.fy	/^#:def debug_code(code)$/;"	m	language:Fypp
get	input.fy	/^  interface get$/;"	i	language:Fortran	module:test_interface
get_1d	input.fy	/^  subroutine get_1d(/;"	s	language:Fortran	module:test_interface
get_2d	input.fy	/^  subroutine get_2d(/;"	s	language:Fortran	module:test_interface
helloworld	input.fy	/^program helloworld$/;"	p	language:Fortran
repeat_code	input.fy	/^#:def repeat_code(code, repeat)$/;"	m	language:Fypp
test_interface	input.fy	/^module test_interface$/;"	m	language:Fortran
```

I guess fypp can be used as a preprocessor for another language.
My fypp parser can handle other than fortran if a user gives a hint.
In the following example, my fypp parser parses fypp powered java source code:

```
[yamato@master]~/var/ctags-github/Units/parser-fypp.r/run-alt-guest.d% cat input.fy                                                                            
#! This test case is derived from #1810, a pull request submitted by @p-vitt
#! and parser-java.r/enum.java.d/input.java
#! Callable needs only string argument
#:def debug_code(code)
  #:if DEBUG > 0
    $:code
  #:endif
#:enddef debug_code
public enum e {
	A,    // should be 'e', not 'f'
	B(),  // should be 'e', not 'm'
	C(1), // should be 'e', not missing

#! Pass code block as first positional argument
#:call debug_code
  if (size(array) > 100) then
    print *, "DEBUG: spuriously large array"x
  end if
#:endcall debug_code
	D,    // should be 'e', not 'f'
	E(),  // should be 'e', not 'm'
	F(1), // should be 'e', not missing
	;
#! Callable needs also non-string argument types
#:def repeat_code(code, repeat)
  #:for ind in range(repeat)
    $:code
  #:endfor
#:enddef repeat_code

        e(int x) {}
        e() {}

	public String string;
	public final Shape shape;
	public final boolean twoKeywordsInARow;

#! fypp preprocessor comments  here, and
	public String getString() {
		return string;
	}

#! there
	public final Shape getShape() {
		return shape;
	}
#! Pass code block as positional argument and 3 as keyword argument "repeat"
#:call repeat_code(repeat=3)
this will be repeated 3 times
#:endcall repeat_code
}
[yamato@master]~/var/ctags-github/Units/parser-fypp.r/run-alt-guest.d%  u-ctags -o - --fields=+l --param-Fypp:guest=Java --extras=+g input.fy
A	input.fy	/^	A,    \/\/ should be 'e', not 'f'$/;"	e	language:Java	enum:e	file:
B	input.fy	/^	B(),  \/\/ should be 'e', not 'm'$/;"	e	language:Java	enum:e	file:
C	input.fy	/^	C(1), \/\/ should be 'e', not missing$/;"	e	language:Java	enum:e	file:
D	input.fy	/^	D,    \/\/ should be 'e', not 'f'$/;"	e	language:Java	enum:e	file:
E	input.fy	/^	E(),  \/\/ should be 'e', not 'm'$/;"	e	language:Java	enum:e	file:
F	input.fy	/^	F(1), \/\/ should be 'e', not missing$/;"	e	language:Java	enum:e	file:
debug_code	input.fy	/^#:def debug_code(code)$/;"	m	language:Fypp
e	input.fy	/^        e() {}$/;"	m	language:Java	class:e
e	input.fy	/^        e(int x) {}$/;"	m	language:Java	class:e
e	input.fy	/^public enum e {$/;"	g	language:Java
getShape	input.fy	/^	public final Shape getShape() {$/;"	m	language:Java	class:e
getString	input.fy	/^	public String getString() {$/;"	m	language:Java	class:e
repeat_code	input.fy	/^#:def repeat_code(code, repeat)$/;"	m	language:Fypp
shape	input.fy	/^	public final Shape shape;$/;"	f	language:Java	class:e
string	input.fy	/^	public String string;$/;"	f	language:Java	class:e
twoKeywordsInARow	input.fy	/^	public final boolean twoKeywordsInARow;$/;"	f	language:Java	class:e
```
Works fine. My fypp parser doesn't expand macros and eval the conditions for if/elif/else. So this is not perfect but I think much better than nothing.

Newly introduced technique behind the parser can by applied to parse yaml input stacked on jinja2 input for example. Quite fruitful.

masatake:
Thank you for opening this pull request.
I merged my implementation. Now ctags can parser fypp input.

masatake:
Tracking included files as reference tags is a TODO item.

p-vitt:
Thank you for implementing this in an easier way. :)

PR #1811, opened 2018-08-06
Improvements on the AutoIt parser

b4n:
See #1598, #1599 and https://github.com/geany/geany/pull/1752.

@Skif-off @masatake I'm not sure of what I'm doing here as I have no clue about AutoIt, but well :smile: 
@masatake please check the usage of roles here, I have no idea if it's correct.  Well, I should check the docs, but still.

### DIFF
--- a/Tmain/list-fields-with-prefix.d/stdout-expected.txt
+++ b/Tmain/list-fields-with-prefix.d/stdout-expected.txt
@@ -4,6 +4,7 @@ e       UCTAGSend           no      NONE             -i-    no    end lines of v
 p       UCTAGSscopeKind     no      NONE             s--    no    Kind of scope as full name
 r       UCTAGSroles         no      NONE             s--    no    Roles
 x       UCTAGSxpath         no      NONE             s--    no    xpath for the tag
+-       UCTAGSproperties    no      AutoIt           s--    no    properties (static, volatile, ...)
 -       UCTAGSproperties    no      C                s--    no    properties (static, inline, mutable,...)
 -       UCTAGScaptures      no      C++              s--    no    lambda capture list
 -       UCTAGSname          yes     C++              s--    no    aliased names
--- a/Tmain/list-fields.d/stdout-expected.txt
+++ b/Tmain/list-fields.d/stdout-expected.txt
@@ -22,6 +22,7 @@ s	NONE	yes	NONE	s--	no	Scope of tag definition (`p' can be used for printing its
 t	typeref	yes	NONE	s--	no	Type and name of a variable or typedef
 x	xpath	no	NONE	s--	no	xpath for the tag
 z	kind	no	NONE	s--	no	Include the "kind:" key in kind field (use k or K) in tags output, kind full name in xref output
+-	properties	no	AutoIt	s--	no	properties (static, volatile, ...)
 -	properties	no	C	s--	no	properties (static, inline, mutable,...)
 -	captures	no	C++	s--	no	lambda capture list
 -	name	yes	C++	s--	no	aliased names
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -4,6 +4,8 @@
 #
 #LANGUAGE      KIND(L/N)         NAME                ENABLED DESCRIPTION
 Asm            s/section         placement           on      placement where the assembled code goes
+AutoIt         S/script          local               on      local include
+AutoIt         S/script          system              on      system include
 Automake       c/condition       branched            on      used for branching
 Automake       d/directory       data                on      directory for DATA primary
 Automake       d/directory       library             on      directory for LIBRARIES primary
@@ -62,6 +64,8 @@ Vera           h/header          system              on      system header
 #
 #LANGUAGE      KIND(L/N)         NAME                ENABLED DESCRIPTION
 Asm            s/section         placement           on      placement where the assembled code goes
+AutoIt         S/script          local               on      local include
+AutoIt         S/script          system              on      system include
 Automake       c/condition       branched            on      used for branching
 Automake       d/directory       data                on      directory for DATA primary
 Automake       d/directory       library             on      directory for LIBRARIES primary
--- a/Units/parser-autoit.r/simple-au3.d/args.ctags
+++ b/Units/parser-autoit.r/simple-au3.d/args.ctags
@@ -1,0 +1,2 @@
+--fields=+r
+--extras=+r
--- a/Units/parser-autoit.r/simple-au3.d/args.ctags
+++ b/Units/parser-autoit.r/simple-au3.d/args.ctags
@@ -1,2 +1,3 @@
 --fields=+erS
+--fields-AutoIt=+{properties}
 --extras=+r
--- a/Units/parser-autoit.r/simple-au3.d/args.ctags
+++ b/Units/parser-autoit.r/simple-au3.d/args.ctags
@@ -1,2 +1,2 @@
---fields=+r
+--fields=+rS
 --extras=+r
--- a/Units/parser-autoit.r/simple-au3.d/args.ctags
+++ b/Units/parser-autoit.r/simple-au3.d/args.ctags
@@ -1,2 +1,2 @@
---fields=+rS
+--fields=+erS
 --extras=+r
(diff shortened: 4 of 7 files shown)


masatake:
"script" is better than "include" as the kind name.

```
diff --git a/Tmain/list-roles.d/stdout-expected.txt b/Tmain/list-roles.d/stdout-expected.txt
index 264285df..aec3c3b6 100644
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -4,6 +4,8 @@
 #
 #LANGUAGE      KIND(L/N)         NAME                ENABLED DESCRIPTION
 Asm            s/section         placement           on      placement where the assembled code goes
+AutoIt         s/script          asLibrary           on      included as a library
+AutoIt         s/script          fromCurdir          on      included from the current directory
 Automake       c/condition       branched            on      used for branching
 Automake       d/directory       data                on      directory for DATA primary
 Automake       d/directory       library             on      directory for LIBRARIES primary
@@ -62,6 +64,8 @@ Vera           h/header          system              on      system header
 #
 #LANGUAGE      KIND(L/N)         NAME                ENABLED DESCRIPTION
 Asm            s/section         placement           on      placement where the assembled code goes
+AutoIt         s/script          asLibrary           on      included as a library
+AutoIt         s/script          fromCurdir          on      included from the current directory
 Automake       c/condition       branched            on      used for branching
 Automake       d/directory       data                on      directory for DATA primary
 Automake       d/directory       library             on      directory for LIBRARIES primary
diff --git a/parsers/autoit.c b/parsers/autoit.c
index 973839bf..33c00cce 100644
--- a/parsers/autoit.c
+++ b/parsers/autoit.c
@@ -32,13 +32,26 @@ typedef enum {
        K_REGION,
        K_GLOBALVAR,
        K_LOCALVAR,
+       K_SCRIPT,
 } AutoItKind;
 
+typedef enum {
+       R_SCRIPT_AS_LIBRARY,
+       R_SCRIPT_FROM_CURDIR,
+} AutoItScriptRole;
+
+static roleDefinition AutoItScriptRoles [] = {
+       { true, "asLibrary", "included as a library" },
+       { true, "fromCurdir", "included from the current directory" },
+};
+
 static kindDefinition AutoItKinds [] = {
        { true, 'f', "func", "functions" },
        { true, 'r', "region", "regions" },
        { true, 'g', "global", "global variables" },
        { true, 'l', "local", "local variables" },
+       { true, 's', "script", "scripts",
+         .referenceOnly = true, ATTACH_ROLES (AutoItScriptRoles) },
 };
 
 /*
@@ -107,6 +120,29 @@ static void findAutoItTags (void)
                        }
                        else if (match (p, "endregion"))
                                nestingLevelsPop (nls);
+                       else if (match (p, "include"))
+                       {
+                               p += 7; /* strlen("include") = 7 */
+                               while (isspace ((int) *p))
+                                       ++p;
+                               if (*p == '<' || *p == '"')
+                               {
+                                       int role = (*p == '<')
+                                               ? R_SCRIPT_AS_LIBRARY
+                                               : R_SCRIPT_FROM_CURDIR;
+                                       ++p;
+                                       while (*p != '\0' && *p != '>' && *p != '"')
+                                       {
+                                               vStringPut (name, (int) *p);
+                                               ++p;
+                                       }
+                                       if (vStringLength(name) > 0)
+                                       {
+                                               makeSimpleRefTag (name, K_SCRIPT, role);
+                                               vStringClear (name);
+                                       }
+                               }
+                       }
                }
                else
                {
```

Your change cannot capture the script name in
```
#include<GUIConstantsEx.au3>
```
match() expects a whitespace after keywords.
(See Units/parser-autoit.r/simple-au3.d/input.au3.)

For testing you have to pass `--extras=+r` and `--fields=+r`.
``--extras=+r`` means capturing reference tags.
``--fields=+r`` means print the roles for tags.

I'm not sure the role names: `fromCurdir` and `asLibrary`.
C parser does:
```
[yamato@master]/tmp% u-ctags --list-roles=C
u-ctags --list-roles=C
#KIND(L/N) NAME   ENABLED DESCRIPTION
d/macro    undef  on      undefined
h/header   local  on      local header
h/header   system on      system header
```
So "local" (local script) and "system" (system script) are another candidates.

b4n:
@masatake done, and I also fixed handling a directive not followed by one of the expected delimiters.
I also added a few extra fancy features while playing with it :)

masatake:
@b4n, I also added an extra fancy features. I made a pull request to your branch. Could you merge it?

b4n:
@masatake haha I had the exact same thing waiting on my local copy for me to commit :)  I committed a mix of both our works which were virtually identical with slightly nicer details in each.

@Skif-off bah, I just expanded upon your work :)  BTW, it's my sweet spot: feed me a parser for a language I don't know and I'll learn the syntax and add a gazillion fancy features. I love it :]
Sadly AutoIt doesn't seem to have a proper syntax reference, so there's a good dose of guessing :disappointed:

Skif-off:
@b4n I think it's not just "just" and "expanded" :)
> Sadly AutoIt doesn't seem to have a proper syntax reference, so there's a good dose of guessing

What do you mean? The documentation is quite detailed, with examples.

b4n:
@Skif-off I mean that I couldn't find a formal answer to e.g. the question "what is a valid function identifier" and this kind of formal definitions of what might seem like a detail, but is important for a conforming implementation.

Skif-off:
[User functions](https://www.autoitscript.com/autoit3/docs/intro/lang_functions.htm):
> function names are case insensitive
> ...
> Function names must start with either a letter or an underscore, and the remainder of the name can contain any combination of letters and numbers and underscores.

Namespace is typical, declaring is [here](https://www.autoitscript.com/autoit3/docs/keywords/Func.htm). Is it not what you mean?

b4n:
Hum, good point, mae culp -- moreover because I might actually have found that one.  It's a little confusing not to have the formal definition of what `funcname` can be with what looks like the formal definition of the `func` keyword, but OK.

However, it's still not thorough, for example I just discovered that a user function can have a `Volatile` qualifier, which is not mentioned in the `Func` reference -- and thus. This makes it hard to discover the formal syntax.

b4n:
OK I think I'm done here, @masatake maybe you'd like to look at the last changes before merging?

Skif-off:
@b4n, one little trifle, maybe you delete these two lines?
```
*   Functions: (?i)^func[ \t]{1,}([a-z0-9_]{1,}) >> \1
*   Regions:   (?i)^#region[ \t]*(.*?)           >> \1
```
I wrote them as a reminder, but now it can parse more than functions and regions.

masatake:
C++ parser stores "static", "inline", "mutable", etc. to "properties:" field.
See the output of "--list-fields".
If you are interested in you can implement the field in AutoIt parser.
This is NOT must for merging.

"propperties" field is a parser specific field. If you are interested in implementing the field ofr AutoIt, talk to me.

Skif-off:
@masatake sorry, I don't understand what is "properties:" and how to use it in practice. Where I can found (or how get) detailed example?

masatake:
@Skif-off, currently only C/C++ parser emits properties field.
```
[yamato@master]/tmp% cat bar.c
cat bar.c
extern int f(void)
{
  return 0;
}

static int g(void)
{
  return 0;
}

inline int h(void)
{
  return 0;
}
[yamato@master]/tmp% u-ctags -o - --fields-C='+{properties}' bar.c
u-ctags -o - --fields-C='+{properties}' bar.c
f	bar.c	/^extern int f(void)$/;"	f	typeref:typename:int	properties:extern
g	bar.c	/^static int g(void)$/;"	f	typeref:typename:int	file:	properties:static
h	bar.c	/^inline int h(void)$/;"	f	typeref:typename:int	properties:inline
```
I cannot answer the field is useful or not.
How to utilize is up to client tools.

masatake:
@b4n, feel free to merge this.

Skif-off:
@masatake thanks, it's interesting feature, but on first sight I can not see how it can be useful in AutoIt. Maybe later.

masatake:
@b4n, can I merge this?

b4n:
I just pushed the properties field for `static` and `volatile`, which I implemented as much as a PoC than because it seemed useful.  But it looks simple and nice enough to me to be included either way.
@masatake Apart from that I'm done, and your or I can merge this once you checked the last commit I just pushed.

masatake:
About the new commit for implementing  `properties` field, it also looks fine to me.
Please, go ahead after appveyor says o.k.
Thank you.

PR #1812, opened 2018-08-06
Fix list of supported flags in the help for "fields" and "extras" options

b4n:
Update the lists with the output of

	$ ctags --list-fields | grep -o '^[^-#]' | sort | tr -d '\n'

and

	$ ctags --list-extras | grep -o '^[^-#]' | sort | tr -d '\n'

respectively.


masatake:
Thank you.

Instead of enumerating all the flags, I think printing following messages is better from the view of  maintainability:
```
(flags: see the output of --list-fields option)
```
and
```
(flags: see the output of --list-extras option)
```
.

TODO: I will work on following two items after merging this pull request.
* This is applicable to the help message for `--kinds-<LANG>'.
* Remove --<LANG>-kinds from help message.

b4n:
@masatake I think it's quite nice for the user checking the help to see a summary, instead of having to make a second call with another option.
I guess ideally the list would be auto-generated just like the output of `--list-fields` is.  Not sure how hard it'd be, but that would keep it up to date while keeping the current summary.

b4n:
@masatake we could do something like that https://github.com/b4n/fishman-ctags/commit/671e23220ef6e8f25638116eba85fb0af2f18be5
But it might be a little overcomplicated for the improvement, although I'm pretty sure my quick implementation is fairly easy to improve and simplify.

masatake:
I would like to take over b4n@671e232.

b4n:
@masatake sure, go ahead

PR #1814, opened 2018-08-09
C++,QtMoc: initialize uninitialized local variable

masatake:

(Reported by valgrind.)


masatake:
I wonder why gcc doesn't report this.

Issue #1816, opened 2018-08-09
CMake parser


hadrielk:
Hi, I've written a CMake files parser using the optlib multi-table regex options. It only tags basic things like build target names, cmake variables, cmake function names, etc. My questions are:

1. Is this something useful that you want?
2. If yes, should I create a PR for this repo, or the [ctags-xparsers](https://github.com/universal-ctags/ctags-xparsers) one?
3. Should I use optlib2c and include the generated C-code in the PR? Or should I just start with the `cmake.ctags` file and wait for feedback first?

Thanks!

p.s. I really appreciate the new multi-table regex stuff. With it we were finally able to create tags for things generated by C/C++ macros in my employer's codebase. Thank you!

masatake:
Today is one of the most delightful day in my free software programming life.
As far as I can remember, other than I, you are the first person who uses the multi-table regex feature.
I'm suppressed at there is a person who can use it because the feature is not documented enough.
I guess you read my puppetManifest.ctags.  Any comment for the feature itself and pull request for its documents are really welcome.


> Hi, I've written a CMake files parser using the optlib multi-table regex options. It only tags basic things like > build target names, cmake variables, cmake function names, etc. My questions are:
> 1.  Is this something useful that you want?

YES. Do you know I implemented the multi-table regex feature was for writing cmake parser?
After implementing the feature, I had to read puppet manifests in my day job. So I changed the
target to puppet manifets. "writing cmake parser" has been in my queue. It seems that I can dequeue the item:) In addtion, I think build-scriptts are the important area for u-ctags. cmake becomes popular, so we need a parser for it.

> 2. If yes, should I create a PR for this repo, or the ctags-xparsers one?
Please, forget ctags-xparsers repo. When we started this project, I expected we get many parsers written with `--regex-<LANG>` option. At that time we don't have optlib2c translator. So I have a plan to distribute
.ctags files in a package separated from ctags itself. ctags-xparsers was planed as a place to maintain separated .ctags files. Now we have the translator so we don't need the separate repository.
Please, make a pull request for this repo.

> 3. Should I use optlib2c and include the generated C-code in the PR? Or should I just start with the cmake.ctags file and wait for feedback first?

Please, add generated C-code, too. When I merge your change, the C-code must be included.

> p.s. I really appreciate the new multi-table regex stuff. With it we were finally able to create tags for things generated by C/C++ macros in my employer's codebase. Thank you!

[...]

Issue #1798, opened 2018-07-17
lua: function declared inside an object is not tagged correctly


doronbehar:
The name of the parser: `lua`
The command line you used to run ctags:
```
$ ctags --options=NONE test.lua
```
The content of input file:

```lua
myVar = {
	obj = 12,
	myMethod = function()
	end,
	str = "bafsg"
}

return myVar
```

The tags output you are not satisfied with:
```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	/5a4b6d04/
myMethod	test.lua	/^	myMethod = function()$/;"	f
```

The tags output you expect:

```
!__ COMMENTS __!
myVar:myMethod	test.lua	/^	myMethod = function()$/;"	f
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(5a4b6d04), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Jul 11 2018, 12:14:12
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml
```
`ctags` was build from source from this repository.

masatake:
Thank you for contacting.

The lua parser is incomplete to implement what you want quickly. For example, it doesn't capture
a variable like `myVar` in your example.
The current implementation is line oriented.  It must be rewritten in token oriented style.
Rewriting it is a small thing.

Difficulties are in what kind of output do we want for a dynamic language like lua.
JavaScript looks similar to lua for me.
In addition, the JavaScript parser of ctags is written by @b4n, an experienced developer.
So I would like to reuse the output design used in the JavaScript parser.
(I hope you may know the syntax of JavaScript.)

Js input:
```
[...]

PR #1818, opened 2018-08-11
CMake: new parser

hadrielk:
As mentioned in issue #1816, this is a CMake file parser using the optlib multitable regex feature.


win32/ctags_vs2013.vcxproj.filters
```diff
@@ -660,4 +663,4 @@
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
masatake: This change for the end of file should not be added. git add -p may help you.

masatake:
Other than one minor point about the build file for windoes. everything looks good to me.
I made one pull request to you to support end field. My pull request supports scope, too. However, I'm not sure the scope filed makes sense in cmake syntax. If my pull request is acceptable, could you merge and push the changes here?
(https://github.com/hadrielk/ctags/pull/1)

After visiting your profile page I know why you needs cmake parser; you are a wireshark hacker.
Uuniversal-ctags can offer more attractions to you, visit https://github.com/universal-ctags/ctags/issues/1798 :-P

masatake:
( I run "make chop LANGUAGES=CMake", "make slap LANGUAGES=CMake", and "make fuzz LANGUAGES=CMake". All are passed.)

masatake:
The techniques used in the cmake.ctags are impressive. I have to consider applying them to my PuppetManifest parser.

win32/ctags_vs2013.vcxproj.filters
hadrielk: Oops. That's my editor being too smart for its own good. I'll fix it.

hadrielk:
> I'm not sure the scope filed makes sense in cmake syntax

There's no namespace concept in CMake - they do have a concept of "scope", but not in the sense that ctags would care I think (other than the `end:` field). Most of their scoping is based on directory information. Except variables (created with `set()`), which if set in a function/macro are in that function's scope only, unless they specify the `PARENT_SCOPE` argument.

Do you think I should try to add the scope for variable tags within functions/macros? (ie, use `{scope=ref}`) If so I'll have to also look for `PARENT_SCOPE` and if it's there then not set the scope reference.

optlib/cmake.ctags
```diff
@@ -0,0 +1,158 @@
...
+# capture a required trailing '[# \t\n\)]', to verify the full name token matched
+# the name's pattern, but then we advanceTo=2start for the next round, so that
+# we don't go past a potential '#' comment token but instead match it again in
+# the main table as a comment (or whitespace if it was whitespace).
+#
+
+#
+# variable
+#
+--_mtable-regex-CMake=variable/([A-Za-z0-9_.-]+)([# \t\n\)])/\1/v/{tleave}{advanceTo=2start}
```
hadrielk: bug: all these `{advanceTo=2start}` should be `{_advanceTo=2start}`

masatake:
Could you squash 08c4658  to acdd8 ? 
Could you squash the changes for optlib/cmake.ctags in 66c62e0 to 729059f  ? 
Could you squash the changes for  Units/parser-cmake.r/cmake-comments.d/expected.tags and  Units/parser-cmake.r/cmake-comments.d/input.ctags in 66c62e0 to d3803b4 ?
Could you sqash the changes for optlib/cmake.c in  66c62e0 to acdd8 ?

hadrielk:
Would you mind if I just squashed all my changes down to one commit? It would be easier.

masatake:
Merging it one ("CMake: new parser")  is o.k.

hadrielk:
Done.

masatake:
Thank you!

PR #1824, opened 2018-08-12
Git ignore .DS_Store files generated by Mac-OSX

hadrielk:
Mac OSX generates `.DS_Store` files in directories, which most projects add to `.gitignore` to ignore. So this does that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 *~
 .deps
 .dirstamp
+.DS_Store
 a.out
 aclocal.m4
 autom4te.cache


masatake:
Do you think we should add .DS_Store to the built-in ignore list of ctags defined in
initOptions() in main/options.c?

hadrielk:
> Do you think we should add .DS_Store to the built-in ignore list of ctags defined in
initOptions() in main/options.c?

Huh, I didn't even know ctags had that. Yes it should be added to that too. Want a separate PR, or me to add it to this one?

masatake:
Could you add a commit here?

PR #1828, opened 2018-08-15
main: introduce TRACE_PRINT_{PREFIX|FMT|NEWLINE} macros

masatake:
TRACE_PRINT prints a line at once; it inserts a prefix at the start and a newline at the end.
Sometimes I want to print a trace line incrementally.

TRACE_PRINT_{PREFIX|FMT|NEWLINE} are for the purpose.

TRACE_PRINT_PREFIX just prints a prefix.
TRACE_PRINT_FMT does the same as TRACE_PRINT but it doesn't print a prefix and a newline.
TRACE_PRINT_NEWLINE just prints a newline.

You can use them like:

		    TRACE_PRINT_PREFIX();
    #ifdef DO_TRACING
		    if (ctx->if_tracker)
		    {
			    for (int i = 0; i < intArrayCount (ctx->if_tracker); i++)
				    TRACE_PRINT_FMT("%d ", intArrayItem(ctx->if_tracker, i));
		    }
    #endif	/* DO_TRACING */
		    TRACE_PRINT_NEWLINE();

### DIFF
--- a/main/trace.c
+++ b/main/trace.c
@@ -54,7 +54,35 @@ void traceLeave(const char * szFunction,const char * szFormat,...)
 	fprintf(stderr,"\n");
 }
 
-void tracePrint(const char * szFunction,const char * szFormat,...)
+static void tracePrintFmtVa(const char * szFormat, va_list va)
+{
+	vfprintf(stderr,szFormat,va);
+}
+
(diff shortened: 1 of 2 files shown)


masatake:
(@pragmaware, you may be interested i this change.)

PR #1831, opened 2018-08-16
CMake: fix cmake.c generated file

hadrielk:


hadrielk:
> Could you add a case for this pattern?

Done.

PR #1838, opened 2018-08-20
Asciidoc: new parser

hadrielk:
This ports the Asciidoc parser from geany, fixes a bug in it, and adds tagging of anchors. (other than anchors inline in one-line titles)

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -38,6 +38,7 @@ The following parsers have been added:
 
 * Ada
 * AnsiblePlaybook *libyaml*
+* Asciidoc
 * Autoconf
 * Automake
 * AutoIt
--- a/Units/parser-asciidoc.r/anchor-asciidoc.d/expected.tags
+++ b/Units/parser-asciidoc.r/anchor-asciidoc.d/expected.tags
@@ -1,0 +1,15 @@
+Chapter 1 (Level 0)	input.asciidoc	/^= Chapter 1 (Level 0)$/;"	c
+Chapter 2	input.asciidoc	/^Chapter 2$/;"	c
+Level 3 Section 1.1.1.1 Title	input.asciidoc	/^==== Level 3 Section 1.1.1.1 Title$/;"	t	subsection:Subsection 1.1.1
+Section 1.1	input.asciidoc	/^== Section 1.1$/;"	s	chapter:Chapter 1 (Level 0)
+Section 1.2	input.asciidoc	/^== Section 1.2$/;"	s	chapter:Chapter 1 (Level 0)
+Section 2.1	input.asciidoc	/^Section 2.1$/;"	s	chapter:Chapter 2
+Subsection 1.1.1	input.asciidoc	/^=== Subsection 1.1.1$/;"	S	section:Section 1.1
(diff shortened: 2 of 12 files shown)


masatake:
I didn't do this because the version of GPL of the orignal code is not written explictly.
@b4n, should we contact to the orignal author?

b4n:
@masatake yeah probably, not sure what is legal when there's no version specified.
Anyway, fortunately he's easy to reach and likely not to care too much… @elextr ? :)

b4n:
@masatake However, as the PR doesn't change the licensing header, I'd think it just makes this file whichever version, doesn't it?  Or is it a problem because v3 is incompatible with v1 in a way that cannot be mixed in an executable?

elextr:
Well, as the comment at the top says, I copied most of it from the rest parser, including the license section.  As that was explicitly included in

masatake:
@elextr, thank you.
Each source file of Exuberant-ctags didn't show the actual version of GPL. However, COPYING file at the top leve of the source tree is Version 2. So we updated the header of each source files derived from Exuberant-ctags like:

```
/*
*   Copyright (c) 1996-2003, Darren Hiebert
*
*
*   This module contains functions for parsing and scanning C, C++, C#, D and Java
*   source files.
*/
```

> As that was explicitly included in Geany that would make it GPL v2+.

It seems that the siltation is similar to old Universal-ctags.
The COPYING file at the top of source tree shows the actual version of GPL.
We I think we can expect we can use the Ascii parser in GPL v2.

However, the asciidoc parser submitted here leaves the original Geany source tree when this
pull request is merge. So it is not obvious that the relation between the source code and Geany's COPYING file. Updating the original header at Geany's side is ideal for me.

elextr:
Actually Geany itself is licensed GPL V2 but most of the source has the usual "V2 or later" text.

> Each source file of Exuberant-ctags didn't show the actual version of GPL. 

Given that Nick (the copyright holder of the rest parser I copied and modified) committed the rest parser to Geany himself back in 2007, he was just copying the then standard Exuberant-ctags approach and clearly intended the Geany license (GPL v2) to cover it.  And GPL V2 states "...If the Program does not specify a version number of this License, you may choose any version...".

As the file is being removed from Geany @masatake is right this PR just needs to make the version explicit as above, and also add
```
Copyright (c) 2007-2011, Nick Treleaven
```
to cover the material originally copied from the rest parser.

I note that [ctags rest parser](https://github.com/universal-ctags/ctags/blob/master/parsers/rst.c) is still copyright Nick and has the V2 or later added.

masatake:
@hadrielk, thank you. I will update rst.c side.

masatake:
> This ports the Asciidoc parser from geany

Could you add this sentence to the commit log for 1eb07f9.

docs/news.rst
```diff
@@ -38,6 +38,7 @@ The following parsers have been added:
 
 * Ada
 * AnsiblePlaybook *libyaml*
+* Asciidoc
```
masatake: Fine!

masatake:
LGTM. 
You can add a comment to #1566 after merging this:-)

PR #1847, opened 2018-08-21
[HEADS-UP,WIP] peg based new java parser

masatake:
This will be the biggest change in the history of fishman/ctags and Universal-ctags.

This is an experimental branch for testing "packcc" parser generator
and peg based "NewJava" parser.  Only building with Autotools is
supported. Volunteers for the other build-system are welcome.

packcc is registred as a git submodule. It is taken from https://github.com/enechaev/packcc.
The peg definition for Java 1.7 is taken from https://github.com/pointlander/peg/blob/master/grammars/java/java_1_7.peg.

git submodule handling is done in ./autogen.sh.

How to switch the branch:

```
[yamato@master]/tmp% git clone https://github.com/masatake/ctags.git
Cloning into 'ctags'...
remote: Counting objects: 38867, done.
remote: Compressing objects: 100% (48/48), done.
remote: Total 38867 (delta 18), reused 43 (delta 16), pack-reused 38802
Receiving objects: 100% (38867/38867), 11.80 MiB | 4.41 MiB/s, done.
Resolving deltas: 100% (24486/24486), done.
[yamato@master]/tmp% cd ctags
[yamato@master]/tmp/ctags% git checkout peg-based-java-parser
Branch 'peg-based-java-parser' set up to track remote branch 'peg-based-java-parser' from 'origin'.
Switched to a new branch 'peg-based-java-parser'
```

How to try the new java parser:
```
[yamato@master]~/var/ctags-peg% ./ctags --fields=+l -o - --languages=+NewJava --language-force=NewJava foo.java
A	foo.java	/^    A;$/;"	e	language:NewJava	enum:e
e	foo.java	/^public enum e {$/;"	g	language:NewJava
getString	foo.java	/^    public String getString() {$/;"	m	language:NewJava	enum:e
```

TODO:

* Improve or modify packcc to make ctags pass the test case parser-java.r/accented-latin1-identifiers.java.d,
* Build a ctags executable on non-Autotools platform,
* Support Java-1.9 (See http://www.romanredz.se/Mouse/index.htm#parsing), and
* Add a parser for peg itself.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 *.TMP
 *.tmp
 *~
+/packcc
 .deps
 .dirstamp
 .DS_Store
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "misc/packcc"]
+	path = misc/packcc
+	url = https://github.com/masatake/packcc.git
+	branch = ctags
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,14 +79,17 @@ nodist_ctags_SOURCES = $(REPOINFO_HEADS)
 BUILT_SOURCES = $(REPOINFO_HEADS)
 CLEANFILES = $(REPOINFO_HEADS)
 $(REPOINFO_OBJS): $(REPOINFO_SRCS) $(REPOINFO_HEADS)
+repoinfo_verbose = $(repoinfo_verbose_@AM_V@)
+repoinfo_verbose_ = $(repoinfo_verbose_@AM_DEFAULT_V@)
+repoinfo_verbose_0 = @echo REPOINFO "  $@";
 if BUILD_IN_GIT_REPO
 GEN_REPOINFO = $(srcdir)/misc/gen-repoinfo
 $(REPOINFO_HEADS): FORCE
-	$(GEN_REPOINFO) $@
+	$(repoinfo_verbose)$(GEN_REPOINFO) $@
 FORCE:
 else
 $(REPOINFO_HEADS):
-	echo > $@
+	$(repoinfo_verbose)echo > $@
 endif
 
 if RUN_OPTLIB2C
--- a/Makefile.am
+++ b/Makefile.am
@@ -99,8 +99,10 @@ optlib2c_verbose = $(optlib2c_verbose_@AM_V@)
 optlib2c_verbose_ = $(optlib2c_verbose_@AM_DEFAULT_V@)
 optlib2c_verbose_0 = @echo OPTLIB2C "  $@";
 OPTLIB2C = $(srcdir)/misc/optlib2c
-%.c: %.ctags $(OPTLIB2C) Makefile
+SUFFIXES += .ctags
+.ctags.c:
 	$(optlib2c_verbose)$(OPTLIB2C) $< > $@
+$(OPTLIB2C_INPUT): $(OPTLIB2C) Makefile
 endif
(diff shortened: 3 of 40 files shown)


b4n:
Just a few thoughts and silly questions:

* Why did you choose packcc instead of the more widespread lex/yacc, especially if packcc is not maintained?  Similarly, if you think PEG is a better solution anyway, why the choice of packcc specifically? I have no experience there, but for example [peg](http://piumarta.com/software/peg/) is packaged in Debian.  Not sure what's available in which distributions, but from a quick search packcc is not packaged in any distribution I could find.
* Using parser generators is attracting, but an interesting aspect of CTags for me always has been dealing with somewhat incorrect input, for better support of files being under edition (for using ctags as a parser for IDEs). How well can formally defined parser deal with invalid input?

masatake:
https://github.com/antlr/antlr4 and https://github.com/antlr/grammars-v4 are originally I would like to use.
However, it seems that code generator for C language target is not available yes. So I throw away this plan.
I don't have enough strong reason not to use bison and flex. 
I don't say my choice is the best. I choose packcc in this pull request for following 3 reasons.

1. build system related concern
packcc is very small. So we can put its source code to ctags source tree. It is written in C and don't require any more libraries. I expected integrating running packcc to our build system is easy even on Windows.
 (However, the result is much harder than I expected.)
I have wasted my time much for Windows related build script. So I didn't want to use external tool like flex and bison. 

2. Hackable
 I expect it is enough small to understand the source code. So we can modify it for our purpose.
https://github.com/enechaev/packcc/commit/71d51c53a33747b016637e7172eedc47dee8e94e is very impressive to me.

3. Existing Java parser
I felt improving Java parser in c.c is hard. As @pragmaware did, I thought I would like to rewrite Java parser when I got two critical bug reports about Java parser.
I agree that there are only a few example of peg. However,  Luckily, I found a peg file for Java grammer.
I tried it to use ctags. It take just a half day, peg based Java parser works. I got very good impression on packcc.

>     Using parser generators is attracting, but an interesting aspect of CTags for me always has been dealing with somewhat incorrect input, for better support of files being under edition (for using ctags as a parser for IDEs). How well can formally defined parser deal with invalid input?

Good point. peg based parser cannot handle invalid input. So it is not suitable for your purpose.
As I wrote before, I'm not a professional programmer. The most of time, I'm reading source code of products.
In my situation, I can expect source code files are stable. None will edit the file during my reading.
So I don't have to think about invalid input in my use-case.
This is very different from your side.

peg parser cannot handle invalid input. In stead, we can retrieve  interesting information easily.
What I'm thinking about is reference tags. I introduced reference tags related to  #569.
However, I have limited to use the feature. Implementing reference tags in a language will take much time. However, I cannot convince whether it is useful or not. If a parser written in peg, implementing reference tags in the parser doesn't take much time. 

Anyway, merging this pull request has no side effect. The java parser written in peg is named "NewJava".
It is disabled by default. A user can use the new parser only when using  new magical option `--_pretend-NewJava=Java`.

Thank you for reading.

masatake:
New Java parser should not use packcc.
As @b4n wrote, CTags has been dealing with somewhat incorrect input, for better support of files being under edition. I should not ignore this for popular real languages like Java.
In addition peg based parser is too slow. It is not acceptable.

So I will withdraw this PR.
However, I would like to use packcc in more minor languages.
My current target is IDL of varlink (https://varlink.org/).

PR #1848, opened 2018-08-22
tex: Fix reporting non-alphanumeric tag contents

### LINKED ISSUE #1846: Umlaute in tags
The name of the parser: Latex

```
$ ctags --options=NONE mwe.tex
```

The content of input file:

```C
\documentclass[11pt]{article}

\begin{document}
\section{ABC}
\section{ZÖZÜZÄZßZ}
\end{document}
```

```
ABC	mwe.tex	/^\\section{ABC}$/;"	s
Z   Z   Z   Z   Z	mwe.tex	/^\\section{ZÖZÜZÄZßZ}$/;"	s
...
```

The tags output you expect:

```
ABC	mwe.tex	/^\\section{ABC}$/;"	s
ZÖZÜZÄZßZ	mwe.tex	/^\\section{ZÖZÜZÄZßZ}$/;"	s
...
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(2258b24b), Copyright (C) 2015 Universal Ctags Team
  Compiled: Aug 19 2018, 13:11:28
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json,
```

How do you get ctags binary:

(
AUR: universal-ctags-git
)

Hello! I wanted to ask whether I miss some option which disables utf8-encoding in the tag names? I already tried the --input-encoding=utf-8 option. Im using vim and tagbar to write latex, this was the first time I had an umlaut in a tag. Thanks for your help!

b4n:
Fixes #1846.


parsers/tex.c
```diff
@@ -35,7 +35,7 @@
 #define isType(token,t)		(bool) ((token)->type == (t))
 #define isKeyword(token,k)	(bool) ((token)->keyword == (k))
 #define isIdentChar(c) \
-	(isalpha (c) || isdigit (c) || (c) == '$' || \
+	(isalpha (c) || isdigit (c) || (((unsigned char) c) > 0x80) || (c) == '$' || \
```
hadrielk: Does this work with `À`? I think that's the two bytes `0xC380` in utf-8, so it would fail this check for the second byte. So I think it should be ` >= 0x80`.

b4n: Oops, you're totally right, just a typo on my part. Fixed -- and it even improved my test results further, as they used some 0x80 bytes.

mightymicha:
@b4n I am definitely no tex expert, but I never saw someone using the \label inside of section names. Normally, if you want to reference sections, you do something like `\section{Foo} \label{bar}`. See [here](https://en.wikibooks.org/wiki/LaTeX/Labels_and_Cross-referencing#Sections). I doubt that there other use cases for using commands inside a section command.

masatake:
@b4n, are you o.k. to merge this?

PR #1849, opened 2018-08-22
vstring: Use memcpy() instead of strncpy()

b4n:
We don't need the extra strncpy() semantic in stringCat() as we always
know the length we want, so it's slightly faster to use memcpy(), about
2% in some cases.

Additionally, this would improve support for NUL bytes embedded in a
vString, avoiding an extra truncation.  However, no consumer currently
rely on NUL byte support so it currently has no impact on this end.

### DIFF
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -107,7 +107,7 @@ static void stringCat (
 	if (string->length + length + 1 > string->size)
 		vStringResize (string, string->length + length + 1);
 
-	strncpy (string->buffer + string->length, s, length);
+	memcpy (string->buffer + string->length, s, length);
 	string->length += length;
 	vStringPut (string, '\0');
 }


main/vstring.c
```diff
@@ -107,7 +107,7 @@ static void stringCat (
 	if (string->length + length + 1 > string->size)
 		vStringResize (string, string->length + length + 1);
 
-	strncpy (string->buffer + string->length, s, length);
+	memcpy (string->buffer + string->length, s, length);
```
hadrielk: This *feels* dangerous, but yeah I searched through the code and the only thing that looked like it might have a behavior change with this is `readLine()`, which does this:

```
		if (File.allLines)
			vStringCat (File.allLines, File.line);
```

...and it got that `File.line` from `fgets()` which I believe does copy NUL bytes from the file... but the change in behavior *should* be benign. 🤞

Issue #1851, opened 2018-08-22
Using '--[extras|kinds|fields]-<LANG>=...', for a langdef before adding regexes yields no scoping


hadrielk:
If you define a new language using `--langdef=Foo`, and add a kind/field/extra to it called `bar`, and then do a `--extras-Foo=+{bar}`, or `--kinds-Foo=+{bar}` or `--fields-Foo=+{bar}`, and you do that *before* you define the regexes for it, then the resultant tags no longer have the scope information.
If you do those `--[extras|kinds|fields]-<LANG>=...` commands *after* you define the regexes, then it works fine.

The reason for this is that `initializeParser()` is invoked when `--[extras|kinds|fields]-<LANG>=...` options are processed, and `initializeParser()` checks to see if any of the regexes use scopes to determine whether to enable Cork or not. Since this happens *before* the regex options are processed, the answer is no, so Cork is disabled and scoping info is never saved/used.

I think the correct fix is to make the `lazyInitialize()` function re-check `hasLanguageScopeActionInRegex()` and if so then set `useCork = true`. I'll submit a PR doing that if no one objects.

PR #1853, opened 2018-08-24
Check again if we need to use cork after initializing parsers

hadrielk:
Fix for issue #1851.

### DIFF
--- a/Units/mtable-simple-with-table-extending.d/args.ctags
+++ b/Units/mtable-simple-with-table-extending.d/args.ctags
@@ -6,6 +6,8 @@
 --kinddef-X=v,var,variables
 --kinddef-X=f,fun,functions
 
+--kinds-X=+{var}
+
 --_tabledef-X=main
 --_tabledef-X=comment_sharp_sa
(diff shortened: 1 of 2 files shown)


Units/mtable-simple-with-table-extending.d/args.ctags
```diff
@@ -6,6 +6,8 @@
 --kinddef-X=v,var,variables
 --kinddef-X=f,fun,functions
 
+--kinds-X=+{var}
+
```
masatake: Can I ask you to make a separate test case?
This test case will be important when I rewrite the code for parser initialization.
I think this should go to Tmain.

hadrielk: Yes I'll make it a separate test case and update the PR.

masatake:
@hadrielk, could you look at #1858 ?

masatake:
#1858 closes the original issue.
@hadrielk, thank you for taking your time for reporting the original issue and this pull request.

Issue #1852, opened 2018-08-23
Disabling a tag kind can screw up scoping in optlib regex and some built-in parsers


hadrielk:
If a tag would have normally been added to the scope info, but you disable that tag kind (`--kinds-<LANG>=-...`) for an optlib language using regex, then it won't be added to the scope info anymore, for other later tags.

For example if the tag was for namespaces, and normally it would push a new namespace scope... then if that tag's kind is disabled the namespace won't be pushed and anything tagged later that should have that namespace's scoping info, won't have it; and instead those later tags won't have *any* scope info.

The reason for this is the Cork API queues tags and uses their queued position as the scope locator. But if a tag kind is disabled, then lregex's `matchTagPattern()` doesn't invoke `makeTagEntry()` so it won't be enqueued.

Most built-in parsers don't have this problem, because they either don't use the Cork API for scoping info, or they enqueue *all* tags, and leave the filtering of disabled kinds up to cork to handle (which it does when it uncorks). That seems like the correct thing for lregex to do as well, I think.

Also, other non-regex built-in parsers might have the same problem. The `makeSimpleTag()` checks the kind and does not enqueue it to cork if the kind is disabled, so any built-in parser using `makeSimpleTag()` and saving the returned integer will have the same problem; though I did not verify this. It looks like: asm, automake, diff, rpmspec, sh, tcl, and vim parsers all might have this issue. I think  `makeSimpleTag()` should not be doing the check, but just enqueuing.

PR #1855, opened 2018-08-24
Disabled kinds for builtin parsers should still enqueue tags

hadrielk:
See issue #1852.
The `makeSimpleTag()` function shouldn't be filtering out tags, because
that can screw up scoping. Note the tags still won't be written to the
file - they will be filtered out later, by `isTagWritable()`.

### DIFF
--- a/Units/parser-tcl.r/namespace-disabled.d/args.ctags
+++ b/Units/parser-tcl.r/namespace-disabled.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=no
+--extras=+q
+--fields=+eE
+--kinds-Tcl=-n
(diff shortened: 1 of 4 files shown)


hadrielk:
One side-note: there's another function, `makeSimpleRefTag()`, right after `makeSimpleTag()`, that seems to also do something similar - it does not enqueue if the extra flag or role kind is disabled for the tag. Currently only one parser (Automake's) saves the return value of `makeSimpleRefTag()`, and it doesn't seem like it breaks anything for it if `makeSimpleRefTag()` returns without enqueuing the tag. But perhaps I should change `makeSimpleRefTag()` as well anyway?

masatake:
I'll take a look at this next. Please, wait for a while.

masatake:
> But perhaps I should change makeSimpleRefTag() as well anyway?

Yes. No test case is needed for the change for makeSimpleRefTag().
Thank you.

----

BTW, I would like to introduce you one secret (if you don't find it yet).
Sometimes we have to introduce a change that is difficult for us to test with a crafted command line or a built-in parser. In that case, you can use CTagsSelfTest parser defined at the end of main/parser.c.
I guess you may want to use the CTagsSelfTest parser in the future.

PR #1826, opened 2018-08-14
Update the docs: extending.rst and optlib.rst sections

hadrielk:
This is a first pass at updating the optlib.rst doc (and minor stuff for extending.rst).
There's still more to do, but I thought this is a big enough change to get some feedback on.
There are also a few open questions I had, which I added as comments in optlib.rst, marked `.. Q:`.
Feedback would be great. Even if it's: "this is crap".

PR #1856, opened 2018-08-24
Add options/flags to help users develop multi-table regex parsers

hadrielk:
This PR adds a couple new options and flags to help users debug and optimize optlib multi-table regex parsers.

Documentation for them will be added to `optlib.rst` after PR #1826 is merged into master. (otherwise I'll get conflicts with that one)

### DIFF
--- a/main/lregex.c
+++ b/main/lregex.c
@@ -130,6 +130,12 @@ typedef struct {
 
 	char *pattern_string;
 	char *unescaped_pattern_string;
+
+	struct {
+		errorSelection selection;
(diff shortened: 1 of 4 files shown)


hadrielk:
Thinking about this change on my drive home, I realized I shouldn't have bothered with the new option for copying the tables - instead I'll just separate the statistics from the pattern stuff in the tables, so the state are kept per-table instead of per-pattern. I was trying to avoid that originally, but in hindsight it would be cleaner.

PR #1858, opened 2018-08-27
main: fill parser->useCork only once at doesParserUseCork()

### LINKED ISSUE #1851: Using '--[extras|kinds|fields]-<LANG>=...', for a langdef before adding regexes yields no scoping

masatake:
Fix #1851.

This fix is based on #1853 submitted by @hadrielk.

The change of #1853 gives the chance ctags to fill the parser->useCork field one more time.

This change unifies all the code for filling the field to doesParserUseCork() function.
Mine is a bit easier to understand the code.


masatake:
@hadrielk, could you see or try this?
I thnks #1851 is fixed with this change.

PR #1863, opened 2018-08-29
Remove dead code related to xcmd feature

masatake:


masatake:
Let's see what happens.

PR #1866, opened 2018-09-03
Robot: remove empty expression .{0} in parsers/robots.c to fix unit test tmain and units on FreeBSD

derekschrock:
Removing empty expression .{0} fixes unit test on FreeBSD due to FreeBSD libc regcomp not liking the following (a|b|.{0}).   Even though .{0} is valid removing it shouldn't create any issues.


masatake:
Thank you.

PR #1872, opened 2018-09-13
unit: accept input_[0-9].* and input_[0-9]_*.* as input file names

masatake:
Java language expects a class and the file where the class is defined
have a same name. On the othre hand, the language doesn't allow to use
`-` in a class name.

This means that we cannot write a test input file as a valid Java program.
This change solves this situation.

### DIFF
--- a/misc/units
+++ b/misc/units
@@ -752,8 +752,8 @@ run_dir ()
 	    continue
 	fi
 
-	extra_inputs=$(for extra_tmp in $(dirname $input)/input-[0-9].* \
-					$(dirname $input)/input-[0-9]-*.* ; do
+	extra_inputs=$(for extra_tmp in $(dirname $input)/input[-_][0-9].* \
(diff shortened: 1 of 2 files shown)


misc/units
```diff
@@ -752,8 +752,8 @@ run_dir ()
 	    continue
 	fi
 
-	extra_inputs=$(for extra_tmp in $(dirname $input)/input-[0-9].* \
-					$(dirname $input)/input-[0-9]-*.* ; do
+	extra_inputs=$(for extra_tmp in $(dirname $input)/input{-,_}[0-9].* \
+					$(dirname $input)/input{-,_}[0-9]{-,_}*.* ; do
```
b4n: why not use `[-_]`?

masatake: Oh, I forgot I can use the notation even in glob match.
Thank you. I will update my change.

Issue #1885, opened 2018-09-23
Compiler warnings


blueyed:
```
  CC       parsers/ctags-ada.o     
parsers/ada.c: In function ‘storeAdaTags’:
parsers/ada.c:2126:9: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
         strncpy(currentScope, parentScope, strlen(parentScope));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/ada.c:2128:9: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
         strncpy(&currentScope[strlen(parentScope) + 1], token->name,
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                 strlen(token->name));
                 ~~~~~~~~~~~~~~~~~~~~
```

```
  CC       parsers/ctags-html.o
parsers/html.c: In function ‘readTag.part.3’:
parsers/html.c:438:6: warning: ‘lineOffset_tmp[1]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
      makePromise ("JavaScript", startLineNumber, startLineOffset,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          endLineNumber, endLineOffset, startSourceLineNumber);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/html.c:438:6: warning: ‘line_tmp[1]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

```
OPTLIB2C   optlib/RSpec.c                                         
Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.32), passed through in regex; marked by <-- HERE in m/^--langdef=([^\{]+)((({ <-- HERE base=([^\{]+)})({(dedicat
ed|shared|bidirectional)})?)?)/ at ./misc/optlib2c line 46.   
[...]

PR #1886, opened 2018-09-23
Ada: suppress compiler warning

masatake:
See #1885.


masatake:
@b4n, thank you for reviewing this.

Issue #1897, opened 2018-10-02
Objective-C++ parser output strips leading underscore from C function and C++ method names.


siegel:
The name of the parser: Objective-C++

The command line you used to run ctags:

```
$ ctags --options=NONE foo.mm
```

The content of input file:

```cpp
static
void _MyFunction(void)
{

}

static
void MyOtherFunction(void)
{

}

class Foo
{
	int _Mumble(void);
	int Grumble(void);
}

int	Foo::_Mumble(void)
{

}

int Foo::Grumble(void)
{

}
```

The tags output you are not satisfied with:

```
Grumble	foo.mm	/^int Foo::Grumble(void)$/;"	f
Mumble	foo.mm	/^int	Foo::_Mumble(void)$/;"	f
MyFunction	foo.mm	/^void _MyFunction(void)$/;"	f
MyOtherFunction	foo.mm	/^void MyOtherFunction(void)$/;"	f
```

The tags output you expect:

```
Grumble	foo.mm	/^int Foo::Grumble(void)$/;"	f
_Mumble	foo.mm	/^int	Foo::_Mumble(void)$/;"	f
_MyFunction	foo.mm	/^void _MyFunction(void)$/;"	f
MyOtherFunction	foo.mm	/^void MyOtherFunction(void)$/;"	f
```

Remarks: the basic problem is that the underscore is being stripped from the tag name for "_Mumble" and "_MyFunction" in the tags output, but only for Objective-C++ files.

However, you can see that the search pattern contains the correct name of the symbol (since that's just an excerpt of the line containing the symbol).

If I change the file's extension to "cpp", the C++ tags output includes underscores in the generated tag names, which is what I would have expected it to do for Objective-C++.

The version of ctags:

```
Universal Ctags 0.0.0(d0807887), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Oct  2 2018, 15:26:35
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames
```

How do you get ctags binary:

macosx build from source, clone of universal-ctags/ctags.

siegel:
Here's a patch that I think will fix this:

```
286c286
< 	if (isAlpha (*st->cp))
---
[...]

PR #1898, opened 2018-10-02
Fix for issue #1897 (generated tags omitting leading underscore).

siegel:


masatake:
Can I ask you to add a test case under Units/parser-objectivec.r/ ?
Please, see http://docs.ctags.io/en/latest/units.html#how-to-write-a-test-case .

masatake:
Merged this directly to master after modifying. Thank you.

PR #1899, opened 2018-10-02
tex: Fix infinite loop parsing an identifier at EOF

### LINKED ISSUE #1896: ctags -R hangs when encountering a TeX file which is missing a line break at EOF
Summary: when using `ctags -R` to recursively process a directory, `ctags` will hang when it encounters a TeX source file in which the last line of a file does not end with a line break.

I have attached the example file.

Steps:

- place the attached TeX file in a directory by itself. (There may be other files there, they will not matter.) [junk.txt](https://github.com/universal-ctags/ctags/files/2439438/junk.txt)

- change its filename extension from `txt` to `tex` so that it is recognized as a TeX file.

- check and make certain that the last line in the file does not end with a line break.

- `cd` to the directory containing the TeX file.

- `ctags -R .`

Expected:

=> ctags runs and generates an output file.

Actual:

=> ctags hangs, I have attached the output of a macOS `sample` command: 
[ctags_2018-10-02_153200_k1Fb.sample.txt](https://github.com/universal-ctags/ctags/files/2439451/ctags_2018-10-02_153200_k1Fb.sample.txt)

Regression:

If `-R` is *not* specified on the command line, ctags does not hang on this file.

`ctags --version` reports:

```
Universal Ctags 0.0.0(d0807887), Copyright (C) 2015 Universal Ctags Team
  Compiled: Oct  2 2018, 15:26:35
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames
```

b4n:
This was introduced in 5d5afe64ca5517c50897164f72db174428961e9f which started accepting as an identifier character any byte with a value that cast as unsigned char is greater than or equal to 0x80.  The common value for the EOF constant being -1, which cast to unsigned char gives a value of 0xff, hence leading to accepting EOF as an identifier character.

Fix the loop accordingly to explicitly guard against EOF.

Fixes #1896.

### DIFF
--- a/parsers/tex.c
+++ b/parsers/tex.c
@@ -35,7 +35,7 @@
 #define isType(token,t)		(bool) ((token)->type == (t))
 #define isKeyword(token,k)	(bool) ((token)->keyword == (k))
 #define isIdentChar(c) \
-	(isalpha (c) || isdigit (c) || (((unsigned char) c) >= 0x80) || (c) == '$' || \
+	(isalpha (c) || isdigit (c) || (c) >= 0x80 || (c) == '$' || \
 		(c) == '_' || (c) == '#' || (c) == '-' || (c) == '.' || (c) == ':')
 
 /*
@@ -264,7 +264,7 @@ static void parseIdentifier (vString *const string, const int firstChar)
 	{
 		vStringPut (string, c);
 		c = getcFromInputFile ();
-	} while (isIdentChar (c));
+	} while (c != EOF && isIdentChar (c));
 
 	if (c != EOF)
 		ungetcToInputFile (c);		/* unget non-identifier character */


masatake:
Instead, how about defining isIdentChar like:

```
#define isIdentChar(c) \
	(isalpha (c) || isdigit (c) || (c >= 0x80) || (c) == '$' || \
		(c) == '_' || (c) == '#' || (c) == '-' || (c) == '.' || (c) == ':')
```
?

b4n:
@masatake this works only if `c` is an `int` or similar *and* that it's value isn't one of a `signed char`; if it's effectively a `signed char` it won't work as anything above 0x7F will be negative.
Normally the proper handling in such cases is to use an int, check it's not EOF and then cast to `unsigned char` (see for example the definitions of the *ctype.h*'s `is*()` family: *[…] which must have the value of an unsigned char or EOF […]*).
The Right Thing™ would likely be either:
1) assume `isIdentChar()` doesn't handle `EOF` (what I did here, and the rest of the code is safe against that)
2) check if `c` isn't `EOF` in `isIdentChar()`, and then perform other checks casting `c` to an unsigned char.

Actually in our case we could just drop the cast indeed, because `getcFromInputFile()` does return character values as `unsigned char`s inside an `int`, so it'd be safe.  I also just checked the C standard, and `EOF` is indeed guaranteed to be negative (not necessarily `-1`, but negative), so `>= 0x80` would do.
I'd be in favor of actually doing both: remove the cast, and keep the explicit `EOF` check which also fits nicely with the similar check at the end of that function.

b4n:
@masatake PR updated

masatake:
Please, give me time. I would like to understand the change completely. I will review the other parsers based on the understanding.

b4n:
@masatake No problem, take your time.

masatake:
LGTM. Thank you.

masatake:
Thank you.

PR #1908, opened 2018-10-07
PuppetManifest: Add documentation on how to verify puppet files.

ahakanbaba:
Ideally the `make check` should execute this.
Until then let's document the steps to manually verify the input files.

PR #1909, opened 2018-10-07
Unit test input files are verified during `make verify-units-inputs` 

ahakanbaba:
This is an early work in progress. 

I am sending this PR out to discuss the implementation approach for incorporating the verification of the unit test input files to make check. 

I have used some makefile metaprogramming. http://make.mad-scientist.net/the-eval-function/ I am by no means a makefile expert. So let me know please if there is a better approach to this. 

Some missing pieces: 

1. The CATEGORIES and UNITS env vars are not taken into consideration. So every time all puppet files are verified regardless. 
2. The circileci config file needs to be updated to install puppet dependencies. The puppet command will fail in the containers. 
3. The input.pp files have not been fixed to pass the puppet syntax checking. 
4. If we want to go into this direction, the testing.mak should possibly be further parameterized or more meta-programming should be used. Right now the implementation is specific to the puppetManifest files. 
5. Better variable names can be used in testing.mak

### DIFF
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -88,28 +88,41 @@ slap: $(CTAGS_TEST)
 		--with-timeout=$(TIMEOUT)"; \
 	$(SHELL) $${c} $(srcdir)/Units
 
+TEST_DIRS =
+
 # TODO: Possibly the bmake does not support the metaprogramming similar to gnu make.
 if !USING_BMAKE
+
+# TODO: A disgusting hack
+# TODO: Explain more why this is necessary
+set = $(eval $(1) = $(2))
+
+# $(1): Language Name. For example: PUPPET
+# $(2): Extension. For example: pp
+# $(3): Directory Name. For example: parser-puppetManifest.r
+define VERIFY_GIVEN_UNITS_TEST_DIR
+
 # Find the test directories where ctags is expected to succeed. Only tests with
 # an expected.tags file present are required to have valid input.
-PUPPET_TEST_DIRS = $(foreach path, \
-    $(shell find $(srcdir)/Units/parser-puppetManifest.r -name expected.tags), \
-    $(shell dirname $(path)))
-VERIFY_PUPPET_TEST_DIRS_TARGETS := $(PUPPET_TEST_DIRS:%=%/.input.pp.verified)
+$$(call set,$(1)_TEST_DIRS,$$(foreach path, \
+    $$(shell find $$(srcdir)/Units/$(3) -name expected.tags), \
+    $$(shell dirname $$(path))) )
+$$(call set,VERIFY_$(1)_TEST_DIRS_TARGETS,$$($(1)_TEST_DIRS:%=%/.input.$(2).verified))
 
-#TODO: We shuold make this an empty target rather than a phony target
-define VERIFY_ONE_PUPPET_TEST_DIR
-$(1)/.input.pp.verified: $(1)/input.pp
-	puppet apply --noop $$< && \
-	touch $$@
+verify-units-inputs: $$(VERIFY_$(1)_TEST_DIRS_TARGETS)
+
+define VERIFY_ONE_$(1)_TEST_DIR
+$$(1)/.input.pp.verified: $$(1)/input.pp
+	puppet apply --noop $$$$< && \
+	touch $$$$@
 endef
 
-$(foreach puppet_test_dir,$(PUPPET_TEST_DIRS),$(eval $(call VERIFY_ONE_PUPPET_TEST_DIR,$(puppet_test_dir))))
+$$(foreach $(1)_test_dir,$$($(1)_TEST_DIRS),$$(eval $$(call VERIFY_ONE_$(1)_TEST_DIR,$$($(1)_test_dir))))
 
-#
-# verify-units-inputs Target
-#
-verify-units-inputs: $(VERIFY_PUPPET_TEST_DIRS_TARGETS)
+endef # define VERIFY_GIVEN_UNITS_TEST_DIR
+
+$(eval $(call VERIFY_GIVEN_UNITS_TEST_DIR,PUPPET,pp,parser-puppetManifest.r))
+
 
 endif  # if !USING_BMAKE
 
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -112,7 +112,7 @@ verify-units-inputs: $(VERIFY_PUPPET_TEST_DIRS_TARGETS)
 #
 # UNITS Target
 #
-units: $(CTAGS_TEST) verify-units-inputs
+units: $(CTAGS_TEST)
 	$(V_RUN) \
 	if test -n "$${ZSH_VERSION+set}"; then set -o SH_WORD_SPLIT; fi; \
 	if test x$(VG) = x1; then		\
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -89,7 +89,7 @@ slap: $(CTAGS_TEST)
 	$(SHELL) $${c} $(srcdir)/Units
 
 # TODO: Possibly the bmake does not support the metaprogramming similar to gnu make.
-ifneq ($(MAKE),bmake)
+if !USING_BMAKE
 # Find the test directories where ctags is expected to succeed. Only tests with
 # an expected.tags file present are required to have valid input.
 PUPPET_TEST_DIRS = $(foreach path, \
@@ -110,7 +110,8 @@ $(foreach puppet_test_dir,$(PUPPET_TEST_DIRS),$(eval $(call VERIFY_ONE_PUPPET_TE
 # verify-units-inputs Target
 #
 verify-units-inputs: $(VERIFY_PUPPET_TEST_DIRS_TARGETS)
-endif  # ifneq ($(MAKE),bmake)
+
+endif  # if !USING_BMAKE
 
 #
 # UNITS Target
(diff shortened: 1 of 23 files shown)


masatake:
02188ce looks good to me but could you rename the file to README.md(?) or something reflecting the syntax of markup language.

0c50a74 is very interesting... I would like to take over this item. It should not be integrated to units target.

masatake:
Oh, sorry, I should make a comment about 02188ce on #1908.

ahakanbaba:
Unfortunately having bmake build problems in fedora build

```
bmake: "/root/universal-ctags/Makefile" line 7323: Need an operator
bmake: "/root/universal-ctags/Makefile" line 7327: Need an operator
bmake: "/root/universal-ctags/Makefile" line 7329: Need an operator
```

The line of the makefile in the fedora container looks like this 
```
7319                 --with-timeout=$(TIMEOUT)"; \
7320         $(SHELL) $${c} $(srcdir)/Units
7321
7322 #TODO: We shuold make this an empty target rather than a phony target
7323 define VERIFY_ONE_PUPPET_TEST_DIR
7324 $(1)/.input.pp.verified: $(1)/input.pp
7325         puppet apply --noop $$< && \
7326         touch $$@
7327 endef
7328
7329 $(foreach puppet_test_dir,$(PUPPET_TEST_DIRS),$(eval $(call VERIFY_ONE_PUPPET_TEST_DIR,$(puppet_test_dir))))
7330
7331 #
7332 # verify-units-inputs Target
7333 #
```

I would not be surprised if the bmake does not have the define (metaprogramming) feature . Do you know @masatake  ?

ahakanbaba:
@masatake This PR is about 70% there of what I had originally in mind. 

1. The metaprogramming in the makefile needs to be further parameterized. Right now it has hardcoded strings with the `puppetManifest.r` in them . It *should* be possible to have only 1 more generic define block. Then generate input language checkers for many other languages with using only that block.  
2. The build steps and the output of the make verify-units-inputs can be improved, maybe. 
3. Also if you have better names, I am open to renaming. 

Regardless, I think this PR is mature enough to be reviewed. Let me know if this approach is acceptable to you.

ahakanbaba:
The error messages fixed from the input.pp files are: 

```
Error: Could not find node statement with name 'default' or 'mbp-005063.ad.whirl.net, mbp-005063.ad.whirl, mbp-005063.ad, mbp-005063' on node mbp-005063.ad.whirl.net
Error: Could not parse for environment production: The operator '+=' is no longer supported. See http://links.puppet.com/remove-plus-equals (file: /Users/hbaba/box/src/universal-ctags/ctags/Units/parser-puppetManifest.r/puppet-append.d/input.pp, line: 4, column: 10) on node mbp-005063.ad.whirl.net
Error: Could not parse for environment production: The parameter $name redefines a built in parameter in the 'define' expression (file: /Users/hbaba/box/src/universal-ctags/ctags/Units/parser-puppetManifest.r/puppet-fqdefinition.d/input.pp, line: 1, column: 25) on node mbp-005063.ad.whirl.net
Error: Class 'one' is already defined (file: /Users/hbaba/box/src/universal-ctags/ctags/Units/parser-puppetManifest.r/puppet-multipleclass.d/input.pp, line: 1); cannot redefine (file: /Users/hbaba/box/src/universal-ctags/ctags/Units/parser-puppetManifest.r/puppet-multipleclass.d/input.pp, line: 5) on node mbp-005063.ad.whirl.net
Error: Parameter mode failed on File[/tmp/scopetest]: The file mode specification must be a string, not 'Integer' (file: /Users/hbaba/box/src/universal-ctags/ctags/Units/parser-puppetManifest.r/puppet-scopetest.d/input.pp, line: 5) 
```

makefiles/testing.mak
```diff
@@ -88,6 +88,31 @@ slap: $(CTAGS_TEST)
 		--with-timeout=$(TIMEOUT)"; \
 	$(SHELL) $${c} $(srcdir)/Units
 
+# TODO: Possibly the bmake does not support the metaprogramming similar to gnu make.
```
b4n: I'd rather avoid using non-portable Make stuff if possible, assuming the rest of our makefiles work with BSD make.  Here it seems fairly easy to simply make it the content of the target itself, it doesn't seem very important to build target, nor even only build once -- rebuilding seems unlikely and innocent enough.

So, I guess I'd just do something like this (not tested at all, just for the argument):
```make
verify-units-inputs: verify-units-inputs-puppet

verify-units-inputs-puppet:
    find $(srcdir)/Units/parser-puppetManifest.r -name expected.tags | $(SED) 's%/[^/]*$$%%' | while read -r dir; do \
        puppet apply --noop "$$dir/input.pp" || exit 1; \
    done

ahakanbaba: I see couple of more problems with this  approach in the comment. 

1) One cannot use the make job control mode (`-j`) for `verify-units-inputs`. All `input.XX` file verification has to be done serially. 
 2) What @b4n mentions: Cannot avoid re-execution. What I have seen, re-execution can be quite expensive. The input language toolchain does not necessarily have fast startup time. In this change, puppet runs take several minutes if executed serially. Consider the need to run interpreters, compilers for each `input.xx` file. Hence, for fast test and development time both (1) and (2) are important. 
2) The complexity of the testing.mak is bound to grow. ctags supports 10's of different input languages. Each having their `verify-untis-inputs-<language_name>` target is not easy to maintain. Compared to that, the conditional of `if !USING_BMAKE ` could be acceptable. 

These three are important drawbacks. 

With using metaprogramming, we gain on all 3 of them. (See my next push to this PR) . What we lose is missing `verify-units-inputs` target in BSD builds. That will not prevent BSD platforms from building and using ctags.  Developers on BSD only won't be able to execute `make verify-units-intput` locally, if they are submitting PRs from BSD platforms. 

I have not done development on BSD before. But that does not seem like a big blocker either https://unix.stackexchange.com/a/127955/212862

ahakanbaba:
In the latest version of this PR (16 commits) we enable input language checking while 

1) Supporting parallel execution with job control mode in make (`-j`).
2) Verifying only the input.XX files that have changed since the last execution of `make verify-units-inputs`
3) Start testing new input languages with only a one line change similar to [this](https://github.com/universal-ctags/ctags/pull/1909/files#diff-a891f9ca940820ee450df097dc2b14c1R131)

With this PR, we check not only the puppet files but also the json files.

ahakanbaba:
@masatake  @bn4
Any further comments on this ?

masatake:
I would like to move the logic to misc/units though it makes running parallel impossible.
I will work on this item as the next task. I need this feature for #1847.
NewJava parsre expects the input is syntatically correct. However, the test cases under parsers-java.r are not. I need the infrastructure veryfing input for Java.

I will borrow many codes from this pull request. However, I will design the infrastructure very differently based on my experience implementing misc/units, the backend of make units.

ahakanbaba:
Let me know if you need any help @masatake.
 
In the meantime: It may be throwaway but the single line to support java tests in this testing.mak looks like [this](https://github.com/ahakanbaba/ctags/pull/2/files#diff-a891f9ca940820ee450df097dc2b14c1R133):

makefiles/testing.mak
b4n: Fair enough, and you're right that this is probably not a critical enough part of the build system that it can't have its own requirements. Also, we can remember that GNU make is fairly portable by itself, and although not everybody likes it, it is usually possible to get it on most systems; AFAIK it's available on all BSDs under `gmake`, yet probably not in a stock installation.

Anyway, given your arguments I only have one concern now: we should check for GNU make, not against BSD make.  There are *many* Make implementations that aren't compatible with GNU make, and most Make implementations don't have those metaprogramming features.  So unless somebody feels like tesing every single Make implementation around and see what exact subset it supports, I strongly suggests we have a whitelist instead of a blacklist, and only have GNU make in it at first.

b4n: PS: moreover, most Make implementations are installed as `make`, not as other aliases like `bmake`, `gmake`,  `pmake` or whatnot.  Checking for the name of `$MAKE` is *not* a solid way of cheking which Make implementation is in use.

ahakanbaba: Thanks for the input. I agree on both aspects. The whitelist-blacklist and checking the name of $MAKE. I will hold off on the implementation, though. @masatake mentioned that he will implement this in /misc/units script. If anything changes, I am more than willing to address your comments in this PR.

masatake:
> Let me know if you need any help @masatake.
Great offering!

I always need help for documentation. Especially #1737.
However, I'm not sure whether you are interested in topics other than PuppetManifest.
If you are interested in improving whole ctags, I would like to know your area of interest and area you are familiar with.

All unclosed issues you opend will take time to solve. All the items are cross-parser issues. (Therefore, they are important.)

ahakanbaba:
> > Let me know if you need any help @masatake.
> > Great offering!
> 
> I always need help for documentation. Especially #1737.
> However, I'm not sure whether you are interested in topics other than PuppetManifest.
> If you are interested in improving whole ctags, I would like to know your area of interest and area you are familiar with.
> 
> All unclosed issues you opend will take time to solve. All the items are cross-parser issues. (Therefore, they are important.)

@masatake my biggest priority now is to improve ctags experience with puppet code. I have opened several tickets that should improve exactly that. I would be very interested to help with that purpose. 

Unfortunately, I do not think I have the bandwidth to contribute to many other exciting areas in ctags.

masatake:
@ahakanbaba, I see.

ahakanbaba:
Closing in favor of https://github.com/universal-ctags/ctags/pull/1921

Issue #1912, opened 2018-10-08
PuppetManifest: ctags command never finishes on long puppet file 


ahakanbaba:
I have a ~40K line long puppet file. ctags never finishes processing it. (I left it at lest 3-4 hours). Ctags is stuck at 100% cpu. 

The example.pp file is attached. 
[example.pp.zip](https://github.com/universal-ctags/ctags/files/2454788/example.pp.zip)

The file looks like this 
```

class some::name::path::segment::list {

  ::some::custom::resource { 'some_name':
    param_1              => 'cccccccccccccccccccc',
    param_2              => "dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd",
    param_3              => 'eeeeeeeeeeeeeeeeeeeeeeeeeeeee',
    param_4              => 'kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk',
    param_5              => 'z',
    param_6              => true,
    param_7              => '"pppppppppppppppppppppppppp"',
    param_8              => 'jjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj',
  }

........ repeat enough times to have 40000 lines 

  }
}

```

A half certain observation: If I remove all the params from all the declarations (param_1 to param_8) ctags does indeed complete. ( The owner of the issue needs to reproduce this. I think I am correct but It is worth to retry this one more time) 

The ctags process looks like this in my mac's top -o cpu` output

```
PID    COMMAND      %CPU TIME     #TH   #WQ  #PORT MEM    PURG   CMPRS  PGRP  PPID  STATE    BOOSTS          %CPU_ME %CPU_OTHRS UID       FAULTS    COW     MSGSENT    MSGRECV    SYSBSD      SYSMACH    CSW        PAGEINS IDLEW    POWE
44256  ctags        99.4 08:01.77 1/1   0    12    3960K  0B     0B     44256 23487 running  *0[1]           0.00000 0.00000    980771313 2090      117     66         25         702         95         139624+    128     0        99.4
```

example.pp is indeed valid puppet file 

```
$ /opt/puppetlabs/bin/puppet apply --noop example.pp
Notice: Compiled catalog for mbp-005063.ad.whirl.net in environment production in 3.31 seconds
Notice: Applied catalog in 0.14 seconds
```

[...]

PR #1913, opened 2018-10-08
PuppetManifet: optimizing for #1912

masatake:
See #1912.
About 3 times faster than before.


ahakanbaba:
I have seen these changes to be extremely useful. 
I would love if this merges. 

Thanks again.

PR #1915, opened 2018-10-11
Fix a typo in docs.

ahakanbaba:
The text refers to the 3rd line of this file

```
/* BLOCK COMMENT
var dont_capture_me;
*/
var a /* ANOTHER BLOCK COMMENT */, b;
```

On the 3rd line the the characters are  `*/`

I think there is a typo.


masatake:
Thank you very much. In the next time, when making a pull request for files under docs/, please, use "docs(web):" as prefix for the commit.

PR #1922, opened 2018-10-19
Allow fullqualified function definition

razum2um:
in clojure `clojure.core/defn` is also a valid way to define a function. 

p.s. the reason why I'm not using short `defn` - I have sources, which are generated, but need to be indexed. I those files all functions are full qualified including all clojure.core functions


masatake:
LGTM. Thank you.

BTW, is this change also applicable to `ns`?

razum2um:
@masatake thanks for such quick response, you're right, ns may be also qualified - https://github.com/universal-ctags/ctags/pull/1923 

p.s. quote - not, it's a reserved special

PR #1923, opened 2018-10-20
Allow fullqualified namespace definition

razum2um:
Followup of https://github.com/universal-ctags/ctags/pull/1922


masatake:
Thank you!

PR #1924, opened 2018-10-23
Defining interface for parsers (part 2 of N)

masatake:
As the result of adding features randomly, the main part of ctags is a bit mess.
I would like to start the process for cleaning.

The most dirty part is that we cannot know whether an extern'ed function is for parsers or the main part itself.

I would like to split header files in the main part into two groups.
1. common between the parsers and the main part. This is the interface for the parrsers.
2. used only the main part.

Header files in group 2 should have following file name "foo_p.h".

The changes are so many. Therefore, I would like to do this cleaning up incrementally. 

I would like to introduce `ctags_` or something prefix in the symbols in the interface for the parsers.
However, splitting the header files must be the first.

The most of changes have no impact on the parsers. They just change the name of header files and the order of inclusion.


masatake:
The changes are broken. I withdraw this.

PR #1931, opened 2018-11-04
Fix xMalloc usage

edigaryev:
No need to allocate `sizeof(CXXToken) * sizeof(CXXToken)` bytes for a single token structure when `1 * sizeof(CXXToken)` bytes is enough.

This fixes ctags memory usage issue mentioned here: https://github.com/oracle/opengrok/issues/2364


masatake:
LGTM.

pragmaware:
Yep, good catch indeed.

PR #1938, opened 2018-11-22
added seccomp_rule_add for fstat64 and _llseek (support for 32-bit)

andreasbaumann:


andreasbaumann:
See #1938.

masatake:
Applied. Thank you.

PR #1945, opened 2018-12-08
Various warnings

b4n:
Various small fixes for various things detected by some mean GCC warning options.


b4n:
Fixup: don't use `#ifdef` on [`AC_CHECK_DECLS`](https://www.gnu.org/software/autoconf/manual/autoconf.html#AC_005fCHECK_005fDECLS)-defined macros, see the docs. My bad.

PR #1941, opened 2018-12-03
[snap] Add 'snap/snapcraft.yaml'

bboymimi:
A snap is a bundle of your app and its dependencies that works without
modification across Linux. Snaps are discoverable and installable from
the Snap store.

Signed-off-by: Gavin Guo <gavin.guo@canonical.com>

PR #1947, opened 2018-12-10
Revert "[snap] Add 'snap/snapcraft.yaml'"

masatake:
Reverts universal-ctags/ctags#1941

Though I merged this but the snapfile can be distributed at another repository like universal-ctags/ctags-docker.


masatake:
Snapfile now has its own repo.

PR #1951, opened 2018-12-15
 Allow mio_new_mio() to have size 0

techee:
Consider for instance the following HTML code where a subparser is called
on the contents of the <script> tags.

```
<script language="Javascript" src="./test.js"></script>
<script language="JavaScript">
...
</script>
```

Since the first <script> is empty, mio_new_mio() is called with 0 size
but with the current semantics of the API it creates MIO until the end of
the file so the whole source file to EOF is parsed. While it seems
no extra tags are generated in this case, the parsing happens
unnecessarily.

Instead of using 0, make size signed and use -1 when we want to create a
MIO till the end of the source MIO.

---

I actually noticed this problem in Geany rather than in ctags where tags started getting duplicated because of the multiple parsing. I haven't investigated enough to learn why there aren't duplicate tags in ctags in this case but I added some traces to the parsers and the extra parsing really happens when 0 is passed as the size in io_new_mio().

Also, I think it's cleaner to allow 0-sized MIO from the API point of view,

### DIFF
--- a/parsers/yacc.c
+++ b/parsers/yacc.c
@@ -102,22 +102,25 @@ static bool leave_union (const char *line CTAGS_ATTR_UNUSED,
 
 static void make_promise_for_epilogue (void)
 {
-	const unsigned char *tmp, *last;
+	const unsigned char *tmp;
 	long endCharOffset;
 	unsigned long c_start;
 	unsigned long c_source_start;
 	unsigned long c_end;
 
-	c_start = getInputLineNumber ();
-	c_source_start = getSourceLineNumber();
+	/* We are at line with %% so the next line is the start of epilogue */
+	c_start = getInputLineNumber () + 1;
+	c_source_start = getSourceLineNumber() + 1;
 
 	/* Skip the lines for finding the EOF. */
 	endCharOffset = 0;
-	last = NULL;
 	while ((tmp = readLineFromInputFile ()))
-		last = tmp;
-	if (last)
-		endCharOffset = strlen ((const char *)last);
+	{
+		/* We want to get strlen() of the last line only but because
+		 * readLineFromInputFile() invalidates the previous value, get
+		 * endCharOffset here while the tmp variable is valid. */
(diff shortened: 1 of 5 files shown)


techee:
Funny - notice how coverage of the javascript parser went down because of this patch. It's probably because a lot of HTML was parsed as javascript before and these code paths aren't covered now.

parsers/yacc.c
```diff
@@ -102,22 +102,20 @@ static bool leave_union (const char *line CTAGS_ATTR_UNUSED,
...
+	c_source_start = getSourceLineNumber() + 1;
 
 	/* Skip the lines for finding the EOF. */
 	endCharOffset = 0;
-	last = NULL;
 	while ((tmp = readLineFromInputFile ()))
-		last = tmp;
-	if (last)
-		endCharOffset = strlen ((const char *)last);
+		endCharOffset = strlen ((const char *)tmp);
```
b4n: The trickyness of that loop might warrant a comment for nobody to come by it and "improve" it like it was before.

masatake:
I will review this to understand how I did wrong.

masatake:
LGTM. Sorry for taking your time for fixing my broken code.

masatake:
BTW, I would like ctags to make a tag for ./test.js. It can be captured as a reference tag.

Since I introduced reference tag, I have tried to capture language objects specified with import, include, require, source, etc. as reference tags without no serious use case.

However, now I find a good and critical use case.
New option -r, semantic-recursion. This is similar to -R. -R makes ctags parse given directory tree recursively. -r makes ctags parse import-include-require-source relation recursively.

Let show me example (foo.html):
```
<html>
  <body>
    <script language="Javascript" src="test.js"/>
    <script language="Javascript" src="http://example.com/bar.js"/>
    </body>
</html>
```

For the about input, run ctags with "-r -rpath . foo.html".
ctags parsrs foo.html. Then it parsers test.js in the current directory.
Finally ctags does download  http://example.com/bar.js and parser the downloaded file.

PR #1952, opened 2018-12-15
build-sys: fix "make dist" target 

masatake:
To include all test cases to .tar.gz distribution, misc/dist-test-cases
generated a Makefile.am fragment in which EXTRA_DIST list is.

Now we have so many test cases, EXTRA_DIST becomes too long.
As the result, 'make dist' reports an error.

To avoid the error, this change makes dist-test-cases
generate multiple Makefile.am files. Each Makefile.am
has much shorter EXTRA_DIST list.


b4n:
Any reason why you create one Makefile per directory?  You're allowed to list files in subdirectories from

PR #1965, opened 2019-01-01
CXX: Improve parsing of private/public/protected

pragmaware:
Improve parsing of public/protected/private keywords so it does not need specialized code in token extraction path. This makes it a bit faster and additionally fixes a bug that caused corrupted inheritance information for base classes with names starting with an underscore (a common case in STL).


masatake:
LGTM. QtMoc subpaser also works fine. Thank you.

PR #1968, opened 2019-01-06
Cxx capture parameter s default value as part of signature

### LINKED ISSUE #1967: C++: Incorrectly parsed function parameter's default value
The C++ class member function has the following signature:

` void SetValues(const Foo& lst, const QChar sep = ',');`

The _--_xformat={signature}_ option used to to extract functions signatures, i.e.:

`ctags -R -x --_xformat={signature} --sort=no --extra=*-q --c++-kinds=*-zlm --fields=*-lne --language-force=C++ `

As a result, the signature appears as:

`(const Foo & lst,const QChar sep='')`

Note the missing comma: `''` vs `','` for the default value of `sep` parameter.

Universal Ctags version:

```
Universal Ctags 0.0.0, Copyright (C) 2015 Universal Ctags Team
  Compiled: Nov 13 2016, 23:07:25
  URL: https://ctags.io/
  Optional compiled features: +win32, +wildcards, +regex, +internal-sort, +multibyte
```
However the same issue occurs with the latest release too.

masatake:
Close #1967.

The change doesn't handle whitespaces in string and char literals. 
```
void f(char c = '\t', char d='	', char e = 'a')
{
}
void g(const char *s = "a\tb	c")
{
}
```

### DIFF
--- a/Units/parser-cxx.r/signature.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/signature.cpp.d/expected.tags
@@ -2,3 +2,10 @@ BAR::bar	input.cpp	/^char *BAR::bar (char *c, double d[]) const {}$/;"	f	class:B
 bar	input.cpp	/^char *BAR::bar (char *c, double d[]) const {}$/;"	f	class:BAR	typeref:typename:char *	signature:(char * c,double d[]) const
 foo	input.cpp	/^void foo (int a, char b) {}$/;"	f	typeref:typename:void	signature:(int a,char b)
 foobar	input.cpp	/^void foobar __ARGS ((int a, char b));$/;"	p	typeref:typename:void	file:	signature:(int a,char b)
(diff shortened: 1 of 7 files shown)


masatake:
(See #1969. I will revise the condtion later.)

The default value can be a field for a parameter.
Let me show an example.

```
void f (char c = ',')
{
...
}
```
In this case tags output can be:
```
c input.c /^void f (char c = ',')$/;" z defaultValue:','
```

About a protoype declaration, ctags doesn't capture parameters in it now.
Capturing them may be better. However, I wondher what kind and role should be used for them.
My idea is `z' kind and `inPrototype` role.

```
c input.c /^void f (char c = ',')$/;" z defaultValue:',' typeref:typename:char scope:function:f role:inPrototype
```
I like this output. However, I wonder who needs this?

masatake:
@pragmaware, could you review this one?

pragmaware:
Great!
I have a few optimizations. Will try to make a pull request to your branch.

pragmaware:
Please see masatake/ctags#5

PR #1970, opened 2019-01-08
SystemTap: new parser

masatake:


masatake:
TODO: handle more broken input.

PR #1971, opened 2019-01-10
more complete support for $ in C/C++ identifiers

pskocik:
ctags seemed to have some support for $ in identifiers, like it would index $BAR from `#define $BAR` but not `$name` from `extern int $name;`. Similarly it wouldn't let me define replacement macros e.g., `'-D$my$struct=struct` and then have foo from  `$my$struct foo` indexed. This minipatch appears to fix these issues.

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -1565,7 +1565,7 @@ static void saveMacro(const char * macro)
 		return;
 	}
 
-	if(!(isalpha(*c) || (*c == '_')))
+	if(!(isalpha(*c) || (*c == '_' || (*c == '$') )))
 	{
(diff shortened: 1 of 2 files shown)


parsers/cpreprocessor.c
```diff
@@ -1565,15 +1565,15 @@ static void saveMacro(const char * macro)
 		return;
 	}
 
-	if(!(isalpha(*c) || (*c == '_')))
+	if(!(isalpha(*c) || (*c == '_' || (*c == '$') )))
```
pragmaware: Could you fix the parenthesis so (*c == '$') is at the same nesting level of (*c == '_') ?

pragmaware:
Great! Thank you!

PR #1878, opened 2018-09-17
[DON'T MERGE] Go: fill more fields

masatake:

PR #1974, opened 2019-01-11
Go: fill more fields (revisied)

masatake:
This is revised version of #1878.

In #1878, I tried to make names for scope field shorten.
In this pull request, I don''t do so; this change focuses only on fillling more fields of  Go parser tag entries.
If shorter scope names are better, I expect users may let us know. I will rework if we get such request.

### DIFF
--- a/parsers/go.c
+++ b/parsers/go.c
@@ -968,6 +968,8 @@ static void parseConstTypeVar (tokenInfo *const token, goKind kind, const int sc
 	// VarSpec     = IdentifierList ( Type [ "=" ExpressionList ] | "=" ExpressionList ) .
 	bool usesParens = false;
 	int member_scope = scope;
+	intArray *corks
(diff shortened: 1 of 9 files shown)


parsers/go.c
```diff
@@ -968,6 +968,8 @@ static void parseConstTypeVar (tokenInfo *const token, goKind kind, const int sc
 	// VarSpec     = IdentifierList ( Type [ "=" ExpressionList ] | "=" ExpressionList ) .
 	bool usesParens = false;
 	int member_scope = scope;
+	intArray *corks
+		= (kind == GOTAG_VAR || kind == GOTAG_CONST)? intArrayNew (): NULL;
```
masatake: I can avoid this array allocation with using file static variable (or function static variable).

masatake: I decide to keep this allocation as is.

masatake:
Updated. I filled more typeref field and provide more interesting information to client tools.
Also I studied "what is typeref field". 
Based on this study, I will work on closing  #862.

PR #1976, opened 2019-01-13
ObjectiveC: support cateogries

### LINKED ISSUE #1487: Objective C: inherits field is not impelemented, and categories are not captured.

masatake:
About category of Objective-C, see https://medium.com/ios-os-x-development/an-introduction-to-categories-in-objective-c-d6d55d91569f

Close #1487

### DIFF
--- a/parsers/objc.c
+++ b/parsers/objc.c
@@ -642,7 +642,7 @@ static void parseMethodsNameCommon (vString * const ident, objcToken what,
 				vStringCatS (signature, "id");
 			vStringPut (signature, ')');
 
-			e->extensionFields.signature = eStrdup (vStringValue (signature));
+			e->extensionFields.signature = vStringStrdup (signature);
 
 			vStringClear (signature);
 			vStringPut (signature, '(');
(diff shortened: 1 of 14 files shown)


parsers/objc.c
```diff
@@ -126,6 +131,24 @@ static const keywordTable objcKeywordTable[] = {
...
 
+typedef enum {
+	F_CATEGORY,
+	F_PARTOF,
+} objcField;
+
+static fieldDefinition ObjcFields [] = {
+	{
+		.name = "category",
+		.description = "category attached to th class",
```
b4n: *to th**e** class* (e is missing)

masatake: Thank you. I will update in soon.

masatake:
After thinking again, defining two kinds, categoryIface and categoryImpl looks overkill.
I will revise this.

masatake:
Updated.
categoryImpl and categoryIface kinds are removed. Instead, I introduce category kind.
We can known whether a category is defined in @interface  line or @implementation line from scope information. So we don't have to introduce the two kinds.
Using scope field to know whether a category is defined in @interface  line or @implementation line is
a bit abuse. However, I think it is better than introducing the two kinds.

In addition, I introduced code for parsing protocols spec in @interface line.

PR #1984, opened 2019-01-24
 Go: parse method specifications and inheriting interfaces in an interface

### LINKED ISSUE #1887: Go: tags for methods(?) in an interface
input.go:
```go
package point

type Point interface {
	X() int64
	Y() int64
}
```

X and Y are not tagged.

masatake:
Close #1887.

### DIFF
--- a/parsers/go.c
+++ b/parsers/go.c
@@ -109,6 +109,7 @@ typedef enum {
 	GOTAG_INTERFACE,
 	GOTAG_MEMBER,
 	GOTAG_ANONMEMBER,
+	GOTAG_METHODSPEC,
 	GOTAG_UNKNOWN,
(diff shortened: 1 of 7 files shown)


parsers/go.c
```diff
@@ -274,6 +276,11 @@ static void parseIdentifier (vString *const string, const int firstChar)
 	ungetcToInputFile (c);		/* always unget, LF might add a semicolon */
 }
 
+static bool collectorIsEmpty(collector *collector)
+{
+	return !(vStringLength(collector->str) > 0);
```
masatake: `!vStringLength(collector->str)` is enough.

masatake: Fixed.

PR #1987, opened 2019-01-25
PHP: fill the typeref field for function kind tags

masatake:

### DIFF
--- a/parsers/php.c
+++ b/parsers/php.c
@@ -394,6 +394,7 @@ static void makeClassOrIfaceTag (const phpKind kind, const tokenInfo *const toke
 
 static void makeFunctionTag (const tokenInfo *const token,
 							 const vString *const arglist,
+							 const vString *const rtype,
 							 const accessType access, const implType impl)
 {
 	if (PhpKinds[K_FUNCTION].enabled)
@@ -406,6 +407,11 @@ static void makeFunctionTag (const tokenInfo *const token,
 			e.extensionFields.implementation = implToString (impl);
 		if (arglist)
 			e.extensionFields.signature = vStringValue (arglist);
+		if (rtype)
+		{
+			e.extensionFields.typeRef [0] = "unknown";
+			e.extensionFields.typeRef [1] = vStringValue (rtype);
+		}
 
 		makePhpTagEntry (&e);
 	}
@@ -1043,6 +1049,24 @@ getNextChar:
 	MayBeKeyword = nextMayBeKeyword;
 }
 
+static void readQualifiedName (tokenInfo *const token, vString *name,
+                               tokenInfo *const lastToken)
+{
+	while (token->type == TOKEN_IDENTIFIER || token->type == TOKEN_BACKSLASH)
+	{
+		if (name)
+		{
+			if (token->type == TOKEN_BACKSLASH)
+				vStringPut (name, '\\');
+			else
+				vStringCat (name, token->string);
+		}
+		if (lastToken)
+			copyToken (lastToken, token, true);
+		readToken (token);
+	}
+}
+
 static void enterScope (tokenInfo *const parentToken,
 						const vString *const extraScope,
 						const int parentKind);
@@ -1182,6 +1206,8 @@ static bool parseFunction (tokenInfo *const token, const tokenInfo *name)
 	accessType access = CurrentStatement.access;
 	implType impl = CurrentStatement.impl;
 	tokenInfo *nameFree = NULL;
+	vString *rtype = NULL;
+	vString *arglist = NULL;
 
 	readToken (token);
 	/* skip a possible leading ampersand (return by reference) */
@@ -1200,9 +1226,9 @@ static bool parseFunction (tokenInfo *const token, const tokenInfo *name)
 
 	if (token->type == TOKEN_OPEN_PAREN)
 	{
-		vString *arglist = vStringNew ();
 		int depth = 1;
 
+		arglist = vStringNew ();
 		vStringPut (arglist, '(');
 		do
 		{
@@ -1269,9 +1295,6 @@ static bool parseFunction (tokenInfo *const token, const tokenInfo *name)
 		}
 		while (token->type != TOKEN_EOF && depth > 0);
 
-		makeFunctionTag (name, arglist, access, impl);
-		vStringDelete (arglist);
-
 		readToken (token); /* normally it's an open brace or "use" keyword */
 	}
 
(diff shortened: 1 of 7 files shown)


parsers/php.c
```diff
@@ -401,6 +402,11 @@ static void makeFunctionTag (const tokenInfo *const token,
 			e.extensionFields.implementation = implToString (impl);
 		if (arglist)
 			e.extensionFields.signature = vStringValue (arglist);
+		if (rtype)
+		{
+			e.extensionFields.typeRef [0] = "typename";
```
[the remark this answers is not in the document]
masatake: I see. I will use "unknown".

About "typename", I'm talking with @pragmaware at  #862.
I have a plan to make the subfield empty in another PR

masatake:
appveyor doesn't work...

b4n:
@masatake I just saw that I have a branch doing exactly this laying around in my tree.  It has conflicts, but I'll look at it and see if there's anything good there that would benefit from being added here.

masatake:
I updated my changes based on your review comments.

> @masatake I just saw that I have a branch doing exactly this laying around in my tree. It has conflicts, but I'll look at it and see if there's anything good there that would benefit from being added here.

Oh, I see. How should I do? Should I merge this to the master branch, or will you merge this to the master branch after merging your efforts to this branch?


BTW, I found a minor issue.

http://php.net/manual/en/functions.returning-values.php 
```
Example #7 Nullable return type declaration (as of PHP 7.1.0)
<?php
function get_item(): ?string {
    if (isset($_GET['item'])) {
        return $_GET['item'];
    } else {
        return null;
    }
}
?>
```
Even with my changes, the php parser cannot fill typeref field for "get_item" with "?string".

b4n:
> Oh, I see. How should I do? Should I merge this to the master branch, or will you merge this to the master branch after merging your efforts to this branch?

I'll check and report here.  It looks pretty similar, but possibly with slightly better code reuse.

> BTW, I found a minor issue.
>
>http://php.net/manual/en/functions.returning-values.php

Oh, makes sense, it's missing support for that.  I'll handle this in a subsequent PR after this one is dealt with.

masatake:
@b4n, I would like to review your change to study better code:-)

b4n:
> Just an off topic question. How can you push your change to my branch?
I have found sometimes people use this technique. It is so powerful in collaboration work. I woud like to know its git command line.

I actually wasn't sure I could as I never used that on GitHub, but it's actually trivial: just push to the PR author's repo.  IIUC, if you're owner of the project the PR is against (and possibly if the author didn't uncheck so option) you can push to the author's branch on his repo.

> LGTM. Could you merge this?

Will do.

masatake:
> I actually wasn't sure I could as I never used that on GitHub, but it's actually trivial: just push to the PR author's repo. IIUC, if you're owner of the project the PR is against (and possibly if the author didn't uncheck so option) you can push to the author's branch on his repo.

Oh, I see. This is quite useful and powerful. However, it is too powerful. I should not use it too much:-).

masatake:
```
function get_item(): ?string {
```
I extended the chagens to hanel the above "Nullable return type".

parsers/php.c
```diff
@@ -1282,22 +1362,39 @@ static bool parseFunction (tokenInfo *const token, const tokenInfo *name)
...
 	    (getInputLanguage () == Lang_zephir && token->type == TOKEN_OPERATOR))
 	{
-		do
+		if (arglist)
+			rtype = vStringNew ();
+
+		readToken (token);
+		if (rtype && token->type == TOKEN_QMARK)
+		{
+			vStringPut (rtype, '?');
```
b4n: it probably doesn't change much as all functions should have parentheses and this `rtype` should always be created as `arglist` should always be non-null, but skipping the `?` token should probably be unconditional, and the conditional be moved to the `vStringPut()` line.

masatake: Than you for reviewing.
I wonder I understand what you say correctly.

My understanding:

```
if (arglist)
   rtype = vString ();
```
The condtion can be moved to upper level like:
```
if (arglist && ((getInputLanguage () == Lang_php && token->type == TOKEN_COLON) ||
	    (getInputLanguage () == Lang_zephir && token->type == TOKEN_OPERATOR)))
{
```

So we can remove arglist and rtype related conditions in the block:
```
  rtype = vString (); 
  readToken (token);
  if (token->type TOKEN_QMARK)
  {
     vStrintPut (rtype, '?');
     readToken(token);
   }
...
```
Am I understading what you wrote correctly?

b4n: No, my point would be that the parsing would be the same whether or not we're collecting rtype, so that in the event we'd be parsing a correct declaration and simply don't want to collect the value, the code would still do the same thing.

It currently is not very relevant as the only case I can think of where `rtype` is `NULL` is for invalid code; but my idea is that currently the function behaves the same parsing-wise regardless of whether we're collecting or not, and that seemed nice to me.

However, if we indeed wanted to cut off some parsing when not needed, we could indeed move the check to a higher level.  But that wasn't my point -- which doesn't mean it's not valid :wink:

masatake: > No, my point would be that the parsing would be the same whether or not we're collecting rtype, so that in the event we'd be parsing a correct declaration and simply don't want to collect the value, the code would still do the same thing.

Oh, I see. I think I understand.
So the change will be:
```
 			rtype = vStringNew ();
 
 		readToken (token);
-		if (rtype && token->type == TOKEN_QMARK)
+		if (token->type == TOKEN_QMARK)
 		{
-			vStringPut (rtype, '?');
+			if (rtype)
+				vStringPut (rtype, '?');
 			readToken (token);
 		}
 		readQualifiedName (token, rtype, NULL);
```
I will push the change. I have just found one more crack about rtype. I will push for fixing it, too.

masatake:
The last commit if the condition about deleting rtype.

parsers/php.c
b4n: Yes, that's what I meant :+1:

parsers/php.c
```diff
@@ -1371,14 +1371,15 @@ static bool parseFunction (tokenInfo *const token, const tokenInfo *name)
...
 		{
-			vStringPut (rtype, '?');
+			if (rtype)
+				vStringPut (rtype, '?');
 			readToken (token);
 		}
 		readQualifiedName (token, rtype, NULL);
 
-		if (vStringIsEmpty (rtype))
+		if (rtype && vStringIsEmpty (rtype))
```
b4n: Good catch, as indeed `vStringIsEmpty()` is not null-safe :+1:

b4n:
@masatake do you plan any more changes here?  If not, your or I could merge it.

masatake:
No, I don't change anymore. Please, merge this.

masatake:
Thank you very much.

PR #1989, opened 2019-01-26
JavaScript: introduce getter and setter kinds + computed properties

### LINKED ISSUE #1949: JavaScript: Missing methods inside modernish JavaScript class
The name of the parser:
JavaScript (jscript.c)

```
$ ctags --options=NONE --langmap=javascript:+.mjs -f - models/user.mjs
```

<details>
<summary>The content of input file:</summary>

```javascript
import uuid from "uuid"

import * as emails from "../app/emails"
import Controller from "../controllers/controller"
import Group from "./group"
import Model from "./model"
import Order from "./order"

export default class User extends Model {
  static get tableName() {
    return "users"
  }

  static get ["json-schema"]() {
    return {
      type: "object",
      optional: [
        "passwordHash",
        "passwordResetRequestedAt",
        "passwordResetToken",
        "verificationToken",
        "verifiedAt",
      ],
      properties: {
        name: { type: "string" },
        email: { type: "string", format: "email" },
        passwordHash: { type: "string" },
        verificationToken: { type: "string", format: "uuid" },
        verifiedAt: { type: "string", format: "date-time" },
        passwordResetToken: { type: "string", format: "uuid" },
        passwordResetRequestedAt: { type: "string", format: "date-time" },
        address: {
          type: "object",
          properties: {
            streetAddress: { type: "string", description: "The street address. For example, 1600 Amphitheatre Pkwy." },
            postalCode: { type: "string", description: "The postal code. For example, 94043." },
            locality: { type: "string", description: "The locality. For example, Mountain View." },
            region: { type: "string", description: "The region. For example, CA." },
            country: { type: "string", default: "USA", description: "The country. For example, USA. You can also provide the two-letter ISO 3166-1 alpha-2 country code." }
          }
        }
      }
    }
  }

  static get relationMappings() {
    return {
      orders: {
        relation: Model.HasManyRelation,
        modelClass: Order,
        join: {
          from: "users.id",
          to: "orders.user_id"
        }
      },
      groups: {
        relation: Model.ManyToManyRelation,
        modelClass: Group,
        join: {
          from: "users.id",
          through: {
            from: "users_groups.user_id",
            to: "users_groups.group_id"
          },
          to: "groups.id"
        }
      }
    }
  }

  get isAnonymous() { return false }
  get isAuthenticated () { return true }

  async $beforeInsert(queryContext) {
    await super.$beforeInsert(queryContext)
    this.verificationToken = uuid.v4()
  }

  inspect(_depth, _options) {
    return `${this.name} (${this.id}) <${this.email}>`
  }

  async sendVerificationMail() {
    let verificationLink = [
      process.env.HOST,
      Controller.urlFor("account.confirm_page", { token: this.verificationToken })
    ].join("")

    return emails.verification(this.email, { name: this.name, verificationLink })
  }

  async sendResetPasswordMail() {
    let passwordResetLink = [
      process.env.HOST,
      Controller.urlFor("account.password_reset_page", { token: this.passwordResetToken })
    ].join("")

    return emails.passwordReset(this.email, { name: this.name, passwordResetLink })
  }
}

export class AnonymousUser extends User {

  get isAnonymous() { return true }
  get isAuthenticated () { return false }

  sendVerificationMail() {
    throw new TypeError("Can't send a verification email to the anonymous user.")
  }

  sendResetPasswordMail() {
    throw new TypeError("Can't send a password reset email to the anonymous user.")
  }
}

export const ANONYMOUS_USER = new AnonymousUser({ name: "<anonymous>" })
```

</details>

```
AnonymousUser	models/user.mjs	/^export class AnonymousUser extends User {$/;"	c
User	models/user.mjs	/^export default class User extends Model {$/;"	c
passwordResetLink	models/user.mjs	/^    let passwordResetLink = [$/;"	v
sendResetPasswordMail	models/user.mjs	/^  sendResetPasswordMail() {$/;"	m	class:AnonymousUser
sendVerificationMail	models/user.mjs	/^  sendVerificationMail() {$/;"	m	class:AnonymousUser
verificationLink	models/user.mjs	/^    let verificationLink = [$/;"	v
```

The tags output you expect:

The above, but also tags for the methods/computed properties.

The version of ctags:

```
$ ctags --version                                                                                                                                                                          11:14:25
Universal Ctags 0.0.0(64f7d61), Copyright (C) 2015 Universal Ctags Team
  Compiled: Dec 14 2018, 10:37:50
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames
```

How do you get ctags binary:

homebrew:

```
$ brew info universal-ctags/universal-ctags/universal-ctags                                                                                                                                11:58:48
universal-ctags/universal-ctags/universal-ctags: HEAD
Maintained ctags implementation
https://github.com/universal-ctags/ctags
Conflicts with:
  ctags (because this formula installs the same executable as the ctags formula)
/usr/local/Cellar/universal-ctags/HEAD-64f7d61 (325 files, 4.3MB) *
  Built from source on 2018-12-14 at 10:38:07
From: https://github.com/universal-ctags/homebrew-universal-ctags/blob/master/universal-ctags.rb
```

---

This is taken from an actual project am working on, but I figured the user class should be common enough to demonstrate my point. For some other module/classes there're even less tags outputted.

I think it's the use of computed properties (`get`/`set`), `async` and `["funky-method"]`. The latter is more common then one might think, language level protocols are mostly implemented using [well-known symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Well-known_symbols).

```js
class InfiniteCounter {
  *[Symbol.iterator]() {
    for (let i = 0;; ++i) yield i
   }
}
```

masatake:
Partially close  #1949.
See https://www.ecma-international.org/ecma-262/5.1/#sec-11.1.5


masatake:
I wonder we can remove `signature` field for `getter` kind objects.

masatake:
I introduced `is_dynamic_prop ` local variable to skip `(1+2)*3`. If you did not know this, look at 
9d44448.
However, I didn' handle the nested case. For capturing `subnum` and fill the scope field for `subnum`, we have to put an anonymous tag as the proxy(?) for  `(1+2)*3`.

Anyway I guess we have more things to do about #1949. So I would like to merge this PR.
Is your approve for my changes including 9d44448 ?

b4n:
@masatake yes I've seen it, but as mentioned it's not checked in case of recursing into a subclass; and there we need something to provide scope for the members. Another way could be to ignore the subclass methods in this case.

And yes, my review applies at the time I made it, to the 3 commits.

masatake:
@b4n, thank you!

PR #1990, opened 2019-01-28
Emacs editting mode for ctags optlib file

masatake:


masatake:
None may have interest in this change.
Other developers in this project are abnormal; they develop their editor or IDE :-O

PR #1992, opened 2019-01-29
PHP: use anonGenerate function

masatake:
anonGenerate function defined in main part is for generating names
for anonymous tags. Unlike the current particular implementation
in php.c, the name generated by anonGenerate function is unique in
a tags file. The unique names may help a client tool implement
interesting features.

In addition to using the function, this change marks tags for
anonymous classes XTAG_ANONYMOUS. So a user can enable or disable
the emission of tags for anonymous classes with --extras='-{anonymous}'
option.


masatake:
The impact of this change is small. I will merge this without reviewing.

PR #2003, opened 2019-02-10
Go: handle type aliases and improve signature field of variadic functions

masatake:
This PR does two things.

1. introduce "a,talias,type aliases" kinds for capturing T1 in "type T1 = T2".
   Just using "type" field is not enough for distinguishing the expression from "type T1 T2.
  These were 2 choices; introducing a parser specific field "alias:true" or introducing new kind for the aliase.
  My choice was later one.

2. normalize signature field for a function taking variadic arguments: putting one space before `...` and no space after `...'.


masatake:
TODO: Collecting tokens are well appeared pattern of code. The main part should provide utility functions for the purpose.

PR #2005, opened 2019-02-11
JavaScript: capture function objects in Self-Invoking Anonymous Function Statements

masatake:


parsers/jscript.c
```diff
@@ -2520,6 +2524,78 @@ static bool parseLine (tokenInfo *const token, bool is_inside_class)
...
+				}
+				else if (!isType (token, TOKEN_EOF))
+				{
+					Assert (NextToken == NULL);
+					NextToken = newToken ();
+					copyToken (NextToken, token, false);
+				}
+			}
+		}
+	}
```
b4n: I don't really fancy having all this in `parseLine()` instead of `parseStatement()`, but I must admit that it's a lot easier that way, so why not.

parsers/jscript.c
```diff
@@ -2520,6 +2524,78 @@ static bool parseLine (tokenInfo *const token, bool is_inside_class)
...
+		 *     return z * 1;
+		 * }(1);
+		 */
+		bool started_with_paren = isType (token, TOKEN_OPEN_PAREN);
+
+		readToken (token);
+		is_terminated = parseLine (token, is_inside_class);
+
+		/*
+		 * case C is nothing special; t just returns above `is_terminated'.
```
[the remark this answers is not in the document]
masatake: Yes, clearer. I will update my change.

parsers/jscript.c
masatake: ..I recognize putting all new code to parseLine is not good.
At least I should introduce a new function to handle `(function ()...)`.

I will update my patch.

b4n:
> ..I recognize putting all new code to parseLine is not good.
> At least I should introduce a new function to handle |(function ()...)|.

My remark was simply that this is the kind of things a function called
`parseStatement()` should do, but as it currently is `parseStatement()`
is way too complex and fragile to allow adding this easily, so your
change is probably fine.

I'm not sure it's worth it or if having another function would be
clearer, but if you think so, go ahead :)

masatake:
Sorry to be late. During working on improving my change, I found some inputs the parser doesn't work well.
I'm trying to understand them and improving code for debugging now.

masatake:
> Sorry to be late. During working on improving my change, I found some inputs the parser doesn't work well.
I'm trying to understand them and improving code for debugging now.

I cannot remember the inputs.

masatake:
#3437 is a much better solution.
I withdraw this.

PR #2007, opened 2019-02-12
JavaScript: accept fq name after new operator

masatake:
`a` in
```
var a = new foo.bar();
```
was not captured well though `b` in
```
var b = new baz();
```
was captured well.


parsers/jscript.c
```diff
@@ -1158,6 +1158,33 @@ static void skipArrayList (tokenInfo *const token, bool include_newlines)
...
+					 && isType (token, TOKEN_IDENTIFIER))
+			{
+				expecting_period = true;
+			}
+			else
+				break;
+
+		} while (1);
+	}
+}
```
b4n: What about what I feel is the simpler:
```C
static void skipQualifiedIdentifier (tokenInfo *const token)
{
	while (isType (token, TOKEN_IDENTIFIER))
	{
		readToken (token);
		if (isType (token, TOKEN_PERIOD))
			readToken (token);
		else
			break;
	}
}
```

masatake:
Updated. Thank you.

PR #2008, opened 2019-02-13
Go: don't try to solve the type for "_" variable

masatake:
To solve a type for a variable later, go parser stores the cork id of
the variable to integer array. However, the cork id for "_" variable
becomes CORK_NIL because makeTagFull() doesn't record "_" as a tag.

The original code has a bug that that go parser stores the CORK_NIL
for "_" to the intger array. The parser should not do so. When resolving
the type for the variable, go parser does (1) pick a cork id from the
intger array, (2) get a tagEntry for the cork id, then (3) access
the internal of the tagEntry. If the cork id is CORK_NIL here, ctags
crash at (3) because getEntryInCorkQueue returns NULL when receiving
CORK_NIL as a cork id.


masatake:
This fix a bug I injected in ebf1bd78183e9c5f5d751aeb2dd3b8c191716210.

Issue #2014, opened 2019-02-21
signature went wrong after PR #1968


elecalion:
It seems signature went wrong after PR #1968 if --fields=+S is set.

The command line you used to run ctags:

```
--output-format=e-ctags
--kinds-c++=+p
--fields=+iaS
```

The content of input file:

```C
/* Can NOT be parsed */
int32 test(int32 a)
{
    return 0;
}
/* Can be parsed */
int32 test2(void)
{
    return 0;
}
```

The tags output you are not satisfied with:

```
test2	test.c	/^int32 test(void)$/;"	f	typeref:typename:int32	signature:(void)
```

The tags output you expect:

```
test	test.c	/^int32 test(void)$/;"	f	typeref:typename:int32	signature:(int32 ...)
test2	test.c	/^int32 test(void)$/;"	f	typeref:typename:int32	signature:(void)
```

The version of ctags:

```
Universal Ctags 0.0.0(b60a7b92), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Jan  8 2019, 00:09:52
  URL: https://ctags.io/
  Optional compiled features: +win32, +wildcards, +regex, +internal-sort, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames
```

How do you get ctags binary:
win32 binary taken from Universal-ctags/ctags-win32 project

b4n:
This only happens when using `e-ctags` output format.

b4n:
First bad commit is ba82e4f4382a9aa0acf2a05e1804101639fa6bb3

b4n:
I'm not quite sure we disallow spaces (as opposed to tabs) when rendering E-CTags?  AFAIK it should work there, shouldn't it?

masatake:
@elecalion, thank you for reporting.

This one may be related to https://github.com/universal-ctags/ctags/issues/1969.
```
diff --git a/main/field.c b/main/field.c
index 7e32154e..a32142bf 100644
--- a/main/field.c
+++ b/main/field.c
@@ -437,7 +437,7 @@ static const char *renderFieldNameNoEscape (const tagEntryInfo *const tag, const
 
 static bool hasWhitespaceInName (const tagEntryInfo *const tag, const char *value CTAGS_ATTR_UNUSED)
 {
[...]

Issue #1969, opened 2019-01-06
Wrong condition to detect whether escaping is needed or not


masatake:
This is my todo.

Universal-ctags introduces backslash as the prefix of escape sequence. It allows including tab chars in a tags file. In Exuberant ctags compatible output mode, u-ctags doesn't use the escape sequence. Instead, it doesn't print a tag which includes a tab char.

In  Exuberant ctags compatible output mode, u-ctags verifies a tag entry needs the escaping or not.
The condition used in verifying looks wrong:
```
 strpbrk(tag->anyField, " \t");
```

A whitespace is nothing to do with escaping.
Instead of checking a whitespace, control chars handled in vStringCatSWithEscaping must be searched.

I wonder why I introduced such wrong condition.
I should revise the condition.

PR #2015, opened 2019-02-21
main: accept whiespace in tagEntryInfo even in e-ctags output mode

masatake:
The original code rejects printing a tagEntryInfo that includes whiespace or tab chars
in e-ctags output mode.

This is reported as a bug in #2014 by @elecalion.
I myself also reported in #1969.

What we should reject is just tab char.

NOTE: This char doesn't fix the original bug perfectly.  If --sort=no
is given, e-ctags output mode doesn't reject a tagEntryInfo including
'\t'. The code for rejecting depends on wrong assumption that the
output stream (to tags file) is rewindable. This is not true if the
output is connected to stdout directly. The options combination, "-o
-" and "--sort=no" breaks my assumption.

### DIFF
--- a/Tmain/e-ctags-output.d/input_file.cc
+++ b/Tmain/e-ctags-output.d/input_file.cc
@@ -1,0 +1,28 @@
+/* Could NOT be parsed well.
+   The signature should be  recorded well. */
+int32 test(int32 a)
+{
+    return 0;
+}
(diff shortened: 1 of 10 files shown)


masatake:
@b4n, thank you. Next, I will fix the case "--sort=no" is given.

PR #2016, opened 2019-02-23
Fix the behavior in "-o - --sort=no --output-format=e-ctags" combination (was: Overhaul the code handling tab chars in parser specific fields)

### LINKED ISSUE #1969: Wrong condition to detect whether escaping is needed or not

masatake:
During working on #2015, I found some broken code for handling tab chars in parser specific fields.
It was not well considered.
The test code in this PR explains how the code was broken.

~~"--sort=no" issue I reported in #2015 is not fixed in this PR yet.~~

I also fixed the bug found in the option combination in "-o - --sort=no --output-format=e-ctags".
Close #1969 and #2014.

### DIFF
--- a/Tmain/e-ctags-output.d/run.sh
+++ b/Tmain/e-ctags-output.d/run.sh
@@ -5,22 +5,33 @@
 CTAGS=$1
 
 tmp="input file.cc"
-cp input_file.cc "${tmp}"
-"${CTAGS}" --quiet --options=NONE --output-format=e-ctags \
-		   --kinds-c++=+p --fields=+iaSs \
-		   -o - \
-		   "${tmp}" \
-		   input_tab.rst input_space.rst
-rm "${tmp}"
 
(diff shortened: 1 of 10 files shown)


masatake:
Updated.
The original commits hit a bug about our internal sort routine.

PR #2022, opened 2019-03-01
optlib2c: escape a double quote char in a description string w backslash

dreamtigers:
Hello, I'm interested in picking up the elixir branch (`elixir-optlib`) of this repo, and noticed that said branch has two commits: the first one, is this one I'm doing the pull request for, in which @masatake added the code to escape a double quote char in a string, and the second one, which has the code to parse elixir code.

I noticed that the first commit isn't actually related to Elixir (or at least isn't in a way I could see), so it shouldn't be in the elixir branch, and it's useful enough to be merged into master. That's the reason I took this commit and rebased it to `master`, fixing a small conflict in the process. So now optlib should be able to parse the double quote char in a description string, and I should be free to hack on `elixir-optlib` after rebasing it to `master`.

I'm not 100% sure how to test it, but what I did was 
```
$ ./autogen.sh
$ ./configure
$ make validate-input check
```
And it ran without any error I could notice.

Any help or thought related to the commit or how to test is well received.

### DIFF
--- a/misc/optlib2c
+++ b/misc/optlib2c
@@ -569,6 +569,23 @@ sub emit_patterns {
     emit_list $_[0], "patterns";
 }
 
+sub escape_dquotes {
+    my $input = shift;
+    my $output = "";
+
+    my $c;
+
+    foreach $c (split //, $input) {
+	if ($c eq '"') {
+	    $output = $output . '\\' . '"';
+	} else {
+	    $output = $output . $c;
+	}
+    }
+
+    return $output;
+}
+
 sub emit_roledefs {
     my $opts = shift;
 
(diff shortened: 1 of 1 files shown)


masatake:
> Hello, I'm interested in picking up the elixir branch (elixir-optlib) of this repo,

Thank you. This change is obviously useful.


We don't test optlib2c directly in our CI environment.
Instead, we run tests code generated by optlib2.c as a parser.

You can test the change with
```
$  ./misc/optlib2c /tmp/elixir.tags  | grep operator
```

Without this patch you will see:
```
		  true, 'o', "operator", "operators (e.g. "defmacro a <<< b")",
```

This line causes a syntax error as a C language source code.

dreamtigers:
Oh, I see now. Thank you.

PR #2026, opened 2019-03-04
optlib2c: Change some indentation for aesthetics purposes

dreamtigers:


masatake:
Thank you. It seems that I took a mistake in Makefile.am As the result optlib/*.c files are not regenerated even we modify misc/optlib2c. I will fix it. With my fix, I will merge your change in this PR.

masatake:
I opened #2027 based on this PR.

Issue #2029, opened 2019-03-05
Out-of-source builds fail because packcc could not be found


Myz:
I am automatically build every commit of your repository by running:
`cd $(mktemp -d) && git clone https://github.com/universal-ctags/ctags.git && cd ctags && ./autogen.sh && mkdir mybuild && cd mybuild && ../configure && make -j10`
that will result in:
```
...
  CC       dsl/readtags-qualifier.o
  CCLD     readtags
  CCLD     packcc
PACKCC     peg/varlink.c
/bin/sh: ../packcc: No such file or directory
make[2]: *** [Makefile:5118: peg/varlink.c] Error 127
...
```

The interesting part is `/bin/sh: ../packcc: No such file or directory` where it looks for the `packcc` bin in the parent folder but it is located in the current one which is the sub folder `mybuild`.

I was able to solve this by changing a line in the `Makefile.am` from
```
PACKCC = $(srcdir)/packcc$(EXEEXT)
```
to
```
PACKCC = $(top_builddir)/packcc$(EXEEXT)
```

masatake:
Thank you for reporting. I have waited  for such a report about packcc that I introduced recently.
Could you make a pull request?
I would like to see how travis and appveyor say about the change.

Myz:
Sure, I will :)

PR #2030, opened 2019-03-05
Locate packcc in build directory instead of the source

Myz:
This should fix a Out-of-source or nested build.

Related to #2029.


b4n:
Looks reasonable.  We should have an out-of-tree build on CI I guess, it's always nice to catch these kind of things early.

masatake:
NOTE: out-of-tree build can be done with `cd $(mktemp -d) && git clone https://github.com/universal-ctags/ctags.git && cd ctags && ./autogen.sh && mkdir mybuild && cd mybuild && ../configure && make -j10` as @Myz written in @2029.

masatake:
Thank you!

Myz:
Anytime

PR #2031, opened 2019-03-05
travis-ci: do out-of-tree build

masatake:


masatake:
It works!

masatake:
It seems that the change has no effect.

k-takata:
Coverage status looks incorrect after this PR. It cannot be 100%.

masatake:
@k-takata, thank you. 
I will inspect this.

masatake:
Fixed. Thank you.

PR #2039, opened 2019-03-11
Revert "travis-ci: do out-of-tree build"

masatake:
Reverts universal-ctags/ctags#2031

@k-takata wrote:

     Coverage status looks incorrect after this PR. It cannot be 100%.


masatake:
Withdraw this.

Issue #2043, opened 2019-03-12
Elixir nullary function with multiline keyword syntax loses final letter


endersstocker:
The name of the parser:
**elixir**

The command line you used to run ctags:
```
$ ctags --options=NONE input.ex
```

The content of input file:
```elixir
defmodule SomeModule do
  def nullary_function,
    do: :some_constant

  def another_nullary_function, do: :another_constant

  def parametric_function(x),
    do: x
end
```

The tags output you are not satisfied with:
```
nullary_functio	input.ex	/^  def nullary_function,$/;"	f	module:SomeModule
```

The tags output you expect:
```
nullary_function	input.ex	/^  def nullary_function,$/;"	f	module:SomeModule
```

The version of ctags:
```
$ ctags --version
Universal Ctags 0.0.0(94cb70f), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Mar 12 2019, 09:14:24
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames, +packcc
```

How do you get ctags binary:
```
$ brew install --HEAD universal-ctags/universal-ctags/universal-ctags
```

endersstocker:
Closed by #2044

PR #2044, opened 2019-03-12
Fix bug in elixir function tag missing a char

dreamtigers:
Quick fix for #2043 
* Removed unnecessary atom in the regex for the function kind.
* Expanded `elixir-functions` test case to include case present in the issue above.

The removed atom was there to deal with a problem present when parsing operators, but I forgot to remove it when the operators problem was fixed.

A quick run of the tests shows them all passing, as before.


masatake:
Thank you.

PR #2045, opened 2019-03-14
Add access information to Elixir parser

endersstocker:


masatake:
As far as I can remember correctly, other than I, you are the first person who uses `--_fielddef-<LANG>` option in the universe.

I expect none uses the option, so there is a limitation I have not removed yet.

`access` field is a common between langugaes. C++ paser and Java parser fill the fields, for an example.

Currently, an optlib parser cannot fill such common fields. I must remove this limitation.
Because of the limitation, you had to define a field `access` for Elixir.
An optlib parser can define a parser specific field, and it can fill only parser specific fields in the current implementation.

Strictly speaking, your parser fills `Elixir.access` field.

If I remove the limitation, your don't have to define `access` field in your parser; your parser can fill `access` common field without changing the flag: `{field=access:...}`.

This is the background of the test failure in travis and circleci.

I cannot remove the limitation now. I'm busy for implementing stream unwind API for newly contributed typescript parser. So could you fix the stdout-expected.txt of failed test cases?
What you have to do is simple; mv stdout-actual.txt stdout-expected.

masatake:
Ideally, we can unify similar patters.

Some choices will be.
```
--regex-Elixir=/^[ \t]*def(p?)[ \t]+([a-z_][a-zA-Z0-9_?!]*)/\2/f/{scope=ref}{eval:lisp=(set-field! access (if (< 0 (length \2))  "public" "private"))}
```
```
--regex-Elixir=/^[ \t]*def(p?)[ \t]+([a-z_][a-zA-Z0-9_?!]*)/\2/f/{scope=ref}{eval:ps={\2 length 0 gt {"public"}{"private"} if /access setField}}
```
--regex-Elixir=/^[ \t]*def(p?)[ \t]+([a-z_][a-zA-Z0-9_?!]*)/\2/f/{scope=ref}{eval:lua=...}
--regex-Elixir=/^[ \t]*def(p?)[ \t]+([a-z_][a-zA-Z0-9_?!]*)/\2/f/{scope=ref}{eval:js=...}
```
Which one do you like:-P

masatake:
Thank you!

PR #2047, opened 2019-03-16
Elixir: Fix protocol scope

endersstocker:


masatake:
@endersstocker, could you look at https://github.com/universal-ctags/codebase ?

I would like you to add some LARGE repositories in which code written in Elixir are to the codebase.

PR #2051, opened 2019-03-19
main,Flex,JavaScriupt,SQL,refactor: introduce a helper function to skip two character sequence

masatake:
/* ... */ style comment syntax is used variety languages.
Each parser implements code skipping comments.

This change introduces skipToCharacterInInputFile2 helper function for
unifying the implementations.

The original code handles '\0' char specially.
skipToCharacterInInputFile2 doesn't do so. I don't understand the need.

(About C and C++ parsers they are handled in CPreProcessor parser.
They are out of scope of this change.)

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -948,6 +948,19 @@ extern int skipToCharacterInInputFile (int c)
 	return d;
 }
 
+extern int skipToCharacterInInputFile2 (int c0, int c1)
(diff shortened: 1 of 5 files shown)


masatake:
Go parse has its own implementation, too.
However, it checks weather newline is included or not.

PR #2035, opened 2019-03-08
Use git-subtree instead of git-submodule for packcc

k-takata:
See: https://github.com/universal-ctags/ctags/issues/1452#issuecomment-470759558
With this PR, no need to run git-submodule commands in autogen.sh anymore.

This also add a script named `misc/pull-packcc.sh` which allows to easily update packcc next time.

PR #2052, opened 2019-03-22
doc: Remove description about git-submodule for packcc

k-takata:
Now we use git-subtree for packcc. (#2035)


masatake:
Thank you.

PR #2066, opened 2019-03-30
C++ template minor improvements

### LINKED ISSUE #2060: ctags fails when parsing C++ template templates
Hello! I'm having issues when using ctags with template arguments that use or require templates themselves. It looks like ctags is stops parsing once it reaches one of these arguments. I found a similar closed issue from January 2018 (#1671), and testing ctags using the previous author's examples reveals that ctags handles default initialized templates just fine.

As for an example, consider the following input:

```C++
template <long x, typename enabled=void>
struct tabs { const static long value = x; };

template <class Key, class T>
class success {};

template <class Key=long, class T=std::string, class Hash=long_hash_function>
class success2 {};

template <class Key, class T, class Hash=std::hash<Key>, class KeyEqual=std::equal_to<Key>>
class failure {};

namespace test {
...
}
```

What one would expect is that ctags lists tags for all declarations:
* tabs
* success
* success2
* failure
* test

Instead, ctags stops at success, because it fails at parsing `class Hash=std::hash<Key>`, specifically the `<Key>` part. You would only get tags for:
* tabs
* success
* success2

I am currently on Arch Linux, and I am using the official ctags package from the extra repository. 
`ctags --version` outputs:
```
Universal Ctags 0.0.0(45968eff), Copyright (C) 2015 Universal Ctags Team
  Compiled: Oct 15 2018, 21:55:11
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml
```

Let me know if you need anything else to aid in debugging or if this was previously resolved.

masatake:
Close #2060.

Only the last commit is related to #2060.
@pragmaware, if I should submit the commits in separated pull requests, let me know. I will withdraw this and do as your request.

In this pull request, I use a strlist instead of corkQueue level symbol table. The corkQueue level one is still in design.

### DIFF
--- a/Units/parser-cxx.r/bug-github-pull-972.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/bug-github-pull-972.cpp.d/expected.tags
@@ -4,4 +4,4 @@ A	input.cpp	/^    A(int i ) : m_i(i)$/;"	function	scope:class:A	file:	signature:
 create	input.cpp	/^    static A create(int i)$/;"	function	scope:class:A	typeref:typename:A	file:	signature:(int i)	properties:static
 g_i	input.cpp	/^    static int g_i;$/;"	member	scope:class:A	typeref:typename:int	file:	properties:static
 m_i	input.cpp	/^    int m_i;$/;"	member	scope:class:A	typeref:typename:int	file:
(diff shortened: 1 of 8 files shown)


pragmaware:
Ow.. sorry, in my messy week I've missed #2060...
I'll take a look at the changes.

pragmaware:
LGTM

masatake:
Thanks.

PR #2070, opened 2019-04-01
Moose subparser based on perl parser

### LINKED ISSUE #2067: Feature Request: Perl Attributes with (has)
Currently Perl is extended with Moose, Mouse, Moo, and Mo, and the like. These modules all have one commonality -- their attribute selectors are established with `has`. I believe we can make the assumption that anything that follows the Moose-established convention of, 

[`has $name|@$names => %options`](https://metacpan.org/pod/Moose#has-$name|@$names-=>-%options) is a properly formed attribute for one of the many object build systems out there.

We'd have to cover things like

```
has "foo" => ()
has ("foo" => ())
has [qw/foo bar/] => ()
has ([qw/foo bar/] => ())
```

masatake:
Close #2067.

### DIFF
--- a/parsers/moose.c
+++ b/parsers/moose.c
@@ -1,0 +1,510 @@
+/*
+ *   Copyright (c) 2019, Masatake YAMATO
+ *
+ *   This source code is released for free distribution under the terms of the
+ *   GNU General Public License version 2 or (at your option) any later version.
+ *
+ *   This module contains functions for generating tags for Moose perl extension.
+ *   https://metacpan.org/pod/Moose
+ */
+
+/* NOTE about kind/role design:
+ *
+ * sub foo { ... }
+ * after 'foo' => sub { ...}
+ *
+ * There were two ideas to capture 'foo':
+ *
+ * A: capturing 'foo' as a reference tag with 'method' kind and 'wrapped' role, and
+ * B: capturing 'foo' as a definition tag with 'wrapper' kind.
+ *
+ * This implementation takes idea B. */
+
+/*
+ *   INCLUDE FILES
+ */
+#include "general.h"  /* must always come first */
+
+#include "debug.h"
+#include "entry.h"
+#include "kind.h"
+#include "parse.h"
+#include "perl.h"
+#include "read.h"
+#include "routines.h"
+#include "trace.h"
+
+/*
+ *   DATA DECLARATIONS
+ */
+
+enum MooseKind {
+	K_CLASS,
+	K_METHOD,
+	K_ATTR,
+	K_WRAPPER,
+};
+
+static kindDefinition MooseKinds[] = {
+	{ true, 'c', "class", "classes"  },
+	{ true, 'm', "method", "methods" },
+	{ true, 'a', "attribute", "attributes" },
+	{ true, 'w', "wrapper", "wrappers" },
+};
+
+typedef enum {
+	F_WRAPPING,
+} MooseField;
+
+static fieldDefinition MooseFields [] = {
+	{
+		.name = "wrapping",
+		.description = "how a wrapper wrapping the method (around, after, or before)",
+		.enabled = true,
+	},
+};
+
+enum Wrapping {
+	W_UNKNOWN,
+	W_AROUND,
+	W_BEFORE,
+	W_AFTER,
+};
+
+static char * WrappingStrings[] = {
+	"unknown",
+	"around",
+	"before",
+	"after",
+};
+
+struct mooseSubparser {
+	perlSubparser perl;
+	bool notInMoose;
+	bool inPod;
+	int packageCork;
+	int classCork;
+	bool notContinuousExtendsLines;
+	const char *superClass;
+};
+
+/*
+ *   FUNCTION PROTOTYPES
+ */
+static void inputStart (subparser *s);
+static void inputEnd (subparser *s);
+static void makeTagEntryNotify (subparser *s, const tagEntryInfo *tag, int corkIndex);
+static void enteringPodNotify (perlSubparser *perl);
+static void leavingPodNotify  (perlSubparser *perl);
+
+/*
+ *   DATA DEFINITIONS
+ */
+
+static struct mooseSubparser mooseSubparser = {
+	.perl = {
+		.subparser = {
+			.direction  = SUBPARSER_BI_DIRECTION,
+			.inputStart = inputStart,
+			.inputEnd   = inputEnd,
+			.makeTagEntryNotify = makeTagEntryNotify,
+		},
+		.enteringPodNotify = enteringPodNotify,
+		.leavingPodNotify  = leavingPodNotify,
+	},
+};
+
+
+/*
+ *   FUNCTION DEFINITIONS
+ */
+
+static void inputStart (subparser *s)
+{
+	struct mooseSubparser *moose = (struct mooseSubparser *)s;
+
+	moose->notInMoose = true;
+	moose->packageCork = CORK_NIL;
+	moose->classCork = CORK_NIL;
+	moose->inPod = false;
+	moose->superClass = NULL;
+	moose->notContinuousExtendsLines = true;
+}
+
+static void inputEnd (subparser *s)
+{
+	struct mooseSubparser *moose = (struct mooseSubparser *)s;
+
+	if (moose->classCork != CORK_NIL)
+	{
+		tagEntryInfo *e = getEntryInCorkQueue (moose->classCork);
+		Assert (e);
+		e->extensionFields.endLine = getInputLineNumber ();
+	}
+
+	if (moose->superClass)
+		eFree ((char *)moose->superClass);
+}
+
+static void makeTagEntryNotify (subparser *s, const tagEntryInfo *tag, int corkIndex)
+{
+	perlSubparser *perl = (perlSubparser *)s;
+	struct mooseSubparser *moose = (struct mooseSubparser *)perl;
+
+	if (isTagExtraBitMarked(tag, XTAG_QUALIFIED_TAGS))
+		return;
+
+	if (tag->kindIndex == KIND_PERL_PACKAGE)
+		moose->packageCork = corkIndex;
+	else if ((!moose->notInMoose) && tag->kindIndex == KIND_PERL_SUBROUTINE)
+	{
+		tagEntryInfo moose_e;
+		initTagEntry (&moose_e, tag->name, K_METHOD);
+		moose_e.extensionFields.scopeIndex = moose->classCork;
+		makeTagEntry (&moose_e);
+	}
+}
+
+static void enteringPodNotify (perlSubparser *perl)
+{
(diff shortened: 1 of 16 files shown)


masatake:
A critical internal bug is fixed.

parsers/moose.c
```diff
@@ -0,0 +1,273 @@
...
+							  &mooseSubparser);
+	addLanguageCallbackRegex (language, "^ *no +Moose *;",
+							  "{exclusive}",
+							  leaveMoose, &mooseSubparser.notInMoose,
+							  &mooseSubparser);
+	addLanguageCallbackRegex (language, "^ *extends *'([^']+)';",
+							  "{exclusive}",
+							  extendClass, &mooseSubparser.notInMoose,
+							  &mooseSubparser);
+	addLanguageCallbackRegex (language, "^ *(has|after) *'([^']+)'",
```
EvanCarroll: If you're going to do `after` you may as well do `around` and `before` just to finish them off. They all fit into a class of Aspect Oriented Programming modifiers that wrap methods provided by the parent class.

parsers/moose.c
```diff
@@ -0,0 +1,273 @@
...
+}
+
+static void findMooseTags(void)
+{
+	scheduleRunningBaseparser (RUN_DEFAULT_SUBPARSERS);
+}
+
+static void initializeMooseParser (langType language)
+{
+	addLanguageCallbackRegex (language, "^ *use +Moose(:| *;)",
```
EvanCarroll: I'm not sure how this syntax works, but this sublanguage which we've been referring to in the ticket as "Moose et al" should also trigger on `Moo` which also provides all these same features. https://metacpan.org/pod/Moo#IMPORTED-SUBROUTINES

parsers/moose.c
EvanCarroll: As another special note, it seems from this regex that the `'` is required but in perl, they're not required (and they can and ofter are omitted entirely). As an added complexity they can be `"` (double quotes) too.

EvanCarroll: This seems to only cover the `has 'literal1' =>`, but there is another form the `has [qw/literal1 literal2/]` form which is a shorthand to define many attributes with the same property.

parsers/moose.c
masatake: Maybe just "use Moose;" is enough.  I should remove "Moose:".

I don't know Moo but I guess we should implement is as a different subparser running on Perl.
even if Moo and Moose subparesrs can share C functions and data types definitions.

parsers/moose.c
masatake: I see. I will add more commits for the comments.

parsers/moose.c
masatake: "Moose:" may have a sense when we support Rules.

EvanCarroll: Moo and Mouse essentially follow the same spec as Moose. So much so that I think they both naturally use Moose if it's there. (it becomes a noop) The primary reason for their existence is to shift work into compile time and eliminate work from run-time, but they coat-tail (follow very closely) the Moose syntax.

masatake:
Updated.
This one only supporst Moose.

parsers/moose.c
EvanCarroll: @masatake we're in a situation where I either have to duplicate the patch for a new subparser "Moo" that works that the same on the initial push. Or, we can add Moo to the trigger so it works on both. The difference between the two is pretty minimal. I don't see the purpose of maintaining two subparsers, but we are losing a lot of users of this because it's not yet catching Moo and Mouse. Please advise. If you want me to duplicate the parser (and thus have one subparser for each Moose we'll quickly stack up a few highly duplicate code bases).

EvanCarroll: New issue for support here https://github.com/universal-ctags/ctags/issues/2103

masatake: Do you mean extending following regex patterns is enough to support Moo and Mouse?
```
	addLanguageCallbackRegex (language, "^[ \t]*use +Moose *;",
							  "{exclusive}",
							  enterMoose, NULL,
							  &mooseSubparser);
	addLanguageCallbackRegex (language, "^[ \t]*no +Moose *;",
							  "{exclusive}",
							  leaveMoose, &mooseSubparser.notInMoose,
							  &mooseSubparser);
```
?

masatake: I think gathering common code to somewhere (let's call it perlMx) is better.
We can derive Moose, Mouse, and Moo from perlMx.
I understand Moose, Mouse and Moo has very similar syntax.

The question is the biggest difference betwwen them.
If really no diffrence between them, we should not take much time to make extra subparsers.
If there is, separating multipe subparsers has make sense to make potential cost of maintenance in the future minimal.

masatake: I found an interesting page:
https://blog.jdpfu.com/2011/12/02/of-moose-mouse-moo-and-mo-for-oo-perl

They are not the same.

masatake: As far as reading the page, I think we should build a small library for implementing Moose, Moo, Mouse and Mo(?).

We can write each parsers using the library.
Deciding the name for the library is the start point.

masatake: Having multiple object systems in a language. It looks a bit crazy situation. But I just remember "There's More Than One Way To Do It."
This is the perl way.

EvanCarroll: @masatake  they are not the same because of their implementation. The API into them is the same. For example, there would be no difference in the patch you've provided to support Moose and Moo other than the enter language line. In fact, in my use case here I don't even care about `Moose`. I've verified your code works by changing that Moose to `(Moose|Moo)`.

masatake: > I've verified your code works by changing that Moose to (Moose|Moo).

How about Mouse?

EvanCarroll: I think we can and **should** add Mouse to it. None of these libraries have any significant change in API. The differences are **extremely** minor, and all of these people benefit from your code.

Really 99% of the functional advantage of this patch is simply in the attribute tag with `has`. That's cool. It'll work on any Moose-like module build system. The remaining useful functionality is mostly the around/before/after method modifiers.

Mouse hasn't been patched since August of last year. Mouse exists so people can write modern perl objects into actual 1992 CGI scripts. It's not a very high visibility project and Perl as a whole is on the decline. I would enable the subparser for Mouse/Moo/Moose and it'll be good enough.

masatake: I see. Just extending the regex pattern may be enough. However, as a field value, `language:Moose` is reported for all 3 object systems. "Different things should be tagged differently", is the motto of u-ctags.
Reporting the language field differently for 3 object systems like "language:Moose", "language"Moo", and "language:Moo",  will be better in this case
See also #2079 and 2076.

How about the patch at the end of this page?

masatake:
How about this one?
```
diff --git a/parsers/moose.c b/parsers/moose.c
index 1d7aaf10..69ac4cb2 100644
--- a/parsers/moose.c
+++ b/parsers/moose.c
@@ -34,6 +34,16 @@
 #include "routines.h"
 #include "trace.h"
 
+#include <string.h>
+
+/*
+*   MACROS
+*/
+
+#define MOOSE_NAME "Moose"
+#define MOO_NAME   "Moo"
+#define MOUSE_NAME "Mouse"
+
 /*
  *   DATA DECLARATIONS
  */
@@ -447,7 +457,22 @@ static void findMooseTags (void)
 
 static void initializeMooseParser (langType language)
 {
-	addLanguageCallbackRegex (language, "^[ \t]*use +Moose *;",
+	const char *langName = getLanguageName (language);
+	const char *use_pattern;
+
+	Assert (langName);
+
+	if (strcmp (langName, MOOSE_NAME) == 0)
+		use_pattern = "^[ \t]*use +Moose *;";
+	else if (strcmp (langName, MOO_NAME) == 0)
+		use_pattern = "^[ \t]*use +Moo *;";
+	else if (strcmp (langName, MOUSE_NAME) == 0)
+		use_pattern = "^[ \t]*use +Mouse *;";
+	else
+		AssertNotReached ();
+
+
+	addLanguageCallbackRegex (language, use_pattern,
 							  "{exclusive}",
 							  enterMoose, NULL,
 							  &mooseSubparser);
@@ -485,16 +510,10 @@ static void initializeMooseParser (langType language)
 							  &mooseSubparser);
 }
 
-extern parserDefinition* MooseParser (void)
+static parserDefinition* perlMxParserNew (const char *name)
 {
-	parserDefinition* const def = parserNew("Moose");
-
-	static parserDependency dependencies [] = {
-		[0] = { DEPTYPE_SUBPARSER, "Perl", &mooseSubparser },
-	};
+	parserDefinition* const def = parserNew(name);
 
-	def->dependencies = dependencies;
-	def->dependencyCount = ARRAY_SIZE (dependencies);
 
 	def->kindTable = MooseKinds;
 	def->kindCount = ARRAY_SIZE(MooseKinds);
@@ -508,3 +527,36 @@ extern parserDefinition* MooseParser (void)
 
 	return def;
 }
+
+extern parserDefinition* MooseParser (void)
+{
+	parserDefinition* const def = perlMxParserNew (MOOSE_NAME);
+	static parserDependency dependencies [] = {
+		[0] = { DEPTYPE_SUBPARSER, "Perl", &mooseSubparser },
+	};
+	def->dependencies = dependencies;
+	def->dependencyCount = ARRAY_SIZE (dependencies);
+	return def;
+}
+
+extern parserDefinition* MooParser (void)
+{
+	parserDefinition* const def = perlMxParserNew (MOO_NAME);
+	static parserDependency dependencies [] = {
+		[0] = { DEPTYPE_SUBPARSER, "Perl", &mooseSubparser },
+	};
+	def->dependencies = dependencies;
+	def->dependencyCount = ARRAY_SIZE (dependencies);
+	return def;
+}
+
+extern parserDefinition* MouseParser (void)
+{
+	parserDefinition* const def = perlMxParserNew (MOUSE_NAME);
+	static parserDependency dependencies [] = {
+		[0] = { DEPTYPE_SUBPARSER, "Perl", &mooseSubparser },
+	};
+	def->dependencies = dependencies;
+	def->dependencyCount = ARRAY_SIZE (dependencies);
+	return def;
+}
```

EvanCarroll:
@masatake that looks good. I'm unconvinced of the merits of it but if you like that -- go for it. Do we make the same distinction about API-vs-implementation for things like Iron Python and cPython, or SQL (where the differences are more drastic) like PostgreSQL, SQL Server, and MySQL? It seems like this level of purity (though not without merit) is going to be a LOT of work.

masatake:
I found very different approach.
The reason I don't like just extending the regex pattern is that it hides whether input file uses Moose or Moo though ctags itself knows which module is used.
At the first time, I've tried to use "language:" field to let client tools know which module is used.
I'm not sure whether the original approach is over engineering or not.
However, I found I should do before using the field.
If perl parser captures modules specified in use built-in function, a client tool can know which module is used in the source file. Introducing module kind solves my concern indirectly.

PR #2071, opened 2019-04-03
add support for TTCN-3 Language Extensions: Advanced Parameterization

akrzyz:
add support for TTCN-3 Language Extensions: Advanced Parameterization
extension is described in: ETSI ES 202 784 V1.6.1

new Units are added:
 Units/parser-ttcn.r/ttcn-template-type.d
 Units/parser-ttcn.r/ttcn-template-template.d
 Units/parser-ttcn.r/ttcn-template-function.d
 Units/parser-ttcn.r/ttcn-template-variables-and-constants.d


parsers/ttcn.c
```diff
@@ -894,7 +897,7 @@ static int parseTypeDefBody (void)
 			/* SubTypeDef */
 			ungetToken();
 			/* Stop after type name, no need to parse ArrayDef, StringLength, etc */
-			return (parseType() && parseID(K_TYPE));
+			return parseType() && parseID(K_TYPE);
```
masatake: Do we need this change for supporting advanced parameterization?
At least this chnage should go to a separated commit.

parsers/ttcn.c
```diff
@@ -933,7 +936,9 @@ static void parseTTCN (void)
 			case T_TEMPLATE:
 				/* A.1.6.1.3 TemplateDef ::= "template" (Type | Signature) ID ... */
 				if (parseType() || parseSignature())
+                {
```
masatake: Do we need this change for supporting advanced parameterization?
At least this chnage should go to a separated commit.

parsers/ttcn.c
```diff
@@ -933,7 +936,9 @@ static void parseTTCN (void)
...
 					parseID(K_TEMPLATE);
+                }
```
masatake: Do we need this change for supporting advanced parameterization?
At least this chnage should go to a separated commit.

parsers/ttcn.c
akrzyz: we don't need this, i removed this changed and squashed commits into one.

masatake:
Thank you!

Issue #2072, opened 2019-04-04
ctags fails sandbox tests when compiled without json support


latricewilgus:
The sandbox tests fail if ctags was compiled without json support:

./autogen.sh
./configure --disable-json
make
make tmain UNITS=sandbox

failing tests are

 Tmain/gcov-sandbox.d
 Tmain/sandbox-crash.d
 Tmain/sandbox-default-req.d
 Tmain/sandbox-no-eager-guessing.d
 Tmain/sandbox-unknown-submode.d
 Tmain/sandbox-with-eager-guessing.d
 Tmain/sandbox.d

The failing message the tests fail with is: "Unknown option: --_interactive"
Indeed, the tests call ctags with the --_interactive option. Moreover, as stated in the ctags documentation for the feature interactive [0]:

    This feature [--_interactive] needs ctags to be built with json support and this requires libjansson to be installed at build-time.

Possible solution:
Check for the feature ``interactive`` in the above mentioned tests and skip them if not present.

[0] http://docs.ctags.io/en/latest/interactive-mode.html

PR #2073, opened 2019-04-04
Check for interactive feature in sandbox tests

latricewilgus:
The sandbox tests pass the --_interactive option to ctags, which is only present if ctags was built with json support. Add check for the interactive feature in the tests skipping the tests if the feature is not present.

Closes: #2072


masatake:
Good catch!

masatake:
Thank you!

latricewilgus:
Thank you for the quick response and the prompt merge!

Issue #1889, opened 2018-09-26
Incompatibity in command line handling


masatake:
E-ctags:
```
[jet@localhost]~/var/ctags% ctags  -o - foo.c --list-kinds=Sh          
f  functions
```

U-ctags:
```
[jet@localhost]~/var/ctags% ./ctags  -o - foo.c --list-kinds=Sh
ctags: Warning: cannot open input file "--list-kinds=Sh" : No such file or directory
a	foo.c	/^void a () {}$/;"	f	typeref:typename:void
b	foo.c	/^void b () {}$/;"	f	typeref:typename:void
```

hirooih:
This issue was fixed in #2859.

Issue #1883, opened 2018-09-22
ctags -L: not possible to specifiy files with leading dashes (handled as option)


blueyed:
The man page for `-L file` says:

> Options are also accepted in this input.

This is a problem with real files with uncommon names like `-X`:

```
% printf '%s\n'  '-X' | ctags -L -
ctags: Unknown option: -X
```

I've tried passing in `--` in front to ignore filenames like `-X`, but this results in:

```
% printf '%s\n' '--' '-X' | ctags -L -
ctags: Unknown option: --
```

I think `-L -` should either not accept/handle options (backward incompatible?), or skip option parsing after `--`.

My use case is using `rg --files | ctags --links=no -L -` to generate tags.

masatake:
How about using ./ as prefix like:

```
[yamato@master]~/var/ctags-github% cat ./-X                                 
#!/bin/sh
function foo
{
	:
}
[yamato@master]~/var/ctags-github% printf "./%s\n" -X | ./ctags -G -o - -L -
foo	./-X	/^function foo$/;"	f
```

blueyed:
My input is coming from `rg --files`, so not easy to change.

Even doing something like `(echo --; rg --files) | ctags --links=no -L -` to work around this (if `--` would be supported) is not really convenient.

I found it rather confusing that `-L -` takes options in the first place.

masatake:
```
diff --git a/main/options.c b/main/options.c
index 4a9d9cc6..f3e14a0b 100644
--- a/main/options.c
+++ b/main/options.c
@@ -115,6 +115,7 @@ typedef const struct sBooleanOption {
 *   DATA DEFINITIONS
 */
 
+/* Set true to this to stop option proessing.*/
 static bool NonOptionEncountered = false;
 static stringList *OptionFiles;
 
@@ -871,7 +872,9 @@ static void cArgRead (cookedArgs *const current)
                item = argItem (current->args);
                current->shortOptions = NULL;
                Assert (item != NULL);
-               if (strncmp (item, "--", (size_t) 2) == 0)
+
+               if ((!NonOptionEncountered)
+                       && (strncmp (item, "--", (size_t) 2) == 0))
                {
                        current->isOption = true;
                        current->longOption = true;
@@ -879,7 +882,8 @@ static void cArgRead (cookedArgs *const current)
                        Assert (current->item != NULL);
                        Assert (current->parameter != NULL);
                }
-               else if (*item == '-')
+               else if ((!NonOptionEncountered)
[...]

PR #2079, opened 2019-04-11
Cobol: introduce a parser specific parameter named "format" to switch cobol dialects about source code (based on #2067) format

masatake:
@b4n, could you look at the last commit of this PR?

I introduced a parameter for switching a format in a Cobol parser.
I introduced a CobolFixed parser.  There are Cobol, CobolFixed, CobolFree, and CobolVariable parsers.

Cobol parser is for u-ctags users.
The rest parsers are for Geany.

./ctags --list-languages just prints Cobol only. The rest is invisible. u-ctags users can specify an input format with `--param-Cobol:format=fixed|free|variable` option.

You can run one of three directly with /ctags --language-force=CobolX where X is one of Fixed, Free, or Variable. This means Geany can specify one of the explicitly.

I don't write a test for my change yes.

If it looks acceptable, please, cherry-pick the last one.


b4n:
Thinking about this a little more, I'm wondering if the argument is convenient at all for a library consumer… the thing is that it kind of affect *which* language is parsed, not really details of how like could be the `-D` for C.
So a library consumer would have to set the options *every single time* it calls the parser, in case another call changed the option: if the consumer want to parse e.g. fixed and free formats because the user has opened one file of each, then it has to set that parameter each time it switches from one to the other, or more simply each time it calls the parser.

It's not necessarily a problem, it's just an extra initialization that the caller ought to do in all cases, which to me sounds kind of like separate parsers.  From what I gathered grepping in the current code, the parser-specific arguments we have seem not quite similar to that.

What do you think?

masatake:
Even with the changes that I proposed in #2076 (or without them), the parsers for each format are there.
I didn't remove any parser in  #2076.
So the client like Geany can choose a proper Cobol parser from 3 format specific parsers.


Next, I will write about the command line interface (CLI).
I have not inspected the details but in the current u-ctags, but there is no way to switch values for the parameter per input file. But let's forget it.
* https://github.com/universal-ctags/ctags/issues/1889
* https://github.com/universal-ctags/ctags/issues/1883
* https://github.com/universal-ctags/ctags/pull/1884


I think having a format-neutral COBOL parser is better. Here, the format-neutral means the format used in the parser can be switched with the "format" parser specific parameter.
It seems that the format used in a COBOL source file can be specified in the file.
See "1.3.16. Format of Program Source Lines" in https://open-cobol.sourceforge.io/HTML/gnucobpg.html

I think CobolFixed, CobolFree, and CobolVariale should not handle the format specification in the file.
e.g. Even if ">> FORMAT IS FIXED" is specified, CobolVariable parser should ignore it because "Variable" is specified by the user running ctags.

The format-neutral COBOL parser can handle the source line format specification.
That means I have to add one more parameter: ```--param-Cobol:inline-source-format=[yes]|no```.
The parameter allows the COBOL parser to override the format that a user-specified with --param-Cobol:format.

We can remove the format-neutral COBOL parser by adding the inline-source-format parameter to the format-specific parsers ( CobolFixed, CobolFree, and CobolVariale). Instead of introducing the inline-source-format parameter, statically allowing the format-specific parsers to switch the format by the source line format specification is technically possible, of course.

b4n:
> I didn't remove any parser in #2076.
So the client like Geany can choose a proper Cobol parser from 3 format specific parsers.

I know they are still there for the moment, but I though I understood you planned on removing them once we had library API to set parameters.  And my comment was solely on this, whether the option made sense outside of the ctags CLI.  And my conclusion for the moment is that I'm not sure it makes much sense for a library, but I don't mind either.

> It seems that the format used in a COBOL source file can be specified in the file.
See "1.3.16. Format of Program Source Lines" in https://open-cobol.sourceforge.io/HTML/gnucobpg.html

Interesting, I didn't know that.

> I think CobolFixed, CobolFree, and CobolVariale should not handle the format specification in the file.

Whether a parser should not handle that if it was selected I don't know, it depends what a compiler would do.  I would myself expect the format lines to always be handled and have precedence over whatever is specified on the command line, and the command line acts just like a format line before the input.  But I'd have to check what's actually happening.
Depending on this, we should or should not have an extra option to ignore those.  Well, we could always have one, but it'd be a lot less useful.

masatake:
> So a library consumer would have to set the options every single time it calls the parser, in case another call changed the option: if the consumer want to parse e.g. fixed and free formats because the user has opened one file of each, then it has to set that parameter each time it switches from one to the other, or more simply each time it calls the parser.

A user has to choose a parser from 3 or has to set a format from 3. A user has to do either anyway.

I don't like to provide 3 parsers for Cobol in user-visible way.
I think this is not abnormal idea.
GNU Cobol compiler provides `-fixed` option. It doesn't provide a separate command for compiling fixed formatted input.  https://open-cobol.sourceforge.io/HTML/gnucobpg.html#index-Fixed-Format-Mode
A cobol editing mode for GNU emacs has a parameter named cobol-source-format
(https://gist.github.com/Edward-H/6768e7dc53ea3dd2adca ).

Even if we remove three format specific parsers, I think there is no impact on Geany users.
Following figures show what I mean.

cobol

masatake:
Rebased.

PR #2084, opened 2019-04-23
ActionScript: Fix and improve the parser

b4n:
The Flex parser has 2 parts: ActionScript (which is close to JavaScript) and MXLM (which is an XML dialect).  The Flex parser was built around an older version of the JavaScript parser, before the latter got a lot of fixes and overhauls.
This means I had to port a *lot* of fixes over from the JavaScript parser, so the approach I took in https://github.com/universal-ctags/ctags/commit/8bbb9d0d4177c07985ab4af12d746c33707dcaaa was to simply compare the two parsers at their current state and sync what can be.  This was a lot less painful than applying every single fix, with the conflicts due to the divergence and such.  So if one want to review the changes, it's interesting to check against jscript.c as well (and good luck).

After that "synchronizing" commit (it's actually sill different in some area, but a lot closer, and a lot less buggy), comes simpler improvements, fixing or adding support for specific syntax.  That part is fairly straightforward.

There is still room for improvements here, and the most obvious are:
* Make MXLM a proper separate parser that uses the ActionScript parser as a subparser;
* Share code between ActionScript and JavaScript parsers that are both ECMA-262-based languages (what does that mean when they don't have the same syntax?  I don't know, but the parsers sure do share a lot of code);
* Report more details (signature, visibility, inheritance…);
* Fix scope type properly.

I might work on some or all of those at some point, but I don't have time right now, but needed a working ActionScript parser to replace Geany's previous regex one, which to my knowledge is now the case, within its limits.

### DIFF
--- a/Tmain/list-roles-with-kind-names.d/stdout-expected.txt
+++ b/Tmain/list-roles-with-kind-names.d/stdout-expected.txt
@@ -24,6 +24,7 @@ CPreProcessor  h/header    local     on      local header
 CPreProcessor  h/header    system    on      system header
 CUDA           h/header    local     on      local header
 CUDA           h/header    system    on      system header
+Flex           I/import    import    on      imports
 M4             I/macrofile included  on      included macro
 M4             I/macrofile sincluded on      silently included macro
 Make           I/makefile  included  on      included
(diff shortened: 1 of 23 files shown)


masatake:
(reviewing now)

masatake:
Could you change the prefix of the commit log of 9a9998e  to "flex" from "ActionScript".

masatake:
It seems that my last comment is pointless.

masatake:
Thank you. Sorry to be late.

PR #2085, opened 2019-04-25
Sample client app using ctags as a library plus all the related changes in ctags

techee:
This patch demonstrates how we use ctags in Geany as a library. The
patch tries to be minimalistic to demonstrate only the things we
need in ctags so it just directly replaces the ctags code with what
we need and the generated "ctags" binary is the demo app. For this
to be mergeable to ctags, there have to be some ifdefs specifying
whether ctags should be compiled as a library and mini-geany has
to be a separate binary.

@masatake Here you are - this should give you a pretty detailed idea about how we use ctags in Geany. I'll definitely welcome any suggestions if you think some of the ctags changes can be done in a better way - you are definitely more familiar with it than I am and I could have missed some things.

For anyone interested in the related discussion in Geany, please see https://github.com/geany/geany/issues/2119.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -156,7 +156,7 @@ extern const char *tagFileName (void)
 
 extern void abort_if_ferror(MIO *const mio)
 {
-	if (mio_error (mio))
+	if (mio != NULL && mio_error (mio))
 		error (FATAL | PERROR, "cannot write tag file");
 }
 
@@ -1688,12 +1688,14 @@ extern void invalidatePatternCache(void)
 
 extern void tagFilePosition (MIOPos *p)
 {
-	mio_getpos (TagFile.mio, p);
+	if (TagFile.mio)
+		mio_getpos (TagFile.mio, p);
 }
 
 extern void setTagFilePosition (MIOPos *p)
 {
-	mio_setpos (TagFile.mio, p);
+	if (TagFile.mio)
+		mio_setpos (TagFile.mio, p);
 }
 
 extern const char* getTagFileDirectory (void)
--- a/main/mini-geany.c
+++ b/main/mini-geany.c
@@ -1,0 +1,346 @@
+/*
+*   Copyright (c) 2019, Jiri Techet
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   Provides a simple application using ctags as a library and using the same
+*   set of ctags functions as the Geany editor
+*/
+
+#include "general.h"  /* must always come first */
+
+#include "types.h"
+#include "routines.h"
+#include "mio.h"
+#include "error_p.h"
+#include "writer_p.h"
+#include "parse_p.h"
+#include "options_p.h"
+#include "trashbox_p.h"
+#include "field_p.h"
+#include "xtag_p.h"
+#include "entry_p.h"
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+static int writeEntry (tagWriter *writer, MIO * mio, const tagEntryInfo *const tag);
+static void rescanFailed (tagWriter *writer, unsigned long validTagNum);
+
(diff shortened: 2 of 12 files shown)


main/parse.c
```diff
@@ -3344,6 +3344,7 @@ static bool createTagsWithFallback1 (const langType language,
 			*/
 			setTagFilePosition (&tagfpos);
 			setNumTagsAdded (numTags);
+			writerRescanFailed (numTags);
```
techee: @masatake Am I right here to assume that numTags here contains the number of tags added to the file before the current scan (which failed)? This is how I use it in the client and the latest patches in Geany and if it isn't correct, re-parsing won't work correctly.

main/entry.c
```diff
@@ -156,7 +156,7 @@ extern const char *tagFileName (void)
 
 extern void abort_if_ferror(MIO *const mio)
 {
-	if (mio_error (mio))
+	if (mio != NULL && mio_error (mio))
```
techee: These NULL checks are here because we bypass opening the MIO which is always NULL in our case.

main/writer.c
```diff
@@ -27,6 +27,11 @@ static tagWriter *writerTable [WRITER_COUNT] = {
 
 static tagWriter *writer;
 
+extern void setCustomTagWriter(tagWriter *w)
```
techee: Currently this isn't very clean because we have to re-use some of the predefined parser types - the demo uses WRITER_U_CTAGS but there should be some support for custom writer type.

main/mini-geany.c
```diff
@@ -0,0 +1,346 @@
...
+		{
+			kindDefinition *def = getLanguageKind(lang, kind);
+			enableKind(def, true);
+		}
+	}
+}
+
+
+/* we need to be able to initialize ctags like in main() but skipping some things */
+static void ctagsInit(void)
```
techee: It might make sense to have this function directly in ctags (or at least some part of it).

main/mini-geany.c
```diff
@@ -0,0 +1,346 @@
...
+	{
+		unsigned int kindNum = countLanguageKinds(lang);
+		unsigned int kind;
+
+		for (kind = 0; kind < kindNum; kind++)
+		{
+			kindDefinition *def = getLanguageKind(lang, kind);
+			enableKind(def, true);
+		}
+	}
```
b4n: I actually hit a problem in current Geany master withe way it was done, and that would solve it: in Geany master the kind is not enabled, but the check to not write tags that are disabled is bypassed.  This looks reasonable but isn't always, as some parsers check the value *before* emitting the tags.  Properly enabling the tags like that would fix the issue.

main/mini-geany.c
```diff
@@ -0,0 +1,346 @@
...
+	initXtagObjects ();
+
+	initializeParsing ();
+	initOptions ();
+
+	/* make sure all parsers are initialized */
+	initializeParser (LANG_AUTO);
+
+	/* change default value which is false */
+	enableXtag(XTAG_TAGS_GENERATED_BY_GUEST_PARSERS, true);
```
b4n: I think we should also enable en reference tags (`XTAG_REFERENCE_TAGS`) as those are used for more and more for includes & al.

main/mini-geany.c
techee: Yeah, when I added this in Geany, it really did lead to generating some extra tags: https://github.com/geany/geany/pull/2132/commits/1bdaa522af227e77f653b0d35151e4b5a540ea94

masatake:
(I'm working on this topic.)

main/parse.c
masatake: You are right.

main/entry.c
masatake: We have to recondier the output stream. More abstracted interfaces may be needed. Temporary, I use this code.

masatake:
Merged via #2088.

PR #2086, opened 2019-04-25
Fix warning about unused function

techee:
parseConfigurationFileOptionsInDirectoryWithLeafname() is only used
by the code protected by the ifdefs so when the code inside ifdefs
isn't compiled, we get a warning about unused function. Move it inside
the ifdef section.


masatake:
Thank you,

PR #2087, opened 2019-04-27
Move CTagsSelfTestParser after PARSER_LIST

techee:
Unless this causes some problems for uctags, it would be nice if
CTagsSelfTestParser could be moved after PARSER_LIST because for
Geany this means we have to shift parsers by 1 compared to what
is inside PARSER_LIST.

Without this change we either have to keep a diff against uctags by removing CTagsSelfTestParser or shift lang by 1 like in

https://github.com/geany/geany/pull/2132/commits/3de0df0872e237d955e291a27db913e650e8b8a9

which is a bit error-prone because it has to happen at multiple places.

### DIFF
--- a/main/parse.c
+++ b/main/parse.c
@@ -121,7 +121,6 @@ static void uninstallTagXpathTable (const langType language);
 */
 static parserDefinition *CTagsSelfTestParser (void);
 static parserDefinitionFunc* BuiltInParsers[] = {
-	CTagsSelfTestParser,
 	PARSER_LIST,
 	XML_PARSER_LIST
 #ifdef HAVE_LIBXML
@@ -131,10 +130,11 @@ static parserDefinitionFunc* BuiltInParsers[] = {
 #ifdef HAVE_LIBYAML
 	,
 #endif
-       PEG_PARSER_LIST
+	PEG_PARSER_LIST
 #ifdef HAVE_PACKCC
-       ,
+	,
 #endif
+	CTagsSelfTestParser
 };
 static parserObject* LanguageTable = NULL;
 static unsigned int LanguageCount = 0;


techee:
As an alternative, CTagsSelfTestParser could be added into PARSER_LIST - we wouldn't use it in Geany but the indexes of parsers would be identical.

techee:
After thinking about it a little more, the right thing IMO is to move the test parser into the parser
list. With the previous approach countParsers() would still return the wrong value in Geany and if we in the future started using e.g. XML parsers, we'd run in the same problem.

See the commit message for more details why this might be good for ctags too.

masatake:
CTagsSelfTestParser was put to parse.c to allow it to access internal of parse.c.
Moving it to a file other than parse.c is not good idea.

I think Geany and ctags should manages their own language ids. A design using identical ids between the two softwares makes the two software couple too tightly. As you did in manaing kind in Geany, the management of language ids should be separated.

I think Geany should have a map between ctags language ids and geany language ids.
Sharing names of languages may be o.k.

If there is an issue making such map, let us extend the library API.

> countParsers() would still return the wrong values

I would like to know more.

countParsers() can returns N. Geany may scan parsers of ctags for 0 to N. Geany may choose acceptable ones during parsing. Allow me to use pseudo code.
```
int N = countParsers();
for (int i = 0 ; i < N; i++)
{
   Parser *parser = ctagsGetParser(i);
   if (geanyAcceptableCtagsParser (parrser))
   {
      geanyRegisterCtagsparser(parser);
   }
}

bool
geanyAcecptableCtagsParser (parser)
{
   if (getName(parser) == "CtgsSelfTestParser")
      return 0;
   else if (isXMLBased(parser))
      return 0;
   ...
   else
      return 1;
}
```

techee:
Yeah, it's this mapping I was trying to avoid. We basically just have a different list of parsers inside parsers.h and even for the future it would be easiest for us to have parsers we are using defined this way (and thus introducing a small diff against ctags). Doing some mapping to/from ctags/geany is definitely possible, I'm just not sure if it's worth it, especially when we probably don't want to compile all ctags parsers but only those we need in Geany - then PARSER_LIST will have to be different anyway.

@b4n What do you think about this?

This means when we keep redefining PARSER_LIST, the CTagsSelfTestParser at the first position inside parsers.c will shift the list by 1 and we'll have to remap it inside Geany. Again, to super-big issue, we can do this inside our code, I was just hoping to get rid of it.

masatake:
I think I don't understand the real requirement.
However, exporting PARSER_LIST from ctags may be not good idea.

As a temporary solution, your original change,https://github.com/universal-ctags/ctags/commit/fb0fad8b256144c03901e460cb9569a1cb71ad93, looks good.
How about moving CTagsSelfTestParser at the end of BuiltInParsers?
You moved it to the middle of BuiltInParsers.

techee:
> I think I don't understand the real requirement.

In Geany we identify parsers of individual languages using this enum:

https://github.com/geany/geany/blob/master/src/tagmanager/tm_parser.h#L58

We keep this enum in sync with

https://github.com/geany/geany/blob/master/ctags/main/parsers.h

so the value of the enum is the index of the parser in ctags (currently we don't have CTagsSelfTestParser so this works). With CTagsSelfTestParser at the first position in parse.c, the values get shifted by 1 because of this which then has to be corrected - see the CTAGS_LANG() and GEANY_LANG() macros which do that here:

https://github.com/geany/geany/commit/3de0df0872e237d955e291a27db913e650e8b8a9

Basically I just wanted to avoid this shifting (or some other mapping of Geany parser indexes to ctags parser indexes), that's all.

> As a temporary solution, your original change,fb0fad8, looks good.
> How about moving CTagsSelfTestParser at the end of BuiltInParsers?

Yep, sounds good. This way we'll get rid of most of the CTAGS_LANG() and GEANY_LANG() macros. The only one remaining change we'll have to do will be countParsers() which we'll have to decrement by 1 because of the extra CTagsSelfTestParser but that's it.

We could also add some TM_CTAGS_SELF_TEST at the beginning of our enum in Geany

https://github.com/geany/geany/blob/master/src/tagmanager/tm_parser.h#L58

which we wouldn't use but which would help us ho have correct parser indexes, but the problem is that this would probably break ABI with our plugins if they were compiled against an older version of Geany.

masatake:
...I'm sorry but it seems that I don't understand the essential requirement.

I got a fundamental question.
Which software ctags or geany decides which languages Geany supports?
     If the answer is Geany, Geany should have its own languages list.
     (I think Geany should decide.)
    If the ansser is Ctags, you should the list of languages dynamically via the API interface
    defined in ctags-api.h.
The API interface declares enough functions to get the list of languages ctags supports.

techee:
> Which software ctags or geany decides which languages Geany supports? If the answer is Geany, Geany should have its own languages list.

It's a hybrid - we have our own parser identifier list in

https://github.com/geany/geany/blob/master/src/tagmanager/tm_parser.h#L58

but based on this list we modify PARSER_LIST inside ctags parsers.h to match the enums (since we don't want to compile the parsers of languages for which we don't have support in Geany, we need to modify parsers.h anyway and remove those parsers in which we are not interested). The enum values shouldn't be modified because this would break our plugin ABI.

So the problem is this - for instance, for C the enum identifier is TM_PARSER_C and its value is 0. Inside PARSER_LIST we have the C parser at the first position so without CTagsSelfTestParser its ctags identifier would also be 0 and everything would work. But because of CTagsSelfTestParser at the first position, all the identifiers have to be shifted by 1.

Again, we can do this shifting by ourselves, it would just be nicer to avoid it by e.g. moving CTagsSelfTestParser to the very end of the list.

masatake:
I see. You are talking about building time issue. Give me time. I would like think of what I can do without breaking ABI.

techee:
@masatake For defining only the parsers we are interested in, it is a build time issue. But to "assign" them to our enum values (which is done by just simply having the parsers inside PARSER_LIST in the same order as in our enum), it is a runtime issue so e.g. our enum TM_PARSER_C has to match the C parser index in ctags.

Anyway, I don't think you have to do anything complicated in ctags because of this. I think it will be perfectly enough for us to move CTagsSelfTestParser to the end of the parser list inside parse.c if it's OK with you. I'll prepare a patch with this.

techee:
> @masatake For defining only the parsers we are interested in, it is a build time issue. But to "assign" them to our enum values (which is done by just simply having the parsers inside PARSER_LIST in the same order as in our enum), it is a runtime issue so e.g. our enum TM_PARSER_C has to match the C parser index in ctags.

Just to be clear, yes, the whole assignment is done at the build time.

masatake:
How about following apraoch?

When building libctags in Geany, run configure with a special option:
```
./configure --enable-external-parser-list=\"../geany-parser-list.h\"
```
The value passed via the option is set to EXTERNAL_PARSER_LIST macro.

```
diff --git a/main/parse.c b/main/parse.c
index 8fbb7148..2cb39c04 100644
--- a/main/parse.c
+++ b/main/parse.c
@@ -121,6 +121,9 @@ static void uninstallTagXpathTable (const langType language);
 */
 static parserDefinition *CTagsSelfTestParser (void);
 static parserDefinitionFunc* BuiltInParsers[] = {
+#ifdef EXTERNAL_PARSER_LIST
+#include EXTERNAL_PARSER_LIST
+#else  /* ! EXTERNAL_PARSER_LIST */
 	CTagsSelfTestParser,
 	PARSER_LIST,
 	XML_PARSER_LIST
@@ -135,6 +138,7 @@ static parserDefinitionFunc* BuiltInParsers[] = {
 #ifdef HAVE_PACKCC
        ,
 #endif
+#endif	/* EXTERNAL_PARSER_LIST */
 };
 static parserObject* LanguageTable = NULL;
 static unsigned int LanguageCount = 0;
```

techee:
@masatake That would actually be great! This way we could store our parser list completely outside the ctags directory and could keep parsers_p.h as it is in ctags.

In fact, we don't really run ctags configure - we just add the necessary defines into our own configure and there's just this single configure used for building both Geany and the ctags version of Geany. So the above patch of yours would be just all we need - we'll define EXTERNAL_PARSER_LIST by ourselves.

masatake:
O.K. I will implement what I wrote here.

techee:
Closing in favor of #2101

PR #2090, opened 2019-05-07
Typescript: performance improvements

ksamborski:
Typescript parser was originally about 10x slower than other parsers. Main performance bottlenecks seemed to be memory related. Here is what I did:
- parser states are now merged into one union type
- multiple single char parsers are merged into one
- unwinding markers stack implementation is changed to use only one `ptrArray` so there is much less allocation and deallocation of memory
- accepting any char is done now by simple ignore flag instead of using special parser
- white chars are skipped outside of the parsers

After applying all of those changes typescript parser is now 5x slower than others. @masatake could you look at this?

### DIFF
--- a/main/unwindi.c
+++ b/main/unwindi.c
@@ -203,21 +203,32 @@ CTAGS_INLINE MIOPos uugcGetFilePosition (void)
 		return getInputFilePosition ();
 }
 
-static ptrArray *uwiMarkerStack;
-static ptrArray *uwiCurrentMarker;
-extern void uwiActivate   (void)
+static ptrArray *uwiBuffer;
+static unsigned int *uwiMarkerStack;
+static unsigned int uwiMarkerStackLength;
+static unsigned int *uwiCurrentMarker;
+
+extern void uwiActivate (unsigned int stackLength)
 {
-	Assert (!uwiMarkerStack);
+	Assert (stackLength > 0);
 
 	uugcActivate ();
-	uwiMarkerStack = ptrArrayNew ((ptrArrayDeleteFunc)ptrArrayDelete);
+	uwiBuffer = ptrArrayNew ((ptrArrayDeleteFunc)uugcDeleteC);
+	uwiMarkerStackLength = stackLength;
+	uwiMarkerStack = xMalloc (stackLength, unsigned int);
+	uwiCurrentMarker = NULL;
 }
 
 extern void uwiDeactivate (void)
 {
+	Assert (uwiBuffer);
 	Assert (uwiMarkerStack);
-	ptrArrayDelete (uwiMarkerStack);
+
+	ptrArrayDelete (uwiBuffer);
+	eFree (uwiMarkerStack);
+	uwiBuffer = NULL;
 	uwiMarkerStack = NULL;
+	uwiMarkerStackLength = 0;
 	uugcDeactive();
 }
 
@@ -229,7 +240,10 @@ extern int uwiGetC ()
 	c = chr->c;
 
 	if (uwiCurrentMarker)
-		ptrArrayAdd (uwiCurrentMarker, chr);
+	{
+		*uwiCurrentMarker += 1;
+		ptrArrayAdd (uwiBuffer, chr);
+	}
 	else
 	{
 		uugcCurrentChar = NULL;
@@ -241,7 +255,7 @@ extern int uwiGetC ()
 
 extern void uwiUngetC (int c)
 {
-	//Assert (!uwiCurrentMarker);
+	Assert (!uwiCurrentMarker);
 	uugcInjectC (c);
 }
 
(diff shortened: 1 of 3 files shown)


masatake:
(reviewing in progress)

masatake:
minor issue: initParseWordState is removed in 91c49b6ac4a74cbf76583e96d8927805f8c7d199. However, it should be removed in f658e2de057d071357cc3fab0f01a8ce533faf07.

main/unwindi.c
```diff
@@ -203,21 +203,23 @@ CTAGS_INLINE MIOPos uugcGetFilePosition (void)
 		return getInputFilePosition ();
 }
 
-static ptrArray *uwiMarkerStack;
-static ptrArray *uwiCurrentMarker;
+static ptrArray *uwiBuffer;
+static unsigned int uwiMarkerStack[256];
```
masatake: Is this (256) is safe enough?
If the answer is yes, we don't need
```
Assert (uwiCurrentMarker - uwiMarkerStack < 255);
```

Consider ctags runs as part of a network service. I'm afraid this 256 may help an attacher design an evil input.

main/unwindi.c
```diff
@@ -257,37 +262,33 @@ extern MIOPos uwiGetFilePosition (void)
 
 extern void uwiPushMarker (void)
 {
-	Assert (uwiMarkerStack);
+	Assert (uwiCurrentMarker - uwiMarkerStack < 255);
```
masatake: Assert is enable only when ctags is built with './configure --enable-debugging`.
What happens if a crafted input causes calling uwiPushMaker repeatedly more than 255 times.

How about following code?
```
extern void uwiPushMarker (void)
{
    if (uwiCurrentMarker - uwiMarkerStack >= 255)
   {
      warning (...);
       return;
    ...
}

File: main/unwindi.c [hunk 3]
```diff
@@ -257,37 +262,33 @@ extern MIOPos uwiGetFilePosition (void)
...
-	ptrArrayDelete (uwiCurrentMarker);
-
-	uwiCurrentMarker = NULL;
-	if (ptrArrayCount (uwiMarkerStack) > 0)
-	{
-		uwiCurrentMarker = ptrArrayLast (uwiMarkerStack);
-		ptrArrayRemoveLast (uwiMarkerStack);
-	}
+	if (uwiCurrentMarker == uwiMarkerStack) uwiCurrentMarker = NULL;
+	else uwiCurrentMarker--;
```
masatake: If you introduce the code I showed above comment, push an pop calls become unbalanced.
As the result, `uwiCurrentMarker - uwiMarkerStack` can be less than 0.

Depending Assert in logic is a bad idea.

Let's handle the condition `>=0' explicitly here.

main/unwindi.c
masatake: 256 is hardcoded. At least, it should be defined as a macro.
So you can reuse the macro in other areas in this source file.

An alternative approach is that passing the limit to uwi layer from client side (typescript.c) via uwiActivate like
```
   uwiActivate(512);
```

ksamborski: > Is this (256) is safe enough?

It depends on the parser implementation. Asserts here only aim to help the developer with debugging.

> 
> Consider ctags runs as part of a network service. I'm afraid this 256 may help an attacher design an evil input.

You're probably right but the attack would be more related to how parser works.

main/unwindi.c
ksamborski: I'm not sure. Asserts help only developer but if he couldn't predict such situation the result would be garbage anyway even with your proposed change.

Having said that, a warning still could be useful for user. It could say something like "Fatal error, please consider adding an issue: <url>".

main/unwindi.c
ksamborski: Ok.

main/unwindi.c
ksamborski: > 256 is hardcoded. At least, it should be defined as a macro.
> So you can reuse the macro in other areas in this source file.

Ok, macro is a good idea!

> An alternative approach is that passing the limit to uwi layer from client side (typescript.c) via uwiActivate like
> 
> ```
>    uwiActivate(512);
> ```

And that's even better idea, thanks!

main/unwindi.c
masatake: The result becoming garbage is much better than allowing buffer overflow alike attack.
```
extern void uwiPushMarker (void)
{
    if (uwiCurrentMarker - uwiMarkerStack >= 255)
   {
      warning (...);
       exit (n);
    ...
}
```
is one of an idea. However, nowadays ctags can be used as a library, exiting when parsing unexpected input is not a good idea. even about the command line interface, aborting parsing is unwanted. Consider the following command line:
```
$ ctags badinput.ts goodinput0.ts goodinput1.ts ...
```
If ctags exits during parsing badinput.ts, tags for goodinput files are never generated.

About warning, please include the current input file name to the message. So we can ask the user reporting an issue about overflow to send or show the input file.

ctags must be robust to unexpected input! (This sentence is for myself.)

ksamborski:
@masatake I've added changes that you asked for. Please check them out.

main/unwindi.c
```diff
@@ -262,7 +271,12 @@ extern MIOPos uwiGetFilePosition (void)
 
 extern void uwiPushMarker (void)
 {
-	Assert (uwiCurrentMarker - uwiMarkerStack < 255);
+	if (uwiCurrentMarker - uwiMarkerStack >= ( uwiMarkerStackLength - 1 )) {
+		error (WARNING,
+			"unwindi.c: trying to add too many markers "
+			"(this is a bug, please consider filing an issue)");
+		uwiCurrentMarker = NULL;
+	}
```
masatake:
```
		error (WARNING,
			"unwindi.c: trying to add too many markers during  parsing: %s"
			"(this is a bug, please consider filing an issue)", getInputFileName());
```

main/unwindi.c
```diff
@@ -272,7 +286,15 @@ extern void uwiPushMarker (void)
 
 extern void uwiPopMarker (const int upto, const bool revertChars)
 {
-	Assert (uwiCurrentMarker - uwiMarkerStack >= 0);
+	if (uwiCurrentMarker - uwiMarkerStack < 0) {
+		error (WARNING,
+				"unwindi.c: trying to drop to many markers "
+				"(this is a bug, please consider filing an issue)");
+
```
masatake:
```
	error (WARNING,
				"unwindi.c: trying to drop to many markers: %s"
				"(this is a bug, please consider filing an issue)", getInputFileName());
```

main/unwindi.c
masatake: My comment was wrong. The source file of ctags is not needed. Instead, the INPUT source file must be printed.

masatake:
Feel free to squash, rebase, and push --force.

main/unwindi.c
ksamborski: OK, I'll fix it.

masatake:
7054fe1 is already merged as fc3a3e536181c0db98a6ae82ace8882a198cb023.
If possible, I would like you to unify 5dc5fb and baf568b.

ksamborski:
I did some cleaning in my commits and fixed the error message.

masatake:
Looks strange.

What I do daily is:
```
git stash (If you have changes not commited yet)
git checkout master
git pull upstream master:master
git checkout typescript-perf
git rebase master
git rebase -i master (to rearrange commits)
...
git push origin  typescript-perf: typescript-perf
```

My .git/config:
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[remote "origin"]
	url = git@github.com:masatake/ctags.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[remote "upstream"]
#	url = https://github.com/fishman/ctags.git
	url = git@github.com:fishman/ctags.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
```

ksamborski:
I did more or less the same as you. I moved all of my commits on top. Does it look better now?

masatake:
Yes!

masatake:
Thank you. Please, merge this.

PR #2091, opened 2019-05-10
Make make dist work

masatake:


kraythen:
Is the project maintained? Lots of PRs unacted on back to 2016

Issue #2093, opened 2019-05-12
Use webhook to update documents at ctags.io


masatake:
I found that ctags.io had not been updated for 4 months because 'GitHub service' feature is gone.
We have to use webhook feature instead.

@cweagans, could you look at this? 
https://docs.readthedocs.io/en/latest/webhooks.html#webhook-creation

I tried updating but I could not because I am not the maintainer.
If possible, I would like you to add me as a member of the maintainer group.

cweagans:
@masatake What is your Read The Docs username? I can add you.

cweagans:
I don't have admin access to the Github repo so I can't set up the webhook, but I'm happy to hand out access to the RTD account so that you can do it!

masatake:
> I  don't have admin access to the Github repo so I can't set up the  webhook, but I'm happy to hand out access to the RTD account so that you  can do it!

I login RTD via github. So I guess my account name is "masatake".

cweagans:
I have added you.

masatake:
Fixed! Thank you.

PR #2094, opened 2019-05-13
Added new parser for Inko

YorickPeterse:
This adds a new ctags parser that can parse Inko
(https://inko-lang.org/) source code.

I based this on the commit that added the systemtap parser (https://github.com/universal-ctags/ctags/commit/6bc1a8acb6ee788cfa21a39a8d6b8f4496d53f2b#diff-3dd5375cb7338d3815ab63cd09ddff93), as the documentation about adding a parser (http://docs.ctags.io/en/latest/optlib.html?highlight=optlib#submitting-an-optlib-file-to-the-universal-ctags-project) appears to be a bit out of date (e.g. the PRELOAD variables appear to not be used).


YorickPeterse:
@masatake What do you think of these changes? Should I take a different approach for adding the parser, or is this how they are supposed to be added?

masatake:
Please, add following two items to the header of inko.ctags:
(1) Copyright notice, and
(2) the reference to the language (https://inko-lang.org ?)

Units/parser-inko.r/everything.d/expected.tags
```diff
@@ -0,0 +1,18 @@
...
+global	input.inko	/^def global {}$/;"	m
+SomeTrait	input.inko	/^trait SomeTrait {$/;"	t
+Socket	input.inko	/^object Socket impl Foo {$/;"	o
+quack	input.inko	/^  def quack {}$/;"	m	object:Socket
+Socket	input.inko	/^impl Bar for Socket {$/;"	r	implements:Bar
+moo	input.inko	/^  def moo !! Integer {}$/;"	m	reopen:Socket
+Chickens	input.inko	/^impl Chickens {$/;"	r
+bok_bok	input.inko	/^  def bok_bok {}$/;"	m	reopen:Chickens
+NUMBER	input.inko	/^let NUMBER = 10$/;"	c
+@number	input.inko	/^let @number = 20$/;"	a
```
masatake: Do you think `@` is the part of the attribute name?
(I'm not sure.)

optlib/inko.ctags
```diff
@@ -0,0 +1,57 @@
+--langdef=Inko
+--map-Inko=+.inko
+--kinddef-Inko=o,object,Object definition
+--kinddef-Inko=m,method,Method definition
+--kinddef-Inko=t,trait,Trait definition
+--kinddef-Inko=a,attribute,Attribute definition
+--kinddef-Inko=c,constant,Constant definition
+--kinddef-Inko=r,reopen,Reopen object
+
+--_fielddef-Inko=implements,Trait being implemented
```
masatake: `ctags` has a built in field named `inherits`.
`inherits` is not suitbale here?

Java has an interface. However, it uses inherits to store the "implements" information to "inherits" lile:


```
[yamato@slave tmp]$ cat /tmp/foo.java
cat /tmp/foo.java

class foo extends Object implements Cloneable {
}
[yamato@slave tmp]$ u-ctags -o - --fields='*' foo.java 
u-ctags -o - --fields='*' foo.java 
foo	foo.java	/^class foo extends Object implements Cloneable {$/;"	kind:class	line:2	language:Java	inherits:Object,Cloneable	roles:def	end:3
```

YorickPeterse:
@masatake Oops! I initially had the copyright in place, but must have accidentally removed it before committing my changes. I added it, along with a summary plus link similar to the systemtap parser.

optlib/inko.ctags
```diff
@@ -0,0 +1,79 @@
...
+--_fielddef-Inko=signature,Method signature
```
masatake: It seems that this signature field is not used anywhere.
If you may have a plan to implement it. you can add this definition in the future.

masatake:
If you add a new parser specific field, you must update the expected output of some test cases.

masatake:
https://docs.ctags.io is not updated. See https://github.com/universal-ctags/ctags/issues/2093 .
So read rst files under docs/ directory directly.

masatake:
Other than some minor issues, it looks good for me.

Units/parser-inko.r/everything.d/expected.tags
YorickPeterse: Yes, the language treats them as being part of the name (`@foo` and `foo` are thus different attributes).

optlib/inko.ctags
YorickPeterse: I suppose this depends a bit on personal tastes. When a trait is implemented, it is not inherited from. Instead, any default methods it may provide are copied into the object that implements the trait. This means "implements" would be more accurate than "inherits".

YorickPeterse:
@masatake 

> https://docs.ctags.io is not updated. See #2093 .
> So read rst files under docs/ directory directly.

What exactly needs to be updated? I didn't see any references there for systemtap, so I'm not sure what to use as a reference.

masatake:
>> https://docs.ctags.io is not updated. See #2093 .
    So read rst files under docs/ directory directly.

> What exactly needs to be updated? I didn't see any references there for systemtap, so I'm not sure what to use as a reference.

Sorry, no update is needed. It is just information.

optlib/inko.ctags
masatake: I see. What you wrote may be applicable to other languages. I wonder what I should do.
Anyway, it is nothing to do with your parser.

masatake:
```
impl Bar for Socket {
  def moo !! Integer {}
}
```
I'm thiking about what should be tagged in above code.
I remember #1604 and #1604.

masatake:
Can a real Inko processor accept your test cases as valid inputs?
See http://docs.ctags.io/en/latest/input-validation.html?highlight=input%20validation

Don't you want full qualified tag?
http://docs.ctags.io/en/latest/optlib.html#generating-fully-qualified-tags-automatically-from-scope-information

Running ctags with --verbose option, you can see statistics about which regex patterns in tables are matched (and not matched) frequently. You can use it for optimization. 
https://github.com/universal-ctags/ctags/issues/1912

You can work on the above items after merging :-)

YorickPeterse:
> I'm thiking about what should be tagged in above code.
I remember #1604 and #1604.

A trait can only be implemented once for an object, so I don't think this would be necessary.

> Can a real Inko processor accept your test cases as valid inputs?
See http://docs.ctags.io/en/latest/input-validation.html?highlight=input%20validation

Inko does not provide a way of _just_ syntax checking a file at the moment, but all existing input examples are valid syntax.

> Don't you want full qualified tag?

I personally haven't really had a use for it so far, and only `optlib/puppetManifest.ctags` seems to be using it.

> Running ctags with --verbose option, you can see statistics about which regex patterns in tables are matched (and not matched) frequently. You can use it for optimization.

I tested it with a file that contains 24815 lines. This file takes about 1.2 seconds to process, which seems pretty reasonable to me given the size.

masatake:
> I tested it with a file that contains 24815 lines. This file takes about 1.2 seconds to process, which seems pretty reasonable to me given the size.
Merge state

Is the software having 24815 available as an open source software?
If yes, I would like to add to https://github.com/universal-ctags/codebase.

masatake:
(I started writing this comment to make a question to you. However, during writing I got an answer by myself. The answer I got is important for improving Rust parser. So I put what I wrote here as notes for myself. )

```
[yamato@slave]~% cat /tmp/foo.rs 
impl SomeTrait for SomeObject {
     fn bar(&self) {
     }
}
[yamato@slave]~% cat /tmp/foo.inko
impl SomeTrait for SomeObject {
  def bar {}
}
[yamato@slave]~% u-ctags --sort=no --fields='*'-lnr -o - /tmp/foo.rs 
SomeObject	/tmp/foo.rs	/^impl SomeTrait for SomeObject {$/;"	kind:implementation
bar	/tmp/foo.rs	/^     fn bar(&self) {$/;"	kind:method	scope:implementation:SomeObject	signature:(&self)
[yamato@slave]~% u-ctags --sort=no --fields='*'-lnr -o - /tmp/foo.inko
SomeObject	/tmp/foo.inko	/^impl SomeTrait for SomeObject {$/;"	kind:reopen	end:3	implements:SomeTrait
bar	/tmp/foo.inko	/^  def bar {}$/;"	kind:method	scope:reopen:SomeObject	end:2
```

I'm comapring Inko and Rust.  Consistency between languages are not the most important things in ctags.
However, making them consistent is better.

If we introduce "implements" field to Inko, I would like to use it in Rust. It looks for me that the two languages have similar grammars.

A. SomeObject implements SomeTrait.
B. SomeTrait implements SomeObject.

Which sentence is correct?
The answer is A. So the field name should be "implements:".

masatake:
Thank you.

YorickPeterse:
@masatake 

> Is the software having 24815 available as an open source software?
If yes, I would like to add to https://github.com/universal-ctags/codebase.

It's just the same content repeated over and over and over again, not actual software you'd find anywhere. It's as follows: [big.inko.txt](https://github.com/universal-ctags/ctags/files/3179615/big.inko.txt)

Thanks for merging!

masatake:
> It's just the same content repeated over and over and over again, not actual software you'd find anywhere. It's as follows: big.inko.txt

I see.

PR #2099, opened 2019-05-17
main: extend --total optoin to gather parser specific statistics

masatake:
@ksamborski, I'm working on providing the way to gather parser specific statistics.
I will extend --total option. I hope this helps you to optimize your parser.

```
[jet@living]~/var/ctags% ./ctags --totals=yes --language-force=CTagsSelfTest main/main.c 
1 file, 604 lines (13 kB) scanned in 0.0 seconds (29530 kB/s)
8 tags added to tag file
8 tags sorted in 0.00 seconds
longest tag line = 78

STATISTICS of CTagsSelfTest
==============================================
The number of handled chars: 2
```


masatake:
I will introduce "extended" or something like as the new value for `--total=yes|no`.
However, the interface for parser will not be changed. So you can fill initStats and printStats from now on if you think this feature is useful.

PR #2100, opened 2019-05-19
Allow compilation when either scandir() or dirent.h aren't available

techee:
At the moment, when you comment-out HAVE_SCANDIR or HAVE_DIRENT_H in
config.h (which is the configuration we use in Geany because we don't
need this functionality), one gets a compilation error.

This patch enables all possible combinations of HAVE_SCANDIR and
HAVE_DIRENT_H:
1. It removes the error from portable-dirent.h when neither HAVE_DIRENT_H
is defined nor _MSC_VER is used (the header does nothing in this case).
2. Inside portable-scandir.c, the checks are more precise - scandir() from
dirent.h is only used when both HAVE_SCANDIR and HAVE_DIRENT_H are
defined and the scandir() implementation is only compiled when
HAVE_DIRENT_H is available or _MSC_VER (which implies dirent from
portable-dirent is used) is defined. This means we have
scandir() -- either the system one or the ctags one -- when
HAVE_DIRENT_H is available or _MSC_VER is defined.
3. Inside options.c, the #ifdef checks can be reduced to
#if defined (HAVE_DIRENT_H) || defined (_MSC_VER) because of the above.
4. scanDirectory() extern declaration is moved from routines_p.h
inside options.c because it's the only place where it's used and it can
be protected here by the corresponding ifdefs.
5. Inside main.c ifdef HAVE_OPENDIR checks are extended by HAVE_DIRENT_H
or _MSC_VER because of the DIR structure declaration (either declared
inside dirent.h or portable-dirent_p.h).

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -85,7 +85,7 @@ static bool createTagsForEntry (const char *const entryName);
 *   FUNCTION DEFINITIONS
 */
 
-#if defined (HAVE_OPENDIR)
+#if defined (HAVE_OPENDIR) && (defined (HAVE_DIRENT_H) || defined (_MSC_VER))
 static bool recurseUsingOpendir (const char *const dirName)
 {
 	bool resize = false;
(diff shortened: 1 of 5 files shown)


techee:
We could actually live without this patch in Geany and check for dirent.h in configure.ac (I did this in my pull request against Geany for now). But I still think the checks are a little broken in uctags so it still makes sense to have it in uctags.

masatake:
> We could actually live without this patch in Geany and check for dirent.h in configure.ac (I did this in my pull request against Geany for now). But I still think the checks are a little broken in uctags so it still makes sense to have it in uctags.

I see. Thank you.

PR #2102, opened 2019-05-21
main: allow specifying alternative parser list

techee:
Allow applications which use ctags as a library provide their own
list of parsers. This list has to be declared as a define called
EXTERNAL_PARSER_LIST inside an external header. ctags will use the
EXTERNAL_PARSER_LIST_FILE define to detect the presence of such a
header and it should contain path to the header. When not defined,
internal parsers are used.

I've tested this against Geany and it seems to work.


techee:
@masatake OK to merge? (Or would you merge this?) This is the last missing piece for https://github.com/geany/geany/pull/2132 which I'd like to finalize.

masatake:
Yes, please, merge this.

PR #2105, opened 2019-05-27
Add missing Geany parsers

techee:
This pull request adds the extra parsers we have in Geany except those defined in c.c and except the markdown parser which is readLine() based and not really superior to the regex-based parser ctags uses (we should probably use the ctags parser in Geany instead).

The added parsers are:
* Abaqus (some obscure hardware definition language)
* Abc (https://en.wikipedia.org/wiki/ABC_notation)
* Docbook (https://en.wikipedia.org/wiki/DocBook)
* Haskell (https://en.wikipedia.org/wiki/Haskell_(programming_language))
* Haxe (https://en.wikipedia.org/wiki/Haxe)
* Nsis (https://en.wikipedia.org/wiki/Nullsoft_Scriptable_Install_System)
* Powershell (https://en.wikipedia.org/wiki/PowerShell)
* Txt2tags (https://en.wikipedia.org/wiki/Txt2tags)

For more details see the individual commits.


parsers/abc.c
```diff
@@ -28,8 +28,8 @@
 */
 
 static kindDefinition AbcKinds[] = {
-	{ true, 'm', "member", "sections" },
-	{ true, 's', "struct",  "header1"}
+	{ true, 's', "section", "sections" },
+/*	{ true, 'h', "header",  "header1"} */
```
masatake: I wonder why this one is commented out.

parsers/abaqus.c
```diff
@@ -0,0 +1,126 @@
+/*
+ *   This source code is released for free distribution under the terms of the
```
masatake: Is it possible to write the copyright holder explicitly?

parsers/abc.c
```diff
@@ -0,0 +1,124 @@
...
+#include "read.h"
+#include "vstring.h"
+#include "routines.h"
+#include "entry.h"
+
+/*
+*   DATA DEFINITIONS
+*/
+
+static kindDefinition AbcKinds[] = {
```
masatake: If 's' is not used, let's remove it.
Event there is only one kind, I would like you to define enum for kinds like:
```
enum AbcKind {
   K_MEMBER,
};
```

parsers/haskell.c
```diff
@@ -0,0 +1,355 @@
...
+	def->kindCount  = ARRAY_SIZE(HaskellKinds);
+	def->extensions = extensions;
+	def->parser     = findNormalHaskellTags;
+	return def;
+}
+
+extern parserDefinition* LiterateHaskellParser (void)
+{
+	static const char *const extensions [] = { "lhs", NULL };
+	parserDefinition* def = parserNew ("Literate Haskell");
```
masatake: Let's avoid to use whitespace in a parser name.

parsers/abc.c
```diff
@@ -0,0 +1,124 @@
...
+			return false;
+	}
+	return true;
+}*/
+
+
+static void makeAbcTag (const vString* const name, bool name_before)
+{
+	tagEntryInfo e;
+	initTagEntry (&e, vStringValue(name), 0);
```
masatake: I would like you to use 'K_MEMBER' instead of 0.

masatake:
I will add a selector for .hx file.

masatake:
Thank you very much.

parsers/abc.c
techee: I've removed all the commented-out code from the parser which is unmaintained and there doesn't seem to be anything of a great value. I've also made some cleanups in the parser.

parsers/haskell.c
techee: Done. By the way, this variant of the parser is unused right now (and we don't use it in Geany either, just the ordinary HaskellParser).

techee:
> Could you add the pages about the languages in the commit log:
> 
> > 
> 
> ```
> Abaqus (some obscure hardware definition language)
> Abc (https://en.wikipedia.org/wiki/ABC_notation)
> Docbook (https://en.wikipedia.org/wiki/DocBook)
> Haskell (https://en.wikipedia.org/wiki/Haskell_(programming_language))
> Haxe (https://en.wikipedia.org/wiki/Haxe)
> Nsis (https://en.wikipedia.org/wiki/Nullsoft_Scriptable_Install_System)
> Powershell (https://en.wikipedia.org/wiki/PowerShell)
> Txt2tags (https://en.wikipedia.org/wiki/Txt2tags)
> ```
> 
> to each parser source file?

Done, including Abaqus as well (which actually isn't a hardware-definition language as I thought but some simulation software language).

docs/news.rst
```diff
@@ -67,11 +72,13 @@ The following parsers have been added:
 * Maven2 *libxml*
 * Moose *perl basesd subperser*
 * M4
+* Nsis
 * ObjectiveC
 * Passwd *optlib*
 * PuppetManifest *optlib*
 * Perl6
 * Pod *optlib*
+* Powershell
```
masatake: Very minor issue. Powershell should be PowerShell.

techee: Fixed, including a few more capitalization problems.

masatake: Thank you.

masatake:
If there is no trouble at Geany side, I would like you to do one of following two:
A. squash the commits into one then merge it, or
B. convert the commits into per-parser commits and merge them.

techee:
>If there is no trouble at Geany side, I would like you to do one of following two:
>A. squash the commits into one then merge it, or
>B. convert the commits into per-parser commits and merge them.

No problem, I guess A will be easier to do.

techee:
I opened a separate pull request with the rebased version in #2106 so the individual commits are still visible if e.g. @b4n wants to have a look at them. Closing this pull request.

PR #2109, opened 2019-06-01
Perl,Moose: support Moo extension in Moose parser

### LINKED ISSUE #2103: Solve problems for Moose-like parsers.
Previously in #2070 we got support for a subparser for Moose. There are other modules like Mouse and Moo that follow the same API with Moose as a reference-spec (or strong-influence).

We need to decide if we want to 

* Implement subparsers that share a lot in a common,
* Implement one subparser for all Moose-like variants.

I strongly believe the second tactic is better.

masatake:
Close  #2103.
The design discussion was done at #2070.

To support Moo extension in Moose parser, I made perl parser captures module names specified in use build-in function and no built-in function. (I don't touch `require` in this commits because it is not needed to support Moo. However, we can add it to our TODO list.)

Module names are captured as reference tags. So if we don't pass --extras=+r and --kinds-Perl=+m options, the module names are never printed.

### DIFF
--- a/parsers/perl.c
+++ b/parsers/perl.c
@@ -36,6 +36,7 @@ typedef enum PerlModuleRoleType perlModuleRole;
 
 static roleDefinition PerlModuleRoles [] = {
 	{ true, "used",   "specified in `use' built-in function" },
+	{ true, "unused", "specified in `no' built-in function" },
 };
 
 static kindDefinition PerlKinds [] = {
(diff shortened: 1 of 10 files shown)


parsers/perl.c
```diff
@@ -384,15 +398,30 @@ static void findPerlTags (void)
...
-			cp += 8;
+
+			vString *module = NULL;
+			while (isalnum(*cp) || *cp == ':' || *cp == '.') {
+				if (!module)
+					module = vStringNew();
+				vStringPut(module, *cp);
+				++cp;
+			}
+			if (module) {
```
masatake:
```
if (!module)
  continue;
...
```
will be better.

PR #2110, opened 2019-06-07
Markdown: ignore GFM codeblock

masatake:
This is a bug fix discussed in https://github.com/Shougo/denite.nvim/issues/434.
@delphinus explain me the bug at the issue page.

### DIFF
--- a/Units/simple-markdown.d/input.md
+++ b/Units/simple-markdown.d/input.md
@@ -28,7 +28,29 @@
 ##### r #
 ##### s ######
 
+```sh
+# A COMMENT LINE in SHELL SYNTAX
+funciton x
+{
+	echo 'Hello, World!'
+}
+```
 ###### t #
+```sh
+# THE OTHER COMMENT LINE in SHELL SYNTAX
+funciton x
+{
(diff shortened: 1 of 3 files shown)


delphinus:
Thanks a lot! But I'm afraid [GFM spec for this][gfm] is extremely complicated and your code are insufficient... I leave it to you how detailed spec ctags can pase because even this PR is an great advantage for GFM-lovers!

[gfm]: https://github.github.com/gfm/#fenced-code-blocks

Units/simple-markdown.d/input.md
```diff
@@ -28,6 +28,13 @@
 ##### r #
 ##### s ######
 
+'''sh
```
delphinus: GFM recognizes not `''' ... '''`, but `｀｀｀ ... ｀｀｀` or <code>~~~ ... ~~~</code> as the [Fenced Code Blocks][fcb]. Correct this only, at least.

[fcb]: https://github.github.com/gfm/#fenced-code-blocks

masatake: Thank you. I will update my change.

Issue #1682, opened 2018-02-11
JSON: the parser overflows stack on deeply nested files (modified by @masatake)


codebrainz:
Not sure if this is considered a bug, but the JSON parser uses recursion while parsing (as I'm sure other parsers do), so if a large enough, deeply-nested enough file is processed, the parser runs out of stack and crashes.

It would be great if it used iteration instead of recursion, or maybe limited the recursion depth so it doesn't crash at least. This came up as [a Geany bug](https://github.com/geany/geany/issues/1768).

Backtrace:

```
#0  0x00007ffff756a938 in _IO_vfprintf_internal (s=s@entry=0x7fffff7ff570, format=format@entry=0x7ffff7d49353 "%u", ap=ap@entry=0x7fffff7ff6f0)
    at vfprintf.c:1320
#1  0x00007ffff763bc79 in ___vsnprintf_chk (s=0x7fffff7ff7e0 "", maxlen=<optimized out>, flags=1, slen=<optimized out>, format=0x7ffff7d49353 "%u", args=args@entry=0x7fffff7ff6f0) at vsnprintf_chk.c:63
#2  0x00007ffff763bba5 in ___snprintf_chk (s=s@entry=0x7fffff7ff7e0 "", maxlen=maxlen@entry=32, flags=flags@entry=1, slen=slen@entry=32, format=format@entry=0x7ffff7d49353 "%u") at snprintf_chk.c:34
#3  0x00007ffff7d334f9 in snprintf (__fmt=0x7ffff7d49353 "%u", __n=32, __s=0x7fffff7ff7e0 "") at /usr/include/x86_64-linux-gnu/bits/stdio2.h:64
#4  0x00007ffff7d334f9 in parseValue (token=token@entry=0x555555b638e0) at ../../ctags/parsers/json.c:330
#5  0x00007ffff7d33535 in parseValue (token=token@entry=0x555555b638e0) at ../../ctags/parsers/json.c:335
#6  0x00007ffff7d33535 in parseValue (token=token@entry=0x555555b638e0) at ../../ctags/parsers/json.c:335
[snip many more identical lines]
```

I generated a test file to reproduce the crash by piping the output of the following script to a file and opening it in Geany:

```python
#!/usr/bin/env python3
import sys
limit = 75000
for i in range(0, limit):
    sys.stdout.write('[')
sys.stdout.write('\n"foo"\n')
for i in range(0, limit):
    sys.stdout.write(']')
```

[...]

PR #2111, opened 2019-06-11
JSON: limit the depth of recursion by tracking character pair balancing

### LINKED ISSUE #2107: JSON DOS
(

The name of the parser:

json

```
./ctags  /tmp/borer/core/.jvm/src/test/resources/n_structure_100000_opening_arrays.json
```

The content of input file: https://github.com/sirthias/borer/blob/master/core/.jvm/src/test/resources/n_structure_100000_opening_arrays.json

(this is a sequence of 100k '[' chars)

heap increases to many Gb until it crashes with segfault.

It would be nice if this could either exit with an error, or use less memory and succeed. 

```
!_THIS_IS_AN_EXAMPLE
mainVoid	foo.c	/^main(void)$/;"	kind:function	line:2	language:C	typeref:typename:int	signature:(void)	roles:def
...
```

The tags output you expect:

```
!_THIS_IS_AN_EXAMPLE
main	foo.c	/^main(void)$/;"	kind:function	line:2	language:C	typeref:typename:int	signature:(void)	roles:def
...
```

The version of ctags:

```
$ ctags --version

Universal Ctags 0.0.0(3fdf28bc), Copyright (C) 2015 Universal Ctags Team
  Compiled: May 29 2019, 15:19:17
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +json, +interactive, +sandbox, +packcc
hanwen@hanwen:~/vc/ctags$ 
```
How do you get ctags binary:

local build, linux.

masatake:
Close #2107.
Related to #1682.

To avoid stack overflow, this change provides the way to terminate
the parsing.

The recursion is very related to square brackets and curly brackets.
Instead of limit the function recursion, this change tracks the depth
of brackets. Here I assume deep brackets in input stream causes
deep function call recursion.

readTokenFull is changed to return EOF when it detects too deep (> 512)
brackets in the current input stream. The EOF terminates the parsing.


masatake:
TODO: adding a test case

masatake:
A test case is added.

masatake:
Oh, sorry. I have just merged this.

masatake:
I will reflect your comment.

PR #2113, opened 2019-06-12
Add support for special Inko method names

YorickPeterse:
Inko allows you to use certain special characters for method names. For
example: +, -, /, [], and []=. This adds support for creating tags for
such methods, which would previously not be tagged.


masatake:
They look like operators. Are you o.k. to use "method" as kind for them?

YorickPeterse:
@masatake In Inko, operators _are_ methods. This means that `A + B` is the same as `A.+(B)`, with `+` being a method defined on `A`.

masatake:
I see.

masatake:
BTW, next time when you make a pull request for the Inko parser, could you add "Inko: " as prefix for commit logs like "Inko: add support for special method names" ?

(I should write this in our pull request template.)

masatake:
Thank you.

Issue #2089, opened 2019-05-01
TypeScript interface methods not recognized when semicolon ; is omitted


chrismwendt:
The name of the parser: TypeScript

The command line you used to run ctags:

```
$ ctags --options=NONE example.ts
```

The content of input file:

```TypeScript
interface Foo {
    bar1(): void;
    bar2(): void
    bar3(): void
    bar4(): void;
    bar5(): void
    bar6(): void;
}
```

The tags output you are not satisfied with:

```
ctags: Notice: No options will be read from files or environment
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	/9b28d8ce/
Foo	example.ts	/^interface Foo {$/;"	i
bar1	example.ts	/^    bar1(): void;$/;"	m	interface:Foo
bar2	example.ts	/^    bar2(): void$/;"	m	interface:Foo
bar5	example.ts	/^    bar5(): void$/;"	m	interface:Foo
```

The tags output you expect: should list `bar[1-6]`

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(9b28d8ce), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: May  1 2019, 13:26:09
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc
```

How do you get ctags binary:

Built from master.

masatake:
@ksamborski, could you look at this?

ksamborski:
Yes, typescript parser won't work if lines don't end with semicolon. It's
still on my todo list...

czw., 2 maj 2019, 02:49 użytkownik Masatake YAMATO <notifications@github.com>
napisał:

Issue #1229, opened 2016-12-05
SAPUI5 / OpenUI5 support


dfishburn:
The javascript parser supports SAP's UI5 framework, which was also open sourced as OpenUI5 ([](http://openui5.org/).

I added the initial support for it in Exuberant tags and I work with it daily.  I have updated the support for it in universal tags.  I will include a pull request for my changes.  Fixed a few other issues along the way while I was testing.

I do have some questions about adding tests and what not.  I see a bunch of directories under the Units directory.  Could someone provide me some instruction on how to use these directories.  First I would like to run all the javascript tests to make sure it is has no regressions.

I would also like to understand how to add my own set of tests for UI5.

Thank you.
David

masatake:
Could you look at http://docs.ctags.io/en/latest/units.html ?

masatake:
After getting your pull request. I would like you to allow me to rearrange your code. I would like to introduce language named OpenUI5 based on your pull request. 

Backgroupd:

Some of languages has "dialects". For example taken from your issue, Javascript language has OpenUI5 as a dialect. Another example is in #1227; C++ has CUDA as a dialect.
Python has CPython. M4 has autoconf. Make has Automake....

I wonder how ctags deals dialects. The classical way is just extending kinds and file name patterns of the base language. #1220 is typical case.
However, this approach has an issue. This approach consumes more kind letters. A namespace of kind letters are   not small[a-zA-Z]. However, it is not enough large for adding many dialects. Not only us but also users will add dialects locally, soon we will see conflicts of kind letters.

Look at https://github.com/universal-ctags/ctags/issues/453
One wants to add 'c' (constant) kind to Ruby. However, it is used already for Rspec dialect.

From these experience I think dialect should has its own parser derived from the base language.
The dialect has its own namespace. The namespace of the base language can be kept clean.
Further more language: field can show the name of dialect.

How can we implement a dialect? I have tired to answer this question for long time.
I have developed some facilities. I have to explain them.

masatake:
(http://openui5.org/getstarted.html)

masatake:
I misunderstood your question.
To run test, do
```
$ make units
```

Only about javascript do
```
$ make units LANGUAGES=JavaScript
```

How to write a test case, please read  http://docs.ctags.io/en/latest/units.html .

masatake:
After thinking I rename what I called "dialect" here to "subparser".
See #1326, especially
[...]

PR #2117, opened 2019-06-17
TypeScript: parser no longer needs semicolons

ksamborski:
I updated the parser to not depend on semicolons. Now output should be the same regardless of using semicolons as statement separator. So one can use or not use semicolons in his/her code or even mix those two styles. Generated tags should stay the same.

Closes: #2089

Thanks to new optimizations the current implementation is even a bit faster than the previous one.


Units/parser-typescript.r/ts-class-fq-white.d/args.ctags
```diff
@@ -0,0 +1,3 @@
+--sort=no
+--typescript-kinds=fcigemnzpvCgGal
```
masatake: You can use a wildcard character here (if you want):
```
--typescript-kinds=*
```
--<LANG>-kinds option is obsoleted.
Use --kinds-<LANG> instead.
So the option can be written as:
```
--kinds-typescript=*
```

ksamborski: OK, I'll change it.

ksamborski: Done.

parsers/typescript.c
```diff
@@ -518,35 +536,37 @@ CTAGS_INLINE void parseString(const int c, tokenInfo *const token, const char qu
 			*prev = c;
 			result->status = PARSER_NEEDS_MORE_INPUT;
 		}
-		else
-		{
-			result->status = PARSER_FAILED;
+		else result->status = PARSER_FAILED;
```
masatake:
```
else
   result->status = PARSER_FAILED;
```
I prefer putting a newline char after `else`.

ksamborski: I agree that it would look nicer but it's easier to forget about curly braces when you'd like to add another statement there. One line kinda protects you from that mistake.

masatake: In that csae,
```
					if (!parsingType)
					{
						member = newToken ();
						copyToken (member, token, false);
						member->scope = scope;
						if (visibility)
							member->accessKeyword = visibility;
						else
							member->accessKeyword = KEYWORD_public;
	
```
should be
```
					if (!parsingType)
					{
						member = newToken ();
						copyToken (member, token, false);
						member->scope = scope;
						if (visibility)
							member->accessKeyword = visibility;
						else member->accessKeyword = KEYWORD_public;
```,
shouldn't it?

Comment (ksamborski) [reply parsers/typescript.c]: You got me :) The same applies to that if. I'll try to make my code more consistent.

Comment (masatake) [reply parsers/typescript.c]: I searched "else" in typescript.c:
```
19 matches for "else" in buffer: typescript.c
    445:	else if (*parsed == 0)
    492:		else if (result->status == PARSER_FINISHED)
    505:	else if (state->isBlock)
    515:	else if (c == '\n')
    539:		else result->status = PARSER_FAILED;
    543:	else if (c == EOF)
    557:	else if (c == quote && *prev != '\\')
    585:	else if (c == '}')
    590:	else if (c == end)
    592:	else if ((state->curlyLevel < 1 && c == ';') || c == EOF)
    736:	else if (result.unusedChars > 0)
    738:	else
    902:						else
   1026:					else if (! parsingType && shouldEnd)
   1501:		else
   1610:					else
   1709:							else
   1808:						else
   1828:	else if (member)
```

In the most of all cases, you use stand-alone  'else' .

ksamborski: This is what happens when you suddenly change your mind... :) I'll change all of them one way or the other.

masatake: Howeer, this is a minor point.
Supporiting semicolons-less typescript input is much important, of course.
Feel free to merge the change.

ksamborski: Thanks, I will fix formatting today and then merge.

masatake:
Thank you.

BTW, I'm working hard on supporting modules in JavaScript.
I wrote a document about it in broken English:-P

You may have interests on it. 
I just found you use "alias" kind in typescript.
I named the same things "name" in JavaScript in the document.

https://github.com/universal-ctags/ctags/issues/1229

masatake:
(... typescript also can be used in .vue file. So we should use the similar kind names within JavaScript and TypeScript. )

ksamborski:
"alias" here in TypeScript means type alias which doesn't exist in JavaScript. I don't think that it should be the same as an alias for imported module in JavaScript.

masatake:
Oh, I see.

masatake:
`make slap LANGUAGES=TypeScript `
 and
`make chop LANGUAGES=TypeScript `
are successfully done.

PR #2122, opened 2019-06-22
Reject printing newline in ectags output format

masatake:


masatake:
hasUnwantedChar is better than hasTabOrNewlineChar.

PR #2130, opened 2019-07-03
Import NSIS parser from Geany

masatake:
This pull request adds one of the extra parsers we have in Geany except
those defined in c.c and except the markdown parser which is readLine()
based and not really superior to the regex-based parser ctags uses
(we should probably use the ctags parser in Geany instead).

The added a parser is for Nsis (https://en.wikipedia.org/wiki/Nullsoft_Scriptable_Install_System).

The original version is written for Geany by Enrico Tröger <enrico.troeger@uvena.de>:
commit 61f89d5c259c2538ceff815e8f2e5226343e89ff
Author: Enrico Tröger <enrico.troeger@uvena.de>
Date:   Tue Apr 21 20:52:19 2009 +0000

    Add a trivial symbol parser for NSIS files.

    git-svn-id: https://geany.svn.sourceforge.net/svnroot/geany/trunk@3716 ea778897-0a13-0410-b9d1-a72fbfd435f5

This log message is edited by @masatake.


masatake:
@techee, as I wrote, I would like to merge the original big pull request incrementally.
I also want to record the name of the original author to the commit log.

Other than the commit log, I don't change anything.

Units/simple-nsis.d/expected.tags
```diff
@@ -0,0 +1,20 @@
+Answer	input.nsi	/^Var Answer$/;"	v
+UserName	input.nsi	/^Var UserName$/;"	v
+StartmenuFolder	input.nsi	/^Var StartmenuFolder$/;"	v
+UNINSTDIR	input.nsi	/^Var UNINSTDIR$/;"	v
+\x21Program Files	input.nsi	/^Section "!Program Files" SEC01$/;"	s
```
masatake: ! at the head of tag means displaying the section name with bold style.
I wonder this should be part of the section name or not.
https://nsis.sourceforge.io/Reference/Section

Units/simple-nsis.d/expected.tags
```diff
@@ -0,0 +1,20 @@
...
+Plugins	input.nsi	/^Section "Plugins" SEC02$/;"	s
+Language Files	input.nsi	/^Section "Language Files" SEC03$/;"	s
+Documentation	input.nsi	/^Section "Documentation" SEC04$/;"	s
+Autocompletion Tags	input.nsi	/^Section "Autocompletion Tags" SEC05$/;"	s
+GTK 	input.nsi	/^Section "GTK ${GTK_VERSION} Runtime Environment" SEC06$/;"	s
+Context Menus	input.nsi	/^Section "Context Menus" SEC07$/;"	s
+Desktop Shortcuts	input.nsi	/^Section "Desktop Shortcuts" SEC08$/;"	s
+Development files	input.nsi	/^Section "Development files" SEC09$/;"	s
+-AdditionalIcons	input.nsi	/^Section -AdditionalIcons$/;"	s
```
masatake: The name started from - means "hidden section".
I wonder this should be part of the section name or not.
https://nsis.sourceforge.io/Reference/Section

k-takata:
Do you have a plan to support also `SectionGroup`, `!define` and `!macro`?

masatake:
@k-takata, `!define` is done. I will work on the rest.

masatake:
SectionGroup and !macro are captured now.
!import should captured as reference tags.
end: field and signature: field can be filled.

But I should do them later.
I will merge these in soon.

Units/simple-nsis.d/expected.tags
masatake: After thinking, I consider the characters (- and !) as  parts of sections.

sections are not identifiers. So none jumps to a section from an identifier. One may jump to the
section from an item in the navigation bar of one's editor or soemthing. It means strict name maching is not needed. Instead, informative tag names help users.

k-takata:
Does it support LangString for section name?
E.g. `Section "$(message)"`
https://nsis.sourceforge.io/Docs/Chapter4.html#langstring

masatake:
>  Does it support LangString for section name? E.g. Section "$(message)", Section $(message) https://nsis.sourceforge.io/Docs/Chapter4.html#langstring

Do you expect that ctags expands "$(message)" when tagging section names?
I need a bit more concrete example.

k-takata:
This is an example of Vim's installer:
https://github.com/vim/vim/blob/3dabd718f4b2d8e09de9e2ec73832620b91c2f79/nsis/gvim.nsi#L425-L426
```
SectionGroup $(str_group_icons) id_group_icons
	Section "$(str_section_desktop)" id_section_desktop
```

I'm not sure if we should tag `$(...)`, but `"${var}"` will be tagged, right?
(At least, section IDs should be tagged, though.)

k-takata:
Should `LangString` be also tagged?
E.g. https://github.com/vim/vim/blob/3dabd718f4b2d8e09de9e2ec73832620b91c2f79/nsis/lang/english.nsi#L65-L66
Should `str_group_icons` be tagged here?

parsers/nsis.c
```diff
@@ -0,0 +1,249 @@
...
+
+static const unsigned char* parseSection (const unsigned char* cp, vString *name,
+										  int kindIndex, int scopeIndex, int *corkIndex)
+{
+	bool in_quotes = false;
+	cp = skipWhitespace (cp);
+	cp = skipFlags (cp);
+	while (isalnum ((int) *cp) || isspace ((int) *cp) ||
+		   *cp == '_' || *cp == '-' || *cp == '.' || *cp == '!' || *cp == '"'
+		   || (in_quotes && (*cp == '$' || *cp == '{' || *cp == '}' )))
```
k-takata: As far as I know, any characters can be used for a section(group) name when they are quoted.
Even Japanese characters can be used.

masatake:
@k-takata, I think NSIS parser captures all you suggested, now.
Next some weeks, I am not sure I have enough time work on ctags. So I will merge this without your verifying. Feel free open an issue for the Nsis parser.

k-takata:
I tested this with [Vim's NSIS scripts](https://github.com/vim/vim/tree/master/nsis).
It worked like a charm. Thank you.

masatake:
Thank you for testing.

PR #2136, opened 2019-07-09
win32: Remove extra CR

k-takata:
The last line was CRCRLF.


masatake:
Thank you.

This is not the first time to merge this kind change.
Don't you know any good git's config to avoid extra CR at the end of file?
(Or config for vim?)

k-takata:
> Don't you know any good git's config to avoid extra CR at the end of file?

Unfortunately, no.

PR #2139, opened 2019-07-12
 renamed parser to BibTeX and added recognition of @string entries

TwlyY29:
Hi @masatake ,

here the new pull request. 

Best
Mirco


k-takata:
https://github.com/universal-ctags/ctags/blob/master/docs/news.rst#new-parsers
I think this should be also renamed to BibTeX.

masatake:
I would like to take over this.

masatake:
I made a pull request base on this pull request. See #2142. I will merge #2142 in soon.

PR #2140, opened 2019-07-14
Check eof chars of vcxproj files at circleci

masatake:
make codecheck runs misc/src-check.
I added a function detecting CRCRLF sequence at the eof win32 build scripts to misc/src-check.
I made the codecheck target is run on fedora_bmake combination run at circlci,


masatake:
A function to run spellcheck should be added to codecheck target.

PR #2142, opened 2019-07-15
BibTeX: renamed parser to BibTeX and added recognition of @string entries (based on work done by @TwlyY29 in #2139)

masatake:


TwlyY29:
Thanks!

PR #2147, opened 2019-07-18
Refine AppVeyor

k-takata:
* Reduce time to install packages.
* Make CI fail if fail to generate docs.

(Both are mainly for ctags-win32 project.)


masatake:
LGTM though I don't have much experience and knowledge about windows.

PR #2149, opened 2019-07-18
buildsys,mvc: Support packcc

k-takata:
Support packcc on MSVC.

### DIFF
--- a/mk_mingw.mak
+++ b/mk_mingw.mak
@@ -2,7 +2,7 @@
 
 include source.mak
 
-REGEX_DEFINES = -DHAVE_REGCOMP -D__USE_GNU -Dbool=int -Dfalse=0 -Dtrue=1 -Dstrcasecmp=stricmp
+REGEX_DEFINES = -DHAVE_REGCOMP -D__USE_GNU -DHAVE_STDBOOL_H -Dstrcasecmp=stricmp
 
 CFLAGS = -Wall -std=gnu99
 # sizeof (size_t) == sizeof(unsigned long) == 4 on i686-w64-mingw32-gcc.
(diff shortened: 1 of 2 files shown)


k-takata:
Hmm, it seems that packcc causes runtime error when compiled by VC2013.
It works fine when compiled by VC2015.

masatake:
Do you have VC2013 on your local PC?
Debugging only on Appveyor is really painful...

k-takata:
I tested locally and I found that a null pointer access occurs inside fprintf. So, it might be a problem of VC2013.
I also found another problem. Packcc uses "z" flag to print size_t variables, but VC2013 doesn't support the "z" flag of printf. VC supports it from VC2015. So, it might be better to support packcc only from VC2015.

k-takata:
Note that https://github.com/universal-ctags/packcc/pull/1 is needed for building with VC2015.

masatake:
> So, it might be better to support packcc only from VC2015.

...I don't have any objection for this.

k-takata:
Ah, defining `SIZE_T_FMT_CHAR` to `"I"` fixes the problem.

k-takata:
Okay, now the build succeeded and the varlink tests passed:
https://ci.appveyor.com/project/universalctags/ctags/builds/26092806/job/2hjnwt7vi6tdd8k5#L3221

masatake:
Excellent! Feel free to merge the change.

k-takata:
I'm going to change some more.
* Remove an unused variable.
* Make mk_mvc.mak and mk_mingw.mak more similar. Some variables are defined differently. Make them the same.

k-takata:
I also removed the boolean definitions and included stdbool.h like #2151.
It was needed for making some variable definitions the same between mk_mvc.mak and mk_mingw.mak.

k-takata:

PR #2150, opened 2019-07-18
buildsys,mingw: Remove git-submodule commands

k-takata:
They are not needed anymore.

Also update the `clean` target.


masatake:
LGTM. Thank you.

PR #2151, opened 2019-07-18
win32: Remove boolean definitions

k-takata:
VC2013 or later has stdbool.h. No need to define `bool` by ourself.
This suppresses the following warnings:
```
C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\stdbool.h(10): warning C4005: 'bool' : macro redefinition [C:\projects\ctags\win32\ctags_vs2013.vcxproj]
          command-line arguments :  see previous definition of 'bool'
```


masatake:
LGTM.

PR #2152, opened 2019-07-19
appveyor: Make only html for the man pages

k-takata:
Creating PDF fails on AppVeyor:
https://ci.appveyor.com/project/universalctags/ctags-win32/builds/26091558
Only html files are needed.


k-takata:
This is needed for the ctags-win32 project.

PR #2153, opened 2019-07-19
Support VC2015

k-takata:
* Import the latest packcc.
* Fix regex.

### DIFF
--- a/gnu_regex/regcomp.c
+++ b/gnu_regex/regcomp.c
@@ -2527,7 +2527,7 @@ parse_dup_op (bin_tree_t *elem, re_string_t *regexp, re_dfa_t *dfa,
     old_tree = NULL;
 
   if (elem->token.type == SUBEXP)
-    postorder (elem, mark_opt_subexp, (void *) (long) elem->token.opr.idx);
+    postorder (elem, mark_opt_subexp, (void *) (intptr_t) elem->token.opr.idx);
 
   tree = create_tree (dfa, elem, NULL, (end == -1 ? OP_DUP_ASTERISK : OP_ALT));
   if (BE (tree == NULL, 0))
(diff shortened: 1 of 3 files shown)


k-takata:

masatake:
Could you squash
5d11a90 and 
96527a1 into one?

k-takata:
Squashed.
I also updated the commit message and explained why I used `size_t` for this.

masatake:
In https://github.com/universal-ctags/ctags/blob/master/main/htable.h#L54
I use intptr_t. Is this can be a trouble on building ctags on MinGW?

masatake:
Intead, how about defining intptr_t in mk_mingw.mak or somewhere?
So we can use intptr_t in our code.

k-takata:
Hmm, at least current CI passes even you uses `intptr_t`.
Is any header needed for using `intptr_t` on a old MinGW?

k-takata:
Maybe `stdint.h`? You include it from `htable.h`.

masatake:
Yes!
```
[yamato@slave]~/var/ctags-github% ./ctags -o - /usr/include/stdint.h | grep ^intptr_t
intptr_t	/usr/include/stdint.h	/^typedef int			intptr_t;$/;"	t	typeref:typename:int
intptr_t	/usr/include/stdint.h	/^typedef long int		intptr_t;$/;"	t	typeref:typename:long int
```

masatake:
If stdint.h is not available on the platform, we have to define it in the .mak.
It seems that intptr_t is introduced in c99. So I expect it is available on any platform.
If not, u-ctags cannot be built in many various reasons.

k-takata:
I think we should add `-DHAVE_STDINT_H` in mk_mingw.mak. I will try it later.

k-takata:
The CI succeeded.

masatake:
Fine! LGTM.

PR #2155, opened 2019-07-23
NSIS: Fix typos

k-takata:


parsers/nsis.c
```diff
@@ -216,7 +216,7 @@ static const unsigned char* parseLangString (const unsigned char* cp, vString *n
 {
 	cp = skipWhitespace (cp);
 
-	/* `^' is not explaned the nsis reference manual. However, it is used
+	/* `^' is not explained the nsis reference manual. However, it is used
```
b4n:
```suggestion
	/* `^' is not explained in the nsis reference manual. However, it is used
```
or maybe "mentioned" instead of "explained"?

k-takata: Updated.

masatake:
LGTM. Thank you.

PR #2156, opened 2019-07-23
win32: Update copyright and license information

k-takata:
Explicitly state that it is GPL 2 or later.


masatake:
LGTM. Thank you.

Issue #2176, opened 2019-08-15
TypeScript parser broken when a function is on the RHS of `=`


chrismwendt:
The name of the parser: TypeScript

The command line you used to run ctags:

```
$ ctags --options=NONE a.ts
```

The content of input file:

```typescript
var foo = function() {
  console.log();
};
```

The tags output you are not satisfied with:

```
console	a.ts	/^  console.log();$/;"	v
foo	a.ts	/^var foo = function() {$/;"	v
```

See the extra `console` tag? That shouldn't be there. The same happens for arrow functions `var foo = () => { ... }`.

The tags output you expect:

```
foo	a.ts	/^var foo = function() {$/;"	v
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(9787a6c0), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Aug  6 2019, 11:36:53
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc
```

How do you get ctags binary:

Building it locally

ksamborski:
Thanks for the issue, I've prepared a quick fix for it: #2177

PR #2177, opened 2019-08-15
Typescript: fixes #2175, #2176

### LINKED ISSUE #2175: The typescript parser does not handle properly expressions that end with a curly brace not separated by a space
The typescript parser stops parsing the file at the line with the call to `g`.
Adding a space before the closed brace in the argument of `g` makes the parser work correctly.

The name of the parser: TypeScript

```
$ ctags --options=NONE file.ts
```

The content of input file:

```TypeScript
class X {
    public f(a: T): number {
        return g({ a: a});
    }

    public g(a: T): number {
        return 1;
    }
}
```

```
x.ts,52
class X {^?X^A1,0
    public f(a: T): number {^?f^A2,10
```

The tags output you expect:

```
x.ts,88
class X {^?X^A1,0
    public f(a: T): number {^?f^A2,10
    public g(a: T): number {^?g^A6,74
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(b818539), Copyright (C) 2015 Universal Ctags Team
  Compiled: Aug 14 2019, 16:54:57
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +packcc
```

How do you get ctags binary:

Building it locally, via GNU/Linux distribution, as BSD's package,

ksamborski:
Both issues are related to functions but in different way. I added test cases for them and this PR should now fix all of them.  @masatake please take a look if you have a moment

### DIFF
--- a/Units/parser-typescript.r/ts-class-white.d/expected.tags
+++ b/Units/parser-typescript.r/ts-class-white.d/expected.tags
@@ -54,3 +54,8 @@ options	input.ts	/^  longArgsFun(options: {$/;"	z	method:A.longArgsFun
 closure	input.ts	/^  closure($/;"	m	class:A
 x	input.ts	/^    x: number,$/;"	z	method:A.closure
 normalizedPath	input.ts	/^    const normalizedPath = path === '\/*' ? '' : path$/;"	C	method:A.closure
+X	input.ts	/^class X {$/;"	c
+f	input.ts	/^    public f(a: T): number {$/;"	m	class:X
+a	input.ts	/^    public f(a: T): number {$/;"	z	method:X.f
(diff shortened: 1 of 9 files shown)


masatake:
LGTM.

ksamborski:
OK, I've changed commit messages for those with `close #n`.

masatake:
Well... I expected just adding "Close #n" like:

```
Typescript: fixed parsing variables that are functions
Close #2176.
```

However, you did
```
Typescript: close #2176 
Fixed parsing variables that are functions
```

I like the former style.

ksamborski:
Thanks, fixed.

PR #2179, opened 2019-08-18
Typescript: parsing multiple interface names in implements part

### LINKED ISSUE #2178: TypeScript parser does not parse correctly a list with multiple `implements`
The name of the parser: TypeScript

```
$ ctags --options=NONE -e x.ts
```

The content of input file:

```TypeScript
class C extends A implements B, D {
    x: number;
}
```

```
x.ts,42
class C extends A implements B, D {^?C^A1,0
```

The tags output you expect:

```
x.ts,61
class C extends A implements B, D {^?C^A1,0
    x: number;^?x^A2,33
```

The version of ctags:

```
Universal Ctags 0.0.0(b818539), Copyright (C) 2015 Universal Ctags Team
  Compiled: Aug 14 2019, 16:54:57
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +packcc
```

How do you get ctags binary:

Building it locally.

ksamborski:
Typescript parser now handles properly multiple interface names in `implements` part of class declaration. Additionally outputs `inherits` field with all class and interface names used by the class.

Fixes #2178 

@masatake Class declarations using templates like `class A<T> implements I<T>` IMHO should output only class/interface name without parameters. They can be very complicated and long and I can't think of anything they could be used for (when putting them also in  `inherits` field).

### DIFF
--- a/Units/parser-typescript.r/ts-class-white.d/args.ctags
+++ b/Units/parser-typescript.r/ts-class-white.d/args.ctags
@@ -1,2 +1,3 @@
 --sort=no
 --typescript-kinds=fcigemnzpvCgGal
+--fields=+i
--- a/Units/parser-typescript.r/ts-class-white.d/expected.tags
+++ b/Units/parser-typescript.r/ts-class-white.d/expected.tags
@@ -59,3 +59,5 @@ f	input.ts	/^    public f(a: T): number {$/;"	m	class:X
 a	input.ts	/^    public f(a: T): number {$/;"	z	method:X.f
 g	input.ts	/^    public g(a: T): number {$/;"	m	class:X
 a	input.ts	/^    public g(a: T): number {$/;"	z	method:X.g
+C	input.ts	/^class C extends A implements B, D {$/;"	c
+x	input.ts	/^    x: number$/;"	p	class:C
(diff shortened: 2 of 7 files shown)


masatake:
Just a question.

Is the next code valid?

```
function a<T>(x: T) {
  alert(x);
}

function a(x: string) {
  alert(x);
}
```

If yes, (ideally) we should consider template parameters more deeply.
Current implementation captures two `a`. Users cannot distinguish the one for the other.

If the parser can include template parameters to tag names, the parser can capture
`a<T>` and `a`. The difference helps the user understand code.

Anyway the change in this pull request is useful. So deciding how we handle template parameters in TypeScript parser is NOT blocker for merging this pull request.

 I have ever seen a similar issue with this issue. https://github.com/universal-ctags/ctags/pull/1604

ksamborski:
The code is invalid. In Typescript we can't have two definitions of a function with the same name. But even if it was valid there would be still a problem if we emit those functions as `a` and `a<T>` tag. We could call those functions in two different ways:

```
a('string');
a<string>('string');
```

Both would call the same specialized version of the function (`function a(x: string)`) but the user might not get right function definition in the second case (I assume that only `a<T>` tag would be shown).

It might be even more complicated as template parameters can narrow down the number of types that it could get, e. g.:

```
function a<T extends Person>(x: T)
```

so if Typescript would allow generic function specialization (but again, it doesn't) we could have two mutually exclusive definitions:

```
function a<T extends Person>(x: T)
{
}

function a<T extends Car>(x: T)
{
}
```

And then having these long tag names with all the parameters would cause more trouble to the editors (or IDEs).

But we could probably save template parameters in some different field so the user could at least see that the function is generic.

PR #2188, opened 2019-08-26
C++: print real name of popped scope

mymedia2:
* Move call of cxxScopeGetFullName before cxxTokenChainTakeLast.
 * Also do not double check CXX_DO_DEBUGGING - explicitly via #ifdef and inside CXX_DEBUG_PRINT macro.

### DIFF
--- a/parsers/cxx/cxx_scope.c
+++ b/parsers/cxx/cxx_scope.c
@@ -227,11 +227,7 @@ void cxxScopePushTop(CXXToken * t)
 	cxxTokenChainAppend(g_pScope,t);
 	g_bScopeNameDirty = true;
 
-#ifdef CXX_DO_DEBUGGING
-	const char * szScopeName = cxxScopeGetFullName();
-
-	CXX_DEBUG_PRINT("Pushed scope: '%s'",szScopeName ? szScopeName : "");
-#endif
+	CXX_DEBUG_PRINT("Pushed scope: '%s'",cxxScopeGetFullName() ?: "");
 }
 
 CXXToken * cxxScopeTakeTop(void)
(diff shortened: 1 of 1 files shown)


mymedia2:
@coveralls, this is completely normal situation as I touched already covered code. Within the modified lines code coverage level hasn't changed.

parsers/cxx/cxx_scope.c
```diff
@@ -227,11 +227,7 @@ void cxxScopePushTop(CXXToken * t)
 	cxxTokenChainAppend(g_pScope,t);
 	g_bScopeNameDirty = true;
 
-#ifdef CXX_DO_DEBUGGING
-	const char * szScopeName = cxxScopeGetFullName();
-
-	CXX_DEBUG_PRINT("Pushed scope: '%s'",szScopeName ? szScopeName : "");
-#endif
+	CXX_DEBUG_PRINT("Pushed scope: '%s'",cxxScopeGetFullName() ?: "");
```
k-takata: `?:` is GNU extension.

masatake: Yes, it is a GNU extension. I like it but it should not be used in u-ctags.

masatake: The double check requires only in built-time. So the double check is acceptable.
With the following code, we can avoid the double check but it is a bit harder to read.

```
#ifdef CXX_DO_DEBUGGING
	const char * szScopeName = cxxScopeGetFullName();
#endif
	CXX_DEBUG_PRINT("Pushed scope: '%s'",szScopeName ? szScopeName : "");
```

parsers/cxx/cxx_scope.c
```diff
@@ -240,15 +236,11 @@ CXXToken * cxxScopeTakeTop(void)
...
 	CXXToken * t = cxxTokenChainTakeLast(g_pScope);
 	g_bScopeNameDirty = true;
 
-#ifdef CXX_DO_DEBUGGING
-	const char * szScopeName = cxxScopeGetFullName();
-
-	CXX_DEBUG_PRINT("Popped scope: '%s'",szScopeName ? szScopeName : "");
-#endif
 	return t;
 }
```
masatake: @pragmaware, do you think this change (moving CXX_DEBUG_PRINT before calling cxxTokenChainTakeLast) is good?

pragmaware: Well... yet it's probably a bit clearer to the reader. The popped scope is really the one that was there before.
I suggest we apply this one and your non-GNU-restricted version of the previous diff.

PR #2194, opened 2019-09-10
appveyor: Use rst2html instead of rst2html3

k-takata:
The rst2html3 command from the python3-docutils package was renamed to
rst2html a few weeks ago. No need to set RST2HTML variable anymore.

An error occurs without this fix:
https://ci.appveyor.com/project/universalctags/ctags-win32/builds/27311853/job/bu0ignl4x15w473g#L1177


masatake:
LGTM.

PR #2106, opened 2019-05-28
Add Geany parsers which are missing in ctags (rebased version)

techee:
This pull request adds the extra parsers we have in Geany except
those defined in c.c and except the markdown parser which is readLine()
based and not really superior to the regex-based parser ctags uses
(we should probably use the ctags parser in Geany instead).

The added parsers are:

* Abaqus (https://en.wikipedia.org/wiki/Abaqus)
* Abc (https://en.wikipedia.org/wiki/ABC_notation)
* Docbook (https://en.wikipedia.org/wiki/DocBook)
* Haskell (https://en.wikipedia.org/wiki/Haskell_(programming_language))
* Haxe (https://en.wikipedia.org/wiki/Haxe)
* Nsis (https://en.wikipedia.org/wiki/Nullsoft_Scriptable_Install_System)
* Powershell (https://en.wikipedia.org/wiki/PowerShell)
* Txt2tags (https://en.wikipedia.org/wiki/Txt2tags)

Some small cleanups were made to the Abc parser compared to Geany
version - basically just commented code removal.

The patch also adds simple unit tests from Geany. Because the added
Haxe parser extension "hx" clashes with QemuHX file extension,
--language-force=QemuHX had to be added to the args of QemuHX tests.

PR #2198, opened 2019-09-16
Docbook: new parser

masatake:
After studying an docbook parser implementation proposed at #2106, I decided to take very different approach to implement the parser. See the comments in  parsers/docbook.c.

There are two weaknesses.
* Unlike Geany's one, this parser requires libxml2.
* The input must be well-formed XML.

namespace of XML is not handled well. I have to study how to handle the namespace in u-ctags.


masatake:
@techee, sorry to be late. I'm merging #2106 incrementally.
If you have time, I would like to see the comment at the head of parsers/docbook.c.

masatake:
What a wonderful bug!

masatake:
The mechanism for choosing a parser is not enough to choose a suitable XML subparser.
selectParserForXmlDoc is called too early stage.
It should be called from the XML base parser.

I have to fix this.

PR #2204, opened 2019-09-19
main,win32: Use _getdrive() also on MinGW

k-takata:
absoluteFilename() always returned C: drive when compiled with MinGW.
MinGW also has _getdrive(), so use it.


masatake:
LGTM.

PR #2212, opened 2019-09-28
Cxx: add a field for recording macro definition to macro kind

masatake:
Currently, saveMacro feature can be used from command line (-D option).
I'm thinking about using it from CPreProcessor parser itself.
This is the first step to implement the idea.
With this patch, entries for macros can have "signature" and "macrodef" fields.
It is enough to make arguments passed to saveMacro.


masatake:
I will add some test cases.

masatake:
The coverage is increased.

PR #2213, opened 2019-09-29
Include man pages to the web doc

### LINKED ISSUE #1444: Make man pages avaiable on http://docs.ctags.io/en/latest/
Derived from #1435.

masatake:
Close #1444.


k-takata:
Seeing the [generated page](http://docs.ctags.io/en/latest/man/ctags.1.html), `@CTAGS_NAME_EXECUTABLE@` and so on are embedded as is. Is it possible to replace with `ctags`, etc.?

masatake:
> Seeing the generated page, @CTAGS_NAME_EXECUTABLE@ and so on are embedded as is. Is it possible to replace with ctags, etc.?

I don't know how to replace them.
We must write about this issue to http://docs.ctags.io/en/latest/man-pages.html .

PR #2216, opened 2019-09-30
Docs adding scrollbar in the sidebar

masatake:
Try to close https://github.com/universal-ctags/universal-ctags.github.io/issues/18 .
I'm not sure whether this works well or not.
As far as I tried locally, there is no effect of this change.


masatake:
This will not work well.

PR #2218, opened 2019-10-01
docs: Update some documents

k-takata:
Catch up with the latest status.

* `--enable-iconv` is not needed anymore.
* Update a URL.
* Update the information about scandir().
* Write about mkstemp().


masatake:
LGTM. Thank you.

PR #2221, opened 2019-10-03
appveyor: Copy man/*.rst before building html files

k-takata:
Copy man/*.rst to docs/man/ before building html files, because Git for
Windows doesn't handle symlinks by default.


masatake:
People using u-ctags on Windows must thank @k-takata

PR #2209, opened 2019-09-27
convert docs/format.rst to tags(5) man page

masatake:

PR #2225, opened 2019-10-04
Update .gitignore

k-takata:
Update for #2209.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,8 @@ man/ctags-incompatibilities.7.rst
 man/ctags-optlib.7
 man/ctags-optlib.7.html
 man/ctags-optlib.7.rst
+man/tags.5
+man/tags.5.html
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -80,6 +80,9 @@ man/ctags-incompatibilities.7.rst
 man/ctags-optlib.7
 man/ctags-optlib.7.html
 man/ctags-optlib.7.rst
+man/tags.5
+man/tags.5.html
+man/tags.5.rst
```
masatake: tags.5.rst was added to repository.
Unlike the other man pages, tags.5.rst is not generated from tags.5.rst.in.
So man/tags.5.rst should not be ignored.

k-takata: Oh, I will fix it.

masatake:
LGTM.

masatake:
Thank you.

PR #2228, opened 2019-10-06
PowerShell: new parser imported from Geany

masatake:
This pull request adds one of the extra parsers we have in Geany except
those defined in c.c and except the markdown parser which is readLine()
based and not really superior to the regex-based parser ctags uses
(we should probably use the ctags parser in Geany instead).

The added a parser is for PowerShell (https://en.wikipedia.org/wiki/PowerShell)

The original version is written for Geany by Enrico Tröger <enrico.troeger@uvena.de>:
commit 8a6fbd978691fa324bd5c5ff171bcd508e3b466c
Author: Enrico Tröger <enrico.troeger@uvena.de>
Date:   Sat Apr 25 16:17:03 2015 +0200

    Add PowerShell tag parser

@masatake did when merging:
* edit this message,
* replace -1, the value representing unknown kind with
  KIND_GHOST_INDEX, and
* extend the original test case for testing signature.

Known issue:
* q/qualified extra doesn't work, and
* the version of GPL doesn't written explicitly.

### DIFF
--- a/win32/ctags_vs2013.vcxproj.filters
+++ b/win32/ctags_vs2013.vcxproj.filters
@@ -120,6 +120,9 @@
     <ClCompile Include="..\main\portable-scandir.c">
       <Filter>Source Files\Main</Filter>
     </ClCompile>
+    <ClCompile Include="..\main\powershell.c">
+      <Filter>Source Files\Main</Filter>
+    </ClCompile>
     <ClCompile Include="..\main\promise.c">
       <Filter>Source Files\Main</Filter>
     </ClCompile>
--- a/Units/parser-powershell.r/simple-powershell.d/args.ctags
+++ b/Units/parser-powershell.r/simple-powershell.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=no
+# "qualified" extra doesn't work.
+# --extras=+q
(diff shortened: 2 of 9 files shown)


win32/ctags_vs2013.vcxproj.filters
```diff
@@ -120,6 +120,9 @@
     <ClCompile Include="..\main\portable-scandir.c">
       <Filter>Source Files\Main</Filter>
     </ClCompile>
+    <ClCompile Include="..\main\powershell.c">
+      <Filter>Source Files\Main</Filter>
+    </ClCompile>
```
k-takata: `powershell.c` is in the `parser` folder, not `main`.

masatake: Thank you.

Doesn't test running at appveyor use win32/ctags_vs2013.vcxproj.filters ?

Ideally, this kind of mistake should be detected by a test.

k-takata: *.vcxproj.filters is only used by the Visual Studio IDE for showing the (virtual) directory tree.
It is not used for building, so it cannot detect by CI.

PR #2231, opened 2019-10-07
buildsys: Add some files to EXTRA_DIST

k-takata:
Packcc's documents were not included in the distribution.
Some files in man/ were also missing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,13 +7,15 @@ EXTRA_DIST   = README.md autogen.sh \
 	       .indent.pro .uncrustify.cfg \
 	       gnu_regex/README.txt gnu_regex/regcomp.c gnu_regex/regexec.c \
 	       gnu_regex/regex_internal.c gnu_regex/regex_internal.h \
+	       misc/packcc/.gitignore misc/packcc/LICENSE.txt \
+	       misc/packcc/README.md \
 	       $(WIN32_HEADS) $(WIN32_SRCS) mk_mingw.mak mk_mvc.mak \
 	       win32/mkstemp/COPYING.MinGW-w64-runtime.txt \
 	       win32/ctags_vs2013.sln \
 	       win32/ctags_vs2013.vcxproj win32/ctags_vs2013.vcxproj.filters \
 	       win32/gen-repoinfo.bat \
-	       man/tags.5.rst \
-		$(PEG_INPUT) $(OPTLIB2C_INPUT)
+	       man/Makefile man/README man/tags.5.rst \
+	       $(PEG_INPUT) $(OPTLIB2C_INPUT)
 
 bin_PROGRAMS = ctags
 noinst_LIBRARIES = libctags.a


k-takata:
Actually, there are more files that are missing from EXTRA_DIST:

```
*.yml
.github/
dictfiles/
docs/
main/
	exclusion_p.h
	Makefile
misc/
	validators/
	badinput.c
	budge
	budge.ctags
	cspell
	ctags-optlib-mode.el
	dist-test-cases
	enumstr.sh
	gen-repoinfo
	gencxxtypedumper.sh
	make-dictfiles.sh
	mk-interactive-request.sh
	pull-packcc.sh
	review
	roundtrip
	spelling.ctags
	src-check
	tinst
	travis-check.sh
old-docs/
parsers/
	cxx/
		Makefile
	Makefile
peg/
	.gitignore
win32/
	license/
	appveyor.bat
```

Not sure all the files should be added.
At least, it seems that `main/exclusion_p.h` is not used.

masatake:
Do you think we should add .gitignore to EXTRA_DIST?
I think the file makes sense only when it is in a git repo.
.tar.gz archive is disconnected(?) from the git repo.
How do you think?

masatake:
At least, "make dist; tar zxvf universal-ctags*.tar.gz; cd universal-ctags-*; ./configure && make && make check && make dist" must work well.
I have just found Makefiles for Tmake are generated but broken. I will fix.

k-takata:
> Do you think we should add .gitignore to EXTRA_DIST?
> I think the file makes sense only when it is in a git repo.

I'm also wondering this.
.gitignore is a file for a git repo. So basically it is not useful for a tarball.
However some kind of grep-like tools (e.g. ag, pt) read .gitignore to speed up searching.
So is it better to add to EXTRA_DIST for completeness?

Some scripts like dist-test-cases, gen-repoinfo and pull-packcc.sh don't work outside the git repo.

And I'm also wondering about the files for CI (e.g. *.yml, appveyor.bat).
They are not useful for tarball at all.

k-takata:
> At least, "make dist; tar zxvf universal-ctags*.tar.gz; cd universal-ctags-*; ./configure && make && make check && make dist" must work well.

`make distcheck` might be used.

masatake:
(Sorry to be late. I will review your pull requests in this weekend.)

masatake:
Sorry, still I cannot find enough time to think about this important topic. I will try to find next week end.

masatake:
LGTM. As you did I think all packcc files should be included. Please, merge this.

I revised the list you made.

As you wrote, main/exclusion_p.h should be removed from our git repo.

Files under misc/validators should be added to EXTRA_DIST.
So people can run validators locally.

misc/dist-test-cases should be added to EXTRA_DIST.
So people can run autogen.sh locally.

mk-interactive-request.sh should be added to EXTRA_DIST.
So people can try ctags --interactive easily.

roundtrip should be added to EXTRA_DIST.
So people can test readtags.

tinst should be added to EXTRA_DIST.
So people can run make tinst.

About win32/license, I'm not sure. Do you have an idea?

In the future, we should add misc/ctags-optlib-mode.el to EXTRA_DIST.

k-takata:
Okay, I will merge this. Other files should be added later if needed.

> misc/dist-test-cases should be added to EXTRA_DIST.
> So people can run autogen.sh locally.

But `misc/dist-test-cases` uses `git ls-files` command, so it won't work outside the git repository.

> About win32/license, I'm not sure. Do you have an idea?

These files are used when creating binary packages in universal-ctags/ctags-win32, so they are not necessary for a tarball.

masatake:
> But misc/dist-test-cases uses git ls-files command, so it won't work outside the git repository.

It implies we should not add autogen.sh to EXTRA_DIST.

>> About win32/license, I'm not sure. Do you have an idea?
> These files are used when creating binary packages in universal-ctags/ctags-win32, so they are not necessary for a tarball.

I see.

PR #2234, opened 2019-10-08
buildsys: Make dist-test-cases simpler and faster

k-takata:
Use `git ls-tree` and `git ls-files .` and stop using `sed`.
This makes the script faster especially on Cygwin/MSYS2.
Note that fork(2) is very slow on Cygwin/MSYS2, so reducing the
execution of processes from a script makes it much faster.

Before:

	$ time ./misc/dist-test-cases

	real    1m3.455s
	user    0m15.465s
	sys     0m38.575s

After:

	$ time ./misc/dist-test-cases

	real    0m19.584s
	user    0m2.973s
	sys     0m8.444s

### DIFF
--- a/misc/dist-test-cases
+++ b/misc/dist-test-cases
@@ -39,9 +39,11 @@ make_am_sub_with_git ()
     print_am_header
     echo "EXTRA_DIST    ="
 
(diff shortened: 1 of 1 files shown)


masatake:
Great! Please, merge this.

Issue #1496, opened 2017-06-29
"uninitialized  variable" warning


masatake:
Not only on Windows, we got following warning.

See #1490.

```
  CC       parsers/ctags-html.o
parsers/html.c: In function 'readTag':
parsers/html.c:439:6: warning: 'lineOffset_tmp[1]' may be used uninitialized in this function [-Wmaybe-uninitialized]
      makePromise ("JavaScript", startLineNumber, startLineOffset,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          endLineNumber, endLineOffset, startSourceLineNumber);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/html.c:439:6: warning: 'line_tmp[1]' may be used uninitialized in this function [-Wmaybe-uninitialized]
```

PR #2235, opened 2019-10-08
main,win32: Suppress MinGW warnings

k-takata:
There are several warnings when compiled with MinGW:
https://ci.appveyor.com/project/universalctags/ctags/builds/27959476/job/7xggnuxg5wf97wqh#L673

This suppress some of the warnings.

* Check the existence of direct.h for `_getdrive()`.
* Don't redefine `stat`.
* Use proper size for printf format string.

### DIFF
--- a/main/lregex.c
+++ b/main/lregex.c
@@ -25,6 +25,8 @@
 #endif
 #include <regex.h>
 
+#include <inttypes.h>
+
 #include "debug.h"
 #include "colprint_p.h"
 #include "entry_p.h"
@@ -2753,10 +2755,10 @@ static void   guestRequestSubmit (struct guestRequest *r)
 {
 	const char *langName = getLanguageName (r->lang);
 	verbose ("guestRequestSubmit: %s; "
-			 "range: %lld - %lld\n",
+			 "range: %"PRId64" - %"PRId64"\n",
 			 langName,
-			 (long long)r->boundary[BOUNDARY_START].offset,
-			 (long long)r->boundary[BOUNDARY_END].offset);
+			 (int64_t)r->boundary[BOUNDARY_START].offset,
+			 (int64_t)r->boundary[BOUNDARY_END].offset);
 	makePromiseForAreaSpecifiedWithOffsets (langName,
 											r->boundary[BOUNDARY_START].offset,
 											r->boundary[BOUNDARY_END].offset);
--- a/main/lregex.c
+++ b/main/lregex.c
@@ -25,6 +25,8 @@
 #endif
 #include <regex.h>
 
+#include <inttypes.h>
+
 #include "debug.h"
 #include "colprint_p.h"
 #include "entry_p.h"
@@ -2753,7 +2755,7 @@ static void   guestRequestSubmit (struct guestRequest *r)
 {
 	const char *langName = getLanguageName (r->lang);
 	verbose ("guestRequestSubmit: %s; "
-			 "range: %ld - %ld\n",
+			 "range: %"PRIdPTR" - %"PRIdPTR"\n",
 			 langName,
 			 r->boundary[BOUNDARY_START].offset,
 			 r->boundary[BOUNDARY_END].offset);
--- a/main/lregex.c
+++ b/main/lregex.c
@@ -25,8 +25,6 @@
 #endif
 #include <regex.h>
 
-#include <inttypes.h>
-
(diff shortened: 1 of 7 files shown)


main/lregex.c
```diff
@@ -2753,7 +2755,7 @@ static void   guestRequestSubmit (struct guestRequest *r)
 {
 	const char *langName = getLanguageName (r->lang);
 	verbose ("guestRequestSubmit: %s; "
-			 "range: %ld - %ld\n",
+			 "range: %"PRIdPTR" - %"PRIdPTR"\n",
```
k-takata: Hmm, this still causes warnings on 32-bit MinGW.
https://ci.appveyor.com/project/universalctags/ctags/builds/27963569/job/nbs73twybwv5s63l?fullLog=true#L676
It seems that `off_t` is 64-bit on MinGW and 32-bit on Visual C++ regardless of the bitness of the compiler.
Need a something different way...

k-takata: Always casting `off_t` to `long long` might be the easiest way. Then we can use `%lld` here.

main/lregex.c
```diff
@@ -2753,10 +2753,10 @@ static void   guestRequestSubmit (struct guestRequest *r)
 {
 	const char *langName = getLanguageName (r->lang);
 	verbose ("guestRequestSubmit: %s; "
-			 "range: %ld - %ld\n",
+			 "range: %lld - %lld\n",
```
k-takata: MinGW complains about "%lld". Strange.

https://ci.appveyor.com/project/universalctags/ctags/builds/27974782/job/qclxucel15nj4ufp#L676
```
main/lregex.c: In function 'guestRequestSubmit':
main/lregex.c:2755:11: warning: unknown conversion type character 'l' in format [-Wformat=]
 2755 |  verbose ("guestRequestSubmit: %s; "
      |           ^~~~~~~~~~~~~~~~~~~~~~~~~~
main/lregex.c:2756:15: note: format string is defined here
 2756 |     "range: %lld - %lld\n",
      |               ^
```

k-takata: Hmm, it seems that MinGW's `-Wformat` doesn't support "%lld" for compatibility for very old versions of msvcrt.dll:
https://sourceforge.net/p/mingw/bugs/2190/
It might be better to use `PRId64` instead.

masatake:
I will review this next. Give me time.

masatake:
Other than c0a9a4a, they look good to me.

c0a9a4a may be related to #1496 and #1490 (https://github.com/universal-ctags/ctags/pull/1490#issuecomment-311297098). I will look into the item again.

k-takata:
Ah, right. it is exactly same as #1496.
It seems that `lineOffset_tmp[1]` and `line_tmp[1]` are already initialized when they are going to be used, but gcc compiler doesn't seem to detect it correctly. I think it is false positive. Initializing them with 0 is the easiest way to suppress the warning.

k-takata:
Okay, I removed the commit https://github.com/universal-ctags/ctags/commit/c0a9a4a203f4dd7b1a314be79b0461ace164c5a4 from this PR, and created a separate PR for that (#2252).

masatake:
> Okay, I removed the commit c0a9a4a from this PR, and created a separate PR for that (#2252).

Thank you. Please, merge this.

masatake:
@k-takata, just a question. 
Can you rerun "ci/circleci: centos_make" and "ci/circleci: fedora_bmake" with your permission?
It seems that we got errors in preparing the test environments randomly these days. So I had to do "restart workflow" in a circleci page. Doing this is o.k. for me. However, I don't want you to wait for my action for restarting.

k-takata:
> Can you rerun "ci/circleci: centos_make" and "ci/circleci: fedora_bmake" with your permission?

No. It seems that I don't have a permission to rerun them.

masatake:
> No. It seems that I don't have a permission to rerun them.

I see. I'll try to fix this.

masatake:
@ k-takata, can you login (=! signup)  https://circleci.com via your GitHub account?

At the upper-left edge of the page shown after login, there is a menu choosing a role(?) you use.
In my account, I can choose "universal-ctags" role. How about you?
circleci

If there is no menu item, I have to update/arrange universal-ctags side.

k-takata:
It seems that now I can rerun the workflow. Thank you.

PR #2237, opened 2019-10-09
Update .editorconfig

k-takata:
GNU source's tab width is 8. It is different from indent size.

(Should indent size be 2 for GNU style?)


masatake:
LGTM.

masatake:
> (Should indent size be 2 for GNU style?)

I'm not sure.

PR #2238, opened 2019-10-10
units: Make units faster

k-takata:
`units` is really slow on Cygwin/MSYS2.
This makes it faster. (About 3.5x on my environment.)

* Reduce execution of external processes.
* Run `units` and `tmain` in parallel.
  - Use `&` and `wait`.
  - Use temporary files to store the results.
  - Display the result in a single command to avoid the output being mixed.
* Use bash's pattern matching instead of echo + grep. (Not sure this is accepted.)

Additionally, this supports `make units CATEGORIES=ROOT`.

### DIFF
--- a/misc/units
+++ b/misc/units
@@ -47,6 +47,7 @@ VALIDATORS=
 _CMDLINE=
 _CMDLINE_FOR_SHRINKING=
 _PREPERE_ENV=
+_FEATURE_LIST=
 readonly _DEFAULT_CATEGORY=ROOT
 readonly _TIMEOUT_EXIT=124
 readonly _VG_TIMEOUT_FACTOR=10
@@ -206,6 +207,15 @@ check_units ()
     return 1
 }
 
+init_features()
+{
+    _FEATURE_LIST=$( ${CTAGS} --quiet --options=NONE \
+			      --list-features --with-list-header=no \
+			      2> /dev/null \
+		| "${_LINE_SPLITTER}" \
+		| cut -f 1 -d ' ')
+}
+
 check_features()
 {
     local flag="$1"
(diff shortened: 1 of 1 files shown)


misc/units
```diff
@@ -735,9 +764,7 @@ unwanted_file ()
 {
     local file=$1
     # ignore backup files
-    if echo "$file" | grep -q '~$'; then
-	return 0
-    elif echo "$file" | grep -q '\*'; then
+    if [[ "$file" =~ ~$|\* ]]; then
```
k-takata: At least, dash doesn't support this. Is it acceptable?

k-takata: I updated the script so that also works on dash.

masatake:
Sorry to be late.


Even on GNU/Linux, your changes make three times faster than before:
```
before bash make units SHELL=/bin/bash  25.67s user 21.47s system 111% cpu 42.248 total
before dash make units SHELL=/bin/dash  20.63s user 14.69s system 108% cpu 32.448 total
after  bash make units SHELL=/bin/bash  23.28s user 24.65s system 384% cpu 12.450 total
before bash make units SHELL=/bin/dash  20.08s user 14.07s system 337% cpu 10.107 total
---------------------------------------------------------------------------^^^^^^
```

I will review your patches

masatake:
f00bd6f0604c0ed2790ef514668a2f06f6f86d92: LGTM

k-takata:
Updated.

BTW, now I consider replacing `misc/units` with a python script.
Current prototype is about three times faster than this PR on Cygwin on my local PC.

masatake:
> Updated.

I added more comments. Could you look at that?

k-takata:
I rearranged the commits.

masatake:
Perfect!!
Please, merge this.

PR #2240, opened 2019-10-16
units: Add misc/units.py

k-takata:
Create a replacement for `misc/units` by python.
This speeds up `make units` especially on Cygwin/MSYS2.

Related: #2238

### DIFF
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -1,7 +1,7 @@
 # -*- makefile -*-
 .PHONY: check units fuzz noise tmain tinst tlib clean-units clean-tlib clean-tmain clean-gcov run-gcov codecheck cppcheck dicts cspell validate-input
 
-EXTRA_DIST += misc/units
+EXTRA_DIST += misc/units misc/units.py
 EXTRA_DIST += misc/tlib misc/mini-geany.expected
 
 DIST_SUBDIRS = Tmain Units
@@ -108,9 +108,26 @@ units: $(CTAGS_TEST)
 		SHOW_DIFF_OUTPUT=--show-diff-output;		\
 	fi;							\
 	builddir=$$(pwd); \
+	if ! test x$(PYTHON) = x; then	\
+		PROG=$(PYTHON);		\
+		SCRIPT=$(srcdir)/misc/units.py;	\
+		if type cygpath > /dev/null 2>&1; then	\
+			builddir=$$(cygpath -m "$$(pwd)");	\
+			if ! test x$(SHELL) = x; then	\
+				SHELL_OPT=--shell=$$(cygpath -m $(SHELL));	\
+			fi;	\
+		else	\
+			if ! test x$(SHELL) = x; then	\
+				SHELL_OPT=--shell=$(SHELL);	\
+			fi;	\
+		fi;	\
+	else	\
+		PROG=$(SHELL);		\
+		SCRIPT=$(srcdir)/misc/units;	\
+	fi;	\
 	mkdir -p $${builddir}/Units && \
 	\
-	c="$(srcdir)/misc/units run \
+	c="$${SCRIPT} run \
 		--ctags=$(CTAGS_TEST) \
 		--languages=$(LANGUAGES) \
 		--categories=$(CATEGORIES) \
@@ -118,9 +135,10 @@ units: $(CTAGS_TEST)
 		--with-pretense-map=$(PMAP) \
 		$${VALGRIND} --run-shrink \
 		--with-timeout=`expr $(TIMEOUT) '*' 10`\
+		$${SHELL_OPT} \
 		$${SHOW_DIFF_OUTPUT}"; \
 	 TRAVIS=$(TRAVIS) APPVEYOR=$(APPVEYOR) CIRCLECI=$(CIRCLECI)\
-		 $(SHELL) $${c} $(srcdir)/Units $${builddir}/Units
+		 $${PROG} $${c} $(srcdir)/Units $${builddir}/Units
 
 clean-units:
 	$(SILENT) echo Cleaning test units
@@ -155,15 +173,33 @@ tmain: $(CTAGS_TEST)
 		SHOW_DIFF_OUTPUT=--show-diff-output;		\
 	fi;							\
 	builddir=$$(pwd); \
+	if ! test x$(PYTHON) = x; then	\
+		PROG=$(PYTHON);		\
+		SCRIPT=$(srcdir)/misc/units.py;	\
+		if type cygpath > /dev/null 2>&1; then	\
+			builddir=$$(cygpath -m "$$(pwd)");	\
+			if ! test x$(SHELL) = x; then	\
+				SHELL_OPT=--shell=$$(cygpath -m $(SHELL));	\
+			fi;	\
+		else	\
+			if ! test x$(SHELL) = x; then	\
+				SHELL_OPT=--shell=$(SHELL);	\
+			fi;	\
+		fi;	\
+	else	\
+		PROG=$(SHELL);		\
+		SCRIPT=$(srcdir)/misc/units;	\
+	fi;	\
 	mkdir -p $${builddir}/Tmain && \
 	\
-	c="$(srcdir)/misc/units tmain \
+	c="$${SCRIPT} tmain \
 		--ctags=$(CTAGS_TEST) \
 		--units=$(UNITS) \
 		$${VALGRIND} \
+		$${SHELL_OPT} \
 		$${SHOW_DIFF_OUTPUT}"; \
 	TRAVIS=$(TRAVIS) APPVEYOR=$(APPVEYOR) CIRCLECI=$(CIRCLECI)\
-		$(SHELL) $${c} $(srcdir)/Tmain $${builddir}/Tmain
+		$${PROG} $${c} $(srcdir)/Tmain $${builddir}/Tmain
 
 clean-tmain:
 	$(SILENT) echo Cleaning main part tests
--- a/misc/units.py
+++ b/misc/units.py
@@ -1,9 +1,36 @@
 #!/usr/bin/env python3
 
+#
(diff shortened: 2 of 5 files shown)


k-takata:
I almost implemented `tmain` action and `run` action in `misc/units.py`.
It seems that the speed is almost same as #2238 on Linux, but this is really faster on Windows. It is about twice faster than #2238 on AppVeyor, three times faster on my local environment.

k-takata:
Not implemented yet for action `run`:
* --run-shrink
* --with-pretense-map

masatake:
--run-shrink

   If you are o.k., I will implement this AFTER merging this pull request.
  This is a good time for me to study python :-P

--with-pretense-map

   This feature is originally designed for implementing new Java parser.
   We don't need now. 
   

ready to review?

misc/units.py
```diff
@@ -0,0 +1,1044 @@
+#!/usr/bin/env python3
```
masatake: Please, put a copyright header with your name.

k-takata:
Updated.

* Rebased upon master. (Mainly for including #2243.)
* Add copyright comment.
* Implement `clean` and `clean-tmain` actions. (Useful for Windows.)
* Add `--threads` option.
* Add help messages.
* Implement `--with-pretense-map` (even it is not needed).
* Make `units` as an alias of `run` action.

I think it is ready for review.

misc/units.py
k-takata: done.

makefiles/testing.mak
```diff
@@ -107,20 +107,31 @@ units: $(CTAGS_TEST)
...
+	if ! test x$(PYTHON) = x; then	\
+		PROG=$(PYTHON);		\
+		SCRIPT=$(srcdir)/misc/units.py;	\
+	else	\
+		PROG=$(SHELL);		\
+		SCRIPT=$(srcdir)/misc/units;	\
+	fi;	\
+	if ! test x$(SHELL_TEST) = x; then	\
+		SHELL_OPT=--shell=$(SHELL_TEST);	\
+	fi;							\
```
masatake: When I implement shrinker in python, let's remove misc/units!

masatake:
It seems that you are still tuning the details. When I should review units.py, please, remove "WIP" from the subject.
I read files other than units.py, and I think they are ready to merge.

makefiles/testing.mak
k-takata: The problem is that I have only implemented 5 actions: help, run, clean, tmain and clean-tmain.
Other actions (fuzz, shrink, noise, chop|slap, validate-input) are not implemented.

Also, I don't know how to install python3 on CentOS. So, `ci/circleci: centos_make` still uses `misc/units`.

masatake: > The problem is that I have only implemented 5 actions: help, run, clean, tmain and clean-tmain.
Other actions (fuzz, shrink, noise, chop|slap, validate-input) are not implemented.

I see. Let's keep the shell script for awhile.

> Also, I don't know how to install python3 on CentOS. So, ci/circleci: centos_make still uses misc/units.

This is my job.  I will take a look the way to install python3 on CentOS.

masatake: I've tried installing python3 to the environment established by docker.io/centos:latest.
What we have to do is quite simple: just adding python3 to yum command line like:

```
diff --git a/circle.yml b/circle.yml
index e143612a..db184682 100644
--- a/circle.yml
+++ b/circle.yml
@@ -40,7 +40,7 @@ jobs:
            name: Install build tools
            # TODO: enable spell checker
            command: |
-              yum -y --enablerepo=PowerTools install gcc automake autoconf pkgconfig make libxml2-devel jansson-devel libyaml-devel findutils diffutils
+              yum -y --enablerepo=PowerTools install python3 gcc automake autoconf pkgconfig make libxml2-devel jansson-devel libyaml-devel findutils diffutils
        - run:
            name: Build
            command: |
```

I hope this helps you.

k-takata:
It seems that sometimes mkstemp() on Win32 fails with some reason:
https://ci.appveyor.com/project/universalctags/ctags/builds/28237000/job/sbyv7o6j620kfdsc#L2642
> ctags.exe: cannot open temporary file: C:\msys64\tmp/tags.P1kAlM : Permission denied

Perhaps, mkstemp() has some potential problem, when running multiple commands in parallel?

makefiles/testing.mak
k-takata: I confirmed that it worked.

k-takata:
> It seems that sometimes mkstemp() on Win32 fails with some reason:

I created a PR for this: #2249.

k-takata:
Yesterday, the latest commit takes about 26 minutes on AppVeyor:
https://ci.appveyor.com/project/universalctags/ctags/builds/28349965
(Maybe it was crowded.)
But now, the same commit takes only 11 minutes on AppVeyor:
https://ci.appveyor.com/project/universalctags/ctags/builds/28349965
This is about 2x faster than the current master branch.

I think that the tuning is almost done, but I will do the final check on this weekend.

win32/appveyor.bat
```diff
@@ -86,22 +86,22 @@ copy %ICONV_BUILD_DIR%\msvc10\iconv.dll %APPVEYOR_BUILD_FOLDER% > nul
 cd %APPVEYOR_BUILD_FOLDER%
 nmake -f mk_mvc.mak WITH_ICONV=yes ICONV_DIR=%ICONV_DIR% PDB=yes || exit 1
 
-:: Backup VC binaries
-mkdir vc
-move *.exe vc > nul
+rem :: Backup VC binaries
```
masatake: In this commit, you converted some lines to comments with "rem".
I wonder why these conversions are needed.

k-takata: Before this commit, we generated a Makefile by MSYS2 and used it for testing. When generating the Makefile, MSVC binaries were overwritten by MSYS2, so we needed to backup them.
Now I stop generating the Makefile, and use misc/units.py directly instead. This reduces the testing time about 2 minutes (or more).

`rem` is a comment. I can also remove those lines.

k-takata:
I finished tuning. Now the test on AppVeyor takes only about 8 minutes.

win32/appveyor.bat
masatake: I see.
As you proposed, I would like you to remove the comment lines. git may help us when we want to go back to the original code.

k-takata: Okay, updated.

k-takata:
I found a minor issue: https://travis-ci.org/universal-ctags/ctags/jobs/603640589#L899
> /bin/bash: line 15: type: cygpath: not found

This error should not be shown. I will fix this.

k-takata:
> This error should not be shown. I will fix this.

Fixed.
Also adjusted Travis CI. (Not directly related to units.py, though.)
* Stop using -j2 for `make check`.
* Reduce configure and make from 3 times to 2 times.

masatake:
> Also adjusted Travis CI. (Not directly related to units.py, though.)

I would like to make one pull request including all changes not related to units.py, and merge them first. So you can simplify this pull request for units.py.

I would like you to simplify the pull request. e.g. I think there is no benefit in adding 400fa1f 
as a separated commit to the ctags git repo. It may be making a sense as your work log. 
8757c1f  and 2113683 are other examples.
If units.py is already been in our repository, the separated commits make sense.
However, units.py is a newly introduced file. So the history of developing units.py is
not so important.

1e09f50 can be reduced by "git rebase master".

There may be important change for units.py. Making a separated commit for the change will make sense. However, minor changes should be squashed into one commit.

k-takata:
> I would like to make one pull request including all changes not related to units.py, and merge them first.

I created #2274 for this.

> I would like you to simplify the pull request.

I will revise the commit history later.

k-takata:
Revised the history a bit. Still WIP.

k-takata:
Finished revising the commit history.

masatake:
fedora_bmake fails but it is something to do with its yum repository side. Let's ignore it.

k-takata:
Thank you for approving.
Next task might be implementing `--run-shrink` (and some other features).

masatake:
Thank you very much.

> Next task might be implementing --run-shrink (and some other features).

Can I ask you this item? You will see `dd` based crazy code.

In the next two weeks, I cannot find enough time for ctags.

k-takata:
I saw the code of run-shrink a few days ago, but I couldn't understand what it is doing.
I need a time for understanding it. If there is a working example, it might help a lot.

masatake:
The idea of shrinker comes from "git bisect".
The shrinker tries fo find semi-minimum input for crashing ctags.

Assume units detects a crash for ctags process for the input "abcdef" from the exit status code.
units runs ctags again with input "abc". If the crash is not reproduced for the input,
units runs ctags again with input "abcde". If the crash is not reproduced for the input,
units runs ctags again with input "abcd". The crash may be reproduced.
This is the process of tophalf.

Then unts runs ctags again with input "cd". If the crash is not reproduced for the input,
units run ctags again with input "bcd". If the crash is not reproduced for the input,
we can say "abcd" may be the semi-minimum input crashes ctags.
I don't consider mid-bytes in the input like "abd", "acd", "ad", and so on.

As the examples show, byte operations are needed to implelemnt the shrinker. So I used dd to generate the test input derived from the original input.

For implementing shrink, you may have to modify a parser triggers crash. See at the end of main/main.c.
A hidden parser for testing is there.

Of course, you can skip this item. As I wrote, the dd based code is a bit crazy. 
Shrinker is run seldomly. It cannot be a performance bottle neck. So keeping it as a shell script is one of choice. In that case, I would like you to add code running the shrinker shell script to your units.py.

k-takata:
Python can handle bytes object, so run-shrink might be able to be implemented without dd.
The ported code can be simpler than shell script. But I don't think porting is easy.

> So keeping it as a shell script is one of choice. In that case, I would like you to add code running the shrinker shell script to your units.py.

This sounds reasonable.

PR #2241, opened 2019-10-16
Pascal: Parse first `type` identifier

ntrel:
This is a change ported from Geany. Needs tests (submitting as draft so it doesn't get lost).


masatake:
Test cases under review-needed.r were imported from exuberant-ctags.
As the directory name shows, we don't understand what kind of tags output is expected for the input.

If you know Pascal, I would like you to review the test case first.
If you don't want to review, introduce a new test case for your change instead of extending 
review-needed.r/bug612019.pas.t.

ntrel:
> I would like you to review the test case first

Test case looks fine.

> introduce a new test case for your change instead of extending
review-needed.r/bug612019.pas.t

This pull will modify the tags generated by the existing test case. All I did was add a function.

masatake:
> Test case looks fine.
Thank you.

> This pull will modify the tags generated by the existing test case. All I did was add a function.

Your work is nothing to do with bug612019.pas, isn't it?
The name of the test case points a bug report at Exuberant-ctags project run at Sourceforge.

ntrel:
> Test case looks fine

Although the procedures and function (which I added) should maybe have an empty parameter list.

ntrel:
> Your work is nothing to do with bug612019.pas, isn't it?

It has a type declaration.

masatake:
> It has a type declaration.

Do you mean the original test input has a type declaration, and so you reuse the test input for your work?

Units/review-needed.r/bug612019.pas.t/expected.tags
```diff
@@ -1,3 +1,4 @@
-Test1	input.pas	/^    procedure Test1;$/;"	p
```
masatake, line 1: Test1 is not tagged with the change.

PR #2243, opened 2019-10-17
units: Don't test parser-cxx.r/cxx-keywords-as-c-identifiers twice

k-takata:
`parser-cxx.r/cxx-keywords-as-c-identifiers` had two input files: `input.c` and `input.h`.
This causes the test being executed twice. Actually, `input.c` is not used. So, delete it.

(Found while implementing #2240.)


masatake:
LGTM.

PR #2249, opened 2019-10-22
main,win32: Work around that sometimes mkstemp() fails

k-takata:
It seems that mkstemp() (from MinGW-w64) sometimes fails with permission
error or some other reasons, especially when running multiple commands
in parallel.

mkstemp() uses rand() but srand() was not called in our code.
Initialize srand() and retry mkstemp() several times as a workaround.

See https://github.com/universal-ctags/ctags/pull/2240#issuecomment-544528511.

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -28,7 +28,9 @@
 #include <crt_externs.h>
 #endif
 
+#include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


masatake:
LGTM. 
Very mysterious.

PR #2254, opened 2019-10-23
Units: Delete unnecessary filter

k-takata:
`parser-c.r/line_directives.c.d/filter` is not needed anymore because we have merged #2199.


masatake:
LGTM.

PR #2255, opened 2019-10-23
docs: Update units.rst

k-takata:
* Add headings about Test facility and Units facility.
* Fix the filename of `Units/TEST/filter`. `filter-*.*` are not supported.


masatake:
LGTM.

PR #2258, opened 2019-10-24
buildsys: Add --enable-static option to ./configure

k-takata:
Support static linking on MinGW.

Actually, we can enable static linking even without this PR, however, many options are needed for that.
This PR makes it easy.


masatake:
LGTM.

PR #2260, opened 2019-10-24
buildsys: Don't make Makefiles for each *.d directory

k-takata:
Creating many Makefiles slows down `./configure` especially on Cygwin/MSYS2.
Reduce the number of Makefiles to avoid it. Create only in `*.r` directories.

This has a potential risk that it may cause #1952 again.
I don't know how many files are accepted in one Makefile.

### DIFF
--- a/misc/dist-test-cases
+++ b/misc/dist-test-cases
@@ -60,9 +60,15 @@ make_am_with_git ()
     cd "${git_top}"
     for top_d in $(git ls-tree --name-only HEAD | sort); do
 	if [ -d "${top_d}" ]; then
-	    echo "DIST_SUBDIRS +=" "${top_d}"
-	    make_am_sub_with_git ${top_d} > ${top_d}/Makefile.am
-	    echo "AC_CONFIG_FILES([${git_top}/${top_d}/Makefile])" >> dist.m4
+	    if [ "${top_d##*.}" = "r" ]; then
+		echo "DIST_SUBDIRS +=" "${top_d}"
+		make_am_sub_with_git ${top_d} > ${top_d}/Makefile.am
+		echo "AC_CONFIG_FILES([${git_top}/${top_d}/Makefile])" >> dist.m4
+	    else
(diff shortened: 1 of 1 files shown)


k-takata:
This PR is about 3x faster on Cygwin.

Before:

```
$ time ./configure
real    7m5.832s
user    1m18.179s
sys     3m4.709s
```

After:

```
$ time ./configure
real    2m45.256s
user    0m29.296s
sys     1m8.806s
```

masatake:
LGTM.

> This PR is about 3x faster on Cygwin.
Nice!

k-takata:
Some values which might be related:

```console
$ getconf ARG_MAX    # on Cygwin
32000

$ getconf ARG_MAX    # on Ubuntu 16.04 on WSL
2097152

$ (cd Tmain; git ls-files .) | wc -c
24952

$ (cd Units; git ls-files *.d) | wc -c
7807

$ (cd Units; git ls-files .) | wc -c
123992
```

k-takata:
> This has a potential risk that it may cause #1952 again.

What do you think of this risk?
The most risky directory is `Tmain`, because this PR creates only one Makefile under `Tmain`.
The total size of filenames under `Tmain` is now 24952 bytes, and `ARG_MAX` on Cygwin is 32000.
(I'm not sure whether #1952 is actually restricted by `ARG_MAX`, though.)

If this can be accepted, I will merge this.

masatake:
units has the way to categorize test cases with `*.r` dir. So reducing the top-level `*.d` is easy.
train doesn't have the way. It can be trouble. Even we have `*.r` feature in Tmain, I wonder whether we can categorize test cases under Tmain: no good apsect to categorize them.

masatake:
> If this can be accepted, I will merge this.

I would like to consider when it becomes real trouble.
Let's merge this.

k-takata:
OK, I will merge this.

masatake:
Another option is stopping bundling test cases to tar.gz.

PR #2262, opened 2019-10-25
Tmain: Speed up filter_by_column_index

k-takata:
`filter_by_column_index` function is really slow on Cygwin/MSYS2.  This
is used by the test `option-extras-enabling-all` and it takes about 15
seconds in my environment.  Use awk if available.  Now it takes less
than 2 seconds.

### DIFF
--- a/Tmain/utils.sh
+++ b/Tmain/utils.sh
@@ -116,15 +116,19 @@ filter_by_column_index()
 	local column
 	local tmp
 
-	while read line; do
-		tmp=0
-		for column in $line; do
-			if [ $tmp = $index ]; then
-				echo $column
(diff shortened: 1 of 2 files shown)


k-takata:
> Instead, could you add a code to check the availability of awk in tmain target in testing.mak?

Isn't it better to check inside `misc/units` by using `check_availability awk` in the `action_tmain` function?

masatake:
> Isn't it better to check inside misc/units by using check_availability awk in the action_tmain function?

YES!  Though I remembered I wrote a function or command for checking, I could not remember where I wrote.

LGTM. Please, merge this.

PR #2264, opened 2019-10-26
CXX: handle '>>>' as the end of template parameter list

masatake:
Close #2263.

PR #2265, opened 2019-10-27
[WIP] C++:  make tags for type parameters

masatake:
See #2264.

@pragmaware, this is my version that is for capturing type parameters in template prefix.
I used corkIndex to track type parameters unlike your idea: tracking by CXXToken.
cork indexes for type parameters are stored to intArray.
To avoid memory leak, trashbox is used for the intArray object.

I'm thinking about introducing a symbol table by extending the cork queue.
So I think using cork inde for tracking type parameters is quite natutal.

I still get unexpected results for some test cases. So I marked [WIP] on this pull request.

### DIFF
--- a/Units/parser-cxx.r/bug-github-2263.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/bug-github-2263.cpp.d/expected.tags
@@ -1,24 +1,41 @@
+T	input.cc	/^template <class T> class C0 {$/;"	Z
 C0	input.cc	/^template <class T> class C0 {$/;"	c	file:
 t	input.cc	/^  T t;$/;"	m	class:C0	typeref:typename:T	file:
+T	input.cc	/^template <class T, typename T0 = C0<const T>> class C1 {$/;"	Z
+T0	input.cc	/^template <class T, typename T0 = C0<const T>> class C1 {$/;"	Z
 C1	input.cc	/^template <class T, typename T0 = C0<const T>> class C1 {$/;"	c	file:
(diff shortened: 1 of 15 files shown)


masatake:
For the input
```
template<typename T>
struct A {
    template<class X1> void g1(T, X1); // member template
};
```

How the output of `ctags -o - --kinds-C++=*` should be?

What we got now is:
```
A	/tmp/f.cpp	/^struct A {$/;"	s	file:
T	/tmp/f.cpp	/^template<typename T>$/;"	Z	struct:A
X1	/tmp/f.cpp	/^    template<class X1> void g1(T, X1); \/\/ member template$/;"	Z	struct:A
g1	/tmp/f.cpp	/^    template<class X1> void g1(T, X1); \/\/ member template$/;"	p	struct:A	typeref:typename:void	file:
```

The scope of X1 is A. Should this be g1? However, the kind of g1 is p.

masatake:
I found more interesting input in https://en.wikipedia.org/wiki/Template_(C%2B%2B)#Variable_templates  .

pragmaware:
The template part is already tracked in a separate token chain. The purpose is exactly the same you're pursuing here: work around the fact that the template is parsed out of the proper scope. So for the type parameters we can use CXXToken pointers and we don't need to track them via cork queue.
See #2267 which adds template type parameters (with Z letter like in this PR) and also improves template parsing a bit. It even fixes one of our old tests.

masatake:
I see. I will take a look at your pull request.

pragmaware:
I have pushed an update that fixes a naming issue.

PR #2267, opened 2019-10-28
CXX: Improve template parsing. Emit tags for template type parameters.

pragmaware:
Smarter template parsing: less confused by less-than, greater-than, shift-left, shift-right operators.
Also emits tags for template type parameters (with proper scope).


masatake:
I added myself as a reviewer to study the code.

pragmaware:
Some units are still missing but I'll look at them tonight.

pragmaware:
I hate coveralls.

k-takata:
> I hate coveralls.

If you hate that coveralls reporting decreasing of the coverage, we can set a threshold for that. (Not sure what threshold is good. 0.05% or 0.1%?)

We also use codecov and it is easier to read than coveralls, but the result is not posted on this PR. I don't know why.

pragmaware:
Well.. code coverage is good and coveralls is a valuable tool for pointing out the uncovered pieces of code. The problem is that sometimes it's very hard to shut it up.

For instance, in this case there are some bits of cleanup code that are just good practice to have. They are unreachable in the current state of the source (at least via unit tests) but they might become reachable later and their existence is a "safety net" we actually rely on.
In other instances there are pieces of code that are reachable if we feed ctags utterly broken input. It just makes no sense to do so in the units: we can't test every possible broken-code-handling branch.

I've found out that there are special words that will make coveralls skip a range of lines. I have added them. Now coveralls doesn't complain about the specific lines but the coverage percentage is down yet more... I have no clue on how to fix it.

PR #2268, opened 2019-10-28
SQL: handle "as" and "is" in different way when predicating it as a reserved word

### LINKED ISSUE #2266: parsers: Clang warning in sql.c
The following warning is shown when compiled with clang:

```
../parsers/sql.c:379:28: warning: initializer overrides prior initialization of
      this subobject [-Winitializer-overrides]
        [KEYWORD_is]            = 1 & 0&1&1&1 & 1&1 & 1,
                                  ^~~~~~~~~~~~~~~~~~~~~
../parsers/sql.c:347:28: note: previous initialization is here
        [KEYWORD_is]            = 1 & 1&1&1&1 & 1&1 & 1,
                                  ^~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

`[KEYWORD_is]` is initialized twice, and the values are different. I don't know which value is correct.

masatake:
Close #2266.

TODO:
I would like to add a test case for this item.
I should include the original bug report to this commit log.


masatake:
The change is mainly about build-phase. So I recognized making a test doesn't make much sense (after writing the test case after installing a container image for postgresql.)

PR #2270, opened 2019-10-28
buildsys: Add some more files to EXTRA_DIST

k-takata:
See also: https://github.com/universal-ctags/ctags/pull/2231#issuecomment-539025926


masatake:
7aadb35  looks good.

I wonder whether we should bundle the files under docs or not.
The original files were written for developers and potential contributors.
(So I allowed myself to write them in broken English.)

My idea was that instead of bundling all the docs files, converting parts of docs files that are useful for normal users to man pages. e.g. converting docs/parser-cxx.rst to man/ctags-parser-cxx.7.rst.

This is just my idea. I cannot find enough time to work on ctags these days. So if you have a strong opinion, please, go ahead to the way you think better. This is applicable to all the pull requests you opened and will open.

k-takata:
I think the docs files are worth to be included in a archive file. Especially, documents for building ctags are useful for source code package. (It's not a strong opinion, though.)

masatake:
> I think the docs files are worth to be included in a archive file. Especially, documents for building ctags are useful for source code package. (It's not a strong opinion, though.)

I see. As you wrote, the instruction for installation should be in the archive.
Please, merge this. Thank you.

PR #2271, opened 2019-10-28
main: Delete unused file

k-takata:
See also:
https://github.com/universal-ctags/ctags/pull/2231#issuecomment-539025926
https://github.com/universal-ctags/ctags/pull/2231#issuecomment-544234701


masatake:
Could you keep Makefiles?
Sometimes, I would like to run make from a subdirectory.

k-takata:
restored.

PR #2274, opened 2019-10-28
travis: Revise build process

k-takata:
Derived from #2240.

* Stop using -j2 for make check. This makes the log readable.
* Reduce configure and make from 3 times to 2 times. This makes the CI about 30 seconds faster.


masatake:
> Stop using -j2 for make check. This makes the log readable.

I agree with this strongly.

PR #2277, opened 2019-10-28
parser: Remove dead increments

alanbarr:
Detected by clang scan-build


masatake:
I would like to reject this PR.
The results of incrementations are not used NOW.
However, when the parsers are extended, the developers may expect the variables are incremeted.

masatake:
BTW, thank you very much for splitting pull requests into smaller ones.

PR #2281, opened 2019-10-29
appveyor: Use `make -j2` to speed up the builds

k-takata:
Also remove `--enable-iconv`, because now it is the default.

### DIFF
--- a/win32/appveyor.bat
+++ b/win32/appveyor.bat
@@ -155,7 +155,7 @@ bash -lc "for i in {1..3}; do pacman --noconfirm --noprogressbar -S --needed min
 
 bash -lc "./autogen.sh"
 :: Use static link.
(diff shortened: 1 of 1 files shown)


k-takata:
This makes AppVeyor about 1 or 2 minutes faster.
I merge this.

(Not sure why fedora_bmake fails, but it is not related to this.)

```
Last metadata expiration check: 0:00:41 ago on Tue Oct 29 14:45:16 2019.
Package gcc-9.2.1-1.fc31.x86_64 is already installed.
Package pkgconf-pkg-config-1.6.3-2.fc31.x86_64 is already installed.
Error: 
 Problem: conflicting requests
  - nothing provides mk-files needed by bmake-20180512-4.fc31.x86_64
(try to add '--skip-broken' to skip uninstallable packages)
```

PR #2284, opened 2019-10-30
typescript: Suppress clang warning

k-takata:
Clang warns this:

```
../parsers/typescript.c:740:18: warning: passing an object that undergoes default argument promotion to 'va_start' has undefined behavior [-Wvarargs]
        va_start (args, skipUnparsed);
                        ^
../parsers/typescript.c:732:56: note: parameter of type 'bool' is declared here
static bool tryInSequence(tokenInfo *const token, bool skipUnparsed, ...)
                                                       ^
```

See also:
https://wiki.sei.cmu.edu/confluence/display/cplusplus/EXP58-CPP.+Pass+an+object+of+the+correct+type+to+va_start

It seems that passing `bool` to `va_start` has undefined behavior.
I think that the first argument after `skipUnparsed` is mandatory.
So, I added `parser` argument after `skipUnparsed`. Then, this causes
type mismatch between `parseComment()` function. So, I also updated the
function to match with `Parser` type.

### DIFF
--- a/parsers/typescript.c
+++ b/parsers/typescript.c
@@ -479,40 +479,40 @@ CTAGS_INLINE void parseWordToken(const int c, tokenInfo *const token, const char
 	result->status = PARSER_FAILED;
 }
 
-CTAGS_INLINE void parseComment(const int c, tokenInfo *const token, commentState *state, parserResult *const result)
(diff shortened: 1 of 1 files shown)


ksamborski:
Ok, good catch!

PR #2287, opened 2019-10-31
units: Implement run_shrink()

k-takata:
Make run_shrink() fallback to shell script version of misc/units.

Related: https://github.com/universal-ctags/ctags/pull/2240#issuecomment-547760119


masatake:
Thank you.

PR #2289, opened 2019-11-01
docs,man: Commit generated man files into docs/man/

k-takata:
Fix the issue in https://github.com/universal-ctags/ctags/pull/2213#issuecomment-536358592.

Replace the placeholders like `@CTAGS_NAME_EXECUTABLE@` with `ctags` and
so on, so that the man pages can be shown correctly on the Web site.

The files under docs/man/ need to be updated manually:

```
$ make -C man QUICK=1 update-docs
```

So a checker for that is added to Travis CI.

Now the docs/man/ files are not symlinks, so they can be correctly
handled on Windows. A workaround on AppVeyor is not needed.

### DIFF
--- a/docs/man/ctags-incompatibilities.7.rst
+++ b/docs/man/ctags-incompatibilities.7.rst
@@ -1,1 +1,191 @@
-../../man/ctags-incompatibilities.7.rst.in
+.. _ctags-incompatibilities(7):
+
+==============================================================
+ctags
+==============================================================
+--------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


k-takata:
Hmm, it seems that the indentation of `misc/travis-check.sh` is inconsistent.
Some part is indented with space and some other part is indented with tab.

The settings of `.editorconfig` are:

```
[*.{c,h,sh}]
indent_size = 4
indent_style = tab
trim_trailing_whitespace = true
```

So, should we use tab?

masatake:
Nice!

PR #2294, opened 2019-11-02
circleci: add fedora:latest + make combination as a new job

masatake:
https://bugzilla.redhat.com/show_bug.cgi?id=1736126
make is not available. So use make instead.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -26,6 +26,32 @@ jobs:
            command: |
              MAKE=bmake bmake validate-input check roundtrip codecheck CIRCLECI=1
 
+   fedoraLatest_make:
+     working_directory: ~/universal-ctags
+     docker:
+       - image: docker.io/fedora:latest
+     steps:
+       - run:
+           name: Install Git and Gdb
+           command: |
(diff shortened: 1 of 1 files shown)


masatake:
???

circle.yml
```diff
@@ -26,6 +26,32 @@ jobs:
...
+       - run:
+           name: Install build tools
+           command: |
+             dnf -y install gcc automake autoconf pkgconfig make aspell-devel aspell-en libseccomp-devel libxml2-devel jansson-devel libyaml-devel findutils
+             dnf -y install jq puppet
+       - run:
+           name: Build
+           command: |
+             bash ./autogen.sh
+             MAKE=bmake ./configure --enable-debugging
```
k-takata: Why you use bmake here?

k-takata:
> make is not available. So use make instead.

Do you mean "bmake is not available"?

BTW, I'm also trying to use fedora:latest + make on #2298.

masatake:
Maybe I was half asleep. I withdraw this. #2294 implements what I wanted.

PR #2297, opened 2019-11-02
Tmain: Fix `make distcheck`

k-takata:
`make distcheck` sets the source directories readonly. So, temporary files or output files should be written in $BUILDDIR. Otherwise, tests will fail with permission errors.

`make distcheck` is important when we create a release.

PR #2298, opened 2019-11-04
buildsys: Fix `make distcheck`

k-takata:
This is an additional PR to #2297.

This fixes:

* Failed to update peg/varlink.[ch]. Generate them in the builddir instead.
* The generated peg/*.[ch] files were distributed. No need to distribute them.
* optlib/*.c were left behind after `make (dist)clean`. Delete them with `make clean`.
* Use fedora:latest to check `make distcheck`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,7 +87,7 @@ endif
 PARSER_SRCS += $(PEG_SRCS)
 PARSER_HEADS += $(PEG_HEADS) $(PEG_EXTRA_HEADS)
 
-libctags_a_CPPFLAGS = -I. -I$(srcdir) -I$(srcdir)/main -DHAVE_PACKCC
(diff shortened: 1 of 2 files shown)


k-takata:
BTW, I'm thinking whether we can stop adding each file under Tmain/ and Units/ into EXTRA_DIST.
https://www.gnu.org/software/automake/manual/html_node/Basics-of-Distribution.html
> You can also mention a directory in EXTRA_DIST; in this case the entire directory will be recursively copied into the distribution.

Can we use this?
Maybe we need to use [dist hook](https://www.gnu.org/software/automake/manual/html_node/The-dist-Hook.html) to remove unwanted files.

masatake:
> Can we use this?

Sure. I didn't know the feature.

PR #2299, opened 2019-11-04
buildsys: Don't add each file to EXTRA_DIST

k-takata:
Add the directories docs, Tmain and Units to EXTRA_DIST instead.
Also add a dist-hook to remove unwanted files from distribution files.

This makes the Makefiles simpler and faster.

See also: https://github.com/universal-ctags/ctags/pull/2298#issuecomment-549336442
https://www.gnu.org/software/automake/manual/html_node/Basics-of-Distribution.html
https://www.gnu.org/software/automake/manual/html_node/The-dist-Hook.html


k-takata:
Now AppVeyor finished within 8 minutes.
https://ci.appveyor.com/project/universalctags/ctags/builds/28621016

masatake:
> Now AppVeyor finished within 8 minutes.
Fine!

k-takata:
@masatake
As I wrote above, I added the dist hook to remove unwanted files from `docs/`, `Tmain/` and `Units/` when executing `make dist`.
https://github.com/universal-ctags/ctags/blob/9fcdb06503d82f76aebdb84de9448642a0f2d54b/Makefile.am#L193-L200

However, I'm not sure whether this is enough. If you find any other unwanted files, please update this.

masatake:
> As I wrote above, I added the dist hook to remove unwanted files from docs/, Tmain/ and Units/ when executing make dist.

I see.

PR #2300, opened 2019-11-04
parsers: Suppress warnings when configured with --enable-debugging

### LINKED ISSUE #2286: parsers: gcc warnings when configured with --enable-debugging
The following two warnings are shown when u-ctags is configured with `--enable-debugging`:

```
../parsers/html.c:189:13: warning: ‘dumpToken’ defined but not used [-Wunused-function]
 static void dumpToken (tokenInfo *token, const char *context, const char* extra_context)
             ^
```

```
../parsers/jscript.c:2617:13: warning: ‘dumpToken’ defined but not used [-Wunused-function]
 static void dumpToken (const tokenInfo *const token)
             ^
```

We can comment out these two functions with `#if 0 ... #endif` or completely remove them to suppress the warnings.
(Not sure which is better.)

k-takata:
Suppress unused function warnings.

Close #2286


masatake:
Thank you.

PR #2301, opened 2019-11-06
docs(web): Rephrase introduction about --input-encoding

k-takata:
* No need to focus on Japanese. Remove the sentence.
  I, as a Japanese, felt that this sentence was to specific to Japanese.
* Mention UTF-8 and some other actual encodings.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -243,10 +243,11 @@ can be used to disable them.
 
 .. TODO: Review...
 
-Japanese programmers sometimes use the Japanese language in source
-code comments. Of course, it is not limited to Japanese. People may
-use their own native language and in such cases encoding becomes an
-issue.
+People may use their own native language in source code comments (or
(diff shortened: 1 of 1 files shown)


masatake:
Could you use "docs(web)" instead of "docs" as prefix for the first line of commit log?
Other than this, the change looks good for me. THANK YOU.

k-takata:
It's okay, but what is "docs(web)" for and "docs" for?

masatake:
> It's okay, but what is "docs(web)" for and "docs" for?

Well,... docs(web) means  "documents published as web pages".
Alternative one is docs(man) which means "documents published as man pages".

If you change both web pages and man pages in a commit, "docs" can be used.

k-takata:
Then, when we edit a file in docs/, normally we should use "docs(web)".

k-takata:
I added one more sentence: "These options are useful for such files."

PR #2302, opened 2019-11-07
buildsys: Fix that peg/varlink.[ch] were still distributed

k-takata:
No need to distribute them.
Remove them from PARSER_SRCS and PARSER_HEADS.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,8 +90,7 @@ PARSER_SRCS += $(YAML_SRCS)
 PARSER_HEADS += $(YAML_HEADS)
 endif
 
-PARSER_SRCS += $(PEG_SRCS)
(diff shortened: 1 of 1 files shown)


k-takata:
I found an issue while trying to create this PR. It is that packcc may be executed twice when running `make -j2`; one for varlink.c, one for varlink.h. When I changed the suffix rules `.peg.c:` and `.peg.h:` to a pattern rule `%.c %.h: %.peg`, it worked fine on GNU make, but didn't work on BSD make.
I also tried some other ways, but they didn't work perfectly. So I gave up.

This PR only fixes the distribution issue.

k-takata:
Moved the last commit to #2306.

PR #2304, opened 2019-11-08
buildsys: Suppress messages from type command

k-takata:
They are redundant.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,8 +2,8 @@
 
 set -xe
 
-type autoreconf || exit 1
-type pkg-config || exit 1
+type autoreconf > /dev/null || exit 1
+type pkg-config > /dev/null || exit 1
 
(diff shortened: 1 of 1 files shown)


masatake:
LGTM.


NOTE:

type bult-in of zsh reports nothing even when the command passed as an argument is not found.

```
[yamato@slave]~/var/ctags-github/docs% bash
[yamato@slave docs]$ type foobarbaz > /dev/null
bash: type: foobarbaz: not found
[yamato@slave docs]$ exit
[yamato@slave]~/var/ctags-github/docs% zsh
[yamato@slave]~/var/ctags-github/docs% type foobarbaz > /dev/null
```

However, we assume that a user runs autogen.sh within the command line "$ ./autogen.sh", and /bin/sh is specified in the script. So "no reporting" cannot be a trouble.

PR #2306, opened 2019-11-10
buildsys: Fix that packcc may fail on MSYS2

k-takata:
When shadow builds are used on MSYS2, packcc fails to read *.peg file.
It seems that MSYS's make uses `\` for directory separator on some
conditions. In such situation, a command `./packcc ..\peg\varlink.peg`
is executed, and `\` is considered as an escape character. Then the
command is interpreted as `./packcc ..pegvarlink.peg`, and it fails.

(Derived from #2302.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -151,9 +151,9 @@ packcc_verbose_0 = @echo PACKCC "    $@";
 PACKCC = $(top_builddir)/packcc$(EXEEXT)
 SUFFIXES += .peg
 .peg.c:
-	$(packcc_verbose)$(PACKCC) -o $(top_builddir)/peg/$(*F) $<
+	$(packcc_verbose)$(PACKCC) -o $(top_builddir)/peg/$(*F) "$<"
 .peg.h:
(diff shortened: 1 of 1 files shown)


k-takata:
Even after this commit, packcc is executed 3 times:

```
$ make -j2
...
PACKCC     peg/varlink.c
PACKCC     ../peg/varlink.c
PACKCC     ../peg\varlink.h
```

I have no idea how to fix this.

k-takata:
#2298 caused this.
This is a critical issue on MSYS2 when using shadow builds (using a different builddir from srcdir). So I will merge this.

k-takata:
It turns out that this happens when clang is used instead of gcc on MSYS2.
Gcc outputs depfiles (`*/.deps/*.Po`) in mixed style path (`C:/foo/bar`), but clang outputs them in windows style path (`C:\foo\bar`). `Makefile` includes the depfiles and it caused the strange behavior.
Clang can be used on MSYS2, but gcc might be better.

PR #2259, opened 2019-10-24
man: Write about missing options

k-takata:
There are some options missing from the man page.
This PR adds them.
Some explanations are copied from news.rst.

(@masatake This is still WIP, and I pushed this to this repository (not mine) so that you can also update this branch.)

Issue #1711, opened 2018-03-15
Write acceptable characters that can be used in --langdef,--fielddef-, --kinddef-, --extradef-, etc.


masatake:
Though what kind of character we can use as kind letter ~~is written in FORMAT file~~ is not written in FORMAT file , vim accepts only "isalpha" char as kind letter. So ctags should not allow users to define a kind using non-isalpha char. Currently isalnum is used to verify the char used in the letter.

```
[yamato@master]~/var/ctags-github% ./ctags --kinddef-C='3,three,something three things'
./ctags --kinddef-C='3,three,something three things'
ctags: No files specified. Try "ctags --help".
[yamato@master]~/var/ctags-github% ./ctags --kinddef-C='%,three,something three things'
./ctags --kinddef-C='%,three,something three things'
ctags: the kind letter give in "--kinddef-C" option is not an alphabet or a number
```

masatake:
I have to do the same check in --langdef,--fielddef-<LANG>, --kinddef-<LANG>, --extradef-<LANG>, and --regex-<LANG>=/pat/name/KIND-SPEC/. The man page must be updated, too.
readtags doesn't need such limitation. As far as it can search a tag from given tags file, it should any tags file.

masatake:
- Language names:

  - v2 format spec: nothing specified explicitly. The name is used as a value for "language:" field.
    So the rule for "file value" is applied.
  - u-ctags code:  no check is conducted. However, u-ctags prints field in escaped format. So even `\t` is included in the name, the output can be consistent with v3 format.
  - u-ctags man pages: nothing is written.

- Kind names:
  Fixing in progress. See 1737

- Field names:
   TODO

- Extra names:
   TODO

- Role names:
   TODO

masatake:
About --langdef, everything is done. 
The next one is --kinddef. It seems that optlib2c needs some minor work.

masatake:
--kinddef is mostly done in #2335.
--fielddef is the next one.

masatake:
In the future, we must have a test harness and test cases for optlib2c.

masatake:
The last one is roledef.

masatake:
In the future, I would like to limit using `{` character in documentation of the objects (lang, kind, field, extra, and role). However, I would like to do it in a new issue.

PR #2307, opened 2019-11-10
Limit to use "all" as the name for a language

masatake:
This was one of TODO item for releasing.


masatake:
This is related to #1711.

masatake:
```
cp ctags.1.rst ctags-optlib.7.rst ctags-incompatibilities.7.rst tags.5.rst  ../docs/man

Files under docs/man/ are not up to date.

Please execute 'make -C man QUICK=1 update-docs' and commit them.

The command "./misc/travis-check.sh" exited with 1.

```
Nice!  I don't have to remember the command line. The error message tells what I have to do.

k-takata:
https://travis-ci.org/universal-ctags/ctags/jobs/609874484#L447
Oh, #2259 caused the issue, sorry.

k-takata:
I created #2308 to fix it.

misc/travis-check.sh
```diff
@@ -60,6 +60,9 @@ if [ "$TARGET" = "Unix" ]; then
 		if ! git diff --quiet docs/man; then
```
masatake: I think we don't need --quiet option here.
If everything works fine, nothing is printed even if we remove --quiet option.
If something wrong, git diff without --quiet gives us much hint for finding the root cause.

masatake: I will push a change what I think here though I should make another pull request for the purpose.

k-takata:
I merged #2308. Could you rebase on top of it?

misc/travis-check.sh
k-takata: Without --quiet, git doesn't seem to return the error code even the differences are found.

masatake: Oh, I see.

masatake: It looks a bug of git for me:-P.

masatake:
```
       --exit-code
           Make the program exit with codes similar to diff(1). That is, it exits with 1 if
           there were differences and 0 means no differences.

       --quiet
           Disable all output of the program. Implies --exit-code.
```

So I think we can use --exit-code.

Issue #1877, opened 2018-09-16
scheme: add support for Racket


cwfoo:
Add support for the [Racket](https://en.wikipedia.org/wiki/Racket_%28programming_language%29) language.

The Scheme parser could be reused for Racket. Racket file extensions (e.g. `*.rkt`) could be added to these lines:
https://github.com/universal-ctags/ctags/blob/e522743d873abcfa4997fe6bad5f5cf634f57c6d/parsers/scheme.c#L113-L118

masatake:
Could you open a pull request?

masatake:
I don't know Racket well.

As far as reading its web site quikcly, racket can switch grammer (syntax?) for the input file with `#lang` notation. So choosing the scheme parser for ".rkt" file doesn't make good result.

What we need is a Rapcker parser which recognizes `#lang` notation and run a proper subparser.
http://docs.racket-lang.org/search/index.html?q=H%3A

Which non-S-exp language is popoular as Racker language extentsion?
datalog?

masatake:
I see. Thank you.

masatake:
Solved by #2315. Sorry to be late for solving this issue.
 Scribble language is interesting but it is not the time for working for the language.

PR #2315, opened 2019-11-10
scheme: Add racket support to scheme parser

plv:
Addressing #1877 by adding support for the `.rkt` extensions. Change is simple because syntax is the same. Tested locally on some Racket source.


masatake:
Thank you!

PR #2317, opened 2019-11-11
docs,man: Fix escaping backslash

k-takata:
The commit f89c5b6cccaa5ae892e8ef1738b77a36dc196b19 was not enough.
It seems that rst2man converts two backslashes into one, however,
sphinx-build with html builder converts four backslashes into one.
To work around this, surround a backslash with double backquotes.


masatake:
Can I ask you to write this a bit crazy rule to docs/README.md and man/README?
I'm very afraid that I cannot remember the rule.

k-takata:
Updated READMEs.

Issue #621, opened 2015-10-10
Merge support for the Vala language


LemonBoy:
The anjuta project has its own fork of Ctags [2] and it does provide support for vala via some patches [1] that would be nice to integrate into universal-ctags.

[1] http://abderrahim.arablug.org/GSoC2008/ctags-vala.diff
[2] https://git.gnome.org/browse/anjuta/tree/plugins/symbol-db/anjuta-tags

masatake:
Thank you for requesting.

We already have known it(https://github.com/universal-ctags/ctags/blob/master/docs/tracking.rst).

I studied more.

1. The license is explicitly shown at the top of source code(GPLv2+). Good.
2. It seems that it is written in Vala language. (A translator from vala to C is used.)
3. It seems that is has no test case.
4. It extends typeRef field.

Instead of merging the vala translator to universal ctags, I thinkit looks better to make the vala parser a standalone execuable. So universal ctags can utilize the standalone executable(valatags) as a xcmd.

Who maintains the vala parser? Anjuta project?

craigbarnes:
> Who maintains the vala parser? 

It seems like they reuse the parser from libvala, which is the parser used by the Vala compiler. That's probably why the language support is written in Vala.

What is required to hook up a standalone executable to ctags? Is there some documentation for xcmd?

masatake:
Xcmd is gone. See https://github.com/universal-ctags/ctags/issues/1442.

craigbarnes:
> So we have to write the parser for ctags from scratch.

The libvala library that Anjuta uses is callable from C. Is it acceptable for ctags parsers to depend on external dynamic libraries? If merged, the Vala support could be made optional at compile time.

codebrainz:
Geany's [`c.c`](https://github.com/geany/geany/blob/master/ctags/parsers/c.c) has decent Vala support, though I expect it'd be quite painful to bring it into here.

masatake:
@codebrainz, I looked at c.c. There are only a few conditions/branches about Vala. So I think importing it is not so difficult, I guess.

Kind definition for vala can be taken from ctags-vala.diff and c.c of geany.
The last ciritcal issue is that there is no test case.

codebrainz:
> There are only a few conditions/branches about Vala. So I think importing it is not so difficult, I guess.

I wouldn't know, last time I looked in `c.c` I woke up in cold-sweats after nightmares :)

masatake:
[...]

PR #2320, opened 2019-11-13
WIP: Vala

albfan:
Work in progress for vala parser.

Written from scratch.

Right now is just a plain copy/paste from tcl parser

relates to #621


albfan:
I think this is a wrong way to approach the implementation.

albfan:
Reopening this as is the only acceptable approach to include vala into ctags.

parsers/vala.c
```diff
@@ -0,0 +1,570 @@
...
+static void parseStatement (tokenInfo *const token)
+{
+	tokenInfo *lastToken = newValaToken ();
+	bool foundSignature = false;
+
+	do
+	{
+		tokenCopy (lastToken, token);
+		tokenRead (token);
+		if (tokenEqType (lastToken, 'class'))
```
k-takata: Is `'class'` a mistake of `'c'`?

masatake: You are correct.

albfan: Check please, I think it is fixed right now. Not sure if that is the supposed right approach to keep going.

albfan:
@masatake I have first question on this parsers:

How to detect a vala field (inside a class) and a property (field with `get` and `set`:

```
class Address {
   public string street;
...
}

class Person {
public Address address;
public string name {get; set;}
...
}
```

Understanding this seems the key to me. (maybe I should reread the links you suggest right now?)

masatake:
> How to detect a vala field (inside a class) and a property (field with get and set:

Please show the pair of "input.vala" and "exepected.tags".  The pair is very important for me to understand what you need. If you cannot image "expected.tags" for the input, I cannot help you. In that case, what you have to do is reading ctags(1) man page carefully.

Units/parser-vala.r/class.vala.b/expected.tags
```diff
@@ -0,0 +1,13 @@
+main	input.vala	/^void main(string[] args) {$/;"	method	line:8	language:Vala
+Address	input.vala	/^class Address {$/;"	class	line:15	language:Vala
+Person	input.vala	/^class Person {$/;"	class	line:23	language:Vala
+Address	input.vala	/^   public Address address {get; set;}$/;"	field	line:24	language:Vala
+_name	input.vala	16;"	field	line:16	language:Vala
+d_age	input.vala	17;"	field	line:17	language:Vala
+name	input.vala	16;"	property	line:16	language:Vala
```
albfan: @masatake this is a property.

class Person {
   public string name {get; set;}
}

at same time it is a field. In vala you can do something like

class Person {
   private int d_age;

   public int age {
      get { return d_age;}
      set { 
         if (value > 0) {
            d_age = value;
         } else {
            d_age = 0;
         }
      }
   }
}

`d_age` is a private field and `age` is the property.

Old implementation use `_<fieldname>` when there's no private field. Not sure if that really makes sense (pseudo tag or virtual tag?)

In general a field is:

<scope>* <type> <name>

if there's no scope it is private.

A property (can be more complex but is:

field {get; set;}

A property notifies its changes so you can add a listener, force to fill on constructor... things like that.

albfan:
I added a review comment in `expected.ctags` that will be my approach to reimplement this, and once I see how to understand the scope for a tag (things inside a pair) things that depends on several previous tokens I will figure out myself  the rest.

Thanks for help here!

Units/parser-vala.r/class.vala.b/expected.tags
masatake: No indentation breaks my small brain. Could you insert proper whitespaces?

masatake:
I promised @k-tanaka I will focus on releasing related work till we can find a long term volunteer who drives the documentation related work of this project.

So I will show only sample code. I hope it  helps you.
Ideally, I should consider the work of your side.

masatake:
See #621. I updated my branch.

albfan:
@masatake I think I get it. I rebase this on top of your branch (I rebased on top of master too.

`parseClassBody` is really helpful. Let's see if I can complete the rest on my own.

albfan:
Closing in favor of https://github.com/masatake/ctags/pull/6 where the development is really happening

PR #2322, opened 2019-11-14
units: Fix that units.py doesn't respect LANGUAGES

### LINKED ISSUE #2321: make units do not respect LANGUAGES
Seems make units is using a new script `misc/units.py` instead of `misc\units`.

- Python one miss the filter by LANGUAGE
- Bash one works correctly

I was able to launch bash one using:

    make units PYTHON= LANGUAGES=Vala

But is not obvious.

k-takata:
Close #2321.

It seems that I forgot to implement this feature. Implement it.


masatake:
Thank you!

PR #2326, opened 2019-11-15
Fix typo in units python script

albfan:
@k-takata I found this problem trying the new vala parser.

I guess is related with a buggy implementation as vala parser actually SIGSEV.


masatake:
@albfan, thank you!

k-takata:
Thank you for finding this.

PR #2327, opened 2019-11-16
Reject more non alnum chars in langdef

masatake:
All changes about --langdef option.

The main change is for rejecting characters like @, $, % as a parser name.

### DIFF
--- a/Tmain/broken-langdef.d/input.c
+++ b/Tmain/broken-langdef.d/input.c
@@ -1,0 +1,1 @@
+int f(void) { return 0; }
--- a/Tmain/broken-langdef.d/run.sh
+++ b/Tmain/broken-langdef.d/run.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# Copyright: 2019 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
+run_ctags()
+{
+	echo '#' "$1" 1>&2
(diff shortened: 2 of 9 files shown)


masatake:
optlib2c is not done yet.

k-takata:
This means that if someone create a language named C%, u-ctags doesn't accept it, right?
C++ and C# use symbols in their names. Why should other symbols be rejected?

masatake:
> This means that if someone create a language named C%, u-ctags doesn't accept it, right?

Right. It doesn't accept it with this change.

> C++ and C# use symbols in their names. Why should other symbols be rejected?

I would like to keep some graph characters as metacharacters for extending ctags command line.

Currently, [-:{=]  cannot be used because the command line parser is confused.
So there is no strong reason for disallowing using the graph character other than [-:{=+#].

Do you think this is to restrict?

masatake:
I wrote above comment to the commit log.

masatake:
https://en.wikipedia.org/wiki/List_of_programming_languages

masatake:
Comments are welcome. However, we can shrink the disallowing-character list even after releasing 6.0.0.
So I think we can start the more restricted one and can wait feedback.

k-takata:
I understand your intention.

masatake:
I updated ctags-incompatibilities.7.rst.in.

PR #2334, opened 2019-11-18
man(web): Add links to each man page on the Web site

k-takata:
Add links to ctags(1), ctags-incompatibilities(7), ctags-optlib(7) and
tags(5). This is only useful on ./docs/man/, not useful on ./man/, so convert them
when running `make update-docs`.


masatake:
Nice!

PR #2345, opened 2019-11-24
docs(web): remove some "IN MAN PAGE" markers

masatake:


masatake:
About parser guessing and --man-<LANG>, I think I have to write more.
I would like to provide the way to define a selector from CLI.
Currently, the rules are hard-coded in man/selectors.c. Maybe the option will be --selector-<LANG>=...
Without this new option, it is difficult to understand the intention of introducing --map-<LANG> option.

PR #2349, opened 2019-11-26
docs(man): Add links inside ctags(1) on Web pages

k-takata:
Add internal links in ctags(1) man page.

Section titles automatically become hyperlink targets.
Use ``` `phrase reference`_ ``` style to refer them.

---

~~Sphinx supports "ref" role to link to a tag, however rst2man (docutils) doesn't support it.
Define a stub "ref" role to avoid errors on docutils, and add links inside ctags.1.rst.
These links works fine when the man page is converted by sphinx.~~

~~Note that only ``` :ref:`title` ``` style can be used on docutils.
(``` :ref:`title <tag>` ``` style can be only used on sphinx.)~~

~~optlib.rst needed to be updated, because some tags conflicted with ctags.1.rst.~~


k-takata:
BTW, I find this comment in ctags.1.rst:
https://github.com/universal-ctags/ctags/blob/f17ca7731a235b6950b8f02133ec1ff2656bddc7/man/ctags.1.rst.in#L64-L68

I agree with this comment.
The contents of e-ctags man pages are:

* SYNOPSIS
* DESCRIPTION
* SOURCE FILES
* OPTIONS
* ...

However, the contents of u-ctags man pages are:

* SYNOPSIS
* DESCRIPTION
* SOURCE FILES
  * Determining file language
* TAG ENTRIES
  * Fields
  * Kinds
  * Extras
  * Roles
  * Language-specific fields and extras
* COMMAND LINE INTERFACE
  * Letters and names
  * List options
* OPTION ITEMS
* ...

There are too many contents before "OPTION ITEMS".
I think "Determining file language" and "TAG ENTRIES" might be able to move after "OPTION ITEMS".

k-takata:
Hmm, rst2man fails on Travis CI. (It doesn't fail on my environment.)
Maybe because of the version of rst2man?

masatake:
I agree with you about "Determining file language".
I wonder how "TAG ENTRIES" should be. I think items explained in the section are the heart of ctags.
I would like users to know about the concept.

k-takata:
Totally rewrote the commit.

masatake:
It seems that a reader can jump from a section of a man to another section of another (or the same) man page. Right?

k-takata:
`` `phrase reference`_ `` style is valid only in the same page. When this is used on rst2man, it is shown with underline. rst2html converts this to a hyperlink.
`` :ref:`title <tag>` `` style can jump across files, but this can be used only on sphinx and cause error on old docutils (rst2man).

masatake:
I understand the rule. Jumping from the contents for the web to the man pages is enough.
Basically, the contents of the man pages should be self-contained about the stable features.
Jumping for the opposite direction is not needed. Really nice improvement.

Can I ask you to add one more commit?

```
diff --git a/man/README b/man/README
index da2df787..bd85cac1 100644
--- a/man/README
+++ b/man/README
@@ -26,6 +26,18 @@ Rules for writing man pages
 
 	make QUICK=1 update-docs
 
+Tips for writing man pages
+
+* about hyperlinks
+
+  + `phrase reference`_ style is valid only in the same page.
+    When this is used on rst2man, it is shown with underline.
+    
+  + rst2html (and sphinx) converts this to a hyperlink.
+    :ref:`title <tag>` style can jump across files, but this can
+    be used only on sphinx and cause error on old docutils
+    (rst2man).
 
 Testing and verification
```

k-takata:
Done.

masatake:
@k-takata, I got a minor comment about your commit, "man: Add tips about hyperlinks to README".

We have a parser for "man" page. The commit prefix "man: " is for the parser.
So when we change files under man/*.rst*, we use "docs(man)".

k-takata:
I will do so from next time.

PR #2353, opened 2019-11-29
man: use --kinddef-<LANG> option in man optlib parser

masatake:


k-takata:
Was the title of this PR correct?
The prefix is `man:`, but this PR changed `optlib/passwd.ctags`. It doesn't seem to related to man.

masatake:
Ouch! You are correct. This is my mistake.

k-takata:
What was the intended title then?

masatake:
The title should be "passwd: use --kinddef-<LANG> option in passwd optlib parser ".

PR #2347, opened 2019-11-25
buildsys: Generate sed rules based on (IN_)SOURCE_FILES

k-takata:
Close #2346.

PR #2354, opened 2019-12-02
Revise the options about kinds, fields, and extras

masatake:


masatake:
"letters" and "names" are used so freqently in our man pages. So I would like to introduce more distinguishable words for them.

For letter, I introduced "one-letter" or "one-letter flag".
About "names", I would like to introduce "long-name" or "long-name flag".

masatake:
The sed command line synthesized  by make is the same as that in GNU/Linux.
So I guess t his is not about the implementation of make.

This may be about the implementation of sed.
https://riptutorial.com/ja/sed/topic/9436/bsd---macos-sed%E5%AF%BEgnu-sed%E5%AF%BEposix-sed%E4%BB%95%E6%A7%98 (in Japanese)

masatake:
When building ctags executable, we support using bsd make.
However, generating documents is done in before buidling.
So we can assume GNU make is there.

k-takata:
Oh, BSD sed doesn't support `\|`. It is critical...

k-takata:
Is it better to stop executing `make update-docs` on mac?

k-takata:
If we revert #2347, we can use BSD sed, but you may not like it. ;-)

I confirmed that the same results with macOS are produced when adding `--posix` option to GNU sed.

masatake:
> Is it better to stop executing make update-docs on mac?

My understanding is that person using Mac cannot develop u-ctags.
If my understanding is correct, I would like to make "make update-docs" workable even on Mac.

Currently, you make single complicated pattern in reppat.
I wonder we can convert it to multiple patterns like:
```
  sed -e pat1 -e pat2 -e pat3 ....
```
So we can avoid using '\|'.

masatake:
As you guessed, I like #2347. 
This is good time for me to study what you did.
I will try to fix this issue by myself.

k-takata:
BSD sed doesn't support `\<`. This is also a problem.

masatake:
```
commit 4fe409f9aba2e631e048e0b874b03eb34b4df1c7 (HEAD -> misc-doc-fix)
Date:   Mon Dec 2 17:22:08 2019 +0900

    docs(web), docs(man): don't use '\|' in a pattern for sed when making hyperlinks
    
    sed implementation in MacOS doesn't understand '\|'.
    To avoid using '\|', this change expands 'A\|B' to "-e A -e B".
    
diff --git a/man/Makefile b/man/Makefile
index 74ec29de..d191c626 100644
--- a/man/Makefile
+++ b/man/Makefile
@@ -64,14 +64,26 @@ man: $(man_pages)
 html: $(html_pages)
 pdf: $(pdf_pages)
 
-empty :=
-space := $(empty) $(empty)
-reppat := \<\(${subst ${space},\|,${subst .,(,${addsuffix ),${man_pages}}}}\)
+#
+# Convert the names of man page files to man page citations.
+#
+# ctags.1.rst tags.5.rst ...
+# => ctags(1) tags(5) ...
+#
+cites := ${subst .,(,${addsuffix ),${man_pages}}}
+
+#
+# Convert the man page citations to sed patterns for making hyperlinks.
+#
+# ctags(1) tags(5) ...
+# => -e 's/\<ctags(1)/:ref:`& <&>`/g' -e 's/\<ctags(5)/:ref:`& <&>`/g' ...
+#
+reppat := $(foreach m,$(cites),-e 's/\<$(m)/:ref:`& <&>`/g')
 
 update-docs: $(doc_files)
 
 ../docs/man/%.rst: %.rst
-   sed -e 's/$(reppat)/:ref:`& <&>`/g' < $< > $@
+   sed $(reppat) < $< > $@
 
 %.rst: %.rst.in
 ifeq ($(QUICK),)
```

masatake:
> BSD sed doesn't support \<. This is also a problem.

This is a critical issue. Is there anyway to avoid using it?

k-takata:
I used `\<` to avoid replacing `.. _ctags(1):` in top of the rst files. If it can be skipped, `\<` is not needed.

man/Makefile
```diff
@@ -64,14 +64,26 @@ man: $(man_pages)
 html: $(html_pages)
 pdf: $(pdf_pages)
 
-empty :=
-space := $(empty) $(empty)
-reppat := \<\(${subst ${space},\|,${subst .,(,${addsuffix ),${man_pages}}}}\)
+#
+# Convert the names of man page files to man page citations.
+#
+# ctags.1.rst tags.5.rst ...
```
k-takata: This line should be:
`# ctags.1 tags.5 ...`

masatake:
Though I did nothing for `\<`, the tests are passed.

k-takata:
Oh, does recent version of mac's sed support `\<`?

masatake:
> Oh, does recent version of mac's sed support \<?

Maybe yes.
About Mac, what we can do is only believing the result of travis.

man/Makefile
masatake: Thank you. Updated.

masatake:
4a962cf is needed other pull requests.
I will commit it separately.

masatake:
TODO

* the other man pages
* long-name

masatake:
Introducing the terms, "one-letter" and "long-name", to the other man pages is not done yet.
But I will merge this pull request. I will update the man pages later.

Issue #1837, opened 2018-08-19
Unicode file names aren't recognized


vToMy:
Running ctags on Unicode file names fails to open them.

**Example**
For a file called:
`こんにちは世界.txt`
Running:
`ctags --options=NONE *`
Will produce:
```
ctags: Notice: No options will be read from files or environment
ctags: Warning: cannot open input file "???????.txt" : No such file or directory
```

hadrielk:
Works for me:
```
$ ls
こんにちは世界.adoc

$ ../ctags --options=NONE -f - *
ctags: Notice: No options will be read from files or environment
Chapter 1 (Level 0)	こんにちは世界.adoc	/^= Chapter 1 (Level 0)$/;"	c
Level 3 Section 1.1.1.1 Title	こんにちは世界.adoc	/^==== Level 3 Section 1.1.1.1 Title$/;"	t	subsection:Chapter 1 (Level 0).Section 1.1.Subsection 1.1.1
Level 4 Section 1.1.1.1.1 Title	こんにちは世界.adoc	/^===== Level 4 Section 1.1.1.1.1 Title$/;"	T	subsubsection:Chapter 1 (Level 0).Section 1.1.Subsection 1.1.1.Level 3 Section 1.1.1.1 Title
Section 1.1	こんにちは世界.adoc	/^== Section 1.1$/;"	s	chapter:Chapter 1 (Level 0)
Subsection 1.1.1	こんにちは世界.adoc	/^=== Subsection 1.1.1$/;"	S	section:Chapter 1 (Level 0).Section 1.1
```
What operating system are you running on, and what version ctags?

vToMy:
OS: Windows 10 (Version 10.0.17134 Build 17134)
ctags version:
```
>ctags --version
Universal Ctags 0.0.0(2258b24b), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Aug 18 2018, 00:09:59
  URL: https://ctags.io/
  Optional compiled features: +win32, +wildcards, +regex, +internal-sort, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames
```

k-takata:
This is a long-standing issue on Windows.
We currently use ANSI APIs, but we need to use Unicode APIs to handle Unicode file names.
This is a hard work, though.

[...]

PR #2360, opened 2019-12-04
win32: Add version resource and manifest

k-takata:
This adds version resource and application manifest to the win32 executable file.
`win32/ctags.rc` and `win32/resource.h` are created by VS2013 and edited manually.

The application manifest should fix the issue #1837 on Windows 10 version 1903 or later.
See: https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page
This is a relatively new feature in Windows and it might not be matured enough, but at least I confirmed that ctags.exe can handle Unicode file names on Windows 10 1909.

NOTE: the version number needs to be updated manually.
We have version numbers in several files: `configure.ac`, `docs/conf.py`, `main/ctags.h`, `win32/ctags.exe.manifest` and `win32/ctags.rc`. (The last two files are added in this PR.)
We need to update all of them, when we release a new version.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,5 @@
 include $(top_srcdir)/source.mak
 
-# Test cases are added to EXTRA_DIST by makefiles under Tmain and Units
-# Documents are added to EXTRA_DIST by makefiles/docs.mak
 EXTRA_DIST   = README.md autogen.sh \
 	       makefiles/list-optlib2c-input.mak misc/optlib2c \
 	       .ctags.d/exclusion.ctags .dir-locals.el .editorconfig .gdbinit .gitignore \
@@ -18,6 +16,7 @@ EXTRA_DIST   = README.md autogen.sh \
 	       misc/validators/validator-puppet \
 	       $(WIN32_HEADS) $(WIN32_SRCS) mk_mingw.mak mk_mvc.mak \
 	       win32/mkstemp/COPYING.MinGW-w64-runtime.txt \
+	       win32/ctags.rc win32/ctags.exe.manifest win32/resource.h \
 	       win32/ctags_vs2013.sln \
 	       win32/ctags_vs2013.vcxproj win32/ctags_vs2013.vcxproj.filters \
 	       win32/gen-repoinfo.bat \
@@ -166,6 +165,16 @@ ctags_LDADD += $(LIBYAML_LIBS)
 ctags_LDADD += $(SECCOMP_LIBS)
 ctags_LDADD += $(ICONV_LIBS)
 dist_ctags_SOURCES = $(CMDLINE_HEADS) $(CMDLINE_SRCS)
+
+if HOST_MINGW
+WINDRES = windres
+RES_OBJ = win32/ctags.res.o
+ctags_LDADD += $(RES_OBJ)
+$(RES_OBJ): win32/ctags.rc win32/ctags.exe.manifest win32/resource.h
+	mkdir -p $(builddir)/win32
+	$(WINDRES) -o $@ -O coff $<
+MOSTLYCLEANFILES = $(RES_OBJ)
+endif
 
 mini_geany_CPPFLAGS = $(libctags_a_CPPFLAGS)
 mini_geany_CFLAGS = $(libctags_a_CFLAGS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,9 +170,12 @@ if HOST_MINGW
 WINDRES = windres
 RES_OBJ = win32/ctags.res.o
 ctags_LDADD += $(RES_OBJ)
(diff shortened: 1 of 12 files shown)


win32/ctags_vs2013.vcxproj
```diff
@@ -338,8 +344,12 @@
...
+    <ClInclude Include="resource.h" />
+  </ItemGroup>
+  <ItemGroup>
+    <ResourceCompile Include="ctags.rc" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
k-takata: Visual Studio removes the last CRLF from `*.vcxproj` and `*.vcxproj.filters`.
@masatake I think `src/src-check` needs to be adjusted for that.

k-takata: Done.

masatake:
I don't have enough knowledges to review this. Feel free to merge this when you think you are o.k.

k-takata:
Do you have any comments about the commit messages, or about usage of autotools and makefiles?

Makefile.am
```diff
@@ -167,6 +166,16 @@ ctags_LDADD += $(SECCOMP_LIBS)
...
 dist_ctags_SOURCES = $(CMDLINE_HEADS) $(CMDLINE_SRCS)
 
+if HOST_MINGW
+WINDRES = windres
+RES_OBJ = win32/ctags.res.o
+ctags_LDADD += $(RES_OBJ)
+$(RES_OBJ): win32/ctags.rc win32/ctags.exe.manifest win32/resource.h
+	mkdir -p $(builddir)/win32
+	$(WINDRES) -o $@ -O coff $<
+MOSTLYCLEANFILES = $(RES_OBJ)
```
masatake: I prefer to
```
MOSTLYCLEANFILES += $(RES_OBJ)
```

To do as I write, a line 
```
MOSTLYCLEANFILES =
```
may be needed somewhere at the head of Makefile.am.

masatake:
I have only one very minor suggestion.

masatake:
> NOTE: the version number needs to be updated manually.
We have version numbers in several files: configure.ac, docs/conf.py, main/ctags.h, win32/ctags.exe.manifest and win32/ctags.rc. (The last two files are added in this PR.)
We need to update all of them, when we release a new version.

I see. Instead of improving our build-sys for synchronizing them, verifying the consistency of them at on travis may be easy. I like the technique you introduced for converting  man to docs/man. Implementing verifier is much easier, but effective.

Makefile.am
k-takata: If so, maybe `CLEANFILES` needs to be changed in a similar way?

At the head of Makefile.am:

```
CLEANFILES =
```

and at the middle of the file:

```
CLEANFILES += $(REPOINFO_HEADS) $(PEG_SRCS) $(PEG_HEADS)
```

k-takata: I changed both `CLEANFILES` and `MOSTLYCLEANFILES`.

k-takata:
>  Instead of improving our build-sys for synchronizing them, verifying the consistency of them at on travis may be easy.

We need to verify the following points:

https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/configure.ac#L10
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/docs/conf.py#L55-L58
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/main/ctags.h#L20
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/win32/ctags.exe.manifest#L3
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/win32/ctags.rc#L61-L62
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/win32/ctags.rc#L79
https://github.com/universal-ctags/ctags/blob/105f514e33a328fb4169a16c453a2ba5aa742be3/win32/ctags.rc#L84

Note that Windows uses X.Y.Z.W (or X,Y,Z,W) style, not X.Y.Z. Maybe we should set to X.Y.Z.0.

Makefile.am
masatake: Yes. You have already done. Thank you.

win32/ctags_vs2013.vcxproj
masatake: Thank you.

Issue #2361, opened 2019-12-08
main: simplify the output of --version


albfan:
Testing ctags with other programs, some of them try to check version.

Based on http://ctags.sourceforge.net/ (last version 5.8)

I expected 

```
$ ctags --version
5.9.0
```

But I get something like
```
$ ctags --version
Universal Ctags 0.0.0(4012dfce), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Dec  8 2019, 17:47:51
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml, +packcc
```

`--version` seems like an `--info` option.

Related with #228

albfan:
Example of vim plugin using ctags and the complex check to find ctags version:

https://github.com/xolox/vim-easytags/blob/master/autoload/xolox/easytags.vim#L81

```
    let pattern = '\(Exuberant\|Universal\) Ctags \zs\(\d\+\(\.\d\+\)*\|Development\)'
```

I understand that exuberant ctags and universal ctags will match for a while, but as a dev tool we should ease this check.

masatake:
I agree but the format of output of --version is derrived from Exuberant Ctags.
Universal Ctags follows the format and extends it.

```
> $ ctags --version
5.9.0
```

How people can know this is u-ctags of e-ctags from this output?

masatake:
In addition, I would like to show clearly that u-ctags is derived from e-ctags though u-ctags is not official fork.

masatake:
However, we should handle https://github.com/universal-ctags/ctags/pull/2360#issuecomment-562537046 though this issue.

PR #2362, opened 2019-12-08
Update version

albfan:
Related with #2361

This reveals that versioning break test suite (which doesn't m takes sense).

We need a test suite able to avoid tag comments that breaks test suite or preprocess test fixtures to allow version on it (which seems overengineered)


masatake:
Merged into another pull request #2655.
Thank you.

PR #2366, opened 2019-12-10
win32: Use copyright mark on properties dialog box

k-takata:
Just a cosmetic change. Use copyright mark © instead of (C) in the properties dialog box.

Before this fix:
ctags-prop1

After this fix:
ctags-prop2


masatake:
The result of travis-ci is strange.

k-takata:
I rerun the failed travis job, and now it successes. Strange.

Issue #2364, opened 2019-12-09
universal ctags and c++ atributes


andrejlevkovitch:
The name of the parser: `universal ctags`

The command line you used to run ctags:

```
$ ctags main.cpp
```

The content of input file:

```c++
// main.cpp
#include <cstdlib>
#include <iostream>

void foo();

int main([[maybe_unused]]int argc, [[maybe_unused]]char *argv[]) {
  int alpha;
  int bravo;
  int charlie;

  return EXIT_SUCCESS;
}
```

The tags output you are not satisfied with:

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	//
alpha	main.cpp	/^  int alpha;$/;"	v	typeref:typename:int
bravo	main.cpp	/^  int bravo;$/;"	v	typeref:typename:int
charlie	main.cpp	/^  int charlie;$/;"	v	typeref:typename:int
foo	main.cpp	/^void foo() {$/;"	f	typeref:typename:void
```

The tags output you expect:

```
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
!_TAG_PROGRAM_VERSION	0.0.0	//
foo	main.cpp	/^void foo() {$/;"	f	typeref:typename:void
main	main.cpp	/^int main(int argc, char *argv[]) {$/;"	f	typeref:typename:int
```

The version of ctags:

```
$ ctags --version
[...]

PR #2372, opened 2019-12-13
CXX: Handle C++11 attributes. Fixes bug #2364.

pragmaware:
Don't let C++11 [[attributes]] fool the parser.


pragmaware:
Withdrawing for now. Will try to use a different approach as suggested by @masatake .

PR #2376, opened 2019-12-15
travis: sed command line for macos

masatake:
We tracked the strange result about the code verifying the rst files under docs/man are committed.
fc7e236 is for fixing the issue.

Without fc7e236,  in some cases, "make -C man QUICK=1 update-docs" does nothing.
As the result, the code using "git diff --exit-code docs/man" reports "o.k." though it is not o.k.
I guess it is related to low resolution of timestamp of the file system. (This never happens on GNU/Linux.)

This behaviour hides '\<' used in sed command line is not portable.

c6b9a4c defines an alternative regular expression that can be used in BSD sed.

TODO: We should drop the support for BSD make. We already use enough GNU make features. We cannot go back. We can utilize autotools for fixing part of issues. However, it is impossible to put our man pages to our web site with autotools. GNU make is needed to run "make -C man QUICK=1 update-docs".

### DIFF
--- a/misc/travis-check.sh
+++ b/misc/travis-check.sh
@@ -61,7 +61,7 @@ if [ "$TARGET" = "Unix" ]; then
         )
 
     else
-		make -C man QUICK=1 update-docs
+		make -B -C man QUICK=1 update-docs
 		if ! git diff --exit-code docs/man; then
 			echo "Files under docs/man/ are not up to date."
 			echo "Please execute 'make -C man QUICK=1 update-docs' and commit them."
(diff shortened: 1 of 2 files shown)


masatake:
I did it!

misc/travis-check.sh
```diff
@@ -61,6 +61,11 @@ if [ "$TARGET" = "Unix" ]; then
         )
 
     else
+		# For forcing make command to run update-docs target,
+		# call touch command and sleep command here.
+		touch man/*.rst
+		sleep 1
+		touch man/*.rst.i
```
k-takata: Isn't this a mistake of `touch man/*.rst.in`?

k-takata: Or, how about using `make -B`?

masatake: Thank you. -B looks better. I will update this pull request after some testing with -B option.

masatake: Updated. It works fine.

PR #2377, opened 2019-12-15
circleci: run test cases on Fedora31 instead of Fedora30

masatake:


masatake:
It seems that BSD make package in Fedora 31 is broken.

PR #2367, opened 2019-12-11
Extend tokeninfo API

masatake:
Define some new macros accessing fields of tokenInfo data type.
I found they were useful when prototyping a vala parser with tokenInfo in #621.

PR #2378, opened 2019-12-16
travis: force make command to run update-docs target of man/Makefile

masatake:
On a file system with low resolution of timestamp, the code verifying
rst files under docs/man are commited doesn't work well. On the such
file system, "make -C man QUICK=1 update-docs" does nothing in some
cases. I inspected this issue in #2367.

Suggested by @k-takata in #2376.

### DIFF
--- a/misc/travis-check.sh
+++ b/misc/travis-check.sh
@@ -61,7 +61,7 @@ if [ "$TARGET" = "Unix" ]; then
         )
 
     else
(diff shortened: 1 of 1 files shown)


masatake:
Reproduced well.

masatake:
Reproduced again.

masatake:
The chage is cherry-pick'ed by #2376.

PR #2384, opened 2019-12-19
Support XDG base directory specification

### LINKED ISSUE #89: Use XDG dirs spec for data/config
It would be great if the configuration and other data used by fishman-ctags used the [XDG base dir spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) instead of putting everything into the home directory. This would keep it in line with other more modern software and would prevent polluting the home directory.

itchyny:
This is a poc implementation for checking `$XDG_CONFIG_HOME` and loads the `$XDG_CONFIG_HOME/ctags/*.ctags` (fixes #89). But yes, I know the current implementation is not compliant to the specification (when $XDG_CONFIG_HOME is empty or unset, it should fallback to ~/.config/ctags). Note that we still have to use the .ctags extension (can't use `$XDG_CONFIG_HOME/ctags/config`) but is reasonable to split the config per languages.

**EDIT**: Now the implementation is compliant to the XDG base directory specification (I believe).


main/options.c
```diff
@@ -3519,13 +3520,11 @@ struct preloadPathElt {
...
-	char *full_path = NULL;
-
 	const char* const envval = getenv (envvar);
-	if (envval)
-		full_path = combinePathAndFile(envval, path);
+	if (!envval)
+		return NULL;
 
-	return full_path;
+	return combinePathAndFile(envval, path);
```
itchyny: I believe we should not fallback to `/ctags` when the environment variable is not set.

masatake: Good catch!

I would like to introduce a fix for this important suggestion first.
I made a pull request for the fix. See  #2385.

masatake:
>  not compliant to the specification

So how we should do?

A. this change is not compliant to the specification, therefore this pull request is just for discussion, not for merging.
B. this change is not compliant to the specification, however, we should merge this because this change is useful.

For A, I have no strong comment. Simply say, I don't want to take much time for this topic.
For B, we need much more documents for the usefulness. The documents explain introducing this change is much more valuable than the incompliant. I think in the future we will get a bug report "ctags configuration loader isn't compliant to XDG standard". I think I cannot respond to the bug report.
The document may help me.

I want a test case. I can give you hints for writing. Any way I would like to hear the basic direction, A or B.

itchyny:
I added ~/.config/ctags fallback to follow the specification (I was just unconfident that I can write slash in .path).

Tmain/load-conf-files-under-default-xdg-config-home/run.sh
```diff
@@ -0,0 +1,12 @@
+# Copyright: 2017 Masatake YAMATO
```
masatake: I don't write this script.

masatake:
Thank you for addingthe test case.

> Note that we still have to use the .ctags extension (can't use $XDG_CONFIG_HOME/ctags/config) but is reasonable to split the config per languages.

Does this violate the specification? or Is it acceptable in the specification?

I would like you to add minimum code for satisfying the specification.
.ctags loader is already complicated.

itchyny:
> Does this violate the specification?

No. It's just a note for users (including me) who is not used to the behavior of universal-ctags. I believe the current implementation is compliant to the spec.
I'd like to hear some opinions from those who might be interested in XDG config support. @majutsushi @fishman @blueyed @justinmk @cweagans @dsifford (who commented in #89).

Tmain/load-conf-files-under-default-xdg-config-home/run.sh
```diff
@@ -1,12 +1,10 @@
-# Copyright: 2017 Masatake YAMATO
```
masatake: Copyright notice is neeed, but is not mine.

Tmain/load-conf-files-under-xdg-config-home/run.sh
```diff
@@ -1,11 +1,9 @@
-# Copyright: 2017 Masatake YAMATO
```
masatake: Copyright notice is neeed, but is not mine.

masatake:
Thank you for updating.

majutsushi:
Looks reasonable to me.

masatake:
@majutsushi, thank you for veryfing.

This pull request is mostly ready for merging.
However, I would like to keep the history of our repository simple; I don't want to introduce  try-and-error commits like 569c087 and 78b7a9b. 

So I rearranged the commits in this pull request and made a new pull request. @itchyny, could you look into #2386?

masatake:
Merged via #2386. Thank you.

PR #2385, opened 2019-12-23
main: Handle empty HOME, HOMEDRIVE, and HOMEPATH envvar

masatake:
See https://github.com/universal-ctags/ctags/pull/2384#discussion_r359695228 .

No test case. I cannot find a way to write a test case for Tmain easily.


masatake:
@k-takata, could you look into 
610d1fe  ?

masatake:
I unified this pull request and #2384 into #2386.

PR #2388, opened 2020-01-04
units.py: remove -diff.txt files in tmain when a case runs successfully 

masatake:


masatake:
@k-takata, could you take a look at 2b622da?

masatake:
Thank you.

PR #2392, opened 2020-01-06
Protobuf: fill signature and typeref fields for rpc kind

masatake:


masatake:
This parser doesn't handle scope well. 
Rewriting in peg may be better.

peg maybe fit well for IDLs.

PR #2395, opened 2020-01-09
CXX: Handle non-pointer function typedefs

### LINKED ISSUE #2394: C: function pointer type defined without using '(*)'
I found follwoing code when I read netfilter.h of linux kernel:
```
typedef unsigned int nf_hookfn(void *priv,
			       struct sk_buff *skb,
			       const struct nf_hook_state *state);
```

Our C parser cannot capture nf_hookfn.

An input for test case:
```
$ cat  foo.c
typedef int fnt (int);
typedef int (* fntp) (int);

void assign (fntp *fdst, fnt fsrc)
{
  *fdst = fsrc;
}

$ u-ctags --quiet --options=NONE -o -  foo.c
assign	foo.c	/^void assign (fntp *fdst, fnt fsrc)$/;"	f	typeref:typename:void
fntp	foo.c	/^typedef int (* fntp) (int);$/;"	t	typeref:typename:int (*)(int)	file:
```

fnt is not tagged.

pragmaware:
Fixes #2394 .
typeref is currently missing. Not sure if it's worth to add it as it would require some bits of complexity.

### DIFF
--- a/Units/parser-cxx.r/typedefs.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/typedefs.cpp.d/expected.tags
@@ -15,6 +15,8 @@ T305	input.cpp	/^typedef struct AStruct (T305)(int);$/;"	t	typeref:struct:AStruc
 T306	input.cpp	/^typedef union AUnion & (T306)(int);$/;"	t	typeref:union:AUnion & ()(int)	file:
 T307	input.cpp	/^typedef AClass (T307)(int);$/;"	t	typeref:typename:AClass ()(int)	file:
 T308	input.cpp	/^typedef enum AEnum (T308)(int);$/;"	t	typeref:enum:AEnum ()(int)	file:
+T309	input.cpp	/^typedef int T309(int);$/;"	t	file:
+T310	input.cpp	/^typedef int T310(AUnion *);$/;"	t	file:
 T401	input.cpp	/^typedef int T401 [ 10];$/;"	t	typeref:typename:int[10]	file:
 T501	input.cpp	/^typedef ATemplate1<int > T501;$/;"	t	typeref:typename:ATemplate1<int>	file:
 T502	input.cpp	/^typedef ATemplate1< unsigned short int> T502;$/;"	t	typeref:typename:ATemplate1<unsigned short int>	file:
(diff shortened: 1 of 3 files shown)


masatake:
I find a question.

```
[yamato@slave]/tmp% cat /tmp/foo.h
cat /tmp/foo.h
typedef struct file * (* fn_t) (char *);
[yamato@slave]/tmp% u-ctags -o - /tmp/foo.h
u-ctags -o - /tmp/foo.h
fn_t	/tmp/foo.h	/^typedef struct file * (* fn_t) (char *);$/;"	t	language:C++	typeref:struct:file * (*)(char *)
```

I think following tags output is better:
```
[yamato@slave]/tmp% u-ctags -o - /tmp/foo.h
fn_t	/tmp/foo.h	/^typedef struct file * (* fn_t) (char *);$/;"	t	language:C++	typeref:typename:struct file * (*)(char *)
```
or
```
fn_t	/tmp/foo.h	/^typedef struct file * (* fn_t) (char *);$/;"	t	language:C++	typeref:struct:file * signature:(char *)
```

How do you think?
I guess you followed the behaviour of OldC parser.

pragmaware:
Yep, the typeref:<X>: thing is ugly. 
Currently the parser follows a somewhat "compatible" route that uses the keywoard struct/union/enum as the second field if it is found in the type name. If none of these are present then "typename" is used.

I suggest: why don't we add a new "type" field and keep "typeref" as compatibility only? 
Then it would just be:

```
fn_t	/tmp/foo.h	/^typedef struct file * (* fn_t) (char *);$/;"	t	language:C++	type:struct file * (*)(char *)
```

masatake:
> I suggest: why don't we add a new "type" field and keep "typeref" as compatibility only?
Then it would just be:

My understanding is that the semantic of typeref field is not defined clearly. So we can define as we want as far as we don't break the syntax typeref:something:value.
As I wrote ago another issue, I would like to introduce following definition.
1. If value is tagged somewhere, the field may have following form: typeref:kindForTheValue:Value.
2. If value is not tagged, the field may have following form: typeref:typename:value or
    typeref::value. Regarding compatibility, using "typename" in the middle is better. However,
    in that case, we have to reserve the word "typename"; any parsers cannot define "typename" as
    its kind name. But I don't want to introduce more reserved words if possible.
    In such point of view, "typeref::value" is better. However, it can break compatibilities.

I would like to choose 2. This rule(?) can be applied to scope: field.

type: you proposed can be 1 or 2. So I think we don't have introduce new field.

pragmaware:
Currently the second field of typeref contains some informations that someone might be using (?).... I don't know. From a compatibility point of view touching it might be troubling. However from my "personal" point of view it's useless and I'm OK with killing it. typeref::value is OK.

masatake:
I read cxxTagCheckAndSetTypeField() and I understand the limitation of typeref.

I expected the rule 1 is applicable to most of all cases. However, it is completely misunderstanding.
```
typedef struct *foo foop;
```

In this case, the rule 1 cannot be applicable because `*` is used.
So the field becomes like `typeref::struct *foo`.

masatake:
Good news is that `T` letter for fields is not used yet.

PR #2406, opened 2020-01-27
Markdown: fill scope fields

masatake:
TODO:

* regex table optimization
* performance evaluation with codebase


masatake:
Unlike RST, scope fields have full querified tags.
People may say they shouldn't have.

masatake:
The new parser cannot capture B in:
```
# A
### B
```
though the original parser can.

masatake:
...I have to add more tables.

masatake:
I added many tables.

PR #2411, opened 2020-02-04
main: fix field name combination

### LINKED ISSUE #2410: `--fields=z` doesn't work but `--fields=+z` works
Just found a strange bug.

If I run ctags with `ctags -f tmp.tag --fields=z code/code.c`, there is no `kind:value` in the output. 

But if I run it with `ctags -f tmp.tag --fields=+z code/code.c`, it works correctly.

Seems only `z` has this problem, when I use `--fields=nSe` it also works correctly.

version info:
```
Universal Ctags 0.0.0(81fa5b1), Copyright (C) 2015 Universal Ctags Team
  Compiled: Feb  4 2020, 22:42:40
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames, +packcc
```

masatake:
Close #2410.


masatake:
`--fields=z` prints nothing. This was expected behavior. However, it was not user-friendly.
With this pull request, ctags warns specifying only the z field. In addition, ctags turns on K field with more warning.

masatake:
A similar issue can be found in the behavior of Z field.
The combination of Z, s, and p fields are hard to understand.

I will stack the fix for Z field on this pull request.

masatake:
ctags warns the combination of fields only in ctags output mode.
It should not in xref and json output mode.

PR #2425, opened 2020-02-14
Lisp: introduce new kinds: variable, macro, and consant

### LINKED ISSUE #2423: Lisp: ctags seems to think all symbols are functions
The name of the parser: lisp (I guess)

```
$ ctags --options=NONE --fields=K -f- ./example.el
```

The content of input file:

```emacs-lisp
(defun some-func () nil)
(defmacro some-macro () nil)
(defvar some-var nil)
(defconst some-const nil)
```

```
some-const	./example.el	/^(defconst some-const nil)$/;"	function
some-macro	./example.el	/^(defmacro some-macro () nil)$/;"	function
some-var	./example.el	/^(defvar some-var nil)$/;"	function
```
(I tried `$ ctags --list-kinds=lisp`, and there is only a "function" kind. So this is understandable.)

The tags output you expect:

```
some-const	./example.el	/^(defconst some-const nil)$/;"	const
some-macro	./example.el	/^(defmacro some-macro () nil)$/;"	macro
some-var	./example.el	/^(defvar some-var nil)$/;"	variable
```

(Actually I don't need it to be accurate like this. I only need to distinguish callable and non-callable symbols.)

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(a3c87ab5), Copyright (C) 2015 Universal Ctags Team
  Compiled: Jan  7 2020, 22:54:55
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml, +packcc
```

How do you get ctags binary: From the official repo of Manjaro Linux.

masatake:
Close #2423.


masatake:
TODO

* adding a test case for QUOTE, and
* adding a test case for namespace (package?) of common list.

AmaiKinono:
This is amazing. Seems most things in elisp are covered. Here are some more that may be reasonable to add:

- `define-global-minor-mode`: This is an alias of `define-globalized-minor-mode`.
- `define-inline`: Inline function.

and things offerd by `cl-lib`:

- `cl-defun`: Function. An alias of it is `defun*`.
- `cl-defgeneric` and `cl-defmethod`: Define generic function and new method for a generic function.
- `cl-defsubst`: Inline function. An alias is `defsubst*`.
- `cl-defmacro`: Macro. An alias is `defmacro*`.
- `cl-define-compiler-macro`: Macro.

You can eval this in Emacs to get a list of callable symbols start with `def` or `cl-def`, with their documentation:

``` emacs-lisp
(let ((collection nil))
  (dolist (sym
           (append
            (all-completions "def" obarray)
            (all-completions "cl-def" obarray)))
    (when (fboundp (intern sym))
      (push sym collection)))
  (dolist (elt collection)
    (insert "\n")
    (insert (format "%-35s" elt))
    (when (documentation (intern elt))
      (insert (car (split-string
                    (documentation (intern elt)) "\n"))))))
```

masatake:
Thank you.

In this cycle, I will not work on cl- prefix. The original lisp parser heavily replies token starting from "def".

I will add following items to this request as you suggested:
* define-global-minor-mode
* define-inline
* defun*
* defsubst*
* defmacro*

The rest of them will be supported in planed Sexp meta parser.

masatake:
Private memo:

```
--langdef=EmacsLisp{base=Sexp}
--map-EmacsLisp=.el
--kinddef-EmacsLisp=f,function,functions
--kinddef-EmacsLisp=a,advise,advises
--fielddef-EmacsLisp=function,fuction the advice attaches to
# (defun foo (a &rest r)
--Sexp-EmacsLisp=defun/car/f/{iftype=symbol}{field=signature:cadr}
# (defun* bar (arg) ...
--Sexp-EmacsLisp=defun*/car/f{iftype=symbol}{field=signature:cadr}
# (defadvice foo (class name ...)
--Sexp-EmacsLisp=defadvise/cadadr/a/{iftype=symbol}{field=function:car}

--langdef-Scheme{base=Sexp}
--map-Scheme=.scm
--kinddef-Scheme=v,variable,variables
--kinddef-Scheme=p,proc,procedure
# (define (foo args) ...)
--Sexp-Scheme=define/caar/p/{iftype=symbol}{field=signature:cdar}
# (define bar val)
--Sexp-Scheme=define/car/v/{iftype=symbol}
```

masatake:
Added a test case  including quote.
The changes are for emacs-lisp. So I don't touch test cases for Common Lisp here.

PR #2441, opened 2020-02-27
Use SIZE_T_FMT_CHAR="z" in mk_mingw.mak

mattn:
```
gcc -c -O4 -Os -fexpensive-optimizations -Wall -std=gnu99 -DUSE_SYSTEM_STRNLEN -DSIZE_T_FMT_CHAR='""' -o misc/packcc/packcc.o misc/packcc/packcc.c

misc/packcc/packcc.c: In function 'generate_matching_string_code':
misc/packcc/packcc.c:2117:34: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
 2117 |             fprintf(gen->stream, "pcc_refill_buffer(ctx, %" SIZE_T_FMT_CHAR "u) < %" SIZE_T_FMT_CHAR "u ||\n", n, n);
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~                                                    ~
      |                                                                                                                |
      |                                                                                                                size_t {aka long long unsigned int}
misc/packcc/packcc.c:2117:34: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
misc/packcc/packcc.c:2117:34: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
misc/packcc/packcc.c:2117:34: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
misc/packcc/packcc.c:2120:38: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
 2120 |                 fprintf(gen->stream, "((const char *)(ctx->buffer.buf + ctx->pos))[%" SIZE_T_FMT_CHAR "u] != '%s' ||\n", i, escape_character(value[i], &s));
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                    ~
      |                                                                                                                          |
      |                                                                                                                          size_t {aka long long unsigned int}
misc/packcc/packcc.c:2120:38: warning: format '%u' expects argument of type 'unsigned int', but argument 3 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
[...]

PR #2444, opened 2020-02-27
packcc: Use PRIuPTR instead of SIZE_T_FMT_CHAR

k-takata:
Closes #2441.

Use a standard definition instead of our own definition.


k-takata:
I need to remove [the latest commit](https://github.com/universal-ctags/ctags/pull/2444/commits/4bdf5cca4f969c8887a75399bec81ab2c12094e6) and update packcc by using `misc/pull-packcc.sh` after https://github.com/universal-ctags/packcc/pull/2 is merged.

masatake:
Thank you.

PR #2446, opened 2020-02-29
CXX: More work on templates

pragmaware:
- Don't break template arguments when template specializations are found
- Add support for specialization field
- Slightly refined template angle bracket matching
- Gracefully handle some non-prefix template keyword instances
- Emit types of template parameters
- Add some unit tests

### DIFF
--- a/Units/parser-cxx.r/template-nested-triangle-brackets.d/expected.tags
+++ b/Units/parser-cxx.r/template-nested-triangle-brackets.d/expected.tags
@@ -23,18 +23,40 @@ funcC	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	f	
 T	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
 I	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
 J	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
-funcD	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	f	typeref:typename:>2> K	template:<typename T,int I,int J=I>
+funcD	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	f	typeref:typename:K	template:<typename T,int I,int J=I>>2>
 T	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	Z	function:funcD
 I	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	Z	function:funcD
 J	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	Z	function:funcD
-funcE	input-2.hpp	/^template<typename T,int I,int J = 1 >> V> K funcE(T t)$/;"	f	typeref:typename:> V> K	template:<typename T,int I,int J=1>
-T	input-2.hpp	/^template<typename T,int I,int J = 1 >> V> K funcE(T t)$/;"	Z	function:funcE
-I	input-2.hpp	/^template<typename T,int I,int J = 1 >> V> K funcE(T t)$/;"	Z	function:funcE
-J	input-2.hpp	/^template<typename T,int I,int J = 1 >> V> K funcE(T t)$/;"	Z	function:funcE
-funcF	input-2.hpp	/^template<typename T,int I,int J = I >> V> K funcF(T t)$/;"	f	typeref:typename:> V> K	template:<typename T,int I,int J=I>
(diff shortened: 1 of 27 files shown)


Units/parser-cxx.r/template-nested-triangle-brackets.d/expected.tags
```diff
@@ -11,14 +11,27 @@ failure	input-1.hpp	/^class failure {};$/;"	c	template:<class T,class Hash=std::
...
+K	input-2.hpp	/^typedef int K;$/;"	t	typeref:typename:int
+funcB	input-2.hpp	/^template<typename T,int I,int J = I < V> void funcB(T t)$/;"	f	typeref:typename:void	template:<typename T,int I,int J=I<V>
+T	input-2.hpp	/^template<typename T,int I,int J = I < V> void funcB(T t)$/;"	Z	function:funcB
+I	input-2.hpp	/^template<typename T,int I,int J = I < V> void funcB(T t)$/;"	Z	function:funcB
+J	input-2.hpp	/^template<typename T,int I,int J = I < V> void funcB(T t)$/;"	Z	function:funcB
+funcC	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	f	typeref:typename:K	template:<typename T,int I,int J=I<V>
+T	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
+I	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
+J	input-2.hpp	/^template<typename T,int I,int J = I < V> K funcC(T t)$/;"	Z	function:funcC
+funcD	input-2.hpp	/^template<typename T,int I,int J = I >> 2> K funcD(T t)$/;"	f	typeref:typename:>2> K	template:<typename T,int I,int J=I>
```
masatake: Is the value for typeref field is o.k.?

pragmaware: You're right, it's not.
I'm still working on it.

masatake:
I found "meta" in the changes. Though it is not a kind, it seems that it fits well with other parts (typename, class) of C++ parser.

For the future extension of ctags, I would like to understand "meta" well.
I wonder it is C++ parser specific or not. As far as my experience in improving other parsers, "typename" you introduced is needed other parsers especially it fits well in dynamically typed languages.

Just a question, if we introduce "type:" field as you proposed, how it is used to represent  ... e.g. "typeref:meta:class" ?

pragmaware:
> I found "meta" in the changes.

Well, the templates can accept both data arguments and type arguments. For the data arguments a middle field of ``typename/struct/enum`` was OK. For the type arguments a middle field of ``typename`` was not fitting well because then we ended having ``typeref:typename:typename`` or ``typeref:typename:class``, which was kind of ugly. In fact the type here is the "type of a type".... so a "metatype". That's the reason for meta.

But it's just a placeholder. I'm OK with changing it, removing it or keeping typename.

By the way, C++20 is going to introduce concept related syntax here. One will be able to write

``template<ConceptName Name>...``

Now with typeRef we would like to write "meta" in the middle field also in this case but I'm afraid that we will not be able to figure out concepts from plain type names. So later even our "meta" may become inconsistent...

> Just a question, if we introduce "type:" field as you proposed, how it is used to represent ... e.g. "typeref:meta:class" ?

I'm all-in for the new "type:" field without the middle field at all.

masatake:
> I'm all-in for the new "type:" field without the middle field at all.

Do you mean "type:meta class" is the same as "typeref:meta:class" ?

pragmaware:
> Do you mean "type:meta class" is the same as "typeref:meta:class" ?

Just ``type:class``

pragmaware:
> Do you mean "type:class" is the same as "typeref:meta:class"?

Yep.

"meta" is there only because we have to fill that field with something (meaningful). But IMHO it's superfluous. For example, my editor [discards the middle field](https://github.com/pragmaware/code3/blob/a787c53c44cfef685b5d55fd796e8c36446012d6/code3/src/lib/symbol/tools/C3SymbolCTagsParser.cpp#L873) and happily builds a complete symbol tree without it.

(by the way, I've finally found some time to publish [code3](https://github.com/pragmaware/code3/tree/master/code3tags). it includes a slightly stripped-down version of universal-ctags).

masatake:
LGTM.

Thank you for the explanation about "meta". I understand it and I refect it to the future design of u-ctags.

I looked into code3 and I found code3 uses ctags as a command unlike geany users it as a library.
To prevent me from introducing unwanted change to u-ctags command-line interface, you can add "code3" directory under Tmain directory. You can add test cases to the new directory for verifying what code3 expects from the ctags command line interface.

pragmaware:
I have used it as an external command since it was the only option at the time. I should probably look into using it as a library too... 

I'll eventually think about the test cases, though unless the changes are really dramatic it shouldn't be a big problem adjusting code3 to match. I have cloned the source also for this reason: no need to track external versions and I can fix the integration once in a while.

PR #2449, opened 2020-03-01
main: extend hashtable api

masatake:
These are spin-off code developed during studying symbol tables before introducing rbtree.

These are not used anywhere, however, they may be useful in the future for u-ctags development.

* a callback function passed to hashTableForeachItem can break the interation.
* a new function hashTableForeachItemOnChain is for handling entries having the same key.

### DIFF
--- a/main/htable.h
+++ b/main/htable.h
@@ -18,7 +18,10 @@ typedef struct sHashTable hashTable;
 typedef unsigned int (* hashTableHashFunc)  (const void * const key);
 typedef bool      (* hashTableEqualFunc) (const void* a, const void* b);
 typedef void         (* hashTableFreeFunc)  (void * ptr);
-typedef void         (* hashTableForeachFunc) (void *key, void *value, void* user_data);
+
+/* To continue interation, return true.
+ * To break interation, return false. */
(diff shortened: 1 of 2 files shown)


main/htable.h
```diff
@@ -44,7 +47,11 @@ extern void       hashTablePutItem     (hashTable *htable, void *key, void *valu
 extern void*      hashTableGetItem     (hashTable *htable, const void * key);
 extern bool    hashTableHasItem     (hashTable * htable, const void * key);
 extern bool    hashTableDeleteItem  (hashTable *htable, void *key);
-extern void       hashTableForeachItem (hashTable *htable, hashTableForeachFunc proc, void *user_data);
+
+/* return true if proc never returns false.
+ * return false if proc returns false in all the calls. */
```
masatake: More comments are needed.

main/htable.h
```diff
@@ -14,6 +14,19 @@
 #include "general.h"
 #include <stdint.h>
 
+/* This hashtable allows adding multiple items for a same key.
```
masatake: s/a same key/the same key/g

PR #2427, opened 2020-02-15
CPreProcessor: macro expansion in a file

masatake:
Close #2413.

```
[jet@localhost]~/var/ctags% cat /tmp/foo.sv
typedef bit[31:0] int32_t;
module mod(
  input bit clk,
  input int32_t a
);
endmodule
[jet@localhost]~/var/ctags% ./ctags --extras=+q  -o - /tmp/input.sv 
a	/tmp/input.sv	/^  input int32_t a$/;"	p	module:mod
clk	/tmp/input.sv	/^  input bit clk,$/;"	p	module:mod
int32_t	/tmp/input.sv	/^typedef bit[31:0] int32_t;$/;"	T
mod	/tmp/input.sv	/^module mod($/;"	m
mod.a	/tmp/input.sv	/^  input int32_t a$/;"	p	module:mod
mod.clk	/tmp/input.sv	/^  input bit clk,$/;"	p	module:mod
```

----

hashtable related changes are proposed separately in #2449.
The rbtree based symbol table is proposed separately in #2450.

----

This pull request still can be converted to smaller pull requests.

1. a pull request dealing with etags output, temporary file, and file position rewinding.
2. pull requests that make parsers use the symbol tables.

The original SystemVerilog issue will be partially fixed (mitigated) in 2.

Issue #2421, opened 2020-02-12
TeX: [enhancement] Patch for LaTeX for beamer presentations


jagjordi:
Implement a similar patch like in this post https://sourceforge.net/p/ctags/patches/76/ so that LaTeX beamer presentations are sorted by slides instead of sections

masatake:
Thank you for making this issue.


Do you have an example input?
I would like to include it as a test case into our repository.

https://github.com/strace/strace-talks/blob/master/2019-ossdevconf-kaluga-linux_design_flaw_story/ossdevconf_2019-slides-linux_design_flaw_story.tex 

is a good example but it doesn't use `\frametitle`.
It uses `\begin{frame}{AN ITEM WE WOULD LIKE TO TAG}`.

I have already implemented functions for captruing `\frametitle`. 
However, `\begin{frame}{...}` is not yet.

jagjordi:
I think this one can be used, I changed the titles to use `\frametitle` keyword

masatake:
Still struggling.

jagjordi:
Is there any way I can help with this?

masatake:
Could you try #2424?

jagjordi:
Hello @masatake 
I have a suggestion to modify the behaviour of the parser in a beamer presentation.
If a arrangement of slides as this is written
```
Frame 1
|-- Subframe 1.1
|-- Subframe 1.2
|-- Subframe 1.3

Frame 2
|-- Subframe 2.1
|-- Subframe 2.2
```
Normally I would have the same title repeated as many times as "sub-slides" exist. For example

```
\frametitle{Frame1}
\framesubtitle{Subframe1.1}

\frametitle{Frame1}
\framesubtitle{Subframe1.2}

\frametitle{Frame1}
\framesubtitle{Subframe1.3}

\frametitle{Frame2}
\framesubtitle{Subframe2.1}

\frametitle{Frame2}
\framesubtitle{Subframe2.2}
```
With the current behaviour of the parser it will generate a tag entry for every time the command `frametitle` is used. I would argue that generally the behaviour sould be that it only generates a tag entry the first tame that the command is used, and subsequent uses of the command (with the same title name) will "point" to the first usage.

Let me know what you think?

Regards,

masatake:
I will work on this item after merging https://github.com/universal-ctags/ctags/pull/2450.
It will take a longer time than you assume.

In #2450, I will introduce a symbol table.
With that we can write the following code:

```c
[...]

PR #2450, opened 2020-03-01
main: rbtree based symbol table

masatake:
This is the result of a study in #2427.
I removed code related to CPreProcessor and SystemVerilog.

The changes for entry.h in 81647c1 excite you, a parser developer.

06a1588 may also be interesting.

TODO:
- [ ] write abot the new API to docs/*.rst.
- [ ] unify the type representing cork indexes to "int", not "unsigned int".

Test cases are not included in this change.


masatake:
#2421 is a good target to apply the new symbol table feature to.
Applying it to CPreProcessor's macro is more ... fantastic but more changes are needed.

masatake:
Though there are still two TODO items. However, I would like to use the symbol table for fixing #2421.
So I will merge this pull request in soon.

Issue #2454, opened 2020-03-06
Cxx: any way to get the end line of a variable


Lennon925:
I use following command to parse the test.cpp and result is as following.
`universal-ctags-x64.exe --output-format=json --langmap=c:+.inc.def --kinds-c=+plz --fields=+ne -o - D:\projects\demo\test.cpp`

test.cpp
```
const ProcItemStru g_strc[5] = {
    {JOB1, JOB1, 0, 0, JOB1}, {JOB1, JOB1, 0, 0, JOB1},
    {JOB1, JOB1, 0, 0, JOB1}, {JOB1, JOB1, 0, 0, JOB1},
    {JOB1, JOB1, 0, 0, JOB1},
};
```
`{"_type": "tag", "name": "g_mcmL2InnerMsgPorc", "path": "D:\\projects\\demo\\test.cpp", "pattern": "/^const McmL2InnerMsgProcItemStru g_mcmL2InnerMsgPorc[MCM_JOBID_BOTTOM] = {$/", "line": 1, "typeref": "typename:const McmL2InnerMsgProcItemStru[]", "kind": "variable"}`

The result contains only the start line of variable "g_strc". Is there some way to get the end line?

masatake:
@pragmaware, could you look at this one?

I also want to have end: fields for variables.

As usual I tried to solve for stduying your parser.  (See the end of this comment).
```
[yamato@slave]~/var/ctags-github% cat -n /home/yamato/var/ctags-github/input.c
cat -n /home/yamato/var/ctags-github/input.c
     1	int n [2] = {
     2		1,
     3		2,
     4	}, m [3] = {
     5		3,
     6		4,
     7		5,
     8	};
[yamato@slave]~/var/ctags-github% git diff
git diff
diff --git a/parsers/cxx/cxx_parser_variable.c b/parsers/cxx/cxx_parser_variable.c
index c76777be..e86bc580 100644
--- a/parsers/cxx/cxx_parser_variable.c
+++ b/parsers/cxx/cxx_parser_variable.c
@@ -102,6 +102,8 @@ CXXToken * cxxParserFindFirstPossiblyNestedAndQualifiedIdentifier(
 //
 bool cxxParserExtractVariableDeclarations(CXXTokenChain * pChain,unsigned int uFlags)
 {
+	int iCorkIndex = CORK_NIL;
+
 	CXX_DEBUG_ENTER();
 
 	if(pChain->iCount < 1)
[...]

PR #2456, opened 2020-03-06
Cxx: fill end fields of variable kind tags

masatake:
Close #2425.

TODO: This change doesn't handle anonymous functions like

```
auto lg5 = [](X<int,int> x,X<long,long>) -> X<double,double> { return X<double,double>(); };
```

well.


pragmaware:
The lambda problem above is really a template-related problem. This is going to fail too:

``auto xx = A<X>();``

The algorithm trying to skip to the next variable bails out when an uncondensed < is found as templates may become too complicated to be analyzed reliably at this level.

pragmaware:
I think that the template problem is a separate and broader issue we can eventually try to fix later.
This changeset looks OK.

I have a couple of thoughts...
- here we're marking the end of the variable initializer, not the variable declaration itself
  it's probably OK... just thinking.
- I'm wondering if it makes sense omitting the end line number if it is equal to the start line number. 
  For variable declarations it would save some space in the output. Again.. just food for later thoughts.

masatake:
> I'm wondering if it makes sense omitting the end line number if it is equal to the start line number.
For variable declarations it would save some space in the output. Again.. just food for later thoughts.

I also thought about it.
This compression can be done for other kinds including function.
These things should be done in a lower layer.
I will push one more change implementing the compression feature to this pull request.

The biggest issue is compatibility. I guess advanced users may use end: feature in their tools already.

masatake:
BTW, whether ``--fields=+{end}`` was passed or not can be known from pseudo tags.
See #2459.

masatake:
...I change my mind. I will implement the compression(?) technique proposed by @pragmaware in another pull request. It may break compatibility and can be implemented indepedently from this pull request.

masatake:
The change for the technique is small:
```
% git diff
diff --git a/main/writer-ctags.c b/main/writer-ctags.c
index 66ef7294..379808c7 100644
--- a/main/writer-ctags.c
+++ b/main/writer-ctags.c
@@ -310,7 +310,9 @@ static int addExtensionFields (tagWriter *writer, MIO *mio, const tagEntryInfo *
    length += renderExtensionFieldMaybe (writer, FIELD_ROLES, tag, sep, mio);
    length += renderExtensionFieldMaybe (writer, FIELD_EXTRAS, tag, sep, mio);
    length += renderExtensionFieldMaybe (writer, FIELD_XPATH, tag, sep, mio);
-   length += renderExtensionFieldMaybe (writer, FIELD_END_LINE, tag, sep, mio);
+
+   if (tag->extensionFields.endLine != tag->lineNumber)
+       length += renderExtensionFieldMaybe (writer, FIELD_END_LINE, tag, sep, mio);
 
    return length;
 }
```
I may change the json writer, too.
However, the impact is not small:
```
  #FAILED (unexpected-output):            52
	parser-autoconf.r/simple2.ac
	parser-autoconf.r/simple.ac
	parser-puppetManifest.r/puppet-classheirarchy
	parser-puppetManifest.r/puppet-collection
	parser-puppetManifest.r/puppet-collection_within_virtual_definitions
	parser-puppetManifest.r/puppet-componentrequire
	parser-puppetManifest.r/puppet-falsevalues
	parser-puppetManifest.r/puppet-fqdefinition
	parser-puppetManifest.r/puppet-multipleinstances
	parser-puppetManifest.r/puppet-namevartest
	parser-puppetManifest.r/puppet-selectorvalues
	parser-puppetManifest.r/puppet-subclass_name_duplication
	parser-puppetManifest.r/puppet-classname
	parser-puppetManifest.r/puppet-definitionname
	parser-puppetManifest.r/unless
	parser-puppetManifest.r/puppet-casestatement
	parser-puppetManifest.r/puppet-classincludes
	parser-puppetManifest.r/puppet-classpathtest
	parser-puppetManifest.r/puppet-deepclassheirarchy
	parser-puppetManifest.r/puppet-filecreate
	parser-puppetManifest.r/puppet-fqparents
	parser-puppetManifest.r/puppet-multipleclass
	parser-puppetManifest.r/puppet-scopetest
	parser-puppetManifest.r/puppet-multisubs
	parser-puppetManifest.r/puppet-simpledefaults
	parser-puppetManifest.r/puppet-singleselector
	parser-c.r/end-field-of-macro
	parser-itcl.r/simple-itcl
	parser-cxx.r/cxx11-lambdas
	parser-cxx.r/cxx11enum.cpp
	parser-cxx.r/function-return-types
	parser-cxx.r/namespace.cpp
	parser-cxx.r/templates-in-labmdas-2.cpp
	parser-java.r/enum.java
	parser-java.r/java_enum.java
	parser-m4.r/m4-comment
	parser-m4.r/disabling-autoconf
	parser-m4.r/disabling-subparser
	parser-m4.r/m4-multi-undef
	parser-m4.r/m4-switch-lang-by-ac
	parser-m4.r/m4-switch-lang-by-m4
	parser-python.r/simple.py
	parser-rpmspec.r/empty-line-in-macro
	parser-rpmspec.r/simple-rpmspec
	parser-asm.r/gas.s
	parser-dtd.r/colons-in-name
	parser-dtd.r/elements
	parser-dtd.r/simple-dtd
	parser-dtd.r/parameter-entities
	parser-tcl.r/comments.tcl
	parser-fypp.r/run-alt-guest
	parser-systemtap.r/functions
```
I wonder how I should do. Providing an option for controlling enabling/disabling the technique is
one of not-so-good idea. 

The end: field is not implemented in some parsers and some kinds.
How can people, including us, know it for a given tag entry?
If end: is omitted, people cannot know whether it is omitted or it is not implemented.

Providing the formal way to report which parser's which kind implements end: may be the best.
However, the command line interface and interaction between clinet tools becomes more complicated. Implementing it with actual demands may be bad idea.

masatake:
The both thoughts are very important for me as input of thinking about consistency of ctags.
Thank you very much.

I will merge this.

masatake:
BTW, I have one more a bit crazy idea: introducing range table.
I showed symbol-table in #2450.
The range table extends it.
The symbol-table provides "name->corkId" mapping.
The symbol-table provides "linenumber->corkId" mapping.

Let me show how range table is useful.

input C code quoted from linux kernel:
```
 1: static const struct file_operations null_fops = {
 2:	.llseek		= null_lseek,
 3:	.read		= read_null,
 4:	.write		= write_null,
 5:	.read_iter	= read_iter_null,
 6:	.write_iter	= write_iter_null,
 7:	.splice_write	= splice_write_null,
 8: };
```

linux.ctags I'm using daily job:
```
--_roledef-C=m.initialized,initialized
--regex-C=/^[[:space:]]*\.([_a-zA-Z][_a-zA-Z0-9]+)[[:space:]]*=[[:space:]]*/\1/m/{_role=initialized}
```

```
null_fops	m.c	/^static const struct file_operations null_fops = {$/;"	kind:variable	line:1	typeref:typename:const struct file_operations	file:	roles:def	extras:fileScope	properties:static
llseek	m.c	/^	.llseek		= null_lseek,$/;"	kind:member	line:2	roles:initialized	extras:reference
read	m.c	/^	.read		= read_null,$/;"	kind:member	line:3	roles:initialized	extras:reference
write	m.c	/^	.write		= write_null,$/;"	kind:member	line:4	roles:initialized	extras:reference
read_iter	m.c	/^	.read_iter	= read_iter_null,$/;"	kind:member	line:5	roles:initialized	extras:reference
write_iter	m.c	/^	.write_iter	= write_iter_null,$/;"	kind:member	line:6	roles:initialized	extras:reference
splice_write	m.c	/^	.splice_write	= splice_write_null,$/;"	kind:member	line:7	roles:initialized	extras:reference
m.c	m.c	1;"	kind:file	line:1	roles:def	extras:inputFile	end:8
```

The regex captures statically initialized structure members.
As I wrote I use this toy-level hack daily, and I found these tags are quite informative.
It reduces many grep invocations. (One of the important goal of Universal-ctags is reducing grep invocations from this Universal :-). 
However, the tags are not perfect. They don't have scope information.
So I can know "write" member is initialized. However, I cannot know "write" member of "null_fops" variable with `const struct file_operations` type.

However, now we have end: field for vairables. That means, logically ctags can tell us "write" is initialized is in "null_fops".  The missing ring is the way to look up the parent scope from the given line number of a tag entry. The range table is the missing component.

In other words, the range table is basic infrastructure for implementing poor-man's reference tags.

masatake:
This closes #2454. I wrote incorrect issue number in the commit log.

pragmaware:
You're right, end field compression is not really feasible. Well, maybe with a new field that explicitly tells about start and end. But it's probably not worth it.

As for the range table, the idea is not crazy. For instance, looking up the context by line number is a must if you're building a smart completion tool. In Code3 it's called [context map](https://github.com/pragmaware/code3/blob/a787c53c44cfef685b5d55fd796e8c36446012d6/code3/src/lib/symbol/tools/C3SymbolFile.cpp#L69). Currently it contains only functions and classes but the concept is the same...

AmaiKinono:
I had a discussion with @masatake on the range table. I'll post it here.

Me:

The example you given in the last is interesting. I think I can understand how ctags will know which scope does a tag live in, but I don't understand how this will help implementing "poor-man's reference tags".

@masatake:

(I think this discussion should be done in the pull request. Feel free to copy and paste what I have written.)

A tag generator can gather reference tags with serious(?) parsing. However, as I wrote I'm in a bit negative about reference tags because I'm not sure how useful they are for the development cost and tag file size.

However, people including I myself should be able to explore how reference tags are useful if ctags supports them.

You can explore the usefulness of reference tags very easily with --regex-<LANG> options.

``` console
$ cat h.c
int
foo(void)
{
  return 0;
}

int
main(void)
{
  return foo ();
}

$ u-ctags --extras=+r --fields=+r \
          --_roledef-C=f.called,called \
          --regex-C='/[^a-zA-Z0-9]([a-zA-Z0-9]+)[ ]*\('/'\1'/f/'{_role=called}' \
          -o - h.c
>>  foo h.c     /^  return foo ();$/;"  f       language:C      roles:called
foo h.c     /^foo(void)$/;" f       language:C      typeref:typename:int    roles:def
main        h.c     /^main(void)$/;"        f       language:C      typeref:typename:int    roles:def
```

Look at the entry ">>". ctags captures a function call well.

This is regex based. So it cannot ignore comments and string literals well. In that mean, I would like to call it "poor-man's reference tags". But I think this is useful if users understand the limitation.

This is still too poor to use in real world. "too poor" means it is not better than using grep directly. The reason I think so is it doesn't have a scope information.

The scope information here is "foo is called from main".

If we can change the current tag entry

```
foo  h.c     /^  return foo ();$/;"  f       language:C      roles:called
```

to

```
foo  h.c     /^  return foo ();$/;"  f       language:C      roles:called    scope:function:main
```

, this is quite useful. I would like to say it is better than grep.

The range table is needed to fill the scope field. To use the range table, the command line will be

``` console
$ u-ctags --extras=+r --fields=+r \
          --range-table=C \
          --_roledef-C=f.called,called \
          --postregex-C='/[^a-zA-Z0-9]([a-zA-Z0-9]+)[ ]*\('/'\1'/f/'{_role=called}{scope=rangeref}' \
          -o - h.c
```

I imagine this will be quite powerful enhancement. It provides "poor but not-too-poor man's reference tags". In many of the cases, I guess "poor but not-too-poor man's reference tags" is enough.

What I'm only afraid of is performance impact.

masatake:
I implemented the feature discussed here.
See #3678. The test cases in [the pull request ](https://github.com/universal-ctags/ctags/pull/3678/commits/9e94cafbdb0f6611862152f5005c29ba63db6c07) illustrate how to use the feature.

Currently, I call it "intervaltab" because the feature is based on an interval tree.

PR #2458, opened 2020-03-06
CXX: Fix type when multiple variables are extracted from a statement

pragmaware:
Fix type when multiple variables are extracted from a statement and an array is present.


masatake:
LGTM. I'm supervised that it can be fixed only one line modification.

pragmaware:
> LGTM. I'm surprised at that it can be fixed only one line modification.

Yep, that one line would have fixed the problem you reported.
But I've found another form of this issue to fix, so I added few more lines now.

PR #2460, opened 2020-03-07
CXX: Fix cpp branching in the namespace statement. 

pragmaware:
Fixes parsing of Qt's qnamespace.h and similar.


pragmaware:
Closing and reopening to re-run the unit tests

PR #2461, opened 2020-03-07
C: fix an unit test that was broken between merges

pragmaware:
Fix an unit test that was broken between out-of-order merges.


pragmaware:
Merging this as a broken unit test blocks all the other pulls.

PR #2462, opened 2020-03-07
CXX: Fix cpp branching in the namespace statement. 

pragmaware:
Fixes parsing of Qt's qnamespace.h and similar.

### DIFF
--- a/Units/parser-cxx.r/namespace-and-preprocessor.cpp.d/args.ctags
+++ b/Units/parser-cxx.r/namespace-and-preprocessor.cpp.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--kinds-C++=*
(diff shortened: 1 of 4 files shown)


pragmaware:
Had to do a few commits to make this work because of a broken unit test by the way.
Squash when merging.

masatake:
What kind of bug do you fix?

pragmaware:
qnamespace.h from Qt contains an ugly construct:

```
#ifdef SOMETHING
  namespace
#else
  class
#endif
  NAME
```

with this change only one #ifdef branch is chosen and parsing does not stop.
This is how cpp in general works, it was just broken in the namespace case.

masatake:
Thank you for the explanation.
I would like you to add it to your commit log.

If the issue is reproduced only when "-o -" or "-f -" arguments, I'm afraid that the issue is related to 
https://github.com/universal-ctags/ctags/pull/2427/commits/75114a494862f0b6bd7d1181d542119e685c3bc1 in #2427.

pragmaware:
I think it never worked properly in the namespace case. I've kind-of-noticed it a long time ago while doing work stuff but somehow never remembered to investigate it.

pragmaware:
Done.

PR #2466, opened 2020-03-10
Use libreadtags repo

### LINKED ISSUE #2451: Make a new repository for libreadtags
To encourage people to write interesting client tools, it may be better to distribute readtags.c (libreadtags) from a separate repository.

The repository may contains readtags.c, readtags.h and Makefile(.am).

As we did for packcc, ctags incorporates libreadtags repo with git subtree.
ctags repo provides readtags command under GPLv2 or later.
However, libreadtags is distributed in public domain style(?) as the developers of Exuberant-ctags wanted.

Unlike u-ctags, libreadtags should be released. 
Before making a release for libreadtags, we have to extend the API for accessing pseudo tags.

I have a plan to link u-ctags to libreadtags. For what? For implementing multi-pass cross input file parsing. If ctags can read a tags file generated in the first pass, ctags can emit higher quality tags file in the 2nd pass.

Making a repository for libreadtags is a step for the multi-pass cross input file parsing.

masatake:
readtags.[ch] are useful even if you don't use ctags; you may use another tags generator.
To make it possible that people install and use readtags.[ch] without buillding/installing ctags, I made a repository, https://github.com/universal-ctags/libreadtags.

With this change, readtags command of u-ctags is linked to the readtags.[ch] available at the new repository. At u-ctags side, nothing should be changed.

Close #2451.


masatake:
Yeah!

PR #2469, opened 2020-03-12
[RFC] lregex: replace gnu_regex with Onigmo regex engine

masatake:
In the much of cases, I hope there is no impact on existing optlib
code using --regex-... option.

If `r` regex flag is given, you can use extended features of Onigmo
with ruby regex syntax.

A demonstration of one of miracle features:

;; input.mylang

    (define (f1) 1)
    (define ((f2)) #t)
    (define (((f3))) "abc")

    --langdef=mylang
    --map-mylang=.mylang
    --kinddef-mylang=f,fun,function, function returing a function, or function returing a function returing function...
    --_fielddef-mylang=symbol,symbol binding to the function
    --fields-mylang={symbol}
    --regex-mylang=/\(define +(([-a-z0-9]+)|\(\g<1>\))/\1/f/r{_field=symbol:\2}

See the r flag passed to --regex-mylang.

    (((f3)))    input.mylang    /^(define (((f3))) "abc")$/;"   f       symbol:f3
    ((f2))      input.mylang    /^(define ((f2)) #t)$/;"        f       symbol:f2
    (f1)        input.mylang    /^(define (f1) 1)$/;"   f       symbol:f1

Look at the name of tags. The pairs of `(' and `)' are balanced well.


masatake:
It seems that using git-subtree is a bad idea.

k-takata:
> It seems that using git-subtree is a bad idea.

How about using `--squash` option? See `misc/pull-packcc.sh`.

Maybe it's better to add also `misc/pull-onigmo.sh` and `misc/pull-libreadtags.sh`?
I don't remember the command line of git-subtree.

masatake:
@k-takata, thank you for the suggestion.
I will add misc/pull-libreadtags.sh first.
I'm thinking about introducing a common script like:

```
misc/pull-subtree.sh [packcc|libreadtags|onigmo]
```

if possible.

masatake:
Instead of replacing gnu_regex, I'm thinking about adding pcre2 to ctags. See #3036.

Issue #2476, opened 2020-03-28
Ruby: deal with mixin?


AmaiKinono:
I don't know what's the exact output I want, but I have some ideas.

Let me explain a bit about mixin in Ruby (however keep in mind I am also new to Ruby). Mixin is a mechanism to "mix" a module into a class. It's done using one of 3 methods: include, prepend and extend.

Let's define a module first:

``` ruby
module X
  def hi
    p "Calling 'hi' in X."
  end
end
```

Include `X` in `A` will make `hi` an instance method of `A`, and `X` will be inserted to the ancestor chain of `A` (just before `A` itself, so methods defined in `A` can override it):

``` ruby
class A
  include X
  def hi
    p "Calling 'hi' in A."
  end
end

a = A.new
a.hi
# => "Calling 'hi' in A."
# A.hi will throw an error.
A.ancestors
=> [A, X, Object, Kernel, BasicObject]
```

Prepend `X` in `B` will make `hi` an instance method of `B`, and `X` will be inserted to the ancestor chain of `B` (after `B` itself, so its methods overrides the ones defined in `B`):

``` ruby
class B
  prepend X
  def hi
    p "Calling 'hi' in B."
  end
end

b = B.new
b.hi
# => "Calling 'hi' in X."
# B.hi will throw an error.
B.ancestors
=> [X, B, Object, Kernel, BasicObject]
```

Extend `X` in `C` will make `hi` a class method of `C` (seems it won't override class methods defined in `C`), and doesn't affect the ancestor chain:

``` ruby
class C
  extend X
  def hi
    p "Calling 'hi' in C."
  end
end

c = C.new
c.hi
# => "Calling 'hi' in C."
C.hi
# => "Calling 'hi' in X."
C.ancestors
# => [C, Object, Kernel, BasicObject]
```

It seems modules can also include variables and classes, but I'm not familiar with that and don't know if it will affect the mixin.

My main concern is to make auto-completion (based on class information) and jumping to definition work. My thoughts are:

- For `A`: generate tags for methods in the module `X` again, but only when they are not overrided by instance methods defined in the class. Their scope should be `class:A`, and line numbers should point to definitions in `X`.

- For `B`: generate tags for methods in the module `X` again, and remove the tags for overrided instance methods in `B`. Their scope should be `class:B`, and line numbers should point to definitions in `X`.

[...]

PR #2478, opened 2020-03-31
Ruby: introduce mixin field

masatake:
See #2476.

@AmaiKinono, plesae look at 205052c. It is a test case showing how new code dealing with "include" mixin.

You may be interested in c4711e8 ..205052c.  
c4711e8 and 7af2181 are preparation for 0912d1e.


masatake:
I thought nestingLevel was not needed because we have cork.
I find it is wrong. To manage information which is not suitable to attach to a tagEntryInfo, nestingLevel is still useful.

masatake:
I didn't implement reference tags for X and Y because I'm not sure whether it is needed or not.

AmaiKinono:
I think I've done the right changes, but now I don't know how to push it to this branch. Maybe you need to open some "push access" for me? Not sure if it can be done on a per branch basis.

masatake:
> I think I've done the right changes, but now I don't know how to push it to this branch. Maybe you need to open some "push access" for me? Not sure if it can be done on a per branch basis.

> Edit: Oh I just got it. I can create a PR on your fork :D Is that the right way?
Merge state

I will merge this one first. Then you can open your pull request.

Before merging this, I would like to think about following case:
```
class A
  def hi
     include X
    p "Calling 'hi' in A."
  end
end
```

AmaiKinono:
The code you given actually doesn't work on my side. A working code should look like:

- Use instance method:

``` ruby
class A
  def hi
    self.class.prepend X
    p "Calling 'hi' in A."
  end
end

# a = A.new
# a.hi
# => "Calling 'hi' in A."
# a.hi
# => "Calling 'hi' in X."
```

- Or use class method:

``` ruby
class A
  def self.hi
    self.include X
  end
end

# a = A.new
# a.hi
# => (undefined method error)
# A.hi
# a.hi
# => "Calling 'hi' in X."
```

I think we could not bother with this, because it can actually happen anywhere:

``` ruby
class A
end
a = A.new

# a.hi
# (undefined method error)
# A.prepend X
# a.hi
# => "Calling 'hi' in X."
```

masatake:
@AmaiKinono , thank you for the explanation.

I found a pattern I would like to support:
```
module X
  def hi
    p "Calling 'hi' in X."
  end
end

class A
  def self.prep
    include X
  end
  def hi
    p "Calling 'hi' in A."
  end

end

a = A.new
A.prep
p A.ancestors
```

Recognizing include in the singleton method is not hard.

masatake:
@AmaiKinono, give me an ack if you think it is ready to merge.
Tell me if you think the last commit should be removed.

AmaiKinono:
Yes I think this is ready to merge.

About `include` in a singleton method: I can see this is useful. I like it. Currently I can't think about similar cases (as I said I am new to ruby), which might be better to add now. But maybe we could extend it in the future when these cases actually come up.

masatake:
@AmaiKinono, thank you for reviewing.

Feel free to ask me questions about the changes here.

> About include in a singleton method: I can see this is useful. I like it. Currently I can't think about similar cases (as I said I am new to ruby), which might be better to add now. But maybe we could extend it in the future when these cases actually come up.

I agree with you.

PR #2479, opened 2020-03-31
Cxx: attach the end: field to a prototype kind tag

### LINKED ISSUE #2473: Question : Can tags be generated for in-memory code file?
It is more like a feature query, whether this is supported with the existing version or whether there is an upcoming release that will support this?

### LINKED ISSUE #2474: can I get the end line of prototype in C/C++?
I want to get the end line of  `computeLen` and `foo` in the following code:
```
extern int computeLen(int number,
                      int size,
                      int width);
class Shape 
{
   public:
      void setWidth(int w)
      {
         width = w;
      }
      void foo(int h, int w,
                int g, int k);
   protected:
      int width;
};
```
I use command ` ctags --output-format=json --langmap=c:+.inc.def.ph --kinds-c=+plz --fields=+nie -o - test.cpp` to analyze and get the start line of `computeLen` and `foo`. But I still want to get the end line of them.
Is there any option I can use to get end line info? Or can you add end line information for the prototype kind symbol?

masatake:
Close #2474.

```
[jet@living]~/var/ctags% ./ctags --fields=+e --kinds-C++=p -o - ./input.h
main	./input.h	/^main$/;"	p	typeref:typename:int	end:8
[jet@living]~/var/ctags% cat -n input.h                                  
     1	int
     2	main
     3	(
     4		int
     5		argc,
     6		char **
     7		argv
     8		);
[jet@living]~/var/ctags% ./ctags --fields=+e --kinds-C++=p -o - ./input.h
main	./input.h	/^main$/;"	p	typeref:typename:int	end:8
```

### DIFF
--- a/parsers/cxx/cxx_parser.c
+++ b/parsers/cxx/cxx_parser.c
@@ -1485,7 +1485,13 @@ check_function_signature:
 			// out its proper scope. Better avoid emitting this one.
 			CXX_DEBUG_PRINT("But it has been preceded by the 'friend' keyword: this is not a real prototype");
 		} else {
-			int iScopesPushed = cxxParserEmitFunctionTags(&oInfo,CXXTagKindPROTOTYPE,CXXEmitFunctionTagsPushScopes,NULL);
+			int piCorkQueueIndex;
+			int iScopesPushed = cxxParserEmitFunctionTags(&oInfo,CXXTagKindPROTOTYPE,CXXEmitFunctionTagsPushScopes, &piCorkQueueIndex);
+			if (piCorkQueueIndex != CORK_NIL)
+			{
+				CXXToken * t = cxxTokenChainLast(g_cxx.pTokenChain);
+				cxxParserSetEndLineForTagInCorkQueue (piCorkQueueIndex, t->iLineNumber);
+			}
 			while(iScopesPushed > 0)
 			{
 				cxxScopePop();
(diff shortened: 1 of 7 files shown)


masatake:
The parameters in a prototype can be tagged.
I showed the change for it at a comment in #2474.
I wonder whether the change should be merged or not.

parsers/cxx/cxx_parser.c
```diff
@@ -1485,7 +1485,10 @@ void cxxParserAnalyzeOtherStatement(void)
 			// out its proper scope. Better avoid emitting this one.
 			CXX_DEBUG_PRINT("But it has been preceded by the 'friend' keyword: this is not a real prototype");
 		} else {
-			int iScopesPushed = cxxParserEmitFunctionTags(&oInfo,CXXTagKindPROTOTYPE,CXXEmitFunctionTagsPushScopes,NULL);
+			int piCorkQueueIndex;
+			int iScopesPushed = cxxParserEmitFunctionTags(&oInfo,CXXTagKindPROTOTYPE,CXXEmitFunctionTagsPushScopes, &piCorkQueueIndex);
+			if (piCorkQueueIndex != CORK_NIL)
+				cxxParserSetEndLineForTagInCorkQueue (piCorkQueueIndex, g_cxx.pToken->iLineNumber);
```
pragmaware, line 1491: Here we are assuming that g_cxx.pToken points to the last token of the chain.
This is currently true. But maybe we could use the cxxTokenChainLast() instead to remove this dependency.

pragmaware:
Sorry for the delay in the review. Been a bit busy.

parsers/cxx/cxx_parser.c
masatake: Thank you. Updated.

masatake: Could you merge this pull request if  the change is o.k.?

pragmaware: Ow, sure!

PR #2481, opened 2020-04-02
Ruby: extend mixin field

AmaiKinono:
The solution developed in https://github.com/universal-ctags/ctags/pull/2478 is extended to deal with "prepend" and "extend" methods.

One thing I've noticed is `include`, `prepend` and `extend` are all methods, so it's valid to write:

``` ruby
class A
  prepend(X)
end
``` 
But it's not handled by current scheme.

### DIFF
--- a/Units/parser-ruby.r/ruby-mixin-field.d/expected.tags
+++ b/Units/parser-ruby.r/ruby-mixin-field.d/expected.tags
@@ -6,3 +6,11 @@ A	input.rb	/^class A$/;"	c	mixin:include:X,include:Y
 hi	input.rb	/^  def hi$/;"	f	class:A
 B	input.rb	/^class B$/;"	c	mixin:include:X
 prep	input.rb	/^  def self.prep$/;"	S	class:B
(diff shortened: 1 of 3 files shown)


masatake:
Thank you!

masatake:
```
class A
  prepend(X)
end
```

This can be captured with:
```
[yamato@slave]~/var/ctags-github% git diff parsers/ruby.c
diff --git a/parsers/ruby.c b/parsers/ruby.c
index f6ac449c..b6d465b2 100644
--- a/parsers/ruby.c
+++ b/parsers/ruby.c
@@ -476,8 +476,13 @@ static void readAndStoreMixinSpec (const unsigned char** cp, const char *how_mix
    if (! (e->kindIndex == K_CLASS || e->kindIndex == K_MODULE))
        return;

-   if (isspace (**cp))
+   if (isspace (**cp) || (**cp == '('))
    {
+       if (isspace (**cp))
+           skipWhitespace (cp);
+       if (**cp == '(')
+           ++*cp;
+
        vString *spec = vStringNewInit (how_mixin);
        vStringPut(spec, ':');

```
I will make a pull request for this after merging #2482.

Issue #2488, opened 2020-04-04
Typescript: tags not found


sparkcanon:
The name of the parser: **Typescript**

The command line you used to run ctags:

```
$ ctags --options=NONE -e src/index.ts
```

The content of input file:

```typescript
import * as three from "three";

let scene: three.Scene,
  camera: three.PerspectiveCamera,
  renderer: three.WebGLRenderer;
const cubeArr = new three.Group();

// TODO: Add lighting
const setupScene = (): void => {
  scene = new three.Scene();
  camera = new three.PerspectiveCamera(
    75,
    window.innerWidth / window.innerHeight,
    0.1,
    1000
  );

  renderer = new three.WebGLRenderer();
  renderer.setSize(window.innerWidth, window.innerHeight);
  document.body.appendChild(renderer.domElement);
  camera.position.z = 20;
};

// TODO: Position the cubes in the center of the screen
const createCubeArr = (): void => {
  let cubes;

  for (let i = 0; i < 5; i++) {
    for (let r = 1; r < 5; r++) {
      cubes = createCube();
      cubes.position.set(i * 3, r * 2, 0);
      cubeArr.add(cubes);
    }
  }
  cubeArr.position.set(-5, -5, 0);
  scene.add(cubeArr);
};

// TODO: Receive shadow
const createWall = (): void => {
  const geometry = new three.PlaneGeometry(20, 20, 20);
  const material = new three.MeshBasicMaterial({
    color: 0x0000ff,
    side: three.DoubleSide,
  });
  const mesh = new three.Mesh(geometry, material);
  scene.add(mesh);
};

const createCube = (): three.Mesh => {
  const geometry, material;
  geometry = new three.BoxGeometry(1, 1, 1);
  material = new three.MeshBasicMaterial({
    color: 0x00ff00,
    side: three.DoubleSide,
  });
  const cube = new three.Mesh(geometry, material);
  return cube;
};

const animate = (): void => {
  requestAnimationFrame(animate);
  renderer.render(scene, camera);
};

setupScene();
createCubeArr();
createWall();
animate();
```

The tags output you are not satisfied with:

```

src/index.ts,134
let scene: three.Scene,scene3,33
const cubeArr = new three.Group();cubeArr6,125
const setupScene = (): void => {setupScene9,183

```

Expected tags:

All the other `const`s should be detected.

The version of ctags:

```
$ ctags --version
[...]

PR #2493, opened 2020-04-04
[typescript] #2488

ksamborski:
A fix for #2488 for better variables parsing. Especially when they contain regexes (/.../) or division (/).

### DIFF
--- a/Units/parser-typescript.r/ts-const.d/input.ts
+++ b/Units/parser-typescript.r/ts-const.d/input.ts
@@ -1,0 +1,70 @@
+// Taken from issue #2488 submitted by @sparkcanon
+import * as three from "three";
+
+let scene: three.Scene,
+  camera: three.PerspectiveCamera,
+  renderer: three.WebGLRenderer;
+const cubeArr = new three.Group();
+
+// TODO: Add lighting
+const setupScene = (): void => {
+  scene = new three.Scene();
+  camera = new three.PerspectiveCamera(
+    75,
+    window.innerWidth / window.innerHeight,
+    0.1,
+    1000
+  );
+
+  renderer = new three.WebGLRenderer();
+  renderer.setSize(window.innerWidth, window.innerHeight);
+  document.body.appendChild(renderer.domElement);
+  camera.position.z = 20;
+};
+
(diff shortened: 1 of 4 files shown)


Units/parser-typescript.r/ts-const.d/input.ts
```diff
@@ -0,0 +1,69 @@
+import * as three from "three";
```
masatake, line 1: I would like you to write the source of the test input file like
```
// Taken from issue #2488 submitted by @sparkcanon
```

(Fix the example if my English is wrong.)

ksamborski: Good point, thanks! I've updated the test file.

masatake: Updated in c8cb4a4 ?
As far as seeing https://github.com/universal-ctags/ctags/pull/2493/files, the test input is not changed.

ksamborski: yeah, I was doing `git commit --amend` and forgot to run `git add` first. It should be ok now.

sparkcanon:
I tested your branch, @ksamborski and noticed that more tags are now listed.
```

src/index.ts,561
const scene: three.Scene,scene3,33
  camera: three.PerspectiveCamera,camera4,59
  renderer: three.WebGLRenderer,renderer5,94
const setupScene = (): void => {setupScene19,401
const createCube = (): three.Mesh => {createCube37,886
const createCubeArr = (): void => {createCubeArr50,1275
const createWall = (): void => {createWall67,1702
const animate = (): void => {animate78,1999
```

Strangely, some other tags are not recognised such as,
`Units/parser-typescript.r/ts-const.d/input.ts` line 42
`Units/parser-typescript.r/ts-const.d/input.ts` line 43
`Units/parser-typescript.r/ts-const.d/input.ts` line 52
`Units/parser-typescript.r/ts-const.d/input.ts` line 27

sparkcanon:
> @sparkcanon yes, but currently the body of lambda functions is ignored. I will look into that in the future.

Makes sense, thank you.

PR #2497, opened 2020-04-07
Ruby: fill signature fields

masatake:

### DIFF
--- a/parsers/ruby.c
+++ b/parsers/ruby.c
@@ -466,6 +466,47 @@ static void parseString (const unsigned char** cp, vString* vstr)
 		++*cp;
 }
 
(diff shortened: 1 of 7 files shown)


masatake:
The change assumes a signature is given in a line.

parsers/ruby.c
```diff
@@ -436,6 +436,34 @@ static rubyKind parseIdentifier (
 	return kind;
 }
 
+
+static void parseSignature (const unsigned char** cp, vString* vstr)
+{
+	int depth = 1;
+
+	/* FIXME: handle string literals including ( or ). */
+	while (! (depth == 0) || (**cp == '\0'))
```
masatake, line 445: This should be:
```
while (! (depth == 0 || **cp == '\0'))
```

PR #2500, opened 2020-04-09
readtags: add upcase operator

AmaiKinono:
To be honest, I don't quite understand the code, and I'm just imitating the existing code. Please see if there's any problem.


Tmain/readtags-qualifier-op-downcase-upcase.d/run.sh
```diff
@@ -27,5 +27,17 @@ ${V} ${READTAGS} -t output.tags -Q '(member "a" (downcase $inherits))' -l
 echo '# (member "b" (downcase $inherits))'
 ${V} ${READTAGS} -t output.tags -Q '(member "b" (downcase $inherits))' -l
 
-echo '# (member "d" (downcase $inherits))'
+echo '# (member "c" (downcase $inherits))'
```
masatake, line 30: This change is not needed. I specified "d" as an object that is not in `(downcase $inherits)`.
I expected no output for ths filter condition.

Tmain/readtags-qualifier-op-downcase-upcase.d/stdout-expected.txt
```diff
@@ -0,0 +1,14 @@
+# (downcase $name)
+C	input.cpp	/^class C: A, B {$/
+# (member "a" (downcase $inherits))
+C	input.cpp	/^class C: A, B {$/
+# (member "b" (downcase $inherits))
+C	input.cpp	/^class C: A, B {$/
+# (member "c" (downcase $inherits))
```
masatake, line 7: This should be `(member "d" (downcase $inherits))`.

Tmain/readtags-qualifier-op-downcase-upcase.d/stdout-expected.txt
```diff
@@ -0,0 +1,14 @@
...
+# (upcase $name)
+C	input.cpp	/^class C: A, B {$/
```
masatake, line 9: C is already being "upcase'ed" in the output.tags. So this cannot test the behavior of "upcase" operator.

What your code tests:
```
(member "C" (upcase "C"))
```

B and C, the super classes of C are also already "upcase'ed'.
See output.tags. 
```
f	input.cpp	/^  int f (void) {$/;"	f	class:A	typeref:typename:int	file:
g	input.cpp	/^  int g (void) {$/;"	f	class:B	typeref:typename:int	file:
h	input.cpp	/^  int h (void) {$/;"	f	class:C	typeref:typename:int	file:
```
You can use f, g, and h for testing upcase operator like:

```
echo '# (upcase $name)'
${V} ${READTAGS} -t output.tags -Q '(eq? "F" (upcase $name))' -l
```

AmaiKinono:
@masatake I don't quite understand run.sh. I think it should print the expression that's actually used, so in my opinion, this

``` sh
echo '# (downcase $name)'
${V} ${READTAGS} -t output.tags -Q '(eq? "c" (downcase $name))' -l
```

should be 

``` sh
echo '# (eq? "c" (downcase $name))'
${V} ${READTAGS} -t output.tags -Q '(eq? "c" (downcase $name))' -l
```

and this

``` sh
echo '# (member "d" (downcase $inherits))'
${V} ${READTAGS} -t output.tags -Q '(member "c" (downcase $inherits))' -l
```

should be 

``` sh
echo '# (member "c" (downcase $inherits))'
${V} ${READTAGS} -t output.tags -Q '(member "c" (downcase $inherits))' -l
```

I understand you want to test a false expression, but it doesn't explain why using `d` in the "comments" and `c` in the expression.

masatake:
> I understand you want to test a false expression, but it doesn't explain why using d in the "comments" and c in the expression.

Good catch. 

At the first case, I should write `'# (eq? "c" (downcase $name))'` as you wrote.
At the second case, I should write 
```
echo '# (member "d" (downcase $inherits))'
${V} ${READTAGS} -t output.tags -Q '(member "d" (downcase $inherits))' -l
```.  As you wrote, I would like to test a false expression.

masatake:
LGTM. Sorry for making you confused.

AmaiKinono:
Never mind :D

masatake:
Thank you!

PR #2511, opened 2020-04-14
readtags: printer

masatake:
This change makes the code for printing tags in readtags command more generic.
--escape option is introduced.

The function for printing a tag entry will be moved to libreadtags.

TODO:

We can not merge this till https://github.com/universal-ctags/libreadtags/issues/17 is fixed.

We have to decide how to handle special characters like tab character in the input file name fields and pattern fields.


masatake:
#2512 fixes https://github.com/universal-ctags/libreadtags/issues/17.

Issue #2498, opened 2020-04-09
c++: end line of a variable is not working for global struct variable in linux source


chongchai:
I find ctags supports end line of variable from https://github.com/universal-ctags/ctags/issues/2454.

I use this command to parse test.c.

## command:
`ctags.exe --output-format=json --langmap=c:+.inc --kinds-c=+plz --fields=+nie -o - d:\test.c`

## test.c
```
static const struct super_operations dax_sops = {
	.statfs = simple_statfs,
	.alloc_inode = dax_alloc_inode,
	.destroy_inode = dax_destroy_inode,
	.free_inode = dax_free_inode,
	.drop_inode = generic_delete_inode,
};

int n[2] = {
  1,
  2
};
```

## result
```
{"_type": "tag", "name": "dax_sops", "path": "D:\\tmp\\test\\al.c", "pattern": "/^static const struct super_operations dax_sops = {$/", "file": true, "line": 1, "typeref": "typename:const struct super_operations", "kind": "variable"}
{"_type": "tag", "name": "n", "path": "D:\\tmp\\test\\al.c", "pattern": "/^int n[2] = {$/", "line": 9, "typeref": "typename:int[2]", "kind": "variable", "end": 12}
```

The variable `dax_sops` does not have end line, while `n` has.

**PS**: `dax_sops` code snippet is copyed from super.c in linux source code (https://github.com/torvalds/linux/blob/master/drivers/dax/super.c#L482)

masatake:
Thank you for reporting. Thank you for writing explicitly the origin of the test input. It helps us add a test case. 

I have tried fixing the issue. The fix is not simple than I expected. So I would like to get feedback from the parser maintainer.

```
[yamato@slave]~/var/ctags-github% cat -n /tmp/bar.c
cat -n /tmp/bar.c
     1	struct point {
     2		int x, y;
     3	};
     4	
     5	struct point P = {
     6		.x = 1,
     7		.y = 2,
     8	};
     9	
    10	/* END HERE */
[yamato@slave]~/var/ctags-github% ./ctags -o - --fields=+'{end}' /tmp/bar.c
./ctags -o - --fields=+'{end}' /tmp/bar.c
P	/tmp/bar.c	/^struct point P = {$/;"	v	language:C	typeref:struct:point	end:8
point	/tmp/bar.c	/^struct point {$/;"	s	language:C	file:	end:3
[...]

PR #2517, opened 2020-04-18
CXX: Improve extraction of variables

pragmaware:
Avoid corrupting the token chain before extracting variables when an assignment is found. This allows us to extract more variables and also to attach end-line tags properly.


masatake:
...Wonderful change. The change is very simple. I must study your code harder and extract the essence.

masatake:
Could you write
"Close #2498"
to the commit log?

masatake:
@ pragmaware, I'll take over this pull request.
I will add "Close #2498" and merge this.

masatake:
@pragmaware, thank you.

PR #2519, opened 2020-04-22
readtags: introduce -S option for customizable sorting

masatake:
Sort by names:
```
readtags -t output.tags -ne -S '(<> $name &name)' -l
```

Sort by input file names
```
readtags -t output.tags -ne -S '(<> $input &input)' -l
```

Sort entries having "schedule" as name by input filenames then line numbers:
```
readtags -t ~/var/linux/tags -ne -S '(<or> (<> $input &input) (<> $line &line))' - schedule
```

Flip the last one:
```
readtags -t ~/var/linux/tags -ne -S '(*- (<or> (<> $input &input) (<> $line &line)))' - schedule

```


masatake:
dsl core becomes enough general.
That means we can use it not only in readtags but also in ctags itself.

masatake:
I will submit more comprehensive test cases for readtags when writing readtags(1).

AmaiKinono:
I think I can understand the examples you gave, but what does `&` stand for? I don't understand the help message:

```
&: (& NAME) -> #f|<string>
```

masatake:
> I think I can understand the examples you gave, but what does & stand for? I don't understand the help message:

Do you know `$`  ?
`&` works as the same as `$` but `&` is for the alternative entry.

In filtering (querying), we have to deal with only one entry within an S expression.
`$` is a generic field accessor for the entry.

In sorting, we have to deal with TWO entries for comparison.
`&` is a generic field accessor for the alternative entry.

Do I make sense?

AmaiKinono:
Thanks. I understand what you said, and I've read the `news.rst`. I think I know how to use it now.

My understanding is what we actually do is to implement a classical "compare function" by the sorter expression, and we can think its arglist is two (generic) lines. `$` is used to get fields from one line, and `&` is from another. Am I right?

masatake:
Yes.
I could not write as you write understandably. Please, help me when writing the man page for readtags.

AmaiKinono:
> Please, help me when writing the man page for readtags.

Of course, please count me in. And, if you are ok, how about letting me write the draft?

masatake:
> Of course, please count me in. And, if you are ok, how about letting me write the draft?

Yes, please. I will prepare a stab. Please, wait for a while.

masatake:
Ready.

Please, edit man/readtags.1.rst.in.
DON'T EDIT man/readtags.1.rst and man/readtags.1. They are generated files.
Feel free to edit man/ctags-client-tools.7.rst.in, too.

Some tips are in man/README.
The most important one is
```
* the files in ../docs/man/ need to be updated after writing man pages.
  Execute the following command to update them:

	make QUICK=1 update-docs
```

To reflect the change to https://ctags.io, you may have to run above command, and add the change of docs/man/ to your commit.

THANK YOU VERY MUCH.

PR #2521, opened 2020-04-25
dsl: introduce true and false as aliases for #t and #f

masatake:
Consider writing a client tool runs on emacs.
The tool may use -Q and -S options of readtags.

The developers of the client tool may want to pass
an S expression built-in emacs lisp to readtags
directly.

However, there was a gap in what we call S expression
in emacs and readtags.

readtags's S expression is scheme based; #t and #f is
used as representation of boolean values. These representation
is not used in emacs. Further more, `#' is special character
(prefix of a reader macro) in emacs. So, it is hard for emacs
to build an S expression passed to readtags if the expression
includes a boolean value.

To fill the gap, this change introduces `true' and `false'
values to the readtags side. So the client tool that runs on
emacs doesn't doesn't have to use #t and #f directly in S
expression building process.

### DIFF
--- a/Tmain/readtags-alias-for-boolean.d/run.sh
+++ b/Tmain/readtags-alias-for-boolean.d/run.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+
+# Copyright: 2020 Masatake YAMATO
+# License: GPL-2
+
+READTAGS=$3
+#V="valgrind --leak-check=full --track-origins=yes -v"
+V=
+
+. ../utils.sh
+
+if ! [ -x "${READTAGS}" ]; then
+	skip "no readtags"
+fi
+
+if ! ( "${READTAGS}" -h | grep -q -e -Q ); then
+    skip "no qualifier function in readtags"
+fi
+
+echo '!_ -Q #t point3d'
+${V} ${READTAGS} -t output.tags -Q '#t' point3d
+
+echo '!_ -Q ture point3d'
+${V} ${READTAGS} -t output.tags -Q 'true' point3d
+
+echo '!_ -Q #f point3d'
+${V} ${READTAGS} -t output.tags -Q '#f' point3d
+
+echo '!_ -Q false point3d'
+${V} ${READTAGS} -t output.tags -Q 'false' point3d
+
+echo '!_ -Q #t -l'
+${V} ${READTAGS} -t output.tags -Q '#t' -l
+
+echo '!_ -Q ture -l'
+${V} ${READTAGS} -t output.tags -Q 'true' -l
+
+echo '!_ -Q #f -l'
+${V} ${READTAGS} -t output.tags -Q '#f' -l
+
+echo '!_ -Q false -l'
+${V} ${READTAGS} -t output.tags -Q 'false' -l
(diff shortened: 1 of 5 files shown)


Tmain/readtags-alias-for-boolean.d/run.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+if ! ( "${READTAGS}" -h | grep -q -e -Q ); then
+    skip "no qualifier function in readtags"
+fi
+
+echo '!_ -Q #t point3d'
+${V} ${READTAGS} -t output.tags -Q '#t' point3d
+
+echo '!_ -Q ture point3d'
```
AmaiKinono, line 23:
```suggestion
echo '!_ -Q true point3d'
```

There are similar mistakes in other places :) But I wonder how the test passed since you spell it right in the code.

masatake: Adding the comments after writing the test code.
Thank you!

AmaiKinono:
I tried and found in elisp, you can use `#` in a symbol name, just escape it use `\`.

The [manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Symbol-Type.html) says:

> A symbol name can contain any characters whatever...Any other characters may be included in a symbol's name by escaping them with a backslash.

masatake:
> I tried and found in elisp, you can use # in a symbol name, just escape it use \.

Is it possible to print in readtags acceptable way?

```
(insert (format "%S" '\#t))
```
This inserts '\#t'.
I think we should not fight against the reader macro.

AmaiKinono:
In my experiment, `(insert (format "%S" '\#t))` inserts `\#t`.

I tried:

``` elisp
(shell-command-to-string
 (format "readtags -Q '%S' -l"
         '\#t))
```

It gives me:

```
GOT ERROR in QUALIFYING: unbound-variable: |\\#t|
```

So you are right, we still need those aliases.

But, Iet me talk some of my thoughts. I thought using `%S` to format a sexp is no problem because:

- It wrap the strings in double quotes, which is what we needed in a readtags filter expression.
- It escapes some characters using backslashes, which, in the shell, also means escaping, so we are good to go.

But this example shows if you put backslashes inside quotes, it is not interpreted as escaping the next char, that causes problems. I think this is a enough reason for not using `%S`, and giving aliases is not a radical solution. Rather, I suggest seeing if `%s` would work. This:

``` elisp
(shell-command-to-string
 (format "readtags -Q '%s' -l"
         '\#t))
```

The main problem is `%s` can't wrap strings. I think we scan the list and wrap all strings by ourselves beforehand.

AmaiKinono:
I thought for a while. I think the main problem is:

- We assume the filter expression is in single quotes.

- In single quotes, most chars doesn't need to be escaped, but some still, like newline, and single quote it self (in a weird way: an escaped `'` is `'"'"'`).

- `%S` escapes all it thinks need escaping, and `%s` escapes none.

So what we need is a function that transforms a list into a string that's suitable for putting in a readtags command, whose behavior is between `%S` and `%s`. Anyway, in my opinion, this is a client side problem, not readtags side problem.

Tmain/readtags-alias-for-boolean.d/run.sh
masatake: It seems that I took a mistake in reflecting the suggestions.

PR #2531, opened 2020-04-30
main: disable all objpools in debug build

masatake:
objpool makes the output of valgrind difficult to be understood.
When ./ctags is built with --enable-debugging, this change disables
all objpools.

### DIFF
--- a/main/debug.h
+++ b/main/debug.h
@@ -51,6 +51,10 @@
 # endif
 #endif
 
+#ifdef DEBUG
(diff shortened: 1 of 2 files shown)


masatake:
..This makes detecting a bug in objpool itself hard on CI/CD environments.
Making controllable from the command line option like --debug is better.

masatake:
I think I must revise --debug option to support disabling objpool.

PR #2539, opened 2020-05-09
CXX: Avoid stack overflow with recursive macros or broken input

### LINKED ISSUE #2538: C/C++: Recursive macro expansion causes a segmentation fault
```
$ cat /tmp/bar.c
int
f(int v)
{
  return v & M;
}
$ u-ctags --options=NONE -D'M=T(M)' /tmp/bar.c
u-ctags: Notice: No options will be read from files or environment
zsh: segmentation fault (core dumped)  u-ctags --options=NONE -D'M=T(M)' /tmp/bar.c
```

pragmaware:
Fixes #2538


Units/parser-cxx.r/bug-github-2538.cpp.d/args.ctags
```diff
@@ -0,0 +1,2 @@
+-D'M=T(M)'
+
```
masatake, line 2: We can remove this empty line.

masatake:
Thank you for the very quick response.

masatake:
I found one more case entering infinite loop.

```
[yamato@slave]~/var/ctags-github% cat /tmp/bar.c
union v {
  struct {
    int dx;
  } d;
};

static
int xd(union v *p)
{
  return p->dx;
}
[yamato@slave]~/var/ctags-github% u-ctags --options=NONE -Ddx=d.dx /tmp/bar.c 
u-ctags: Notice: No options will be read from files or environment
^C

```
If I should report this as a new issue, let me know.

masatake:
During expanding macro X, X appeared in the expansion is never expanded again.
Quoted from https://mkotha.hatenadiary.org/entry/20101016/1287193962 (Japanese):
```
#define A() [B()]
#define B() {A()}

A() /* is expanded to [{A()}] */
```

I guess we may have to change the interaction between Cxx parser and CPreProcessor (semo-)parser.

masatake:
I think following two changes are needed.

* cppBuildMacroReplacement should expand the given macro recursively but should not expand already expanded tokens.
* Cxx parser should not call request to expand to CPreProcessor semi-parser for the tokens tokenized from the string returned from cppBuildMacroReplacement.

I can image the way to change for the former one.
About the later one, I have to study Cxx parser more. 
cxxParserParseNextTokenApplyReplacement expands the macro and pushes back the expanded string to the input stream. Then it reads input with cppGetc(). This is too simple to know the input char is part of already expanded macro or not.

I wonder how about expanding cppGetc like:
```
int cppGetc (boot *is_in_already_expanded_code);
```

With the value `*is_in_already_expanded_code`, Cxx parser can decide whether the parser should request expanding a macro to CPreProecssor semi-parser or not.

masatake:
About the original issue #2538, we have a workaround.  We would like to avoid the segmentation fault.
However, if the user doesn't define such a macro in the command line, the segmentation fault doesn't occur. In the other word, #2538 is not critical in the current ctags.

The issue is critical only when we introduce automatic macro expansion I proposed in #2427.

I'm busy in this week. So I will not touch this issue.

pragmaware:
I have added a workaround for this too. When the token chain grows too large we stop expanding macros.

masatake:
LGTM.

I tried your change proposed here combined with infile-macro-expanding feature I have developed at #2427 against the codebase. The codebase includes the source code of linux kernel and more.

With your change,  ctags --param-CPreProcessor:_expand=1 --fields-C=+{macrodef} --fields=+Ss runs without crashing or entering an infinite loop.

@pragmaware, thank you for preparing the workaround. Could you merge this pull request?
Later I will merge the result of #2427.

After merging #2427, I will try to improve the way to expand the macros as I wrote in https://github.com/universal-ctags/ctags/pull/2539#issuecomment-626587207 .

masatake:
I would like to expand cpp macro in asm parser and ldscript parser.
Therefore I would like to move the macro expansion logic in Cxx parser to CPreProcessor parser as much as possible. So the change for expanding macros in the asm parser and the ldscript parser can be little.

On the other hand, I have to consider the performance impact of the moving.
In the current logic (with the work around proposed here), _expand=1 makes parsing about 2 times slower:

```
[yamato@slave]~/var/codebase% ./codebase  ctags C macro-expand
version: 45c5b3e3
features: +wildcards +regex +iconv +debug +option-directory +xpath +json +interactive +sandbox +yaml +packcc
log: results/45c5b3e3,C...................,macro-expand,time......,default...,2020-05-15-01:57:36.log
tagsoutput: /dev/null
cmdline: + u-ctags --quiet --options=NONE --sort=no --options=profile.d/maps --options=profile.d/macro-expand.ctags --totals=yes --languages=C -o - -R code/linux code/php-src code/r-source code/ruby
28395 files, 20052345 lines (564677 kB) scanned in 51.0 seconds (11074 kB/s)
1197504 tags added to tag file
longest tag line = 6653

real	0m52.279s
user	0m49.144s
sys	0m2.900s
+ set +x
[yamato@slave]~/var/codebase% ./codebase  ctags C             
version: 45c5b3e3
features: +wildcards +regex +iconv +debug +option-directory +xpath +json +interactive +sandbox +yaml +packcc
log: results/45c5b3e3,C...................,..........,time......,default...,2020-05-15-01:58:40.log
tagsoutput: /dev/null
cmdline: + u-ctags --quiet --options=NONE --sort=no --options=profile.d/maps --totals=yes --languages=C -o - -R code/linux code/php-src code/r-source code/ruby
28395 files, 20121318 lines (564677 kB) scanned in 31.5 seconds (17925 kB/s)
1174528 tags added to tag file
longest tag line = 425

real	0m32.488s
user	0m29.822s
sys	0m2.534s
+ set +x
```

pragmaware:
> In the current logic (with the work around proposed here), _expand=1 makes
> parsing about 2 times slower:

I suggest running ctags through callgrind to find the culprit.

```
valgrind --tool=callgrid ctags ...
kcachegrind callgrind.out.<pid>
```

there might be a lot of opportunity for optimisation.

PR #2543, opened 2020-05-16
Cpreprocessor/C/C++/CUDA: infile macro expanding

masatake:
This one may be the biggest result of introducing the symbol table.
This pull request takes over the discussion and efforts in #2427.

Quoted from documents:
```rst
If a CPreProcessor macro defined in a C/C++/CUDA file, the macro invocation in the
SAME file can be expanded with following options:

.. code-block:: text

   --param-CPreProcessor:_expand=1
   --fields-C=+{macrodef}
   --fields-C++=+{macrodef}
   --fields-CUDA=+{macrodef}
   --fields=+{signature}

Let's see an example.

input.c:
.. code-block:: C

	#define DEFUN(NAME) int NAME (int x, int y)
	#define BEGIN {
	#define END }

	DEFUN(myfunc)
	  BEGIN
	  return -1
	  END

The output without options:
.. code-block::

   $ ctags -o - input.c
   BEGIN	input.c	/^#define BEGIN /;"	d	language:C	file:
   DEFUN	input.c	/^#define DEFUN(/;"	d	language:C	file:
   END	input.c	/^#define END /;"	d	language:C	file:

The output with options:
.. code-block::

   $ ctags --param-CPreProcessor:_expand=1 --fields-C=+'{macrodef}' --fields=+'{signature}' -o - input.c
   BEGIN	input.c	/^#define BEGIN /;"	d	language:C	file:	macrodef:{
   DEFUN	input.c	/^#define DEFUN(/;"	d	language:C	file:	signature:(NAME)	macrodef:int NAME (int x, int y)
   END	input.c	/^#define END /;"	d	language:C	file:	macrodef:}
   myfunc	input.c	/^DEFUN(myfunc)$/;"	f	language:C	typeref:typename:int	signature:(int x,int y)

``myfunc`` coded by ``DEFUN`` macro is captured well.


This feature is highly experimental. At least three limitations are known.

* This feature doesn't understand ``#undef`` yet.
  Once a macro is defined, its invocation is always expanded even
  after the parser sees ``#undef`` for the macro in the same input
  file.

* Macros are expanded incorrectly if the result of macro expansion
  includes the macro invocation again.

* Currently, ctags can expand a macro invocation only if its
  definitions are in the same input file. ctags cannot expand a macro
  defined in the header file included from the current input file.

Enabling this macro expansion feature makes the parsing speed about
two times slower.
```

### DIFF
--- a/Tmain/list-params.d/stdout-expected.txt
+++ b/Tmain/list-params.d/stdout-expected.txt
@@ -1,37 +1,37 @@
 # ALL
 #LANGUAGE      NAME   DESCRIPTION
-CPreProcessor  define define replacement for an identifier
+CPreProcessor  define define replacement for an identifier (name(params,...)=definition)
 CPreProcessor  if0    examine code within "#if 0" branch (true or [false])
 CPreProcessor  ignore a token to be specially handled
 Fypp           guest  parser run after Fypp parser parses the original input ("NONE" or a parser name [Fortran])
 
 # ALL MACHINABLE
 #LANGUAGE	NAME	DESCRIPTION
-CPreProcessor	define	define replacement for an identifier
+CPreProcessor	define	define replacement for an identifier (name(params,...)=definition)
 CPreProcessor	if0	examine code within "#if 0" branch (true or [false])
 CPreProcessor	ignore	a token to be specially handled
 Fypp	guest	parser run after Fypp parser parses the original input ("NONE" or a parser name [Fortran])
 
 # ALL MACHINABLE NOHEADER
-CPreProcessor	define	define replacement for an identifier
+CPreProcessor	define	define replacement for an identifier (name(params,...)=definition)
 CPreProcessor	if0	examine code within "#if 0" branch (true or [false])
 CPreProcessor	ignore	a token to be specially handled
 Fypp	guest	parser run after Fypp parser parses the original input ("NONE" or a parser name [Fortran])
 
 # CPP
 #NAME   DESCRIPTION
-define  define replacement for an identifier
+define  define replacement for an identifier (name(params,...)=definition)
 if0     examine code within "#if 0" branch (true or [false])
 ignore  a token to be specially handled
 
 # CPP MACHINABLE
 #NAME	DESCRIPTION
-define	define replacement for an identifier
+define	define replacement for an identifier (name(params,...)=definition)
 if0	examine code within "#if 0" branch (true or [false])
 ignore	a token to be specially handled
 
 # CPP MACHINABLE NOHEADER
-define	define replacement for an identifier
+define	define replacement for an identifier (name(params,...)=definition)
 if0	examine code within "#if 0" branch (true or [false])
 ignore	a token to be specially handled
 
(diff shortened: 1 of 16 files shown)


pragmaware:
I have given it a try. This is a great advancement!

However I think macro expansion should not depend on the macros being saved as symbols in the cork queue. It should be an independent feature: if we're expanding macros we directly save them in Cpp.fileMacroTable.

masatake:
> I have given it a try. This is a great advancement!

I'm always encouraged by your response and your code. In this case, the implementation of `-D` option drives me.

> However I think macro expansion should not depend on the macros being saved as symbols in the cork queue. It should be an independent feature: if we're expanding macros we directly save them in Cpp.fileMacroTable.

Thank you for taking the time to read my change.


Let me explain.
I have been trying to implement multi-pass/multi-input file parsing (MM parsing) for 3 years.
WIth MM parsing, ctags can use the knowledge getting from past parsing for the current parsing.
As you know, this brings us great benefits.
Expanding macro is one of them. However, type-resolving is another one.

In my design, the cork queue is essential for MM parsing.

To implement the MM parsing, I wondered what are the sub-goals.
1. we have to track the import/include/export/use directives.
    I found reference tags can be done for the purpose. 
2. we need an in-memory data store for managing the knowledge of parsing.
    I found the cork queue can be used for the purpose. The original cork queue was not enough; it can do only linear scanning. Faster searching by name was needed. Now we have a tree-based one.
Calling it as cork queue is incorrect technically. However, I will use "cork-queue" as the data-store.
3. we need multiple cork queues for managing the knowledge of past parsing.
   Currently, ctags has only one cork queue. The result of parsing a header file is stored in a
   cork queue. To refer to the result during parsing another file, the cork queue must be kept.
   "Barrel" is the place where cork queues are stored. I designed once in #1495.
    Consider parsing input.c:
    ```
    #include <foo.h>
    DEFUN(f, int, (int x, int y)) { return x + y; }
    ...
    ```
     C/CPreProcessor parsers may ask the Barrel to search DEFUN in foo.h and included files from foo.h.
      The original idea of the barrel is in 
      * https://github.com/universal-ctags/ctags/blob/6ea782371d7dc78b884a7882dd4dedd81a03f63c/docs/barrel.svg
      * https://github.com/universal-ctags/ctags/pull/1495/commits/6ea782371d7dc78b884a7882dd4dedd81a03f63c
4. we need the way to load a tag file into cork queues (and store them to Barrel).
libreadtags can be used. It is public domain software;  we can use it as we want:-P.

pragmaware:
I see your point. A correct multipass macro expansion requires analyzing the include chain. Not an easy job. I think that a lot of additional code and data structures will be needed. We also need to make sure that there is no major impact on the non-multipass mode of operation.

masatake:
> I see your point. A correct multipass macro expansion requires analyzing the include chain. Not an easy job. I think that a lot of additional code and data structures will be needed. We also need to make sure that there is no major impact on the non-multipass mode of operation.

I see. I keep in mind your comment about performance impact.

masatake:
TODO
* handle recursive expansion correctly
* utilize #undef

PR #2553, opened 2020-05-22
Add Kconfig parser

MaximeChretien:
Kconfig is a language used for build configurations used in a lot of
open source projects (ex: linux, u-boot, barebox, ...)

Reference:
https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html


cxw42:
This is part of bootlin/elixir#130, in case anyone wonders about the use case.

parsers/kconfig.c
```diff
@@ -0,0 +1,32 @@
+/*   This module contains functions for generating tags for the Kconfig language
```
masatake, line 1: We need copyright notice like:
```
/*
 *   Copyright (c) 2020, YOURNAME
 *
 *
 *   Reference
 *      https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html
...

**Comment by masatake** [comment]
Thank you for making this pull request.

Could you add "Kconfig" as a new parser to docs/news.rst ?

**Comment by MaximeChretien** [comment]
Hi,
Thanks for the review, I made the requested changes

**Comment by masatake** [comment]
I also had a Kconfig parser that is derrived from linux/scripts/tags.sh.
I have not merged my parser because tags.sh is just GPLv2; it is not  "or later version".

Your version is not based on the tags.sh script. So I can arrange my parser based on your parser.
Please, allow me to stack my parser on your pull request.

My version is not written in C. Both yours and mine just use regex pattern matching. So C is not needed.

**Comment by masatake** [comment]
I made the kconfig parser tags more items like menu.
Such tags may help vim build "tree" structured navigation interface.

If elixir doesn't need it, you can turn off from the command line.
The way for disabling is written in the ctags.1 man page.

**Comment by masatake** [comment]
@MaximeChretien, could you try my changes stacked on you changes?
If I introduced something you don't want, could you let me know?

**Comment by masatake** [comment]
"mainmenu" kind should be renamed to "mainMenu".

**Comment by MaximeChretien** [comment]
Hi, 
It looks good for me, with some command line options it will work fine for Elixir.
But maybe configPrefixed should be disabled by default ? Because when it's activated each element appears twice, one with CONFIG_ and the other without

**Comment by MaximeChretien** [comment]
And is there a way to have only the result with configPrefixed ?

**Comment by masatake** [comment]
> Hi,
It looks good for me, with some command line options it will work fine for Elixir.
But maybe configPrefixed should be disabled by default ? Because when it's activated each element appears twice, one with CONFIG_ and the other without

I can disable it. But... do you really want to disable it?
In scripts/tags.sh shipped as part of linux kernel, it is enabled by default.
As I wrote in the comment of kconfig.ctags, I developed this minor feature and it is really useful.

> And is there a way to have only the result with configPrefixed ?

ctags itself doesn't have a feature to do it. However, ``--fields=+{extras}`` (or ``--fields=+E``) option may help you.
Let me explain.

```
$ cat Kconfig.test
config FOO
	bool
$ ./ctags -o test.tags --fields=+'{extras}' Kconfig.test
$ cat test.tags 
!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/;"	extras:pseudo
!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/;"	extras:pseudo
!_TAG_OUTPUT_FILESEP	slash	/slash or backslash/;"	extras:pseudo
!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/;"	extras:pseudo
!_TAG_PATTERN_LENGTH_LIMIT	96	/0 for no limit/;"	extras:pseudo
!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//;"	extras:pseudo
!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/;"	extras:pseudo
!_TAG_PROGRAM_URL	https://ctags.io/	/official site/;"	extras:pseudo
!_TAG_PROGRAM_VERSION	0.0.0	/4ab20532/;"	extras:pseudo
CONFIG_FOO	Kconfig.test	/^config FOO$/;"	c	language:Kconfig	extras:configPrefixed
FOO	Kconfig.test	/^config FOO$/;"	c	language:Kconfig
```

The readtags command that is part of u-ctags, you can extract tags having the field.

```
$ readtags -Q '(and ($ "extras") (substr? ($ "extras") "onfigPrefixed"))' -t test.tags -ne -l
CONFIG_FOO	Kconfig.test	/^config FOO$/;"	kind:c	language:Kconfig	extras:configPrefixed
```

**Comment by MaximeChretien** [comment]
> I can disable it. But... do you really want to disable it?
> In scripts/tags.sh shipped as part of linux kernel, it is enabled by default.
> As I wrote in the comment of kconfig.ctags, I developed this minor feature and it is really useful.

If it's enabled by default in scripts/tags.sh, I guess it should be enabled here too, it's not that hard to disable it if needed.

> ctags itself doesn't have a feature to do it. However, `--fields=+{extras}` (or `--fields=+E`) option may help you.

Ok I will look at that, currently what we do in Elixir is parsing the output of ctags with `awk '{print "CONFIG_"$1" "$2" "$3}'` so we have the CONFIG_ added at the beginning. I think we will stay with that as we will still need to use awk for formatting. 
So for our needs I will need to disable the extra if it's enabled by default but it's not a problem ^^

**Comment by masatake** [comment]
Thank you for replying and the contribution.

**Comment by masatake** [comment]
BTW, Elixir can find "jiffies" for x86_64?
I guess it cannot.
If you turn on LdScript parser of u-ctags, Elixir may find it.

**Comment by MaximeChretien** [comment]
The problem with jiffies is that this 
https://elixir.bootlin.com/linux/latest/source/include/linux/jiffies.h#L81
is not identified as jiffies prototype and thus not listed in the definitions here :
https://elixir.bootlin.com/linux/latest/ident/jiffies

So I think there is a problem in the .h parsing 
And I can confirm jiffies is not returned by ctags when parsing include/linux/jiffies.h

**Comment by masatake** [comment]
--kinds-C++=+x is needed to capture them.

```
% u-ctags -o - --kinds-C++=+x  ~/var/linux/include/linux/jiffies.h | grep ^jif          
jiffies	/home/yamato/var/linux/include/linux/jiffies.h	/^extern unsigned long volatile __cacheline_aligned_in_smp __jiffy_arch_data jiffies;$/;"	x	language:C++	typeref:typename:unsigned long volatile __cacheline_aligned_in_smp __jiffy_arch_data
jiffies_64	/home/yamato/var/linux/include/linux/jiffies.h	/^extern u64 __cacheline_aligned_in_smp jiffies_64;$/;"	x	language:C++	typeref:typename:u64 __cacheline_aligned_in_smp
jiffies_delta_to_clock_t	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline clock_t jiffies_delta_to_clock_t(long delta)$/;"	f	language:C++	typeref:typename:clock_t
jiffies_delta_to_msecs	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline unsigned int jiffies_delta_to_msecs(long delta)$/;"	flanguage:C++	typeref:typename:unsigned int
jiffies_to_nsecs	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline u64 jiffies_to_nsecs(const unsigned long j)$/;"	f	language:C++	typeref:typename:u64
```

MaximeChretien:
> --kinds-C++=+x is needed to capture them.
> 
> ```
> % u-ctags -o - --kinds-C++=+x  ~/var/linux/include/linux/jiffies.h | grep ^jif          
> jiffies	/home/yamato/var/linux/include/linux/jiffies.h	/^extern unsigned long volatile __cacheline_aligned_in_smp __jiffy_arch_data jiffies;$/;"	x	language:C++	typeref:typename:unsigned long volatile __cacheline_aligned_in_smp __jiffy_arch_data
> jiffies_64	/home/yamato/var/linux/include/linux/jiffies.h	/^extern u64 __cacheline_aligned_in_smp jiffies_64;$/;"	x	language:C++	typeref:typename:u64 __cacheline_aligned_in_smp
> jiffies_delta_to_clock_t	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline clock_t jiffies_delta_to_clock_t(long delta)$/;"	f	language:C++	typeref:typename:clock_t
> jiffies_delta_to_msecs	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline unsigned int jiffies_delta_to_msecs(long delta)$/;"	flanguage:C++	typeref:typename:unsigned int
> jiffies_to_nsecs	/home/yamato/var/linux/include/linux/jiffies.h	/^static inline u64 jiffies_to_nsecs(const unsigned long j)$/;"	f	language:C++	typeref:typename:u64
> ```

Oh nice ! Thanks a lot !

PR #2555, opened 2020-05-23
Ruby: skip "public", "protected", and "private" when tagging methods

masatake:


masatake:
This pull request is inspred by https://chrisarcand.com/inline-scoped-ruby-methods-with-ctags/ .

PR #2575, opened 2020-06-14
Format --list-fields output as console code block

ddrone:
Currently the way output rendered on https://docs.ctags.io/en/latest/output-json.html is not particularly pretty; judging by GitHub preview this PR should change that.


masatake:
Thank you.

PR #2579, opened 2020-06-27
main: --roles-<LANG>.<KIND>, an option for enabling disabling roles

masatake:
Introduces `--roles-<LANG>.<KIND>=[+|-]roles` option.

There was no way to disable roles though all roles are enabled when defining.
With growing the number of roles, we may need the way to turn off (and on) the rules.


masatake:
TODO:

* align the form of `--_roledef-<LANG>=kind.role,desc` to that of `--roles-<LANG>.<KIND>`: `--_roledef-<LANG>.<KIND>=role,deck`.

* align the form of `--_param-<LANG>:param=value` to that of `--roles-<LANG>.<KIND>`: `--_param-<LANG>.param=value`.

* update optlibc2c to handle `--roles-<LANG>.<KIND>` option.

masatake:
The man page must be got reviewing.

Tmain/errors-in-options-roles.d/run.sh
```diff
@@ -0,0 +1,121 @@
+# Copyright: 2020 Masatake YAMATO
+# License: GPL-2
+CTAGS=$1
+
+CTAGS=$1
```
k-takata, line 5: Same as above line. Mistake?

masatake: Mistake. Thank you.

masatake: Fixed. thank you.

PR #2581, opened 2020-06-29
SQL: ignore "CREATE EXTENSION" syntax

### LINKED ISSUE #2580: Ignore CREATE EXTENSION syntax in SQL
Sometimes we use [`CREATE EXTENSION`](https://www.postgresql.org/docs/12/sql-createextension.html) to load some module functions in a SQL file, but the parser generates empty tags as long as a file contains this keyword.

This is postgres-specific and may not be a standard syntax. However, any SQL file could contain some dialect keywords related to its specific DB system specs. IMO a better solution would be to ignore parsing `CREATE EXTENSION` and continue parsing the rest instead of outputting no tags at all.

*****
```sh
ctags --options=NONE example.sql
```

The content of input file:

```SQL
CREATE EXTENSION IF NOT EXISTS "pgcrypto" WITH SCHEMA public;

CREATE TABLE public.foo (
    id bigserial PRIMARY KEY,
);
```

The tags output you are not satisfied with (**Empty when `CREATE EXTENSION` is present**):

```
!_TAG_PROGRAM_VERSION	0.0.0	/62f0144/
```

The tags output you expect (**Works when `CREATE EXTENSION` is removed**):

```
!_TAG_PROGRAM_VERSION	0.0.0	/62f0144/
foo	t.sql	/^CREATE TABLE public.foo ($/;"	t
id	t.sql	/^    id bigserial PRIMARY KEY,$/;"	E	table:foo
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(62f0144), Copyright (C) 2015 Universal Ctags Team
  Compiled: Jun 28 2020, 18:30:06
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +case-insensitive-filenames, +packcc
```

masatake:
Close #2580

Quoted from the issue submitted by @nobrick

    Sometimes we use CREATE EXTENSION to load some module functions in a
    SQL file, but the parser generates empty tags as long as a file
    contains this keyword.

    This is postgres-specific and may not be a standard syntax. However,
    any SQL file could contain some dialect keywords related to its
    specific DB system specs. IMO a better solution would be to ignore
    parsing CREATE EXTENSION and continue parsing the rest instead of
    outputting no tags at all.

This change makes the parser ignore the keyword.
However, capturing the name of extension as a reference tag is another
idea.


masatake:
Looks strange.

masatake:
I wonder whether we should capture the name of extension as a reference tag.
SQLite has "load_extesion" function that can be used in SQL.

PR #2587, opened 2020-07-07
buildsys: utilize GitHub action for testing with valgrind

masatake:


.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: install tools and libraries
+      run: sudo apt-get install valgrind pkg-config automake bash
```
masatake, line 17: more libraries should be here.

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: autogen.sh
+      run: ./autogen.sh
+    - name: configure
+      run: ./configure --enable-debugging
+    - name: make
+      run: make
+    - name: make units
+      run: make units VG=1
+    - name: make distcheck
```
masatake, line 26: distcheck should be done somewhere else.

masatake:
It takes ten minutes to run `make units VG=1`. Not so bad.

Issue #2589, opened 2020-07-08
Vim9: no tag is generated for ":def" functions


lacygoill:
**Describe the bug**

No tag is generated for Vim9 `:def` functions.

**To Reproduce**

Run these shell commands:

    mkdir -p /tmp/vim_plugin \
    && cd /tmp/vim_plugin \
    && cat <<'EOF' >foo.vim
        fu Legacy()
            echo 'legacy function'
        endfu
        def Vim9Function()
            echo 'Vim9 function'
        enddef
    EOF

    ctags --options=NONE -R .

Here is the contents of the generated tags file:

    !_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
    !_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
    !_TAG_OUTPUT_FILESEP	slash	/slash or backslash/
    !_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
    !_TAG_PATTERN_LENGTH_LIMIT	96	/0 for no limit/
    !_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
    !_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
    !_TAG_PROGRAM_URL	https://ctags.io/	/official site/
    !_TAG_PROGRAM_VERSION	0.0.0	/b4574f0a/
    Legacy	foo.vim	/^    fu Legacy()$/;"	f

It does not contain a tag for the `Vim9Function()`.

**Expected behavior**

I would expect this line to be present in the tags file:

    Vim9Function	foo.vim	/^    def Vim9Function()$/;"	f

**Environment**

 - ctags version:

       Universal Ctags 0.0.0(b4574f0a), Copyright (C) 2015 Universal Ctags Team
       Universal Ctags is derived from Exuberant Ctags.
       Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
         Compiled: Jul  8 2020, 09:30:25
         URL: https://ctags.io/
         Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml, +packcc

 - OS: Ubuntu 16.04.6 LTS

I built ctags locally after cloning the source code from the github repository, and making sure I was on the latest commit of the master branch.

**Additional context**

[...]

PR #2590, opened 2020-07-08
vim: support new ":def" functions of vim9

lacygoill:
This PR tries to add support for the new [`:def` functions](https://vimhelp.org/vim9.txt.html#:def) introduced in Vim9.  See issue #2589 for more information.

I briefly tested the patch, and it seems to work.  That is now `ctags(1)` correctly generates tags for `:def` functions.

But I don't know if it's really correct; I don't know C.

If it is correct, I think I could also add support for constants defined with the [`:const`](https://vimhelp.org/eval.txt.html#:const) keyword.

### DIFF
--- a/Units/parser-vim.r/vim9-def-function.d/expected.tags
+++ b/Units/parser-vim.r/vim9-def-function.d/expected.tags
@@ -1,0 +1,1 @@
+Func	input.vim	/^def Func()$/;"	f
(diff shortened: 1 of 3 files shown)


masatake:
@lacygoill, please look at https://github.com/universal-ctags/ctags/pull/2590/commits/1a44f70566389d0c5ef405eb99ba40a756b52ff3 .

A test case consists of two files: an input file and an expected output file.
You prepared only an input file. The expected output is so obvious. So I added it.

When adding code for `:const`, please include both files for new code to your commit.

masatake:
Thank you for your contribution. I'm waiting for the patch for `:const`:-)

lacygoill:
Thank you very much for merging the PR.

Issue #2494, opened 2020-04-06
[LOOKING FOR VOLUNTEERS] Language consultants/advisers 


masatake:
In #2076, we have been working on Cobol.
In #1581, we have been working on Vue.
In #2413, we have been working on SystemVerilog.

It is not possible for limited numbers of developers to support multiple languages.
Do you know languages are updated like C++11, C++17,  Java-1.8, Java-1.9,...?
I didn't recognize this horrible fact when I make the first pull request to fishman-ctags, the origin of universal-ctags.

Ideally, we have a maintainer for each parser. However, I know well that people can't live on developing one's  parser alone.
I always wait for your pull request. However, you can help developing a parser for your favourite language without writing code for ctags.

I'm looking for a person who gives us advices about specified programming languages and tags output for the language.

We have to fix or improve parsers without enough knowledges about the target language, its convention, and community. Even a thing is so obvious for persons who use the language is not obvious for us.
If you give me an advise, it is great help for us.

A typical question and advice: 
* https://github.com/universal-ctags/ctags/issues/2489#issuecomment-609083640
* https://github.com/universal-ctags/ctags/issues/2489#issuecomment-609502181

Without this advice, I cannot merge my pull request though my code fixes the original issue.

If you are interested in being "language adviser", please put your name here.
When we get a question about your language, we may contact with you.
We expect you give a response in a week or so.

We expect a language advise reads ctags(1) man page. It means you know

* what is "kind",
* what is "field",
* what is "extra",
* what is "definition tag", and
* what is "reference tag" (especially what I wrote in #2428).

In addition we expect you that
* you know the concept "subparser" (see http://docs.ctags.io/en/latest/running-multi-parsers.html ), 
* you understand the concept of open source software,
* you know Universal-ctags is distributed under the term of GPLv2 or later.

In addition I expect you that
* you are patient for my broken English,

If you are o.k., tell me:
* language(s) you know well
* frame work(s) or tool(s) written in the language you know well (if you have)
  e.g. Ruby-on-rails, moose (perl)... (see http://docs.ctags.io/en/latest/running-multi-parsers.html
  about the background of this question.)
* OS you are using
* a client tool you are using (e.g. vim+tagbar, emasc+xref, ...)
* skill of your English 
  If you set this field "Good", I will ask you to review the changes for our documents. See #1038.
  In some cases, the changes are nothing to do is the language you know well.
* whether you know C language or not?
[...]

PR #2593, opened 2020-07-08
vim: support ":const" command which sets constants

lacygoill:
This PR tries to add support for the Vim command [`:const`](https://vimhelp.org/eval.txt.html#:const).  The latter is used in Vimscript to set a constant.

---

I did not try to create a new C function to parse the command because:

   1. that's above my level
   2. I don't think it's necessary.  In Vim, `:const` is equivalent to `:let` followed by `:lockvar`; it's just a variable whose value is locked so that it can't change.

---

`:const` can be shortened into `:cons`, which is why I wrote `4` and not `5` here:

    else if (wordMatchLen (line, "let", 3) || wordMatchLen (line, "const", 4))
                                                                           ^

---

I added a simple test consisting of an input file and an expected output tags file.
I made sure all the Vim tests passed:

    make units LANGUAGES=Vim

### DIFF
--- a/parsers/vim.c
+++ b/parsers/vim.c
@@ -46,7 +46,8 @@ typedef enum {
 	K_FUNCTION,
 	K_MAP,
 	K_VARIABLE,
-	K_FILENAME
+	K_FILENAME,
+	K_CONST,
 } vimKind;
 
 static kindDefinition VimKinds [] = {
@@ -56,6 +57,7 @@ static kindDefinition VimKinds [] = {
 	{ true,  'm', "map",      "maps" },
 	{ true,  'v', "variable", "variable definitions" },
 	{ true,  'n', "filename", "vimball filename" },
+	{ true,  'C', "constant", "constant definitions" },
 };
 
 /*
@@ -432,7 +434,7 @@ cleanUp:
 	return cmdProcessed;
 }
 
-static void parseLet (const unsigned char *line, int infunction)
+static void parseVariableOrConstant (const unsigned char *line, int infunction, int kindIndex)
 {
 	vString *name = vStringNew ();
 
@@ -482,7 +484,7 @@ static void parseLet (const unsigned char *line, int infunction)
 			vStringPut (name, (int) *cp);
 			++cp;
 		} while (isalnum ((int) *cp) || *cp == '_' || *cp == '#' || *cp == ':' || *cp == '$');
-		makeSimpleTag (name, K_VARIABLE);
+		makeSimpleTag (name, kindIndex);
 		vStringClear (name);
 	}
 
@@ -596,7 +598,11 @@ static bool parseVimLine (const unsigned char *line, int infunction)
 
 	else if (wordMatchLen (line, "let", 3))
 	{
-		parseLet (skipWord (line), infunction);
+		parseVariableOrConstant (skipWord (line), infunction, K_VARIABLE);
+	}
+	else if (wordMatchLen (line, "const", 4))
+	{
+		parseVariableOrConstant (skipWord (line), infunction, K_CONST);
 	}
 
 	return readNextLine;
--- a/Units/parser-vim.r/vim-const.d/expected.tags
+++ b/Units/parser-vim.r/vim-const.d/expected.tags
@@ -1,0 +1,1 @@
+s:CONSTANT	input.vim	/^const s:CONSTANT = 42$/;"	C
--- a/Units/parser-vim.r/vim-const.d/input.vim
+++ b/Units/parser-vim.r/vim-const.d/input.vim
@@ -1,0 +1,1 @@
+const s:CONSTANT = 42


masatake:
> that's above my level

I will asisst you. So the level is not a matter for us.

We know ctags. You know vim script. I know C. Your knowledge about vim script is very important.
See also https://github.com/universal-ctags/ctags/issues/2494.

You reuse the code for handling "let". However, as far as my short study, let is for defining variables.
I think tagging a language object defined with `:const` as a variable is not good idea.
Instead, I think it should be tagged as a constant.

```
const s:CONSTANT = 42
```
=>
```
s:CONSTANT	input.vim	/^const s:CONSTANT = 42$/;"	v
```

Here `v` means a variable. I don't think `s:CONSTANT` is not a vairable.
How do you think?

masatake:
I think ideal behavior is:
```
[jet@living]~/var/ctags% cat input.vim
cat input.vim
const s:CONSTANT = 42
[jet@living]~/var/ctags% ./ctags --list-kinds=Vim
./ctags --list-kinds=Vim
a  autocommand groups
c  user-defined commands
f  function definitions
m  maps
v  variable definitions
n  vimball filename
C  constant definitions
[jet@living]~/var/ctags% ./ctags -o - input.vim
./ctags -o - input.vim
s:CONSTANT	input.vim	/^const s:CONSTANT = 42$/;"	C
```

The change we need is for above result:
```
diff --git a/parsers/vim.c b/parsers/vim.c
index 0bdd1802..7ee767b6 100644
--- a/parsers/vim.c
+++ b/parsers/vim.c
@@ -46,7 +46,8 @@ typedef enum {
 	K_FUNCTION,
 	K_MAP,
 	K_VARIABLE,
-	K_FILENAME
+	K_FILENAME,
+	K_CONST,
 } vimKind;
 
 static kindDefinition VimKinds [] = {
@@ -56,6 +57,7 @@ static kindDefinition VimKinds [] = {
 	{ true,  'm', "map",      "maps" },
 	{ true,  'v', "variable", "variable definitions" },
 	{ true,  'n', "filename", "vimball filename" },
+	{ true,  'C', "constant", "constant definitions" },
 };
 
 /*
@@ -432,7 +434,7 @@ cleanUp:
 	return cmdProcessed;
 }
 
-static void parseLet (const unsigned char *line, int infunction)
+static void parseVariableOrConstant (const unsigned char *line, int infunction, int kindIndex)
 {
 	vString *name = vStringNew ();
 
@@ -482,7 +484,7 @@ static void parseLet (const unsigned char *line, int infunction)
 			vStringPut (name, (int) *cp);
 			++cp;
 		} while (isalnum ((int) *cp) || *cp == '_' || *cp == '#' || *cp == ':' || *cp == '$');
-		makeSimpleTag (name, K_VARIABLE);
+		makeSimpleTag (name, kindIndex);
 		vStringClear (name);
 	}
 
@@ -596,9 +598,12 @@ static bool parseVimLine (const unsigned char *line, int infunction)
 
 	else if (wordMatchLen (line, "let", 3))
 	{
-		parseLet (skipWord (line), infunction);
+		parseVariableOrConstant (skipWord (line), infunction, K_VARIABLE);
+	}
+	else if (wordMatchLen (line, "const", 4))
+	{
+		parseVariableOrConstant (skipWord (line), infunction, K_CONST);
 	}
-
 	return readNextLine;
 }
```

However, I don't know well about vim script. So, I would like to hear your comments.

lacygoill:
I agree with you; technically, a constant is not a variable since its value can't change.
It should probably have its own kind.  `c` is already taken for commands, so `C` is the next best choice.

I've applied your patch, compiled and installed.  All the Vim tests still pass, and the `ctags(1)` binary correctly generates tags for constants defined with the `:const` keyword.  I haven't had the time to read all of https://github.com/universal-ctags/ctags/issues/2494, but if I understand correctly, the purpose is to give advice on Vimscript.  I'm no expert, but I think I can help.  If you have any question regarding the Vimscript language, let me know.

lacygoill:
I made a mistake in the `expected.tags` file; I need to fix it.

parsers/vim.c
```diff
@@ -594,11 +596,14 @@ static bool parseVimLine (const unsigned char *line, int infunction)
 		parseAutogroup (skipWord (line));
 	}
 
-	else if (wordMatchLen (line, "let", 3))
+	else if (wordMatchLen (line, "let", 3) || wordMatchLen (line, "const", 4))
```
masatake, line 599: Mathcing "const" is not needed here.

parsers/vim.c
```diff
@@ -594,11 +596,14 @@ static bool parseVimLine (const unsigned char *line, int infunction)
...
 	{
-		parseLet (skipWord (line), infunction);
+		parseVariableOrConstant (skipWord (line), infunction, K_VARIABLE);
+	}
+	else if (wordMatchLen (line, "const", 4))
+	{
+		parseVariableOrConstant (skipWord (line), infunction, K_CONST);
 	}
-
```
masatake, line 601: Removng this empty line is not needed to support "const".

lacygoill: I think this is fixed; I added back the empty line.

parsers/vim.c
lacygoill: You are right. I forgot to revert my old patch, sorry.

masatake:
Thank you.

PR #2597, opened 2020-07-15
buildsys: fix a couple english spelling errors in configure.ac

ploxiln:
introduced in 1293d03b1a38
(rbtree: introduce new ifdef condition HAVE_STATEMENT_EXPRESSION_EXT)


masatake:
Thank you!

PR #2612, opened 2020-08-08
Rust: allow arrays in function parameters

simias:
I noticed that ctags wouldn't generate a function entry for a signature like:

```rust
fn foo(arr: [u32; 3]) {}
```

The problem was that the parser stopped as soon as a ';' was encountered, probably assuming that it would be a statement delimiter and not an array type.

The fix is to keep track of brackets and ignore ';' if we have at least one opened bracket.

While I was at it I also reject the signature if the brackets aren't balanced, which means that the parser is now a little more strict than it was before and a (broken) function like this one will now be rejected:

```rust
fn bad(p: &[u32) {}
```

### DIFF
--- a/Units/parser-rust.r/rust-test_input.d/expected.tags
+++ b/Units/parser-rust.r/rust-test_input.d/expected.tags
@@ -24,6 +24,7 @@ a_anteater	input.rs	/^	a_anteater(isize),$/;"	e	enum:Animal
 a_bear	input.rs	/^	a_bear(isize),$/;"	e	enum:Animal
 a_cat	input.rs	/^	a_cat(isize),$/;"	e	enum:Animal
 a_dog	input.rs	/^	a_dog(isize),$/;"	e	enum:Animal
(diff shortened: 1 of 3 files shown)


masatake:
The change looks fine to me.
Could you add a test case for this?
See https://docs.ctags.io/en/latest/units.html to know how to write a test case.
See also files under Units/parser-rust.r.
See also commit 4ab2053259892adc2c807e60303726d5f2b56708 as an example.
You don't have to add Makefile.am, Makefile.in and Makefile. They are generated by autogen.sh and configure.

Thank you.

simias:
Hello @masatake and thank you for the feedback.

The 2nd commit adds a test for array parameters the existing rust-test_input test: https://github.com/universal-ctags/ctags/pull/2612/commits/63002655a6ede406bf19ab04d770dbfc9a02600f

Do you want me to create a new test specifically for this change?

simias:
I just squashed the two commits for simplicity.

masatake:
Thank you.

PR #2614, opened 2020-08-10
main: avoid segmentation fault triggered by passing a name of unreadable source file

### LINKED ISSUE #2609: Hope to add protection when to parse files
When I use "ctags.exe -L C:\ctags-2018-12-26_5728abe4-x64\fileList.files --output-format=json --fields=+ne" to parse files in fileList.files，if program has no permission to open aaa.c, the whole parse will be interrupted. bbb.c can't get to be parsed.
I hope program can add some excpetion protections so that when some exceptions occur, other files still be parsed.

fileList.files content:
D:\Projects\aaa.c
D:\Projects\bbb.c

masatake:
When opening an input file is failed, File.source.tagPath is not set.
As the result, when accessing File.source.tagPath via getSourceFileTagPath(),
ctags crashes.

This change avoids the issue by not calling getSourceFileTagPath()
when opening the input file falure.

Close #2609.


main/parse.c
```diff
@@ -3890,15 +3890,21 @@ extern bool runParserInNarrowedInputStream (const langType language,
...
 
 	Assert (0 <= language  &&  language < (int) LanguageCount);
 
 	if (!openInputFile (fileName, language, mio))
+	{
+		if (failureInOpenning)
+			*failureInOpenning = true;
 		return false;
+	}
+	if (*failureInOpenning)
```
CognDiss, line 3906: Shouldn't this be 'if (failureInOpenning)'  instead of 'if (*fail...)' ?

masatake: You are correct. Thank you very much!

PR #2636, opened 2020-09-06
SystemVerilog: typed parameter support

hirooih:
A typed parameter was reported as register-kind instead of constant-kind.

For example,

```systemverilog
localparam int MY_WIDTH                  = 1;
```

MY_WIDTH was reported as register-kind.  On the line 1200, `token->kind` was overwritten K_REGISTER of `int`.
This fix the issue not to overwrite `token->kind` when it is K_CONSTANT.

### DIFF
--- a/Units/parser-verilog.r/systemverilog-2d-array.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-2d-array.d/expected.tags
@@ -1,5 +1,5 @@
 my_pack	input.sv	/^package my_pack;$/;"	K
-MY_WIDTH	input.sv	/^    localparam int MY_WIDTH                  = 1;$/;"	r	package:my_pack
-MY_DEPTH	input.sv	/^    localparam int MY_DEPTH                  = 2;$/;"	r	package:my_pack
+MY_WIDTH	input.sv	/^    localparam int MY_WIDTH                  = 1;$/;"	c	package:my_pack
+MY_DEPTH	input.sv	/^    localparam int MY_DEPTH                  = 2;$/;"	c	package:my_pack
 my_t	input.sv	/^    typedef logic [MY_DEPTH-1:0][MY_WIDTH-1:0]  my_t;$/;"	T	package:my_pack
 my_t2	input.sv	/^    typedef logic   [MY_DEPTH-1:0] 	 [MY_WIDTH-1:0] 	 my_t2;$/;"	T	package:my_pack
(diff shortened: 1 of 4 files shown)


hirooih:
The coverage decreased (-0.002%), because the line 1202 became uncovered.  It was covered by a wrong test case Units/parser-verilog.r/systemverilog-2d-array.d fixed by this pull-request.

masatake (changes requested):
When checking out any one of all commits of the ctags whole history, "make units" should be passed. (ideally)
However, when I merge this PR and checkout e886c2f, "make units" may fail because 33dfb1c is not in the commit.
Therefore e886c2f and 33dfb1c are squashed into one commit. So "make units" never fail.

The test case name, systemverilog-github2636.d/, is not good name.
If possible, I would like you to change it more meaningful name.

With the change, the block in the condition
```
		token->kind = localKind;
```
never be executed during "make units". If possible, I would like you to add a test case that covers the block.

hirooih:
Hi, 

> When checking out any one of all commits of the ctags whole history, "make units" should be passed. (ideally)
> ...

Let me excuse first :-)

I thought as follows;

- There was systemverilog-github646.d, and verilog-sf_bug*.d.
- If I changed existing test cases, it makes difficult to prove the fix does not break existing test cases.
  - Only if an expected value is wrong and it is changed by the fix, the test can be modified. 

So

- I had to make systemverilog-github2636.d.
- I had to make pull-request to get the ticket number #2636.
- I made a draft pull-request and got #2636.
- add systemverilog-github2636.d
- make a pull-request

***

> If possible, I would like you to add a test case that covers the block.

If I understand the current code correctly, this if-block can be removed.

```systemverilog
				if (token->kind != K_PORT && token->kind != K_CONSTANT && localKind != K_IGNORE)
				{
					token->kind = localKind;
				}
```

The if-block has been existing since long time ago (at least 4 years).  At that time qualifiers,  automatic, local, and static, are included in the KeywordTable[].  The if-block existed for them.  You removed them after that.
The line in the if-block has been covered only by the bug I fixed in this issue.
But I may be missing anything...

masatake:
You don't have to rename existing test case.
I'm talking about the new test case. If you can find more meaningful name, please rename it.
If you cannot find, the current name is acceptable.

Abouf if-block, I see the situation. I expect I receive more pull request later:-).

masatake:
Allow me to take over this pull request.

masatake:
(The commit id of the last commit is 0d8c47c).

masatake:
> If I changed existing test cases, it makes difficult to prove the fix does not break existing test cases.
>
>    Only if an expected value is wrong and it is changed by the fix, the test can be modified.

Understandable. However, I would like you to squash it in the final version of the pull request.
Making the test case separated from the associated code change is o.k. in your development process and in draft versions of pull requests, of course. Keeping our public repository clean(== git-bisect-able)  is more important.

masatake:
It seems that there is trouble in azure for getting deb packages.

masatake:
@hirooih, thank you for your contribution.

PR #2645, opened 2020-09-14
gitignore: add ctags-lang-python generated manpage files

ploxiln:
forgotten when adding the new ctags-lang-python man page recently


masatake:
Thank you.

Issue #1488, opened 2017-06-26
SystemVerilog: variables/class members of custom type not captured


RadekRR:
Hi,
in a simple example test.sv:

class test_class;
        int int_memb;
        t_user t_user_memb;
endclass

ctags output:
$ ./ctags -f - test.sv
int_memb        test.sv /^      int int_memb;$/;"       r       class:test_class
test_class      test.sv /^class test_class;$/;" C

so the t_user_memb was not captured

masatake:
I don't know SystemVerilog. Could you tell me which kind `t_user_memb` should be in the following kinds:

```
SystemVerilog
    c  constants (define, parameter, specparam)
    e  events
    f  functions
    m  modules
    n  net data types
    p  ports
    r  register data types
    t  tasks
    b  blocks
    A  assertions
    C  classes
    V  covergroups
    I  interfaces
    M  modports
    K  packages
    P  programs
    Q  prototypes [off]
    R  properties
    T  type declarations
```

Or do we have to define new kind for `t_user_memb` ?

RadekRR:
The register and net kinds are inherited from Verilog language, and inside the scope of the class declaration they aren't suitable I guess. 
I'd say that class members should get a new "m" kind. The system types such as "int" should also be tagged as "m" when declared as a member.
For variables declared inside functions and tasks, which are local, the "l" kind should be used.

masatake:
I would like to veriy my understanding.

You mean
1. we should introduce new kind called "member",
2. in the example both  `int_memb` and `t_user_memb` should be marked with "member" kind.

Do I understand what you wrote correctly?

You wrote 'm'. However, I guess you don't mean "module". I guess you mean 'm' for "member".
The bad new is that 'm', 'M', and 'f(field)' are already used for the other kinds. 'F' is reserved in ctags
global. We have to use another letter.

RadekRR:
> You mean
> 
>     we should introduce new kind called "member",
>     in the example both int_memb and t_user_memb should be marked with "member" kind.
> 
Yes, exactly.

> You wrote 'm'. However, I guess you don't mean "module". I guess you mean 'm' for "member".
>
Yes you are right, 'm' is already used as 'module', we don't want to use it in this case. 

> We have to use another letter.
>
Yes.

masatake:
@RadekRR, thank you.
[...]

Issue #2413, opened 2020-02-04
[systemverilog] parser mistakes typedef for a module's port


antoinemadec:
The name of the parser: **verilog.c**

The command line you used to run ctags:

```
$ ctags --options=NONE foo.sv
```

The content of input file:
**foo.sv**
```systemverilog
typedef bit[31:0] int32_t;
module mod(
  input bit clk,
  input int32_t a
);
endmodule
```

The tags output you are not satisfied with:

```
clk     foo.sv  /^  input bit clk,$/;"  p       module:mod
int32_t foo.sv  /^  input int32_t a$/;" p       module:mod
int32_t foo.sv  /^typedef bit[31:0] int32_t;$/;"        T
mod     foo.sv  /^module mod($/;"       m
mod.clk foo.sv  /^  input bit clk,$/;"  p       module:mod
mod.int32_t     foo.sv  /^  input int32_t a$/;" p       module:mod
```

The tags output you expect:
```
clk     foo.sv  /^  input bit clk,$/;"  p       module:mod
a       foo.sv  /^  input int32_t a$/;" p       module:mod
int32_t foo.sv  /^typedef bit[31:0] int32_t;$/;"        T
mod     foo.sv  /^module mod($/;"       m
mod.clk foo.sv  /^  input bit clk,$/;"  p       module:mod
mod.a   foo.sv  /^  input int32_t a$/;" p       module:mod
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(81fa5b1a), Copyright (C) 2015 Universal Ctags Team                    
  Universal Ctags is derived from Exuberant Ctags.                                            
  Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert                                 
    Compiled: Feb  4 2020, 14:36:50                                                           
    URL: https://ctags.io/                                                                    
    Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, 
  +interactive, +sandbox, +yaml, +packcc
```

masatake:
Thank you for reporting.
Could you try #2492 ?

antoinemadec:
#2492 fixes my original file.

However, it still fails if I do: ```ctags  --options=NONE a.sv b.sv```

Where:
**a.sv**
```systemverilog
`include "b.sv"

module mod(
  input bit clk,
  input int32_t a
);
endmodule
```

**b.sv**
```systemverilog
typedef bit[31:0] int32_t;
```

[...]

Issue #2480, opened 2020-03-31
Ruby: minor improvements


masatake:
Eventually, we may have to rewrite ruby parser with a completely token oriented way to do more interesting things like stacking subparser for Ruby on Rails.

HOWEVER, there are some areas we can improve the current line-oriented parser.

Easy ones:
- [x] capture accessors defined with attr_accessor
      "attrMethod" language speific extra can be added.
      See #408.
- [ ] introduce script kind for capturing 'require "foo"'.
- [x] capture CONSTANT (is it possible in the current implementation?)
- [x] handle mixin (#2476).
- [ ] ~capture "class << HTTP" with an anonymous tag.~
- [x] end field for classes, modules, and method.
- [ ] alias_method ( #408, #2495)
- [ ] alias ( #408, #2495 )
      ripper-tags defines kind for alias/a.
- [ ] capture autoloaded module as reference tags

Hard ones:
- [ ] signature field for methods (is it possible in the current implementation?)
- [ ] ignore here-doc area ( #454 )

Unknown:
- [ ] capture Class.new with anonymous tag (#2408 )
- [ ] subparser for rails ( #408 )

NOTE:
* Please, read ctags(1) and understand the fundamental concepts in ctags including the definition tags, reference tags, kinds, fields, extras, and pseudo-tags before adding something.
* if possible, I would you to read parsers/ruby.c before adding something.
* @masatake will update, delete or edit your comments for keeping this as our TODO list.

colinlieberman:
Just wanted to +1 this - the issue with signatures including string literals is particularly problematic with rubocop eg having to change

```
class MyClass # rubocop:disable Metrics/ClassLength
...
```

to 

```
# rubocop:disable Metrics/ClassLength
class MyClass
  ...
end
# rubocop:enable Metrics/ClassLength
```

and yes, obviously there's another solution for this, but for utility classes that are eg for looking up a number of items from a large dataset, that's the reason the disable options exist

masatake:
@colinlieberman, please open a new issue for the topic.
I would like you to show an example input. I expect the input is not so large.
Please, remember that I don't know ruby well.

colinlieberman:
> @colinlieberman, please open a new issue for the topic.
> I would like you to show an example input. I expect the input is not so large.
> Please, remember that I don't know ruby well.

Thanks, @masatake - i'm trying to put together a good example for you, but not being able to duplicate the issue. 

[...]

PR #2653, opened 2020-09-27
SystemVerilog: user defined type support

hirooih:
@masatake san,
I've working on #1488 and #2413.  It is not completed yet, but let me share the current status with you.

First I checked uncovered lines in verilog.c and add some test cases.
After that I made several refactoring changes.
Until now I made changes which did not affect on the output.  Next I will work on variable/net type checking routines.

In addition to the Unit Tests I've been using source code of [UVM 1.2](https://www.accellera.org/downloads/standards/uvm) and [an opensource RISC CPU](https://github.com/syntacore/scr1) to ensure output was not changed.

This is a draft pull request.  Your feedback and advice are welcome.
Thank you.


parsers/verilog.c
```diff
@@ -46,9 +46,13 @@
  * If you will separate the definitions for the parsers, you must revise the
  * code related to the symbol table. */
 typedef enum {
-	K_IGNORE = -2,
-	K_UNDEFINED,
-	K_CONSTANT,
+	/* parser private items */
+	K_IGNORE = -16,
+	K_DEFINE,
```
masatake, line 51: s it the best way to tag "foo" of "`define foo ..." with constant kind?
Introducing "d/definition" kind and tagging the "foo" with the definition kind is an alternative way.

Using the unified kind "constant" in many areas reduces the information that client tools get from a tags file.
In my experience, it is better to tag different things with different kinds. If ctags hides the differences, and just reports various language objects as "constant," in some cases, a client tool must parse the raw .sv source file for distinguishing whether a given tag is "`define"'ed or not.

IMHO, ctags should not be too smart. Instead, ctags should be stupid; it should report what it sees in source input files as-is to client tools that will do something smart.

masatake: Sorry, what I wrote here is pointless; introducing the d/definition kind breaks compatibility.
So introducing the kind is not so easy.

masatake:
I reviewed up to    69e311f.

parsers/verilog.c
```diff
@@ -221,6 +225,8 @@ static tokenInfo *tagContents = NULL;
  *   FUNCTION DEFINITIONS
  */
 
+static void updateKind (tokenInfo *const token);
```
masatake, line 306: A minor issue.
`updateKind` here is not a function definition. So I would like you to introduce `FUNCTION PROTOTYPES` section.
See the following example taken from c.c:
```C

/*
*   FUNCTION PROTOTYPES
*/
static void createTags (const unsigned int nestLevel, statementInfo *const parent);

/*
*   FUNCTION DEFINITIONS
*/

/*
*   Token management
*/

static void initToken (tokenInfo* const token)
{
	token->type			= TOKEN_NONE;
```

parsers/verilog.c
```diff
@@ -1306,11 +1316,13 @@ static void findTag (tokenInfo *const token)
...
 	verbose ("Checking token %s of kind %d\n", vStringValue (token->name), token->kind);
 
-	if (currentContext->kind != K_UNDEFINED)
+	if (currentContext->kind != K_UNDEFINED && (token->kind == K_END || token->kind == K_END_DE))
 	{
 		/* Drop context, but only if an end token is found */
 		dropEndContext (token);
 	}
+	if (token->kind == K_END_DE)
+		return;
```
masatake, line 1465: My understanding is that this line is called only for broken (invalid syntax) input.
In that case, putting verbose() here may help us in the future.

masatake:
I reviewed up to "SystemVerilog: mark fork as K_BEGIN and join* as K_END to create a co…" (bc88a0c).

parsers/verilog.c
```diff
@@ -386,18 +453,23 @@ static void buildKeywordHash (const langType language, unsigned int idx)
 		if (p->isValid [idx])
 			addKeyword (p->keyword, language, (int) p->kind);
 	}
+	/* mark unspecified keywords as K_IGNORE */
+	for (i = 0  ;  i < size  ;  ++i) {
+		if (lookupKeyword(keywords[i], language) == KEYWORD_NONE)
```
masatake, line 452: I wonder why this `lookupKeyword` is needed at runtime?
If you want to detect adding the same keyword twice, Assert can be used.
Assert is enabled when you run ./configure with --enable-debugging.
On Travis-CI, ctags is built with the option.

masatake: I withdraw the above comment.  I understand the intention.

parsers/verilog.c
```diff
@@ -582,6 +582,7 @@ static void skipToSemiColon (void)
 	} while (c != EOF && c != ';');
 }
 
+/* read an identifier, keyword, number, compiler directive, or macro identifier */
 static bool readIdentifier (tokenInfo *const token, int c)
```
masatake, line 586: How do you think renaming this function to "readToken"?

parsers/verilog.c
```diff
@@ -599,6 +600,32 @@ static bool readIdentifier (tokenInfo *const token, int c)
 	return (bool)(vStringLength (token->name) > 0);
 }
 
+/* check if an ideitifier:
+ *   simple_identifier ::= [ a-zA-Z_ ] { [ a-zA-Z0-9_$ ] } */
+static bool isIdentifier (tokenInfo* token)
+{
+	if (token->kind == K_UNDEFINED)
+	{
+		for (int i = 0; i < token->name->length; i++)
```
masatake, line 609: You can use `vStringLength (token->name)` instead of `token->name->length`..

parsers/verilog.c
```diff
@@ -599,6 +600,32 @@ static bool readIdentifier (tokenInfo *const token, int c)
...
+		{
+			int c = token->name->buffer[i];
```
masatake, line 611: vStringChar (token->name, i) can be used.

parsers/verilog.c
```diff
@@ -1486,7 +1513,12 @@ static void findTag (tokenInfo *const token)
 	{
 		tagNameList (token, skipWhite (vGetc ()));
 	}
-	else if (token->kind != K_UNDEFINED && token->kind != K_IGNORE)
+	else if (token->kind == K_UNDEFINED)
+	{
+		if (isIdentifier(token))	/* user defined type */
+			tagNameList (token, skipWhite (vGetc ()));
+	}
+	else
```
masatake, line 1521: Comparing with the original code, the condition `(token->kind != K_IGNORE)` is gone.
Is this o.k.?

masatake: I withdraw the above comment. The condition is already handled in the new code. Ignore this.

masatake:
Mostly it looks good to me.
I pointed out some minor issues as a result of the review.

masatake:
BTW, you can make an issue for tracking SystemVerilog issued as I did for Ruby in #2480.

hirooih:
@masatake san,

Thank you for your review.
I'll fix the issues you pointed.

Which way do you prefer?

1. make one commit for these issue.  You can see the difference from code you already reviewed.
2. squash each fix to the original commit.  The change may not be clear for you.
3. make each commit.  Squash them after you reviewed.

I prefer 1 or 2.
1 is easiest.   2 and 3  are almost same for me.  This time changes are so trivial that you don't have to review seriously.  2 may be enough.
Anyway I will follow your way.

---

Let me ask you  some questions.

Q1.  "`//`" type comment vs "`/* */`" type comment
Many other parsers are using "`//`" style comment, but I found verilog.c has been using only "`/* */`" style..
I already use it in my local commits.  Should I change them to "`/* */`" style comment?

Q2. nested if-else vs switch case in findTag()
As you see, the nested if-else in findTag() can be replaced with a switch case.  I prefer switch case because it gives readers important information that all condition depend on one variable (token->kind).  How do you think?

---

Supporting user-defined type is almost working well on my working directory.  I am adding some more tests and make some refactoring.   I hope I can commit it this weekend.

parsers/verilog.c
hirooih: You are right.
And I also agree with you that `define should have its own kind if we are allowed some incompatibility.

parsers/verilog.c
hirooih: This line is called for correct input, endfunction, endmodule, endtask, ..., etc.

***

Your comment reminds me one more question.
I think we should call notice() instead of verbose() for broken input.
And every parser should use same format something like this [this](https://www.gnu.org/prep/standards/standards.html#Errors).

How do you think?

parsers/verilog.c
hirooih: I have been thinking exactly same thing.
I will rename them after this issue is complete.

parsers/verilog.c
masatake: I think notice() is too noisy. Furthermore, verbose is also noise for the purpose. Remember ctags will deal with large codebases.



Printing such information has two purposes.

* reducing pointless bug reports
  We know the motto, "if the input is broken, ctags may emit garbages". However, users may not know.
  As the result, we have chances to receive bug reports of ctags to that broken input is given.
  The notice can notify that broken input is given to ctags.

* debugging
  This is for us, not for users. For this purpose, verbose() and notice() are not suitable.
   We can use macros in trace.h. See parsers/fypp.c.
   The macros can be used only with ctags built with ./configure --enable-debugging.
    --_trace=<LANG> must be passed to ctags at runtime.

masatake:
> 2. squash each fix to the original commit. The change may not be clear for you.

The original commits are readable. They are short enough so I can understand. Please, keep the way.
However, if you can reflect my review comment to the commit that I mentioned to easily, update the commit.
As I wrote before, git rebase -i may help you.

The original commits convince me you are more than a contributor.
About coding style, do as you want. 

About indent-style, see .editorconfig. About C syntax, in most of the areas, we choose stdc99.
Many different people have joined the development of ctags (and left), so the style of coding is not unified.
Ideally, they should be unified. However, I have kept it as is. If a parser has no active maintainer, I adjust the
coding style of the parser as I like. If a parser has an active maintainer,  generally, I say nothing; the maintainer
should do as one wants. And, finally, the SystemVerilog parser gets a new maintainer.

If you get a question about coding, asking me is one of the ways.  Inspecting other parsers is an alternative way.

hirooih:
> 2. squash each fix to the original commit. The change may not be clear for you.

I've done this.
The moving the prototype definition is squashed into 7bbbffb.
vString*() fixes were done by 'git --amend'.

I will do other additional refactoring after fixes which I am working will be done.

> About coding style, do as you want.

OK.  I'll do my best.

masatake:
I sent you an invitation to the u-ctags org.  Please, merge your changes for yourself after accepting it.
There are some methods for merging in GitHub web UI. If there is no special requirement, I choose the "Merge pull request" method.

masatake (approved):

hirooih:
Finally I've committed the fix for user defined type support.
a54e815 is the main fix.

By supporting user defined type correctly some expected.tags must be updated.
I also add some test cases.

masatake:
I found a very minor issue.
About a54e815, 
`'` in the second line of the commit log is not closed:
```
is followed by ',', ':', or ').
-----------------------------^
```

parsers/verilog.c
```diff
@@ -1224,49 +1230,79 @@ static void processTypedef (tokenInfo *const token)
...
+					c =skipDimension(c);
+					vUngetc (c);
+				}
 			} while (c != ')' && c != EOF);
 			c = skipWhite (vGetc ());
-			parameters = popToken (parameters);
+			deleteToken (token);
 		}
 	}
+	vUngetc (c);
```
masatake, line 1284: if c is not '#', the memory object associated with token leaks.

1. install valgrind
2. git checkout a54e815ca8dbfcd96e0eae972191e74e6aa684a2
3. make
4. make units LANGUAGES=SystemVerilog VG=1
5. cat Units/parser-verilog.r/systemverilog-class.d/VALGRIND.tmp 
```
==20931== Memcheck, a memory error detector
==20931== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==20931== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==20931== Command: ./ctags --verbose --options=NONE --optlib-dir=+./Units/parser-verilog.r/systemverilog-class.d/optlib -o - --options=./Units/parser-verilog.r/systemverilog-class.d/args.ctags ./Units/parser-verilog.r/systemverilog-class.d/input.sv
==20931== Parent PID: 19254
==20931== 
==20931== 
==20931== HEAP SUMMARY:
==20931==     in use at exit: 768 bytes in 21 blocks
==20931==   total heap usage: 6,777 allocs, 6,756 frees, 360,392 bytes allocated
==20931== 
==20931== 768 (264 direct, 504 indirect) bytes in 3 blocks are definitely lost in loss record 7 of 7
==20931==    at 0x4C2EE3B: malloc (vg_replace_malloc.c:309)
==20931==    by 0x4149BD: eMalloc (routines.c:220)
==20931==    by 0x471C8E: newToken (verilog.c:380)
==20931==    by 0x472147: processParameterList (verilog.c:1235)
==20931==    by 0x4736A8: processClass (verilog.c:1304)
==20931==    by 0x4736A8: findTag (verilog.c:1577)
==20931==    by 0x4736A8: findVerilogTags (verilog.c:1648)
==20931==    by 0x411292: createTagsForFile (parse.c:3722)
==20931==    by 0x411292: createTagsWithFallback1 (parse.c:3844)
==20931==    by 0x4115E3: createTagsWithFallback (parse.c:3934)
==20931==    by 0x4115E3: parseMio (parse.c:4096)
==20931==    by 0x411746: parseFileWithMio (parse.c:4142)
==20931==    by 0x411873: parseFile (parse.c:4079)
==20931==    by 0x403307: createTagsForEntry (main.c:221)
==20931==    by 0x40364F: createTagsForArgs (main.c:266)
==20931==    by 0x40364F: batchMakeTags (main.c:360)
==20931==    by 0x403AD5: runMainLoop (main.c:332)
==20931==    by 0x403AD5: ctags_cli_main (main.c:585)
==20931== 
==20931== LEAK SUMMARY:
==20931==    definitely lost: 264 bytes in 3 blocks
==20931==    indirectly lost: 504 bytes in 18 blocks
==20931==      possibly lost: 0 bytes in 0 blocks
==20931==    still reachable: 0 bytes in 0 blocks
==20931==         suppressed: 0 bytes in 0 blocks
==20931== 
==20931== For counts of detected and suppressed errors, rerun with: -v
==20931== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

masatake: The GitHub action "run units target under VALGRIND" looks broken. It should report this memory leak. I must fix it.

hirooih:
Thank you for your prompt reply.

> make units LANGUAGES=SystemVerilog VG=1

This is great!

I've fix it and confirm VG=1 test passes.

masatake:
BTW, I found an interesting document: https://docs.ctags.io/en/latest/contributions.html#
I cannot remember when I wrote this.

hirooih:
> Please, merge your changes for yourself after accepting it.

I missed this sentence.  Thank you for your review and approval.

masatake:
Thank you and welcome to the team.

masatake:
Please close #1488 and #2413 if the issues are solved with the commits.

hirooih:
The fixes in this pull request include the following user visible changes;

- a fork-join creates a context as begin-end does
- delays like #foo, #'foo, #10, etc. are handled correctly
- multi-dimensional array (ex. [1:0] [7:0])is supported
- parameter assignment on function definition ignored properly
- parameter list also support user defined types

PR #2655, opened 2020-10-03
GitHub Actions: create a tag to our repo weekly if a new change is committed since last tagging

### LINKED ISSUE #1038: Release plans ?
Are there any plans to do a (stable) release ?

masatake:
Partially close #1038.

~The tagging format is v5.9.YYYYMMDD.0~.
The tagging format is p5.9.YYYYMMDD.0.

### DIFF
--- a/.github/workflows/tagging.yml
+++ b/.github/workflows/tagging.yml
@@ -1,0 +1,26 @@
+name: Create a periodical tag
+
+on:
+  schedule:
+    - cron:  '0 0 * * 0'
+
+env:
+  # Account for creating a tag
+  USER_NAME: "Universal-ctags CI"
+  USER_EMAIL: "Universal-ctags@users.noreply.github.com"
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        path: ctags
+
+    - name: create a weekly tag
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        cd ctags
+        misc/git-tag-maybe.sh
--- a/misc/git-tag-maybe.sh
+++ b/misc/git-tag-maybe.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+base=5.9
+cal=$(date +%Y%m%d)
+chicken=0
+tag="p${base}.${cal}.${chicken}"
+
+if desc=$(git describe --tags --exact-match); then
+	case "${desc}" in
+		p*.0)
+			echo "do nothing because nothing happens since last tagging"
+			exit 0
+			;;
+		*)
+			echo "found a tag but it is not periodical one; create a periodical tag"
+			;;
+	esac
+fi
+
+r=0
+desc=$(git describe --tags --always)
+printf "%s (%s <= %s)..." "create tag" "${desc}" "${tag}"
+if git tag "${tag}"; then
+	echo "done"
+	printf "push the tag..."
+	if git push origin --tags; then
+		echo "done"
+	else
+		echo "failed"
+		r=2
+	fi
+else
+	echo "failed"
+	r=1
+fi
+exit $r
--- a/Tmain/input-encoding-option.d/tags-expected.txt
+++ b/Tmain/input-encoding-option.d/tags-expected.txt
@@ -7,7 +7,7 @@
 !_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
 !_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
 !_TAG_PROGRAM_URL	https://ctags.io/	/official site/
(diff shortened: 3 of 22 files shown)


.github/workflows/tagging.yml
```diff
@@ -0,0 +1,26 @@
+name: Create a periodical tag
+
+on:
+  schedule:
+    - cron:  '0 0 * * 0'
+
+env:
+  # Account for committing
```
masatake, line 8: The comment should be "Account for creating a tag". I will fix this later.

.github/workflows/tagging.yml
```diff
@@ -0,0 +1,26 @@
...
+  USER_NAME: "Universal-ctags CI"
+  USER_EMAIL: "Universal-ctags-ci@users.noreply.github.com"
```
k-takata, line 10:
```suggestion
  USER_EMAIL: "Universal-ctags@users.noreply.github.com"
```
GitHub provides email addresses with the `<username>@users.noreply.github.com` format for uses who don't want to open their actual email addresses.
I'm not sure `<organizationname>@users.noreply.github.com` is a valid address, but, at least, the `-ci` should be removed.

misc/git-tag-maybe.sh
```diff
@@ -0,0 +1,26 @@
...
+r=0
+
+desc=$(git describe --tags --always)
+case "${desc}" in
+	v*)
+		echo "do nothing because nothing happens since last tagging"
+		;;
+	*)
+		printf "%s (%s <= %s)..." "create tag" "${desc}" "${tag}"
+		if git tag "${tag}"; then
```
k-takata, line 17: It looks that `git push` is missing.
May be `git push origin --tags`?

k-takata (commented):
Maybe we should update `src/ctags.h` and `configure.ac` to `5.9.0`?
(If they are updated, tests should be also updated, though...)

masatake:
> Maybe we should update src/ctags.h and configure.ac to 5.9.0?
(If they are updated, tests should be also updated, though...)

Do you mean updating the version strings in the files WHEN tagging?

k-takata:
I think it's better to set the major and the minor numbers to the same as the tag name.
I don't want to say that the patch number should be updated on every tagging.

masatake:
@k-takata, thank you for reviewing. I updated this pull request.
I cherry-pick'ed the @albfan's work.

I have a question.
```
win32/ctags.exe.manifest:  <assemblyIdentity type="win32" name="Universal Ctags" version="0.0.0.0" />
win32/ctags.rc:            VALUE "FileVersion", "0.0.0.0"
win32/ctags.rc:            VALUE "ProductVersion", "0.0.0.0"
``

Should I update these files? If the answer is yes, how? "5.9.0.0" is o.k.?

**Comment by k-takata** [comment]
> Should I update these files? If the answer is yes, how? "5.9.0.0" is o.k.?

Yes, "5.9.0.0" should be okay.

**Comment by masatake** [comment]
@k-takata,
Could you see b34696f, the change for win32?

**Comment by k-takata** [comment]
> Could you see [b34696f](https://github.com/universal-ctags/ctags/commit/b34696fa475eae6d1e11e13646df38bb8d5e7a38), the change for win32?

Looks okay.

File: misc/git-tag-maybe.sh [hunk 4]
```diff
@@ -0,0 +1,33 @@
...
+base=5.9
+cal=$(date +%Y%m%d)
+chicken=0
+tag="v${base}.${cal}.${chicken}"
+
+r=0
+
+desc=$(git describe --tags --always)
+case "${desc}" in
+	v*)
```
k-takata, line 12: This may not work as expected.
If once `v5.9.20201007.0` was tagged and some commits were added, `git describe --tags --always` would return `v5.9.20201007.0-N-gHHHHHHH`. (`N` is a number of commits from the last tag, `HHHHHHH` is the commit ID.)
If `--exact-match` option is specified, `git describe` will fail if the current commit is not tagged.

masatake: Thank you. I tried --exact-match but it seems to me that the optoin doesn't work expectedly:
```
$ git describe --tags --always
9ce35430
$ git tag x
$ git describe --exact-match  
fatal: no tag exactly matches '9ce35430b0866d055fb18090d85bd00c1dea43d0'
$ echo $?
128
$ git describe --tags --always
x
```

How about pattern 
```
   v*.0)
``` 
instead of `v*` ?
Too stingy?

k-takata: How about `git describe --tags --exact-match`?

masatake:
I decide to use `p` instead of `v`. I would like to reserve `v` for the future.
`p` means Peridoical.

masatake:
The change of prefix must be written in a new blog entry.

masatake:
Simplified the commits.

masatake:
Let's see what happens in this Sunday.

masatake:
5.9.0 are hard-coded in many places. I must reduce them.

PR #2659, opened 2020-10-08
main,refactor: don't hard-code per-parser ptags when emitting

masatake:
Per-parser ptags are lexically enumerated in parser.c
However, a ptag itself knows whether it is "Per-parser"
ptag or not. So the enumeration is not needed.


masatake:
Introduce bit flags that replace two boolean fields.

PR #2672, opened 2020-10-23
SystemVerilog,refactor: remove unused code about type tracker

masatake:
The parser has been improved by the new maintainer, and can recognize
user defined types well know.

It doesn't need the type tracker based on the symbol table anymore.
This change removes it.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -855,11 +855,7 @@ static void createTag (tokenInfo *const token, verilogKind kind)
 		attachParserField (&tag, false,
 						   fieldTable [F_PARAMETER].ftype, "");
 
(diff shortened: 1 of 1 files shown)


hirooih:
Thank you.

> It doesn't need the type tracker based on the symbol table anymore.

Yes.

I left it because it may be reused for variable type information.  But we can add again any time.

The decrease of coverage should be OK for this change.
I've merged the request.

masatake:
When you want to modify the tag entry after passing it to makeTagEntry in the future development, remember corkAPI.
See corkAPI in https://docs.ctags.io/en/latest/internal.html?highlight=cork#output-tag-stream .

PR #2673, opened 2020-10-23
main: introduce keyword group API

masatake:
Introduce a new function for adding keywords having the same value in batch.

Inspired from the code in (System)Verilog parser.

I would like to use the technique used in (System)Verilog parser in Vala parser I'm working on.

@hirooih, could you review my change for (System)Verilog parser?

### DIFF
--- a/main/keyword.c
+++ b/main/keyword.c
@@ -256,3 +256,21 @@ extern void dumpKeywordTable (FILE *fp)
 		}
 	}
 }
+
+extern void addKeywordGroup (const struct keywordGroup *const groupdef,
+							 langType language)
+{
+	for (int i = 0; groupdef->keywords[i]; i++)
+	{
+		if (groupdef->addingUnlessExisting)
+		{
+			if (lookupKeyword (groupdef->keywords[i],
+							   language) != KEYWORD_NONE)
+				continue;		/* already added */
+		}
+		else
+			Assert (lookupKeyword (groupdef->keywords[i],
+								   language) == KEYWORD_NONE);
+		addKeyword (groupdef->keywords[i], language, groupdef->value);
+	}
+}
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -242,70 +242,80 @@ static tokenInfo *tagContents = NULL;
 static fieldDefinition *fieldTable = NULL;
 
 // IEEE Std 1364-2005 LRM, Appendix B "List of Keywords"
-static const char *verilogKeywords[] = {
-	"always", "and", "assign", "automatic", "begin", "buf", "bufif0",
-	"bufif1", "case", "casex", "casez", "cell", "cmos", "config",
-	"deassign", "default", "defparam", "design", "disable", "edge",
-	"else", "end", "endcase", "endconfig", "endfunction", "endgenerate",
-	"endmodule", "endprimitive", "endspecify", "endtable", "endtask",
-	"event", "for", "force", "forever", "fork", "function", "generate",
-	"genvar", "highz0", "highz1", "if", "ifnone", "incdir", "include",
-	"initial", "inout", "input", "instance", "integer", "join", "large",
-	"liblist", "library", "localparam", "macromodule", "medium", "module",
-	"nand", "negedge", "nmos", "nor", "noshowcancelled", "not", "notif0",
-	"notif1", "or", "output", "parameter", "pmos", "posedge", "primitive",
-	"pull0", "pull1", "pulldown", "pullup", "pulsestyle_onevent",
-	"pulsestyle_ondetect", "rcmos", "real", "realtime", "reg", "release",
-	"repeat", "rnmos", "rpmos", "rtran", "rtranif0", "rtranif1",
-	"scalared", "showcancelled", "signed", "small", "specify",
-	"specparam", "strong0", "strong1", "supply0", "supply1", "table",
-	"task", "time", "tran", "tranif0", "tranif1", "tri", "tri0", "tri1",
-	"triand", "trior", "trireg", "unsigned1", "use", "uwire", "vectored",
-	"wait", "wand", "weak0", "weak1", "while", "wire", "wor", "xnor", "xor"
+const static struct keywordGroup verilogKeywords = {
+	.value = K_IGNORE,
+	.addingUnlessExisting = true,
+	.keywords = {
+		"always", "and", "assign", "automatic", "begin", "buf", "bufif0",
+		"bufif1", "case", "casex", "casez", "cell", "cmos", "config",
+		"deassign", "default", "defparam", "design", "disable", "edge",
+		"else", "end", "endcase", "endconfig", "endfunction", "endgenerate",
+		"endmodule", "endprimitive", "endspecify", "endtable", "endtask",
+		"event", "for", "force", "forever", "fork", "function", "generate",
+		"genvar", "highz0", "highz1", "if", "ifnone", "incdir", "include",
+		"initial", "inout", "input", "instance", "integer", "join", "large",
+		"liblist", "library", "localparam", "macromodule", "medium", "module",
+		"nand", "negedge", "nmos", "nor", "noshowcancelled", "not", "notif0",
+		"notif1", "or", "output", "parameter", "pmos", "posedge", "primitive",
+		"pull0", "pull1", "pulldown", "pullup", "pulsestyle_onevent",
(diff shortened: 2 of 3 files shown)


hirooih (approved):
Looks good.

You may be surprised by my next pull-request which I am going to push.

hirooih:
After sending my approval, I found a idea.  I will put the comment.

hirooih (commented):
Do we need `struct keywordGroupDefinition`?

Passing `const char *keywords []` and `int value` (or `int kind`) to `addKeywordGroup()` is more straightforward and easy to read (understand).

masatake:
> Passing const char *keywords [] and int value (or int kind) to addKeywordGroup() is more straightforward and easy to read (understand).

When reading from the line where addKeywordGroup() is called, I agree with you what you wrote.

```
addIgnoredKeywords (language, K_IGNORE, systemVerilogKeywords);
```
is easier to understand the intetion than
```
addIgnoredKeywords (language, &systemVerilogKeywords);
```
.

The former one is more flexible as the combination of the 2nd parameter and the 3rd parameter can be
changed call by call.

However, during developing a parser, especially in the early stage, the developer may read and edit the definitions of keywords 
much more frequently than the place where addKeywordGroup() is called.

```
static struct keywordGroupDefinition verilogKeywords = {
	.value = K_IGNORE,
	.keywords = {
		"always", "and", "assign", "automatic", "begin", "buf", "bufif0",
```
is easier to understand the intetion than
```
static const char *verilogKeywords[] = {
	"always", "and", "assign", "automatic", "begin", "buf", "bufif0",
```
From the former one, I can understand the keywords are ignored quickly.
From the later one I can understand the keywords are ignored but after jumping where  verilogKeywords is used.

In other words, the former one is more declrative. This helps when a parsr has more than one group

About the flexiblity, I think we don't need it in the most cases.

How do you think?
If you don't agree, I can withdraw  the change for verilog.c. I merge the change only for keyword.[ch].

main/keyword.c
```diff
@@ -256,3 +256,14 @@ extern void dumpKeywordTable (FILE *fp)
 		}
 	}
 }
+
+extern void addKeywordGroup (struct keywordGroupDefinition *groupdef,
+							 langType language, bool checkDuplication)
```
masatake, line 261: Instead of making `checkDuplicatoin` as a parameter,  I should it a member of keywordGroupDefinition.

main/keyword.c
```diff
@@ -256,3 +256,14 @@ extern void dumpKeywordTable (FILE *fp)
...
+{
+	for (int i = 0; groupdef->keywords[i]; i++)
+	{
+		if ((!checkDuplication)
+			|| (lookupKeyword (groupdef->keywords[i], language) == KEYWORD_NONE))
```
masatake, line 266: Even if `checkDuplication` is false, adding a keyword twice is not a good idea; maybe it is a bug.
I can put Assert here in the case that `checkDuplication` is false.

parsers/verilog.c
```diff
@@ -242,70 +242,78 @@ static tokenInfo *tagContents = NULL;
...
-	"notif1", "or", "output", "parameter", "pmos", "posedge", "primitive",
-	"pull0", "pull1", "pulldown", "pullup", "pulsestyle_onevent",
-	"pulsestyle_ondetect", "rcmos", "real", "realtime", "reg", "release",
-	"repeat", "rnmos", "rpmos", "rtran", "rtranif0", "rtranif1",
-	"scalared", "showcancelled", "signed", "small", "specify",
-	"specparam", "strong0", "strong1", "supply0", "supply1", "table",
-	"task", "time", "tran", "tranif0", "tranif1", "tri", "tri0", "tri1",
-	"triand", "trior", "trireg", "unsigned1", "use", "uwire", "vectored",
-	"wait", "wand", "weak0", "weak1", "while", "wire", "wor", "xnor", "xor"
+static struct keywordGroupDefinition verilogKeywords = {
```
masatake, line 245: +const

parsers/verilog.c
```diff
@@ -242,70 +242,78 @@ static tokenInfo *tagContents = NULL;
...
-	"specparam", "static", "string", "strong", "strong0", "strong1",
-	"struct", "super", "supply0", "supply1", "sync_accept_on",
-	"sync_reject_on", "table", "tagged", "task", "this", "throughout",
-	"time", "timeprecision", "timeunit", "tran", "tranif0", "tranif1",
-	"tri", "tri0", "tri1", "triand", "trior", "trireg", "type", "typedef",
-	"union", "unique", "unique0", "unsigned", "until", "until_with",
-	"untyped", "use", "uwire", "var", "vectored", "virtual", "void",
-	"wait", "wait_order", "wand", "weak", "weak0", "weak1", "while",
-	"wildcard", "wire", "with", "within", "wor", "xnor", "xor"
+static struct keywordGroupDefinition systemVerilogKeywords = {
```
masatake, line 271: +const

hirooih:
I've pushed a fix which I was going to send as pull-request to my fork.  This is the diff.  (I have not commit yet, of course.)

https://github.com/universal-ctags/ctags/commit/3a5ca10b25ca7eae31b4fc860e660c69c2f9aae7

I add an array of compiler directives defined by the standard. This is exactly the case you wrote as below. 

> This helps when a parser has more than one group.

Your new API will significantly simplify the change on the line 4xx and line 5xx.
I thought you could read my code before I pushed :-)

> From the former one, I can understand the keywords are ignored quickly.
> From the later one I can understand the keywords are ignored but after jumping where verilogKeywords is used.

They are almost same.  I just prefer simpler one.

It is up-to you.  Please choose one you think better.
After you merge, I will send my pull-request merged with your fix.

main/keyword.c
hirooih: > Even if checkDuplication is false, adding a keyword twice is not a good idea; maybe it is a bug.

When I used this routine, I expected that new value overwrote the old value.  I think it is more natural behavior. Now checkDuplication gives us freedom to chose the behavior.

I think we should not add Assert.

masatake: addKeywordGroup() has two usages.
One is found in (System)Verilog parser: adding keywords that are not added yet.
One will be found in Vala: grouping keywords by their values.

In the former case, trying to add a keyword more than once is o.k; Assert is not needed.
In the latter case, trying to add a keyword implies a programming bug; Assert is needed.

I updated my change, as I wrote above.

masatake:
Thank you for reviewing.
Two developers think about the similar change independently. Good trend. 
However, I am not surprised.This kind of event happens when we seek the same goal, making ctags better, aggressively.

main/keyword.c
hirooih: > One is found in (System)Verilog parser: adding keywords that are not added yet.
> One will be found in Vala: grouping keywords by their values.
> 
> In the former case, trying to add a keyword more than once is o.k; Assert is not needed.
> In the latter case, trying to add a keyword implies a programming bug; Assert is needed.

I agree with you here.

In the former case: `checkDuplication `is set to false. Assert is not needed.
In the latter case: `checkDuplication `is set to true. Assert is needed.

> I can put Assert here in the case that checkDuplication is false.

You wrote opposite.

I've checked-out your commit.  But I don't see `checkDuplication`. Did you stop to add `checkDuplication`?

hirooih: I found I've been misunderstanding the issue.

What I expected is adding `checkDuplication` argument to `addKeyword()` to stop Assert().

`addKeyword()` is now use only by c.c and verilog.c.  Adding an argument can be acceptable.
How do you think?

masatake: > What I expected is adding checkDuplication argument to addKeyword() to stop Assert().

I think I should move the Assert in addKeywordGroup() to addKeyword(). I would like to keep addKeyword() as is.

> addKeyword() is now use only by c.c and verilog.c. Adding an argument can be acceptable.
How do you think?

addKeyword is used in main/parse.c, too. 

I have changed the code of ctags more declarative style since u-ctags started.
Ideally, a parser should just declare keywords. Calling addKeyword is imperative.
If possible, I would like to avoid calling it. However, c.c and verilog.c must call it because more than one parser are defined in a source file. They build their keyword tables from shared data structures with addKeyword().
I added addKeywordGroup() for making the way of building more declarative.

hirooih: > If possible, I would like to avoid calling it. 

This makes sense to me.  I assumed it was a public API.

PR #2675, opened 2020-10-25
SystemVerilog: improve compiler directive support

hirooih:
All standard compiler directives are supported.
Only `define directive emits a tag.  All other directives skip the whole line included.
The newline character preceded by a backslash is ignored to support a multi-line macro definition.

new tests are added.


hirooih:
I checked uncovered lines (L704-709).

I tried to add testcases for this block in this pull-request, but I found that `skipExprssion()` introduced by 702c8227 and K_DIRECTIVE condition block introduced by this pull-request made it uncovered.

I've added a test-case which covers the lines.

masatake:
I think it is nice to have a case for testing a multiline macro.
Other than this, the change looks good for me.

By the way, I recommend you to explore how our C parser support macros.

```
$ cat /tmp/foo.c

#define ARG_START  (
#define ARG_END    )
#define BEGIN   {
#define END     }
#define R(X)    return X
#define ZERO    R(0)
#define SEP     ;

int
main ARG_START
ARG_END
BEGIN
ZERO SEP
END

$ ./ctags -o - /tmp/foo.c
ARG_END	/tmp/foo.c	/^#define ARG_END /;"	d	file:
ARG_START	/tmp/foo.c	/^#define ARG_START /;"	d	file:
BEGIN	/tmp/foo.c	/^#define BEGIN /;"	d	file:
END	/tmp/foo.c	/^#define END /;"	d	file:
R	/tmp/foo.c	/^#define R(/;"	d	file:
SEP	/tmp/foo.c	/^#define SEP /;"	d	file:
ZERO	/tmp/foo.c	/^#define ZERO /;"	d	file:

$ ./ctags --param-CPreProcessor:_expand=true --fields-C=+'{macrodef}' --fields=+'{signature}' -o - /tmp/foo.c
ARG_END	/tmp/foo.c	/^#define ARG_END /;"	d	file:	macrodef:)
ARG_START	/tmp/foo.c	/^#define ARG_START /;"	d	file:	macrodef:(
BEGIN	/tmp/foo.c	/^#define BEGIN /;"	d	file:	macrodef:{
END	/tmp/foo.c	/^#define END /;"	d	file:	macrodef:}
R	/tmp/foo.c	/^#define R(/;"	d	file:	signature:(X)	macrodef:return X
SEP	/tmp/foo.c	/^#define SEP /;"	d	file:	macrodef:;
ZERO	/tmp/foo.c	/^#define ZERO /;"	d	file:	macrodef:R(0)
main	/tmp/foo.c	/^main ARG_START$/;"	f	typeref:typename:int	signature:()
```

In the latter invocation, main is captured as a function:-)
You may don't want to expand macros in your parser. However, extracting signatures of macros is not a  bad idea:
```
R	/tmp/foo.c	/^#define R(/;"	d	file:	signature:(X)	macrodef:return X
```

hirooih:
> I think it is nice to have a case for testing a multiline macro.

Me, too.  I've added it.

> Other than this, the change looks good for me.

Thanks.  I've merged this request.

> You may don't want to expand macros in your parser.

No, I don't.  If we need to expand macros, I prefer to use external preprocessor and use `__LINE__` and `__FILE__` directive.

> However, extracting signatures of macros is not a bad idea:

Good suggestion.  I will add them on the TODO list.

PR #2679, opened 2020-10-27
Fix and tidy up the Inko parser

YorickPeterse:
This applies various fixes to the Inko regex parser, and updates my
Email address as I no longer use Gmail.

Required methods
================

Inko has a notion of required methods. These are defined without a body
like so:

    trait SomeTrait {
      def required_method(some_argument: Type)
    }

These were not parsed properly. In this commit we fix the parser so both
regular and required methods are parsed.

Nested objects and traits
=========================

The parser had support for nested objects and traits, like so:

    object A {
      object B {}
    }

Inko dropped support for this syntax quite some time ago, so we also
remove ctags support for this.

Attribute definitions
=====================

Inko's syntax for defining attributes changed a while ago. Instead of
using `let @foo = bar` you now define attribute types separately like
so:

    object A {
      @foo: SomeType
    }

This adds support for parsing this syntax, and drops support for the
`let` syntax.

Objects implementing traits
===========================

A test case used the syntax `object A impl B`, which Inko hasn't
supported in quite some time. This commit removes this example, as this
isn't valid Inko code any more.

### DIFF
--- a/Units/parser-inko.r/everything.d/expected.tags
+++ b/Units/parser-inko.r/everything.d/expected.tags
@@ -1,18 +1,17 @@
 A	input.inko	/^object A$/;"	o
+@number	input.inko	/^  @number: X$/;"	a	object:A
 in_a	input.inko	/^  def in_a {$/;"	m	object:A
-another	input.inko	/^  def another {}$/;"	m	object:A
+another	input.inko	/^  def another {}$/;"	m
 foo	input.inko	/^def foo {}$/;"	m
 B	input.inko	/^object B {$/;"	o
 in_b	input.inko	/^  def in_b {$/;"	m	object:B
-C	input.inko	/^  object C {$/;"	o	object:B
-nested_method	input.inko	/^    def nested_method {}$/;"	m	object:B.C
 global	input.inko	/^def global {}$/;"	m
 SomeTrait	input.inko	/^trait SomeTrait {$/;"	t
-Socket	input.inko	/^object Socket impl Foo {$/;"	o
+required_method	input.inko	/^  def required_method$/;"	m	trait:SomeTrait
+Socket	input.inko	/^object Socket {$/;"	o
 quack	input.inko	/^  def quack {}$/;"	m	object:Socket
 Socket	input.inko	/^impl Bar for Socket {$/;"	r	implements:Bar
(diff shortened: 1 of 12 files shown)


masatake:
Thank you.

Could you change the header of the commit log like:
```
Inko: fix and tidy up the parser 
```

We expect  "<LANG>: ..." as the header format.

If you have a note that users of this parser should know, you can add a manual page for Inko.
I wrote one for python parser:  
https://docs.ctags.io/en/latest/man/ctags-lang-python.7.html .

If you are interested in writing the man page for Inko, let me know.

YorickPeterse:
@masatake I've adjusted the commit message.

For the man pages, I noticed that the Python one has two files:

```
ctags $ rg 'A tag for an imported module'
man/ctags-lang-python.7.rst.in
127:A tag for an imported module has ``module`` kind with ``imported`` role.  The

docs/man/ctags-lang-python.7.rst
127:A tag for an imported module has ``module`` kind with ``imported`` role.  The
```

When writing a manual page for Inko, do we need the same setup? Or is the second file a generated copy of the first one?

masatake:
I will write the steps to add a per-language man page:
0. read man/README.
1. write man/ctags-lang-LANGUAGE.7.rst.in
    Write what the users of your parser may want to know.
2. add man/ctags-lang-LANGUAGE.7.rst to the end of configure.ac.
3. add man/ctags-lang-LANGUAGE.7 to man_MANS of Makefile.am.
4. add man/ctags-lang-LANGUAGE.7.rst.in to IN_SOURCE_FILES of man/Makefile.
5. run "cd man; make QUICK=1 update-docs". This step generates the rst file at docs/man/.
6. add ctags-lang-LANGUAGE(7) to docs/man-pages.rst.
7. do "git add" for the changed file and new files (man/ctags-lang-LANGUAGE.7.rst.in and docs/man/ctags-lang-LANGUAGE.7.rst), and do git commit with a log header like: "docs(man): add a man page for LANGUAGE".
8. make the pull request for the man page!

masatake:
I merge the change. I expect you to make another pull request for the man page.
Thank you.

YorickPeterse:
@masatake I'll try to take a look at writing the man page(s) this week. Thanks for the info!

PR #2680, opened 2020-10-27
SystemVerilog: handle an identifier include '$'

hirooih:
`isIdentifierCharacter() ` did not handle '$` properly.
I needed to simplify token handling for the better fix of it.
Delay handling is also improved.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -670,13 +670,14 @@ static int skipDimension (int c)
 	return c;
 }
 
-static void skipToSemiColon (void)
+static int skipToSemiColon (void)
 {
 	int c;
 	do
 	{
 		c = vGetc ();
 	} while (c != EOF && c != ';');
+	return c;	// ';' or EOF
 }
 
 static int skipExpression(int c)
@@ -1436,9 +1437,11 @@ static void processAssertion (tokenInfo *const token)
 {
 	if (vStringLength (currentContext->blockName) > 0)
 	{
+		int c;
 		vStringCopy (token->name, currentContext->blockName);
 		createTag (token, K_ASSERTION);
-		skipToSemiColon ();
+		c = skipToSemiColon ();
+		vUngetc (c);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


hirooih:
All of uncovered lines increased are sanity checks for broken inputs.

parsers/verilog.c
```diff
@@ -420,25 +420,49 @@ static short hasSimplePortList (verilogKind kind)
...
+	vStringDelete (dup->name);
+	vStringDelete (dup->blockName);
+	vStringDelete (dup->inheritance);
+	*dup = *token;
+	dup->name = vStringNew ();
+	dup->blockName = vStringNew ();
+	dup->inheritance = vStringNew ();
+	vStringCopy (dup->name, token->name);
+	vStringCopy (dup->blockName, token->blockName);
+	vStringCopy (dup->inheritance, token->inheritance);
```
masatake, line 462: You does "delete", "new", and "copy" three vStrings.

However, when you call newToken () at the beginning of the funciton,  the `dup` returned from newToken() is already cleared.
So the "delete" and "new" steps are just redundant. What you need is only the the "copy" step.

hirooih: Thank you for your review.

First I did make same mistake as you wrote.  But it does not work.
`dup->name`, `dup->blockName`, and `dup->inheritance` are overwritten by `token->name`, `token->blockName`, and `token->inheritance` via `*dup = *token;`.
vStrings allocated by `dupToken()` are never be freed.

Another solution is copying each element without using  `*dup = *token;`.  But I did not choose it because it might cause missing elements.  I am assuming the number of vString elements are much less than the number of other elements.

masatake: How about the following code?

```
static tokenInfo *dupToken (tokenInfo *token)
{
	tokenInfo *dup = newToken ();

	{
	  tokenInfo tmp = *dup;

	  *dup = *token;

	  dup->name = tmp.name;
	  dup->blockName = tmp.blockName;
	  dup->inheritance = tmp.inheritance;
	}

	vStringCopy (dup->name, token->name);
	vStringCopy (dup->blockName, token->blockName);
	vStringCopy (dup->inheritance, token->inheritance);
	return dup;
}
```

hirooih: Thanks.  Better than mine.
I've committed the following code.

``` C
static tokenInfo *dupToken (tokenInfo *token)
{
	tokenInfo *dup = newToken ();
	tokenInfo tmp = *dup;	// save vStrings, name, blockName, and inheritance
	*dup = *token;
	// revert vStrings allocated for dup
	dup->name = tmp.name;
	dup->blockName = tmp.blockName;
	dup->inheritance = tmp.inheritance;
	// copy contents of vStrings
	vStringCopy (dup->name, token->name);
	vStringCopy (dup->blockName, token->blockName);
	vStringCopy (dup->inheritance, token->inheritance);
	return dup;
}
```

masatake (approved):

Issue #2683, opened 2020-10-29
What's the best way of parsing the fields from tagfile/origin tagline for the tools based on ctags


liuchengxu:
Although ctags already supports the JSON output, the compiled feature `+json` is required and some people does not have this feature, in which case the tools built on ctags like some vim plugins([vista.vim](https://github.com/liuchengxu/vista.vim), [vim-clap](https://github.com/liuchengxu/vim-clap)...) need to parse the fields manually, otherwise they can't work. 

vista.vim is currently using https://github.com/liuchengxu/vista.vim/blob/27889ff58d/autoload/vista/parser/ctags.vim#L91-L134, but not sure it's robust enough. In [this vim-clap PR](https://github.com/liuchengxu/vim-clap/pull/557), we plan to parse the tagfile, hoping to find a good solution for such a parser. Any insights would be appreciated.

See the tips for this task from @masatake in https://github.com/liuchengxu/vim-clap/pull/557#issuecomment-718041196

cc @romgrk

masatake:
We have a parser written in C. I hope it will help to find a better way.

See parseTagLine and parseExtensionFields in [readtags.c](https://github.com/universal-ctags/libreadtags/blob/master/readtags.c).

I read https://github.com/liuchengxu/vista.vim/blob/27889ff58d/autoload/vista/parser/ctags.vim#L91-L134. As fa as `;"` is not included in the pattern, it looks find for me as far as you use u-ctags. I would like to know the detail of ", but not sure it's robust enough".

If you don't use u-ctags, a tab in a field can include `\t`. So parsing will be failed. In such case, ... use u-ctags.

AmaiKinono:
The manpage ctags-client-tools(7) has a "Parse Readtags Output" section. You should be good if using the method provided there.

masatake:
@AmaiKinono, see `--excmd=type` in https://docs.ctags.io/en/latest/man/ctags.1.html#options .

PR #2689, opened 2020-11-06
docs(man): correct the parsing technique in ctags-client-tools(7)

AmaiKinono:
Ref: <https://github.com/universal-ctags/ctags/issues/2683#issuecomment-722644523>

The original technique introduced before doesn't deal with pattern fields that only use line numbers.

One thing I'm not sure is till now we are only dealing with `--format=2`. I think not many people will use `--format=1`, but if that's the case, this parsing technique won't work for them.


AmaiKinono:
For some reason the web version of this doc (https://docs.ctags.io/en/latest/man/ctags-client-tools.7.html#parse-readtags-output) is out of shape...

masatake:
I will fix the out of shaping.

masatake:
See  #2690.

PR #2637, opened 2020-09-06
SystemVerilog: Implement struct/union support

hirooih:
struct and union was not supported.   (They are supported only in typedef declaration.)
This pull-request implement it and provide its test-cases.

PR #2696, opened 2020-11-14
SystemVerilog: introduce processType() for consistent type-handling

hirooih:
processType() that processes variable types is introduced.

Type processing was handled individually and ad hoc in tagNameList(), processPortList(), processTypeDef(), and processEnum().
processType() processes variable types consistently.  It simplifies those functions significantly and makes them reliable and robust.


parsers/verilog.c
```diff
@@ -114,6 +114,7 @@ typedef enum {
  *   DATA DEFINITIONS
  */
 static int Ungetc;
+static bool inSkipPastMatch = false;	// for vGetc() and skipPastMatch()
```
masatake, line 117: If possible, I would like to avoid introducing a variable having a broad scope.
I frequently use the following technique to avoid introducing such a variable.

```c
static int vGetcFull (bool inSkipPastMatch)
{
   ...
  else if (inSkipPastMatch && c == '"')
  ...
}
...
static int vGetc (void)
{
    return vGetcFull (false);
}
...
static int skipPastMatch (const char *const pair)
{
   ...
   do
	{
		c = vGetcFull (true);
		if (c == begin)
			++matchLevel;
		else if (c == end)
			--matchLevel;
	}
	while (c != EOF && matchLevel > 0);
    return vGetc (); /* or vGetcFull (false); */
}
...
```

parsers/verilog.c
```diff
@@ -1603,6 +1603,23 @@ static int skipDelay(tokenInfo* token, int c)
 	return c;
 }
 
+static int skipClockEvent(tokenInfo* token, int c)
+{
+	if (c == '@')
+	{
+		c = skipWhite (vGetc ());
+		if (c == ('@')) {	// coverage_event: @@(block_event_expression)
```
masatake, line 1459: Is the parenthesis  around '@' needed?

Units/parser-verilog.r/systemverilog-net-var.d/expected.tags
```diff
@@ -11,9 +11,10 @@ iBus	input.sv	/^  interconnect [0:1] iBus;$/;"	n	module:top
 rlMod	input.sv	/^module rlMod(input interconnect [0:1] iBus);$/;"	m
 iBus	input.sv	/^module rlMod(input interconnect [0:1] iBus);$/;"	p	module:rlMod
 Net_declarations	input.sv	/^module Net_declarations$/;"	m
+cap1	input.sv	/^  trireg (large) logic #(0,0,0) cap1;$/;"	n	module:Net_declarations
 addressT	input.sv	/^  typedef logic [31:0] addressT;$/;"	T	module:Net_declarations
-ecc	input.sv	/^  wire struct packed { logic ecc; logic [7:0] data; } memsig; \/\/ FIXME: ignored -> memsig$/;"	r	module:Net_declarations
-data	input.sv	/^  wire struct packed { logic ecc; logic [7:0] data; } memsig; \/\/ FIXME: ignored -> memsig$/;"	r	module:Net_declarations
```
masatake, line 16: I don't know SystemVerilog. However, ecc and data look important tags for me.
Am I wrong?

parsers/verilog.c
hirooih: Cool!
This technique can be used many other situations. Thanks.
I've fixed this.  (I renamed vGetcFull() to _vGetc().)

parsers/verilog.c
hirooih: Off course, No.
But I found this condition is never covered and removed the if-clause.

Units/parser-verilog.r/systemverilog-net-var.d/expected.tags
hirooih: Good point, but this is OK.
SystemVerilog parser does not support struct/union members yet, as we discussed before on #2637.  It is still on my TODO list.

The struct members are tagged as a variable declaration (tag `r`) wrongly in this case.

hirooih:
I've fixed following to your advices.
I also add some fixes and add tests to improve the code coverage.

masatake (approved):

PR #2701, opened 2020-11-17
SystemVerilog: support struct and union  member

hirooih:
Struct and union  members are emitted by this fix.
The kind "member" (w) is added.  (m and M were already used.)

Enum items are emitted in the defined order with `--sort=no` option.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -82,7 +82,8 @@ typedef enum {
 	K_TYPEDEF,
 	K_CHECKER,
 	K_CLOCKING,
-	K_SEQUENCE
+	K_SEQUENCE,
+	K_MEMBER
 } verilogKind;
 
 typedef struct {
@@ -154,7 +155,8 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'T', "typedef",   "type declarations" },
  { true, 'H', "checker",   "checkers" },
  { true, 'L', "clocking",  "clocking" },
- { true, 'q', "sequence",  "sequences" }
+ { true, 'q', "sequence",  "sequences" },
+ { true, 'w', "member",    "struct and union members" }
 };
 
 static const keywordAssoc KeywordTable [] = {
@@ -376,6 +378,7 @@ static bool findBlockName (tokenInfo *const token);
 static void processDefine (tokenInfo *const token);
 static int processType (tokenInfo* token, int c, verilogKind* kind);
 static int pushEnumNames (tokenInfo* token, int c);
+static int pushMembers (tokenInfo* token, int c);
 static bool readWordToken (tokenInfo *const token, int c);
 static int skipDelay(tokenInfo* token, int c);
 static int tagNameList (tokenInfo* token, int c);
@@ -404,6 +407,7 @@ static short isContainer (verilogKind kind)
 		case K_SEQUENCE:
 		case K_TYPEDEF:
 		case K_ENUM:
+		case K_STRUCT:
 			return true;
 		default:
 			return false;
@@ -416,6 +420,7 @@ static short isTempContext (tokenInfo const* token)
 	{
 		case K_TYPEDEF:
 		case K_ENUM:
+		case K_STRUCT:
 			return true;
 		default:
 			return false;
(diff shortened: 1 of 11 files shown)


masatake:
...I don't like this change and the original code.
See the field used in popToken and pushToken.  These functions use `scope` field.
These functions are originally written for tracking scope information. 
See the comment for `scope`:
```
	struct sTokenInfo*  scope;         /* context of keyword */
```

However, the change and the original code abuse them for dealing with a sequence.
I think this is a bad idea.


Do you know ptrArray (main/ptrarray.h) ?

You can use ptrArray to store the sequence of tokens.
#2702 illustrates how to use ptrArray.

hirooih:
> ...I don't like this change and the original code.

I don't like it, either.  But I also don't like to reinventing the wheels.
I reused the code which was already in verilog.c.

> Do you know ptrArray (main/ptrarray.h) ?

No, I did not.
AFAIK we don't have a list of API which a parser can use.  Can we use any external functions defined in main/*.h?

I have a question related to this topic.
Is there API to get the filename which a parse is parsing?

I will update verilog.c to use ptrArray.

masatake:
>  AFAIK we don't have a list of API which a parser can use. Can we use any external functions defined in main/*.h?

No, we don't have.
We have one rule; we should not use functions and types declared in main/*_p.h in parsers. _p means "private in main."\

> I have a question related to this topic.
> Is there API to get the filename which a parse is parsing?

Yes, you can use getInputFileName() declared in main/read.h.

> I will update verilog.c to use ptrArray.

I would like you to see #2702 and merge it if it is o.k.

hirooih:
> We have one rule; we should not use functions and types declared in main/*_p.h in parsers. _p means "private in main."

I see.  This means we do have a list of API.

> Yes, you can use getInputFileName() declared in main/read.h.

Thanks.

> I would like you to see #2702 and merge it if it is o.k.

You raised the bar!  It will cause conflicts.
It will be my first experience of review on GitHub.

hirooih:
@masatake san,

How about this?  Please review again.
Thank you.

masatake (approved):

parsers/verilog.c
```diff
@@ -1476,6 +1485,57 @@ static int pushEnumNames (tokenInfo* token, int c)
...
 
+// create a list of struct/union members
+static int pushMembers (tokenInfo* token, int c)
+{
+	if (c == '{')
+	{
+		c = skipWhite (vGetc ());
+		do
+		{
+			verilogKind kind = K_UNDEFINED;
```
masatake, line 1496: Is this "kind" used?
The address for this local var is passed to processType().
processType() may fill the variable with a value.
However, the value for the local variable is not used.
If the above my analysis is correct, could you take one of the following action:
* rename the local variable to "dummy_kind", "tmp_kind", "kind_no_use" or something, or
* eliminate the local variable after making processType accept NULL as the third argument.

hirooih: > Is this "kind" used?

Yes.

```
// input
//   kind: kind of context
// output
//   kind: kind of type
//   token: identifier token (unless K_IDENTIFIER nor K_UNDEFINED)
static int processType (tokenInfo* token, int c, verilogKind* kind)
{
```

masatake: Oh, I see. The value for kind is consumed in processType:
```
		if (c == '{')	// skip enum, struct, or union member
		{
			if (*kind == K_ENUM)
				c = pushEnumNames (token, c);
			else
				c = skipWhite (skipPastMatch ("{}"));
		}

```

masatake (approved):

parsers/verilog.c
masatake: Reducing the local variable contributes readability of the code.
You can eliminate "kind" with the following technique.  But the technique makes the code just unreadable.

```
void
foo (verilogKind *i)
{
}

int
main(void)
{
  foo (&((struct { verilogKind i; }){ .i = 1 }.i));
}
```

hirooih: I've added a comment.

```
			verilogKind kind = K_UNDEFINED;	// set kind of context for processType()
```

> So ignore this comment.

Of course, I do :-)

PR #2702, opened 2020-11-17
SystemVerilog,refactor: convert the type of `tagContents' file private variable to ptrArray

masatake:
The original code implemented set or sequence of tokenInfos with using `scope' member
of tokenInfo. `scope' member is for representing scope, not for set or sequence.

For representing a sequence, ptrArray is suitable.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -29,6 +29,7 @@
 #include "read.h"
 #include "routines.h"
 #include "xtag.h"
+#include "ptrarray.h"
 
 /*
 *   MACROS
@@ -245,7 +246,7 @@ static const keywordAssoc KeywordTable [] = {
 };
 
 static tokenInfo *currentContext = NULL;
-static tokenInfo *tagContents = NULL;
+static ptrArray *tagContents;
 static fieldDefinition *fieldTable = NULL;
 
 // IEEE Std 1364-2005 LRM, Appendix B "List of Keywords"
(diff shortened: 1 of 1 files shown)


parsers/verilog.c
```diff
@@ -556,6 +557,19 @@ static void initializeVerilog (const langType language)
 	buildKeywordHash (language, IDX_VERILOG);
 	addKeywordGroup (&verilogKeywords, language);
 	addKeywordGroup (&verilogDirectives, language);
+	if (tagContents == NULL)
+		tagContents = ptrArrayNew ((ptrArrayDeleteFunc)deleteToken);
+
+}
+
+static void finalizeVerilog (const langType language,
```
hirooih, line 565: I think we don't need finalizeVerilog() and finalizeSystemVerilog().

I am not a Purist who frees everything before `exit()`.
If we add this, we should also free the hash table entries added by `addKeyword()`.

parsers/verilog.c
```diff
@@ -988,16 +1014,13 @@ static void createTag (tokenInfo *const token, verilogKind kind)
 		createContext (kind, token->name);
 
 		/* Include found contents in context */
-		if (tagContents != NULL)
-		{
-			tokenInfo* content = tagContents;
-
+		if (ptrArrayCount(tagContents) != 0)
```
hirooih, line 1017: This if-clause is verbose.

hirooih (changes requested):
I agree with you that the ptrarray functions are much better than before.

I might be wrong on both review comments and GitHub usages.
Please check them.

masatake:
Updated.

hirooih (approved):

hirooih:
Thanks!

Issue #2674, opened 2020-10-24
Verilog/SystemVerilog Parser TODO List


hirooih:
Here is a TODO list of Verilog/SystemVerilog parser.  It is ordered in priority, from higher to lower.

If you find any other missing issues or requests to raise priority, please leave your comment on this issue ticket.
Your feedback helps other users and us.

### Known Bugs

- preprocessor directive related issues
  - `else in port lists
  - multi-line macro definition
- `isIdentifierCharacter()` does not handle '$` properly
- update descriptions of kinds
- `fixme` in [`systemverilog-net-var.d/inputs.sv`](https://github.com/universal-ctags/ctags/blob/master/Units/parser-verilog.r/systemverilog-net-var.d/input.sv)
- `fixme` in [`systemverilog-parameter.d/inputs.sv`](https://github.com/universal-ctags/ctags/blob/master/Units/parser-verilog.r/systemverilog-parameter.d/input.sv)
  - interface class (LRM 8.26)
  - suppress parameter field in generate block (cf. #2537)
- user defined type on enum -> generic type handling

### Features Not Implemented Yet

- `clocking`-`clockingend`, `checker`-`endchecker`
- `import `as prototype (ignore as pure and context)
- interface class (LRM 8.26)
- `implement `field for interface class
- members in structures
- type information like C parser
- element range on enum
- escaped identifiers (LRM 5.6.1)

### Refactoring

- simplify token handling
- divide `processDesignElement()`
- apply `skipExpresson()` on `processPortList()`
- unify `processPortList()` and  `tagNameList()`
  - refactor `skipMacro()` called by only `processPortList()` and  `tagNameList()` 
- introduce `skipType()` (generic type handling)
- use [trace.h](https://github.com/universal-ctags/ctags/pull/2664) 

### NOTE

- Please, read [ctags(1)](https://docs.ctags.io/en/latest/man/ctags.1.html) and understand the fundamental concepts in ctags including the definition tags, reference tags, kinds, fields, extras, and pseudo-tags before commenting.
- If possible, take a look at [`parsers/verilog.c`](https://github.com/universal-ctags/ctags/blob/master/parsers/verilog.c) before adding something.
- @hirooih will update, delete, or edit your comments for keeping this as our TODO list.

PR #2703, opened 2020-11-18
verilog: keep trace with instance in TAGS file

my2817:
add a kind named K_INST
update function tagNameList to update kind from K_IDENTIFIER to K_INST

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -68,6 +68,7 @@ typedef enum {
 	K_REGISTER,
 	K_TASK,
 	K_BLOCK,
+	K_INSTANCE,
 	K_ASSERTION,
 	K_CLASS,
 	K_COVERGROUP,
@@ -128,7 +129,8 @@ static kindDefinition VerilogKinds [] = {
  { true, 'p', "port",      "ports" },
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
- { true, 'b', "block",     "blocks (begin, fork)" }
+ { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance",  "instances of module" }
 };
 
 static kindDefinition SystemVerilogKinds [] = {
@@ -141,6 +143,7 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
  { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance"   "instances of module or interface" },
  { true, 'A', "assert",    "assertions (assert, assume, cover, restrict)" },
  { true, 'C', "class",     "classes" },
  { true, 'V', "covergroup","covergroups" },
@@ -1625,6 +1628,19 @@ static int tagNameList (tokenInfo* token, int c)
 			if (c == '=')
 				c = skipExpression (c);
 		}
+		else if (c == '(' || c == '[')	// should be instance
+		{
+			c = skipDimension(c); // name_of_instance {unpacked_dimension}
+			c = skipPastMatch ("()"); // list_of_port_connections
+
+			// if without the next "if" clause, get a instance named: `add_t from the following example
+			// var `add_t(foo) = '0;
+			if (c == ';' || c == ',')
+			{
+				verbose("find instance: %s with kind %s\n", vStringValue (token->name), getNameForKind(K_INSTANCE));
+				createTag (token, K_INSTANCE);
+			}
+		}
 		c = skipMacro (c);	// `ifdef, `else, `endif, etc. (before comma)
 
 		if (c != ',' || c == EOF)
@@ -1635,10 +1651,6 @@ static int tagNameList (tokenInfo* token, int c)
 			kind = K_UNDEFINED;
 	} while (true);
 
-	/* skip port list of module instance: foo bar(xx, yy); */
-	c = skipWhite (c);
-	if (c == '(')
-		c = skipPastMatch ("()");
 	return c;
 }
 
--- a/Units/parser-verilog.r/systemverilog-assertion.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-assertion.d/expected.tags
@@ -49,6 +49,8 @@ sva_svtb	input.sv	/^module sva_svtb;$/;"	m
 clk	input.sv	/^    bit clk;$/;"	r	module:sva_svtb
 a	input.sv	/^    logic a, b;$/;"	r	module:sva_svtb
 b	input.sv	/^    logic a, b;$/;"	r	module:sva_svtb
+dut	input.sv	/^    dut dut (.*);$/;"	i	module:sva_svtb
+tb	input.sv	/^    tb tb (.*);$/;"	i	module:sva_svtb
 m	input.sv	/^module m (input a, b);$/;"	m
 a	input.sv	/^module m (input a, b);$/;"	p	module:m
 b	input.sv	/^module m (input a, b);$/;"	p	module:m
--- a/Units/parser-verilog.r/systemverilog-checker.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-checker.d/expected.tags
@@ -38,6 +38,7 @@ bus	input.sv	/^module m(wire [31:0] bus, logic clk);$/;"	n	module:m
 clk	input.sv	/^module m(wire [31:0] bus, logic clk);$/;"	n	module:m
 res	input.sv	/^    logic res, scan;$/;"	r	module:m
 scan	input.sv	/^    logic res, scan;$/;"	r	module:m
+check_bus	input.sv	/^    mutex check_bus(bus, posedge clk, res);$/;"	i	module:m
 c1	input.sv	/^checker c1(event clk, logic[7:0] a, b);$/;"	H
 clk	input.sv	/^checker c1(event clk, logic[7:0] a, b);$/;"	p	checker:c1
 a	input.sv	/^checker c1(event clk, logic[7:0] a, b);$/;"	p	checker:c1
@@ -54,7 +55,10 @@ en	input.sv	/^module m(input logic rst, clk, logic en, logic[7:0] in1, in2,$/;"	
 in1	input.sv	/^module m(input logic rst, clk, logic en, logic[7:0] in1, in2,$/;"	p	module:m
 in2	input.sv	/^module m(input logic rst, clk, logic en, logic[7:0] in1, in2,$/;"	p	module:m
 in_array	input.sv	/^         in_array [20:0]);$/;"	p	module:m
+check_outside	input.sv	/^    c1 check_outside(posedge clk, in1, in2);$/;"	i	module:m
 v1	input.sv	/^        automatic logic [7:0] v1=0;$/;"	r	module:m
+check_inside	input.sv	/^            c1 check_inside(posedge clk, in1, v1);$/;"	i	module:m
+check_loop	input.sv	/^                c1 check_loop(posedge clk, in1, in_array[v1]);$/;"	i	module:m
 counter_model	input.sv	/^checker counter_model(logic flag);$/;"	H
 flag	input.sv	/^checker counter_model(logic flag);$/;"	p	checker:counter_model
 counter	input.sv	/^    bit [2:0] counter = '0;$/;"	r	checker:counter_model
--- a/Units/parser-verilog.r/systemverilog-net-var.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-net-var.d/expected.tags
@@ -8,8 +8,13 @@ vw	input.sv	/^  logic vw; \/\/ no initial assignment$/;"	r	module:net_decl
 circ	input.sv	/^  real circ;$/;"	r	module:net_decl
 top	input.sv	/^module top();$/;"	m
 iBus	input.sv	/^  interconnect [0:1] iBus;$/;"	n	module:top
+l1	input.sv	/^  lDriver l1(iBus[0]);$/;"	i	module:top
+r1	input.sv	/^  rDriver r1(iBus[1]);$/;"	i	module:top
+m1	input.sv	/^  rlMod m1(iBus);$/;"	i	module:top
 rlMod	input.sv	/^module rlMod(input interconnect [0:1] iBus);$/;"	m
 iBus	input.sv	/^module rlMod(input interconnect [0:1] iBus);$/;"	p	module:rlMod
+l1	input.sv	/^  lMod l1(iBus[0]);$/;"	i	module:rlMod
+r1	input.sv	/^  rMod r1(iBus[1]);$/;"	i	module:rlMod
 Net_declarations	input.sv	/^module Net_declarations;$/;"	m
 cap1	input.sv	/^  trireg (large) logic #(0,0,0) cap1;$/;"	n	module:Net_declarations
 addressT	input.sv	/^  typedef logic [31:0] addressT;$/;"	T	module:Net_declarations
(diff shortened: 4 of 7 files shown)


masatake:
The way to write test cases is wrong. expected.tags is missing. See https://docs.ctags.io/en/latest/units.html.
You should not use `_` as the part of the kind name. The tests on the CI environment may be failed because of the `_` in the kind name.

hirooih:
Thanks for enhancement.

- Please use terminology used in[ the LRM (Language Reference Manual)](https://ieeexplore.ieee.org/document/8299595), inst -> instance.
- Verilog does not support interface.
- Please check with the following tests.

```
make units LANGUAGES=SystemVerilog,Verilog
make units LANGUAGES=SystemVerilog,Verilog VG=1
```

- Make sure only instance names are tagged.

hirooih:
> Make sure only instance names are tagged.

And;

- Make sure all instance names are tagged.

my2817:
> > Make sure only instance names are tagged.
> 
> And;
> 
> * Make sure all instance names are tagged.

Thanks for your work on this awesome tool and advice.

parsers/verilog.c
```diff
@@ -1577,8 +1580,20 @@ static int tagNameList (tokenInfo* token, int c)
 	{
 		c = processType(token, c, &kind);	// update token and kind
 
-		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`')
+		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`' || c == '(')
 		{
+			if(c == '('){
```
masatake, line 1585: we don't use this style, putting `{` at the end of line IN THIS FILE.

```
if (c == '(')
{
   ...
```
is expected.

parsers/verilog.c
```diff
@@ -1577,8 +1580,20 @@ static int tagNameList (tokenInfo* token, int c)
...
+					token->kind = K_INSTANCE;
+					kind = K_INSTANCE;
+					c = skipPastMatch ("()");
+					if(c != ';'){
```
masatake, line 1589: we don't use this style, putting { at the end of line IN THIS FILE.

if (c == '(')
{
   ...

is expected.

parsers/verilog.c
```diff
@@ -1577,8 +1580,20 @@ static int tagNameList (tokenInfo* token, int c)
...
+							token->kind = K_UNDEFINED;
+							kind = K_UNDEFINED;
+					}
+					else {
```
masatake, line 1593: we don't use this style, putting { at the end of line IN THIS FILE.

else
{
   ...

is expected.

parsers/verilog.c
my2817: Thanks, it's my first time to do this job. I will back again

parsers/verilog.c
```diff
@@ -1577,8 +1580,20 @@ static int tagNameList (tokenInfo* token, int c)
...
+							verbose("find instance: %s with kind %d\n", vStringValue (token->name), token->kind);
```
masatake, line 1594: How about printing the kind in a string?
```
verbose("find instance: %s with kind %s\n", vStringValue (token->name),getNameForKind(token->kind));
```

parsers/verilog.c
masatake: This project is looking for persons like you!

parsers/verilog.c
masatake: (I myself didn't test the change that I showed.)

masatake:
I would like you to squash the commits into one commit.
The idea behind the squashing is described in https://github.com/universal-ctags/ctags/pull/2705#issuecomment-730041445 .

In addition, I would like you to use "SystemVerilog" or "Verilog,SystemVerilog" instead of "[verilog/SystemVerilog]" as the prefix of the commit header like: 

```
cc26cc1d Ruby: fill signature fields for the method kind
65f5bc5b Go: use symbol table for attaching typeref field to members of a structure
bf1206cc SystemVerilog: parse a multi dimensional packed array in typedef correctly
```

parsers/verilog.c
```diff
@@ -1577,8 +1580,23 @@ static int tagNameList (tokenInfo* token, int c)
...
 
-		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`')
+		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`' || c == '(')
 		{
+			if(c == '(')
+			{
+				token->kind = K_INSTANCE;
+				kind = K_INSTANCE;
+				c = skipPastMatch ("()");
+				if(c != ';' && c != ',')
```
masatake, line 1590: We can simplify the code a bit like:
```
if (c == '(')
{
      kind = K_UNDEFINED;
      c = skipPastMatch ("()");
      if (c == ';' || c == ',')
      {
			kind = K_INSTANCE;
            verbose("find instance: %s with kind %s\n", vStringValue (token->name), getNameForKind(kind));        
      }
      token->kind = kind;
}
```

hirooih: I will check this code in detail in this week.
Quick comments.

I propose the followings; 

```
if (c == '(')
{
      c = skipPastMatch ("()");
      if (c == ';' || c == ',')
      {
		kind = K_INSTANCE;
        	verbose("find instance: %s with kind %s\n", vStringValue (token->name), getNameForKind(kind));        
      }
      else  // comment why this condition is required
		kind = K_UNDEFINED;

      token->kind = kind;
}
```

Actually I don't understand why the condition is required. 

From LRM 23.3.2 Module instantiation syntax

```
module_instantiation ::= // from A.4.1.1
	module_identifier [ parameter_value_assignment ] hierarchical_instance { , hierarchical_instance };
...
hierarchical_instance ::= name_of_instance ( [ list_of_port_connections ] )
name_of_instance ::= instance_identifier { unpacked_dimension }
...
```

Please take care of ` { unpacked_dimension }` part.

Please check interface instantiation syntax, too.

my2817: From the units test, the parser will treat "`add_t" as instance if there isn't a "else" condition
``` verilog
    var `add_t(foo) = '0;
```
I will check unpacked_dimension and interface syntax later, thanks.

parsers/verilog.c
```diff
@@ -1577,8 +1580,26 @@ static int tagNameList (tokenInfo* token, int c)
...
-		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`')
+		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`' || c == '(' || c == '[')
 		{
+			while(c == '[')
+			{
+				c = skipPastMatch ("[]");
+			}
+			if(c == '(')
+			{
+				c = skipPastMatch ("()");
```
hirooih, line 1591: We need `c= skipWhite (c);` here.

`m inst ()  ; // a space before semicolon` fails.   (You may want to add a test case for this. )

`skipPastMatch()` should do the job, but it does not now.  I will work on this issue on this weekend.
You may wait for the fix.

parsers/verilog.c
```diff
@@ -127,7 +128,8 @@ static kindDefinition VerilogKinds [] = {
  { true, 'p', "port",      "ports" },
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
- { true, 'b', "block",     "blocks (begin, fork)" }
+ { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance",  "instance(module instance)" }
```
hirooih, line 132: Please be consistent with the line above.  i.e. `"instance (module, instance)"`

parsers/verilog.c
```diff
@@ -140,6 +142,7 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
  { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance"   "instance(module or interface instance)"},
```
hirooih, line 145: ditto

hirooih (changes requested):
Thank you for your proposal.

Instance name support is very important. I completely missed the issue.
Please see my comments.

There is one issue.  If I understand your fix correctly, it tags a void function as a instance.

```
// LRM 13.4.1 Return values and void functions
myprint(a); // call myprint (defined below) as a statement
function void myprint (int a);  // FIXME
...
endfunction
```

Void function is not used often.  Supporting instance tag is much more important.  Leave it as known bug for now.

Please add the test case above, too.

Thanks.

parsers/verilog.c
```diff
@@ -1577,13 +1580,32 @@ static int tagNameList (tokenInfo* token, int c)
 	{
 		c = processType(token, c, &kind);	// update token and kind
 
-		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`')
+		if (c == '=' || c == ',' || c == ';' || c == ')' || c == '`' )
```
hirooih, line 1583: Remove a space you added.

parsers/verilog.c
```diff
@@ -1577,13 +1580,32 @@ static int tagNameList (tokenInfo* token, int c)
...
+				c = skipPastMatch ("()");
+				/* c= skipWhite (c); */
+				if(c == ';' || c == ',')
+				{
+					kind = K_INSTANCE;
+					verbose("find instance: %s with kind %s\n", vStringValue (token->name), getNameForKind(kind));
+					createTag (token, kind);
+				}
+				else
+				{
```
hirooih, line 1604: This else clause is not necessary.

parsers/verilog.c
```diff
@@ -1577,13 +1580,32 @@ static int tagNameList (tokenInfo* token, int c)
...
 		{
 			if (kind != K_UNDEFINED && kind != K_IDENTIFIER)	// ignore procedual assignment: foo = bar;
 				createTag (token, kind);
 			if (c == '=')
 				c = skipExpression (c);
 		}
+		else if (c == '(' || c == '[') // should be instance? TODO: function void foo();
+		{
+			c = skipDimension(c);
+			if(c == '(')
```
hirooih, line 1593: This if-clause is not necessary.

parsers/verilog.c
```diff
@@ -1577,13 +1580,32 @@ static int tagNameList (tokenInfo* token, int c)
...
+			{
+				c = skipPastMatch ("()");
+				/* c= skipWhite (c); */
+				if(c == ';' || c == ',')
+				{
+					kind = K_INSTANCE;
```
hirooih, line 1599: This variable is not necessary.

parsers/verilog.c
hirooih, line 1593: add a space after 'if'.

parsers/verilog.c
```diff
@@ -127,7 +128,8 @@ static kindDefinition VerilogKinds [] = {
  { true, 'p', "port",      "ports" },
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
- { true, 'b', "block",     "blocks (begin, fork)" }
+ { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance",  "instances" }
```
hirooih, line 132: If you chose "instances of module or interface" below, this should be "instances of module".

parsers/verilog.c
```diff
@@ -140,6 +142,7 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'r', "register",  "variable data types" },
  { true, 't', "task",      "tasks" },
  { true, 'b', "block",     "blocks (begin, fork)" },
+ { true, 'i', "instance"   "instances of module or interface"},
```
hirooih, line 145: Add a space before "}".

hirooih (changes requested):
Sorry for many change requests...

parsers/verilog.c
my2817: OK, I'm sorry for my poor coding skill 😭

parsers/verilog.c
```diff
@@ -1630,6 +1633,18 @@ static int tagNameList (tokenInfo* token, int c)
...
 				c = skipExpression (c);
 		}
+		else if (c == '(' || c == '[') // should be instance
+		{
+			c = skipDimension(c);
+			c = skipPastMatch ("()");
+			c = skipWhite (c);
+			if (c == ';' || c == ',')
+			{
+				kind = K_INSTANCE;
```
hirooih, line 1643: It seems that you missed my review comment this morning.
This variable is not necessary.  You can specify K_INSTANCE directly.

parsers/verilog.c
```diff
@@ -1630,6 +1633,18 @@ static int tagNameList (tokenInfo* token, int c)
 			if (c == '=')
 				c = skipExpression (c);
 		}
+		else if (c == '(' || c == '[') // should be instance
+		{
+			c = skipDimension(c);
+			c = skipPastMatch ("()");
+			c = skipWhite (c);
```
hirooih, line 1640: I've sent a pull-request #2718 for this.  I am waiting for the approval by @masatake.
If it is merged earlier than this pull-request, remove this line.

Again add a test case which fails by this issue.  (Add a space in ");" at the end of one of module instantiations.)

hirooih (changes requested):
It is getting simpler, isn't it?

parsers/verilog.c
my2817: Got it, thanks!

parsers/verilog.c
```diff
@@ -1630,6 +1633,20 @@ static int tagNameList (tokenInfo* token, int c)
...
+		{
+			c = skipDimension(c); // name_of_instance {unpacked_dimension}
+			c = skipPastMatch ("()"); // list_of_port_connections
+			c = skipWhite (c);
+
+			// if without the next "if" clause, get a instance named: `add_t from the following example
+			// var `add_t(foo) = '0;
+			if (c == ';' || c == ',')
+			{
+				verbose("find instance: %s with kind %s\n", vStringValue (token->name), getNameForKind(kind));
```
hirooih, line 1646: `kind` is in this line, too.

hirooih (changes requested):
See my comment.

> Again add a test case which fails by this issue. (Add a space in ");" at the end of one of module instantiations.)

For example;
verilog-instance.d/input.v: L57

```
             .b                         (b));
```

Change this line as;

```
             .b                         (b))   ;  // spaces before semicolon
```

Remove;
```
			c = skipWhite (c);
```

and confirm that the test fails.

And #2718 is merged into `universal-ctags:master`.
Merge it into your branch and confirm the test passes.

hirooih:
One more thing.
Have you checked the coverage report?  (It is a really awesome tool.)

https://codecov.io/gh/universal-ctags/ctags/compare/4a9f00a1c199133729819043292aedff7c580ed8...2069733b92a881ec74bd5419d186611e7432e870/changes

Your fix made the following lines unnecessary.

```
1643 | 1660 |   | /* skip port list of module instance: foo bar(xx, yy); */
1644 | 1661 |   | c = skipWhite (c);
1645 | 1662 |   | if (c == '(')
1646 | 1663 |   | {
1647 | 1664 |   | c = skipPastMatch ("()");
1648 | 1665 |   | c = skipWhite (c);
1649 | 1666 |   | }
```

Please remove them.

Your fix added very small number of lines and gave us great functionality.  It sounds great!

hirooih (approved):

hirooih:
Again, thank you for your contribution.

my2817:
> 
> 
> Again, thank you for your contribution.

Thanks for your patience!

masatake:
I'm working on VHDL and I have studied some concepts of this technology area.
I got some ideas to improve SystemVerilog parser.
```
module foo # (parameter
...
endmodule: foo
...

  foo uut7 [1:0][10:0]();
```

for uut7, ctags with this pull request emits:
```
uut7   input.v /^   foo uut7 [1:0][10:0]();$/;"        i       module:top
```

I think we can fill typeref: field for representing the relationship between uut7 and foo.
```
uut7   input.v /^   foo uut7 [1:0][10:0]();$/;"        i       module:top       typeref:module:foo
```

I'm taking the same approach in VHDL.

hirooih:
Thanks.
I've added the feature on the TODO list #2674 .

PR #2704, opened 2020-11-18
docs: fix simple typo, preceeded -> preceded

timgates42:
There is a small typo in parsers/cxx/cxx_parser_template.c.

Should read `preceded` rather than `preceeded`.


masatake:
Thank you.

Issue #2691, opened 2020-11-07
Remove combined pattern field?


AmaiKinono:
`--excmd=combine` is added for the request in ~#1890~#1125, to deal with https://github.com/vim/vim/issues/1057.

But at the end, vim didn't take that approach, but utilize `line:` field instead, see <https://github.com/vim/vim/commit/077b9dd3541339a23ade0cc6a23e804ee39312c5>. From what I saw in <https://github.com/vim/vim/blob/master/src/tag.c>, that's still how things work now.

Neovim takes the same approach: <https://github.com/neovim/neovim/blob/643f4a178751df34b27189f7aebea313b17bad3b/src/nvim/tag.c#L2805>.

Combined pattern field also somewhat denies the purpose of using a search pattern: to tolerate certain (actually a wide range of) source file updates. If we start search from the line above or below the tag, this will be easily broken. Personally I don't think it's meaningful.

So, it's safe to remove `--excmd=combine` since even vim doesn't use it, and that would have benefits:

- Keep the pattern field consistent with Exuberant Ctags.
- Client tools on non-vim editors may also wants to support the pattern field. This will make it easier.

masatake:
> --excmd=combine is added for the request in #1890, to deal with vim/vim#1057.

#1890?

#1125 and #1959 are related to this topic. Do I misunderstand?

It seems that the option becomes popular already: e.g. https://github.com/tmm1/ripper-tags/pull/98

AmaiKinono:
Ah yes, It's #1125.

Obviously, ripper tags implements it since u-ctags has it, but still, no client tools are really using it.

masatake:
@blueyed, do you really use `--excmd=combined` option?

blueyed:
@masatake 
No.  I have this commented in my config:
```
# Works great with https://github.com/vim/vim/issues/1057.
# --excmd=combine
# Smaller files.
# But too often offset mismatches.
# --excmd=number
```

masatake:
@blueyed, @mislav, @sshsu, do you have any ideas/commets about causing a trouble when we remove --excmd=combined?

I myself don't want to remove because I think it can make tags file compact comparing with using `line:' fields.

AmaiKinono:
> it can make tags file compact comparing with using `line:' fields.

As long as I know, you don't care about the search pattern, but only the line number. Then `--excmd=number` is better for you.

If you are talking about client tools in Emacs doesn't support `--excmd=number` , well, I will make that happen ;)

[...]

PR #2705, opened 2020-11-18
main: use actual line number in combined pattern fields

AmaiKinono:
See #2691. This makes it straightforward for client tools to "search the nearest occurence of pattern around the line", making good use of both the line number and search pattern.

I'll update #2698 once this is merged.

### DIFF
--- a/Tmain/excmd-combine-backward.d/stdout-expected.txt
+++ b/Tmain/excmd-combine-backward.d/stdout-expected.txt
@@ -1,2 +1,2 @@
-bar	input.cpp	3;?^bar(void)$?;"	f	typeref:typename:int	file:
-main	input.cpp	11;?^main(void)$?;"	f	typeref:typename:int
+bar	input.cpp	2;?^bar(void)$?;"	f	typeref:typename:int	file:
+main	input.cpp	10;?^main(void)$?;"	f	typeref:typename:int
--- a/Tmain/excmd-combine.d/stdout-expected.txt
+++ b/Tmain/excmd-combine.d/stdout-expected.txt
@@ -1,2 +1,2 @@
-bar	input.cpp	1;/^bar(void)$/;"	f	typeref:typename:int	file:
-main	input.cpp	9;/^main(void)$/;"	f	typeref:typename:int
+bar	input.cpp	2;/^bar(void)$/;"	f	typeref:typename:int	file:
+main	input.cpp	10;/^main(void)$/;"	f	typeref:typename:int
--- a/Tmain/extras-field-for-pseudo-tags.d/stdout-expected.txt
+++ b/Tmain/extras-field-for-pseudo-tags.d/stdout-expected.txt
@@ -1,7 +1,7 @@
 # option:  --format=1
 !_TAG_FILE_FORMAT	1	/original ctags format/
 !_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
-!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combine/
+!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combine(v2)/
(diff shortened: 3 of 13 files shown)


masatake:
Though I cannot find any strong reason, I would like to avoid a special character within the name of excmd: how about naming `combineV2` instead of `combine(v2)`?

AmaiKinono:
Done.

masatake:
It seems that you forgot to run `make -C man QUICK=1 update-docs`.

masatake (approved):

masatake:
@AmaiKinono, when merging, please use the "Squash and merge"  method for this change.

AmaiKinono:
OK.

Do we have a criteria of when to use which kind of merging?

masatake:
About the difference between "Create and merge" or "Rebase and merge", I have no word.
So I will write here about whether we should use "Squash merge" or not.

I would like to you make a pull request only with semantically meaningful commits. /* I cannot good word(s) for this. */
We should not make a pull request including a commit just for recording your development process.

Though I requested updating, 6649d2b  semantically meaningful commit.
acf5e2e represents your (our) development process.

When merging I would like you to unify acf5e2e and 6649d2b into one commit.
You can do it with "git rebsae -i master" from the command line or "Squash merge" on the web.
"fixup!" at the beginning of the commit header of acf5e2e implies you know what I wrote here.
Yes! "git rebase -i --autosquash msater" helps you. "git commit --fixup ..." is also useful. You used "--fixup", don't you?
When doing "git log", you will see I merged some "fixup!" commits. I merged them mistakenly. I should not do so.

About a876022, I wonder how we should do.
If you regard it as part of 6649d2b. In this case, we should unify  a876022, acf5e2e, and 6649d2 into one commit.
You can use the commit log of 6649d2 as is.

If you regard the updating of the man page as a separate work, you can keep a876022 as a separate commit from the unification commit of  acf5e2e and 6649d2. HOWEVER, when keeping a876022 as a separate commit, I would like you to update the commit log.  "fixup: make manpage" is not good. This should be something like "docs(man): update the description for --excmd=combine for v2 pattern". 

 "squashing" tasks can be done just before merging. 
However, I would like to do it incrementally during reviewing. So, during reviewing the commits, I can imagine what kind of commits will be introduced when merging. In that case, you can use "git commit --force push".

The very basic assumption is that one commit does one thing. This policy makes reviewing easier. It also makes rearranging commits and resolving conflicts when rebasing easier. See #2707 as examples.

I guess you already know about much of what I wrote here.
In my experience, when I wrote much to you, you did much much more good things. So I write much to you when I find time:-)

AmaiKinono:
> "git rebase -i --autosquash master" helps you. "git commit --fixup ..." is also useful. You used "--fixup", don't you?

I didn't know these before. Thanks for the tip!

I didn't use `--fixup`, but typed them manually. I did this since I've saw that prefix in other PRs before. So, I use `fixup:` but not `fixup!` for https://github.com/universal-ctags/ctags/commit/a876022693ca0dd51d40fc93ba27793955443574, because I thought it is a hint for human, not git.

> I guess you already know about much of what I wrote here.

Indeed. Actually I didn't put my question clear. I know we want these changes to go into a single commit, and it can only be done by "squash and merge" through the GitHub web UI. But using the command line, we could rearrange this branch first, then merge; or we could interactively rebase master against this branch. The difference is whether:

- we have an "actual" commit, plus a merge commit, and we keep the fact that this work is done on a separate branch. Or,

- we have only one commit, and the history is linear, looks like the work is done directly on `master`.

I use the second style in my projects, but I don't know which is preferred by u-ctags when the change is small. That's my question.

> In my experience, when I wrote much to you, you did much much more good things. So I write much to you when I find time:-)

That does happen when people work together. You are helping me know how are things done in this repo, and that increases the possibility of me trying to work more/deeper on u-ctags ;)

masatake:
>  I don't know which is preferred by u-ctags when the change is small. That's my question.

I don't care either way. Maybe there are merits and demerits. But I don't know them.

PR #2708, opened 2020-11-19
Move EXTERNAL_PARSER_LIST ifdef to the beginnning of BuiltInParsers

techee:
Commit

82791c20ff8d01e7055bcbbb30db4999d2c3d463

moved CTagsParser, FallbackParser and CTagsSelfTestParser above the
ifdef which made these parsers included also when using
EXTERNAL_PARSER_LIST - this is not good for Geany because these three
parsers are "invisible" for it and shift the EXTERNAL_PARSER_LIST by 3.

This patch moves the ifdef back where it was before.

### DIFF
--- a/main/parse.c
+++ b/main/parse.c
@@ -126,12 +126,12 @@ static parserDefinition *FallbackParser (void);
 static parserDefinition *CTagsParser (void);
 static parserDefinition *CTagsSelfTestParser (void);
 static parserDefinitionFunc* BuiltInParsers[] = {
-	CTagsParser,				/* This must be first entry. */
-	FallbackParser,				/* LANG_FALLBACK */
-	CTagsSelfTestParser,
 #ifdef EXTERNAL_PARSER_LIST
 	EXTERNAL_PARSER_LIST
 #else  /* ! EXTERNAL_PARSER_LIST */
+	CTagsParser,				/* This must be first entry. */
+	FallbackParser,				/* LANG_FALLBACK */
+	CTagsSelfTestParser,
 
 	PARSER_LIST,
 	XML_PARSER_LIST


main/parse.c
```diff
@@ -126,12 +126,12 @@ static parserDefinition *FallbackParser (void);
 static parserDefinition *CTagsParser (void);
 static parserDefinition *CTagsSelfTestParser (void);
 static parserDefinitionFunc* BuiltInParsers[] = {
-	CTagsParser,				/* This must be first entry. */
-	FallbackParser,				/* LANG_FALLBACK */
-	CTagsSelfTestParser,
 #ifdef EXTERNAL_PARSER_LIST
 	EXTERNAL_PARSER_LIST
 #else  /* ! EXTERNAL_PARSER_LIST */
+	CTagsParser,				/* This must be first entry. */
```
masatake, line 132: I am wondering the meaning of the comment. I cannot remember why it must be first.
I will try to remember.

masatake: As far as reading b9636b4db0e6c3e3f42ed1bd4be16b5c270315eb, there is no reason being CTagsParser at the fist of the list. Actually, it was not the first on the list. However, I would like to test before merging this pull request.

techee: At least for Geany it works without any problem when all these parsers are skipped.

main/parse.c
```diff
@@ -126,12 +126,12 @@ static parserDefinition *FallbackParser (void);
...
+	FallbackParser,				/* LANG_FALLBACK */
```
masatake, line 133: I confirmed we can move CTagsParser any place in the array.
However, FallbackParser is not. It must be the second in the array because the
parser is accessed with LANG_FALLBACK. LANG_FALLBACK is defined as 1 statically.

masatake:
I made the LANG_FALLBACK position-independent in a change. I tried to push the change to your branch. However, I cannot.
So I made a new pull request including your change. Could you look at  #2373  ?

I will close this one.

masatake:
I will merge this in soon.

masatake:
Thank you.
I must update some comments around the change.

PR #2710, opened 2020-11-19
verilog: Stop using non-ASCII chars

### LINKED ISSUE #2709: verilog.c report error C2220
Hello,

I'm using VC2019 C++ Build Tool to build the lastest version of ctags from Git repository. The commmand I use is "nmake -f mk_mvc.mak WITH_ICONV=yes ICONV_DIR=D:\gnuwin64 " But I got following error:

image

I can fix this issue by save the file verilog.c with utf-8-bom by Vim. But it will be good to solve it from Git repository.

k-takata:
Fix #2709

Suppress MSVC warning C4819. (The file contains a character that cannot
be represented in the current code page.)

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1125,7 +1125,7 @@ static void processFunction (tokenInfo *const token)
 		vUngetc (c);
 }
 
-// ( enum | union ) [ enum_base_type ] { < enum_name_declaration > }  { [ … ] }
+// ( enum | union ) [ enum_base_type ] { < enum_name_declaration > }  { [ ... ] }
 static void processEnum (tokenInfo *const token)
 {
 	int c;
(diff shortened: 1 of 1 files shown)


masatake:
Intersting. It will be nice if we can catch this kind of issues in misc/src-check:-)

hirooih:
Thanks!

I compiled the Verilog grammar on Google Docs for this project.

https://docs.google.com/document/d/1Sb0-iO9HofB7BR52XUarhGYPph6mFMx5oIFcU8eeFj0/edit?usp=sharing

The comment were copied from the page.  Google doc converts "...".

> Intersting. It will be nice if we can catch this kind of issues in misc/src-check:-)

I agree with you.
But I also think we should be able to use UTF8 on comments in 2020.

k-takata:
> But I also think we should be able to use UTF8 on comments in 2020.

Actually, MSVC 2015 or later supports `/source-charset:utf-8` and `/utf-8` options. However, this project still supports MSVC 2013. If we drop support for MSVC 2013, we can use them.
Another way is adding a BOM to each file (but I don't like it).

hirooih:
> However, this project still supports MSVC 2013.

Just a question from curiosity.  Is this documented anywhere?

> Another way is adding a BOM to each file (but I don't like it).

I agree. This is not our choice.

k-takata:
> Just a question from curiosity. Is this documented anywhere?

Here: https://docs.ctags.io/en/latest/windows.html#microsoft-visual-studio

hirooih:
I did not understand that MSVS 2013 meant Microsoft Visual Studio 2013.
And I never thought it did not support UTF-8 properly.

PR #2711, opened 2020-11-19
See #2708

masatake:
This is the continuation of #2708.


techee:
What is the fallback parser good for? Basically in Geany we always specify the parser to be used by ourself so I guess it won't be used in this case. Notice that when EXTERNAL_PARSER_LIST is defined, FallbackParser is in the `#else` part so it won't be used for Geany - then it doesn't really matter whether it's position-independent or not - it's just not used in this case.

masatake:
Your guessing is correct.
The parser is used only when
* ctags cannot guess a proper parser for input, or
* ctags chooses etags output mode.

masatake:
.... I understand what you wrote.
I understand the change I introduced is not needed.
I will merge #2708.

PR #2712, opened 2020-11-19
Add Haskell parser from Geany

techee:
Haskell <https://en.wikipedia.org/wiki/Haskell_%28programming_language%29>


masatake:
Can I ask you to provide a test case for LiterateHaskell?
This is the first time to hear about it.

techee:
> This is the first time to hear about it.

Yeah, same here. Not sure if anyone is crazy enough to use it:

https://www.haskell.org/onlinereport/literate.html

masatake:
If Haskel code is embedded in latex code, I think using Tex parser will be better.
Real lhs language processor may just ignore Latex code part.
However, ctags should make tags for the Latex code part, too.

Give me time.

techee:
The question really is whether the literate haskell part is any useful and whether it shouldn't be removed completely. I can't really imagine that anyone would code this way.

techee:
I've just checked and in Geany we don't use the literate haskell part. Personally, I'd just remove it. Should I do it?

masatake:
> I've just checked and in Geany we don't use the literate haskell part. Personally, I'd just remove it. Should I do it?

No, you should not do.

Instead, could you resolve the conflict?

masatake:
Thank you very much.

PR #2713, opened 2020-11-19
Add abc parser from Geany

techee:
Abc (https://en.wikipedia.org/wiki/ABC_notation)


parsers/abc.c
```diff
@@ -0,0 +1,85 @@
...
+{
+	vString *name = vStringNew();
+	const unsigned char *line;
+
+	while ((line = readLineFromInputFile()) != NULL)
+	{
+		if (line[0] == 'T') {
+			vStringCatS(name, " / ");
+			vStringCatS(name, (const char *) line);
+			makeAbcTag(name);
```
masatake, line 61: You can use makeSimpleTag declared in "parse.h" instead of defining a parser specific helper function.

PR #2714, opened 2020-11-19
Add haxe parser from Geany

techee:
Haxe <https://en.wikipedia.org/wiki/Haxe>


Units/parser-qemuhx.r/enable.d/args.ctags
```diff
@@ -1,4 +1,4 @@
 --sort=no
 --extras-QemuHX=+{funcmap}
 --fields=+E
-
+--language-force=QemuHX
```
masatake, line 4: Though this should be fixed, I would like you to delete this modification from this commit?
Can I ask you?

parsers/haxe.c
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	def->kindTable          = HxKinds;
+	def->kindCount  = ARRAY_SIZE (HxKinds);
+	def->parser             = findHxTags;
+	/*def->initialize = initialize;*/
+	return def;
+}
+
+
+
```
masatake, line 220: Could you remove these empty lines at the end of file?

Units/parser-qemuhx.r/enable.d/args.ctags
techee: Hmm, not really, the test fails without this

masatake: Oh, I see why this change is needed.
Ignore what I wrote.

masatake: I have to provide a command-line interface to define an arbitrator resolving this kind of conflict in the future

masatake:
Thank you.
I'm sorry to be late for work on this parser.

techee:
> I'm sorry to be late for work on this parser.

Don't worry, no rush, we are still 1.5 years behind universal ctags in Geany ;-).

Issue #2434, opened 2020-02-22
Markdown: use "" instead of " as scope separator like Tex parser


masatake:
https://github.com/universal-ctags/ctags/blob/bd612ba303a049ae72b157da63fef0766d1a162f/optlib/markdown.ctags#L29

PR #2716, opened 2020-11-19
Add txt2tags parser from Geany

techee:
Txt2tags (https://en.wikipedia.org/wiki/Txt2tags)


parsers/txt2tags.c
```diff
@@ -0,0 +1,195 @@
...
+#include "parse.h"
+#include "read.h"
+#include "nestlevel.h"
+#include "vstring.h"
+#include "routines.h"
+#include "entry.h"
+
+
+/* as any character may happen in an input, use something highly unlikely */
+#define SCOPE_SEPARATOR "\x3" /* ASCII ETX */
```
masatake, line 31: I would like to change this to `""`. Is this acceptable?

We have used `""` in the parser for markdown and tex.
I would like to use `""` as seperators for languages for documentation.

https://github.com/universal-ctags/ctags/issues/2434

techee: No problem for Geany, I'm just wondering, if `""` is a good choice as it isn't sufficiently unlikely to appear in text IMO. It's an empty string in programming languages so it could appear in texts about programming languages.

masatake: You are right. It isn't an ideal choice. I could not find such a better choice when working on markdown and tex.
However, the consistency between parsers dealing with languages for documents may make developing a client tool a bit easier.

So could you change the separator to `""`?

After merging this, I will introduce a definition:
```
#define DEFAULT_DOCLANG_SCOPE_SEPARATOR "\"\""
```

We have the following option as the last resort:
```
  --_scopesep-<LANG>=[parent_kind_letter]/child_kind_letter:separator
       Specify scope separator between <PARENT_KIND> and <KIND>.
       * as a kind letter matches any kind.
```

techee: Done.

I've also updated other-projects.rst to remove some outdated information.

masatake: Thank you.

masatake:
Thank you.

PR #2721, opened 2020-11-22
SystemVerilog,refactor: consistent input character handling

hirooih:
This fix makes input character handling consistent.
Until now, character input processing has been executed in an ad hoc way.

- Each function which executes character input has argument "c" and returns "c".
- It calls skipWhite() before return.

### DIFF
--- a/Units/parser-verilog.r/systemverilog-net-var.d/input.sv
+++ b/Units/parser-verilog.r/systemverilog-net-var.d/input.sv
@@ -50,7 +50,7 @@ module Net_declarations;
 
   // 6.7.2 Net declarations with user-defined nettypes
   nettype T wT;
-  nettype T wTsum with Tsum;  // FIXME: 
+  //nettype T wTsum with Tsum;  // FIXME: "nettype" and "with" are unsupported
   wT w1;
   wT w2[8];
   wTsum w3;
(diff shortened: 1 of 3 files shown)


hirooih:
This is the first half of a big refactoring.

@masatake san, this pull-request is 5 steps of refactoring changes but it changed basic way of input character handling.
Please review it.

Coverage is decreased due to sanity check.  So it is OK.

Units/parser-verilog.r/systemverilog-net-var.d/input.sv
```diff
@@ -50,7 +50,7 @@ module Net_declarations;
 
   // 6.7.2 Net declarations with user-defined nettypes
   nettype T wT;
-  nettype T wTsum with Tsum;  // FIXME: 
+  //nettype T wTsum with Tsum;  // FIXME: "nettype" and "with" are unsupported
```
masatake, line 53: This is o.k.

FYI
`.b` test can be a tool for recording this kind of situation.
https://docs.ctags.io/en/latest/units.html#gathering-test-cases-for-known-bugs

masatake (approved):

Units/parser-verilog.r/systemverilog-net-var.d/input.sv
hirooih: Thanks!

Issue #2724, opened 2020-11-23
SystemVerilog: entering infinite loop


masatake:
input.sv:
```
struct
{
	s*
```

cmdline:
```
$ ./ctags --language-force=SystemVerilog input.sv
```

hirooih:
Thanks.
I found the bug. 
I will check all loops not to being infinite loops on illegal inputs.

hirooih:
fixed by #2726

PR #2726, opened 2020-11-24
SystemVerilog: pushMembers() returns on illegal inputs

hirooih:
fix for #2724
test is added


hirooih:
Fix for #2726 
I've checked all other loops so that they always read something from input. They should not cause infinite-loop.  
I also add a test for this bug.

Please review this.

masatake (approved):

hirooih:
Thanks.

masatake:
I forgot to write one thing.

I wrote the input for a test case must be valid as in input for the target language.
However, there is an exception. You can put an invalid input as a test case when there is no "expected.tags" file.
No-expected.tags implies the test case is a crash-testing or is about an infinite loop.
(However, I recommend you to add a README to the test case in that case.)

"units" target run in a timer. So you don't afraid of adding a test case triggering an infinite loop.

hirooih:
I guess your comment is for #2733, not for this.

> However, there is an exception. You can put an invalid input as a test case when there is no "expected.tags" file.

I understand and understood what you wrote.  I added wrong input test case for this issue.

`    s* ; // pushMembers() caused infinite loop : #2724
`

But #2733 for #2730 is a different story as I commented on #2733 as;

>　I do not add test cases for this issue, because we need a unit case for each loop.

It doesn't make sense to add a test case for the failed loop first. It might be worth adding a test case for each loop that meight cause an infinite loop. However, we need to add dozens of units tests.

masatake:
I found I misunderstood the situation.
You are correct. Sorry for making the noise.

hirooih:
> Sorry for making the noise.

No problem. I always thank you for your valuable comments.

PR #2727, opened 2020-11-24
Julia: generate reference tags for external entities

AmaiKinono:
This is an unsuccessful try, but at least you can use it as a reference. I put all the julia `using/imported` forms in the comment, and created a test case (with the expected output).

I haven't write C code for 7 or 8 years, so maybe my code is too bad to fix, but if you think this can still be fixed and merged, here are the problems:

*(This part is outdated. See the added part below).*

> 1. I can't put `initRefTagEntry` to work. It just don't generate a tag. Fixed.
> 2. Now the library name is recorded with the `:` after it, and those without a `:` is recorded as unknown kinds. I guess I have to change `advanceToken` somehow to fix this. I don't know how to do it. This part is mostly broken, as you can see when manually run the test case.
> 3. The `roles:` kind are all `def` now, but I guess this could be fixed using `initRefTagEntry`.

Here are some minor considerations:

1. The file uses spaces to indent, which is not the coding style of u-ctags. Should we change this?
2. "Libraries" are actually called "modules" in Julia, but it's not bound to files, and you can define a module just like you define a function. So I choose "library" to not intefere with module definitions.

---

*Update:*

The patch is now partially fixed, but still has some problems. Run

``` console
$ ctags --options=args.ctags -f- input.jl
```

in `Units/parser-julia.r/library.d/` to see them.

1. The lexer doesn't jump accross dots. From other unit tests I guess this is the desired behavior, but then for `Library4`, it captures `Library4.func1` as a whole, which is unwanted.

2. Some pattern fields are generated for the wrong lines. I wonder why this happens...


masatake:
Please, run "make tmain" locally.  You must update "stdout-expected.txt" file(s).
About tmain, see https://docs.ctags.io/en/latest/tmain.html.
misc/review script may help you.

masatake:
What you are missing is `--extras=+r`. This makes ctags gather reference tags.
See https://github.com/universal-ctags/ctags/blob/master/Units/parser-python.r/python-import.d/args.ctags .

I will review this pull request more. You can do "push --force"  anytime you want.

masatake:
> I haven't write C code for 7 or 8 years

If the language is not Sexp based, all the languages are the same for us, aren't they?

Units/parser-julia.r/library.d/args.ctags
```diff
@@ -0,0 +1,2 @@
+--fields=Ks{kind}{roles}{scope}
+--sort=no
```
masatake, line 2: Add
```
--extras=+r
```

Units/parser-julia.r/library.d/expected.tags
```diff
@@ -0,0 +1,22 @@
...
+!_TAG_FILE_SORTED	0	/0=unsorted, 1=sorted, 2=foldcase/
+!_TAG_PROGRAM_AUTHOR	Universal Ctags Team	//
+!_TAG_PROGRAM_NAME	Universal Ctags	/Derived from Exuberant Ctags/
+!_TAG_PROGRAM_URL	https://ctags.io/	/official site/
+!_TAG_PROGRAM_VERSION	5.9.0	/16366c1c/
+!_TAG_OUTPUT_MODE	u-ctags	/u-ctags or e-ctags/
+!_TAG_OUTPUT_FILESEP	slash	/slash or backslash/
+!_TAG_PATTERN_LENGTH_LIMIT	96	/0 for no limit/
+!_TAG_PROC_CWD	/home/kino/uctags/Units/parser-julia.r/library.d/	//
+!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combineV2/
```
masatake, line 11: ptags are not needed in units testing. Please, remove them.

parsers/julia.c
```diff
@@ -846,6 +886,29 @@ static void addTag (vString* ident, const char* type, const char* arg_list, int
...
 
+static void addReferenceTag (vString* ident, int kind, int role, unsigned long line, MIOPos pos, vString* scope)
+{
+    if (kind == K_NONE)
+    {
+        return;
+    }
+    tagEntryInfo tag;
+    // HELP WANTED: for some reason initRefTagEntry doesn't work
+    // initRefTagEntry(&tag, vStringValue(ident), kind, role);
```
masatake, line 897: Use initRefTagEntry. 
`--extras=+r` emits the tag initialized with this function.

parsers/julia.c
```diff
@@ -846,6 +886,29 @@ static void addTag (vString* ident, const char* type, const char* arg_list, int
...
+    initTagEntry(&tag, vStringValue(ident), kind);
+    tag.lineNumber = line;
+    tag.filePosition = pos;
+    tag.sourceFileName = getInputFileName();
+    tag.extensionFields.signature = NULL;
+    //tag.extensionFields.roleBits = makeRoleBit(role);
```
masatake, line 903: You may not need this line.
 initRefTagEntry may set the bit.

parsers/julia.c
```diff
@@ -1180,7 +1253,6 @@ static void parseImport (lexerState *lexer, vString *scope, int parent_kind)
             advanceToken(lexer, true);
         }
     }
-
```
masatake, line 1183: Understandable. However, this deletion is nothing to do with generating reftags.
So
(1) don't delete thelin in this commit,
(2) (optional) make another commit for deletion.

Units/parser-julia.r/library.d/args.ctags
AmaiKinono: Done (is `--extras=+{reference}` acceptable?)

Units/parser-julia.r/library.d/expected.tags
AmaiKinono: Done.

parsers/julia.c
AmaiKinono: Done.

parsers/julia.c
AmaiKinono: Done.

parsers/julia.c
AmaiKinono: Done.

AmaiKinono:
> If the language is not Sexp based, all the languages are the same for us, aren't they?

Haha ;) Mostly true, but the `&`/`*` based one is different.

The patch is now partially fixed, but still has some problems. Run

``` console
$ ctags --options=args.ctags -f- input.jl
```

in `Units/parser-julia.r/library.d/` to see them.

1. The lexer doesn't jump accross dots. From other unit tests I guess this is the desired behavior, but then for `Library4`, it captures `Library4.func1` as a whole, which is unwanted.

2. Some pattern fields are generated for the wrong lines. I wonder why this happens...

masatake:
> 
   1.  I can't put initRefTagEntry to work. It just don't generate a tag.
   2. Now the library name is recorded with the : after it, and those without a : is recorded as unknown kinds. I guess I have to change advanceToken somehow to fix this. I don't know how to do it. This part is mostly broken, as you can see when manually run the test case.
   3. The roles: kind are all def now, but I guess this could be fixed using initRefTagEntry.


About 1. and 3. , see my review comments. 
For fixing 2. We have to inspect advanceToken() but asking the original author is the way of GitHub.
@getzze could you look at this pull request?
@AmaiKinono, let's wait for a while. In the case of the original author being busy, I will take a look.

> 1. The file uses spaces to indent, which is not the coding style of u-ctags. Should we change this?

I decided not to be too strict about the way of style because e-ctags was not so strict.
Doing C-M-\ on all files was one of the approaches. However, it makes reading the output of `git blame` harder.
So, in most of all files, I have kept the style as is. Especially if a file has a maintainer.
When making a new .c file, apply .editorconfig to your emacs.

If you want to edit an existing .C file, you have to do (1) \C-x \C-f the file, (2) M-x c-guess.
c-guess, that I contributed to cc-mode years ago, studies the style from the current buffer and update the style in c-mode.
The command has helped me contribute to many projects.

Units/parser-julia.r/library.d/args.ctags
masatake: YES.

AmaiKinono:
> `M-x c-guess`

Oh yes! This is really cool. Thank you!

masatake:
The following change fixes the pattern.

```
diff --git a/parsers/julia.c b/parsers/julia.c
index 1016cf72..58408e21 100644
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -885,14 +885,16 @@ static void addTag (vString* ident, const char* type, const char* arg_list, int
     makeTagEntry(&tag);
 }
 
-static void addReferenceTag (vString* ident, int kind, int role, vString* scope)
+static void addReferenceTag (vString* ident, int kind, int role, unsigned long line, MIOPos pos, vString* scope)
 {
     if (kind == K_NONE)
     {
         return;
     }
     tagEntryInfo tag;
-    initRefTagEntry(&tag, vStringValue(ident), kind, role);
+    initRefTagEntry(&tag, vStringValue (ident), kind, role);
+    tag.lineNumber = line;
+    tag.filePosition = pos;
     if (scope != NULL)
     {
         tag.extensionFields.scopeKindIndex = K_LIBRARY;
@@ -1221,14 +1223,14 @@ static void parseImport (lexerState *lexer, vString *scope, int parent_kind, int
     /* capture the imported name */
     advanceToken(lexer, true);
     vStringCopy(name, lexer->token_str);
-    addReferenceTag(name, K_LIBRARY, library_role, NULL);
+    addReferenceTag(name, K_LIBRARY, library_role, lexer->line, lexer->pos, NULL);
     if (lexer->cur_c == ':')
     {
         advanceChar(lexer);
         advanceToken(lexer, true);
         while (lexer->cur_token == TOKEN_IDENTIFIER || lexer->cur_token == TOKEN_MACROCALL)
         {
-            addReferenceTag(lexer->token_str, K_UNKNOWN, unknown_role, name);
+            addReferenceTag(lexer->token_str, K_UNKNOWN, unknown_role, lexer->line, lexer->pos, name);
             skipWhitespace(lexer, false);
             if (lexer->cur_c == ',')
             {
```

parsers/julia.c
```diff
@@ -1148,36 +1205,46 @@ static void parseModule (lexerState *lexer, vString *scope, int parent_kind)
...
         advanceToken(lexer, true);
-    }
-
-    while (lexer->cur_token == TOKEN_IDENTIFIER || lexer->cur_token == TOKEN_MACROCALL)
-    {
-        // addTag(lexer->token_str, vStringValue(name), NULL, K_IMPORT, lexer->line, lexer->pos, scope, parent_kind);
-
-        skipWhitespace(lexer, false);
-        if (lexer->cur_c == ',')
+        while (lexer->cur_token == TOKEN_IDENTIFIER || lexer->cur_token == TOKEN_MACROCALL)
```
masatake, line 1231: It seems a tab char may be used here. M-x untabify may be needed.

parsers/julia.c
```diff
@@ -1148,36 +1205,46 @@ static void parseModule (lexerState *lexer, vString *scope, int parent_kind)
...
+        unknown_role = JULIA_UNKNOWN_IMPORTED;
+    }
+    else /* if (token_type) == TOKEN_USING */
+    {
+        library_role = JULIA_LIBRARY_USING;
+        unknown_role = JULIA_UNKNOWN_USING;
+    }
     /* capture the imported name */
     advanceToken(lexer, true);
+    vStringCopy(name, lexer->token_str);
```
masatake, line 1225: `.` may be included in vStringValue (lexer->token_str);

So...
```
char * dot = strchr (vString(lexer->token_str_, '.');
if (dot)
{
     vString *library_part = vStringNewNInit(vString(lexer->token_str), dot -  vString(lexer->token_str));
     vString *unknown_part = vStringNewInit(dot + 1);
     /* make_ref_tag for library_part */
    /* make_ref_tag for unknown_part */
    /* delete the strings */
}
else
....

File: parsers/julia.c [hunk 8]
```diff
@@ -1148,36 +1205,46 @@ static void parseModule (lexerState *lexer, vString *scope, int parent_kind)
...
+            else
             {
                 advanceToken(lexer, true);
             }
-        }
-        else
-        {
-            advanceToken(lexer, true);
         }
     }
```
masatake, line 1250:
```
/*
    import Library4.func3, Library4.func4
-------------------------------^
*/
else if  if (lexer->cur_c == ',')
{
...
}
```

parsers/julia.c
AmaiKinono: OK. Didn't know `untabify` before. Actually I ran `c-guess` on this file and seems it didn't set up the correct style for me.

masatake: As far as I can remember, it doesn't distinguish tab and space.

parsers/julia.c
AmaiKinono: Thanks for the tip! I think I can finish this without touching the lexer now.

masatake:
> 2. "Libraries" are actually called "modules" in Julia, but it's not bound to files, and you can define a module just like you define a function. So I choose "library" to not intefere with module definitions.

This is the most important topic. 
Does `using Library1` mean using something defined with "module Library1"?
If yes, "Library1" in `using Library1` is a module. More precisely, it is a reference for the module.

They do not interfere with each other.
`Library1` in `using Library1` is a reference tag having the `module` kind with the `using` role.
`Library1` in `using Library`   is a definition tag having the `module` kind.

AmaiKinono:
Most of the things work well now. The only problem left is for something like:

``` julia
import Lib1.func1, Lib1.func2
```

It will create 2 tags for `Lib1` because it appears 2 times. This is probably acceptable, because semantically it's the same as 

``` julia
import Lib1.func1
import Lib1.func2
```

and generate `Lib1` for 2 times for the latter one looks acceptable for me.

> Does `using Library1` mean using something defined with "module Library1"?

Exactly.

> They do not interfere with each other.
> `Library1` in `using Library1` is a reference tag having the module kind with the using role.
> `Library1` in `module Library1` is a definition tag having the module kind.

Ok. I'll change it.

masatake:
> It will create 2 tags for Lib1 because it appears 2 times. This is probably acceptable,

Yes. 'Lib1' is referenced twice. Therefore two reference tags are needed.

masatake:
You must update Tmain/list-roles.d/stdout-expected.txt.

parsers/julia.c
```diff
@@ -1305,8 +1398,10 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
                 parseType(lexer, scope, kind);
                 break;
             case TOKEN_IMPORT:
-                parseImport(lexer, scope, kind);
+                parseImport(lexer, scope, TOKEN_IMPORT);
```
masatake, line 1401: Just a question: can we expect "import" and "using"  always appear at the top-level in an input source code?

I don't know Julia but I made a pseudo example:
```
function foo
{
    import Library1
}
```
This is an example that "import" doesn't appear at the top-level.

AmaiKinono: > can we expect "import" and "using" always appear at the top-level in an input source code?

There is no clear answer, but I think it's pretty safe to assume this.

The common use of `import` or `using` "inside something" is inside a module:

``` julia
module MyModule
using Base64
# This will work
println(Base64.base64encode("hi"))
end

# This will throw an error
println(Base64.base64encode("hi"))
```

In the terminology of Julia, "inside a module" is still considered "top level", after all it doesn't have a file-level scope. Current parser works fine on this.

Put it in a function will cause error:

``` julia
function f()
    using Base64
end
# => ERROR: LoadError: syntax: "using" expression not at top level
```

This suggests that julia wants to ensure `using` is at top level. But actually it isn't very strict on this.

``` julia
let a = 1
    using Base64
end
```

`let`, like `function` creates a scope. But the above code runs well.

Even inside a function, we could still use some trick:

``` julia
function f()
    @eval using Base64
end
```

This runs well.

`import` is the same as above.

Unfortunately the language reference is also not very clear on this, and I don't know the implementation of `using` or `import`. But based on common practice, I think it's pretty safe to assume `using` and `import` are at top level.

masatake: I see.

masatake:
Could you squash the commits into one commit, and do push --force ?

AmaiKinono:
Done.

getzze:
Hi,
sorry I am not following everything you are doing but I would like to suggest to put in the `input.jl` file imports with newline:
```julia
using Module6,
      Module7,
      Module8

using Module9: func1,
               func2,
               func3
```
To make sure these are parsed correctly. They are well parsed now but just in case maybe.

Also, ideally, when code is written inside a module:
```julia
module Module1
function func1() end
struct Struct1 end
end
```
the definitions should be in `Module1` scope, like `Module1.func1`, but I commented this functionality in `parseModule`(last three commented lines of the function) because it was not looking good in Geany. But I guess this should be uncommented in ctags.

Please let me know if you need my help.

getzze:
Just to let you know of a possibility in Julia:
```julia
using Module1
import Module1: func1
```
the whole `Module1` is imported with `using` and the specific function `func1` is imported with `import`.

AmaiKinono:
Thanks!

``` julia
using Module6,
      Module7,
      Module8
```

This isn't handled well in my experiment. I must fix it.

> ideally, when code is written inside a module...the definition should be in the module scope.

I haven't tried the commented code you mentioned, but let's do it in a separate patch.

However I do want to point out a problem with file-based tagging: If something is directly defined in a file (with no module wrapping it), and when that file is executed, it's in the module `Main`.

Now let's support we have `module-file1.jl` and `module-file2.jl`. Then another file may include those two and wrap them in a module:

```
module MyModule
include("module-file1.jl")
include("module-file2.jl")
end
```

Then when we `using MyModule`, we expect the symbol defined in these files are under the `MyModule` scope. I suspect this can't be done with ctags at least for now. The best thing we can do is attach a `Main` scope, or don't record a scope with them. I guess this is also a problem for Ruby? It doesn't have a file-level scope too, as far as I know.

> the whole `Module1` is imported with `using` and the specific function `func1` is imported with `import`.
`import` is used if you need to overload a function so it makes sense to have this sometimes.

I guess you could just `using Module1` and implement `Module1.func1`. Does that work? It seems to work in my experiment.

AmaiKinono:
The suggested test by @getzee is added (and passed). I also wrap them in a module defining expression, for 1. make sure they are tagged when inside a module, and 2. for fill the `scope` field of imported/used modules in the future.

AmaiKinono:
How about let the lexer stop at dots? For example, this in the test:

``` julia
function Base.ifelse(a::Int)
    println("bar")
end
```

could be tagged as `ifelse`, with `Base` being the scope. The patch in this PR could also become simpler.

This is also good for client tools. When you `import` a module, the names in the module have their own tags, and the module has its own tag. Suppose we want to jump to definition on `Mod1.func`, we could search tag whose name is `func` and scope is `Mod1`. Then if you extend the function like

``` julia
function Mod1.func ()
  ...
```

then we won't find this, because it's name is `Mod1.func`, not `func`.

getzze:
Thank you for working on this.

>> I haven't tried the commented code you mentioned, but let's do it in a separate patch.
Yes, I think it should go in another patch.

>> Now let's support we have module-file1.jl and module-file2.jl. Then another file may include those two and wrap them in a module

I would say that for now it's better not to put any scope, instead of putting a Main scope, so potentially the function defined in the file could be assigned to any module.

>> I guess you could just using Module1 and implement Module1.func1. Does that work? It seems to work in my experiment.
I just meant that this case is common in Julia, so it should probably be accounted for in `input.jl` and made sure that it's correctly parsed.

getzze:
> How about let the lexer stop at dots? For example, this in the test:
> 
> ```julia
> function Base.ifelse(a::Int)
>     println("bar")
> end
> ```
> 
> could be tagged as `ifelse`, with `Base` being the scope. The patch in this PR could also become simpler.
> 
> This is also good for client tools. When you `import` a module, the names in the module have their own tags, and the module has its own tag. Suppose we want to jump to definition on `Mod1.func`, we could search tag whose name is `func` and scope is `Mod1`. Then if you extend the function like
> 
> ```julia
> function Mod1.func ()
>   ...
> ```
> 
> then we won't find this, because it's name is `Mod1.func`, not `func`.

That's a good idea. Do you think this change should be included in this patch or in a new one (maybe with the correct scope attribution inside a module that we mentioned before)?

AmaiKinono:
My thought is we should introduce this in another PR, without touching the reference tag part. Then after it's merged, we rewrite the patch in this PR to work on the new lexer, and attach `scope` field to imported/used modules.

masatake:
```
function Mod1.func ()
```
This is one of the hard issues in ctags.

The*ideal* parser makes a tag: `func` with `scope:Mod1`.

If `--extras=+q` is given, ctags can make the following tag in addition to the above one:
`Mod1.func`.

See parser-ruby.r/ruby-namespaced-class.d/input.rb and its expected.tags.

@AmaiKinono, I recommend you to study cork API before hacking a parser.
https://docs.ctags.io/en/latest/internal.html#cork-api

With the API, you can modify a tag entry after calling makeTagEntry().

AmaiKinono:
@masatake I've read a bit about the cork queue. I have one question for now: where does the name "cork" come from? I couldn't find this concept on Internet.

It may take me few days before starting work on this.

masatake:
@AmaiKinono, you cannot find the name IN the Internet. You can find CORK in one more upper layer. 
The name comes from TCP CORK.  TCP CORK is well explained in "A quick look at the TCP_CORK API" in https://baus.net/on-tcp_cork/. WIth TCP CORK, you can delay sending out data passed to Linux via write system call till removing TCP CORK.
Like TCP CORK, you can delay writing out tag entries passed to ctags' main part via makeTagEntry().
When thinking about what we can do with the ctags' cork API, I cannot say CORK is the best naming. However, I think giving a unique name is not so bad idea.

masatake:
```
module MyModule
include("module-file1.jl")
include("module-file2.jl")
end
```
> Then when we using MyModule, we expect the symbol defined in these files are under the MyModule scope.

This is similar to "#include <stdio.h>" in C language.

What we have to do as the first is:
* defining "script" kind (or something like that having a better name)
* defining "included" kind 
* make reference tags for the file names with the above kind and role.

You can add more tags in the second pass if you have "multipass parsing".
Even so, tagging the files are needed in the first pass.

masatake:
I could track only some of the comments but...you don't need to put all the ideas into one pull request.

AmaiKinono:
I think our CI has some problems on macOS...

masatake:
> I think our CI has some problems on macOS...

In such a case, trigger "Re-run all jobs". With my account, I can do it. I wonder whether you can do it with your account or not.

parsers/julia.c
```diff
@@ -44,19 +44,54 @@ typedef enum {
...
+
+/*
+*  using X               X = (kind:module, role:using)
+*
+*  using X: a, b         X = (kind:module, role:using)
+*                     a, b = (kind:unknown, role:using, scope:module:X)
+*
+*  import X              X = (kind:module, role:imported)
+*
+*  import X.a, X.b       X = (kind:module, role:imported)
```
masatake, line 69: About the role for X, I think "importedPartially" or something is better than "imported".

Even if c is defined in X, c is not visible in the source where "import X.a, X.b" is written.
"importedPartially" can represent this situation.

If we use "imported" role here, about "X" we cannot distinguish "import X" and "import X.a, X.b" from a tag file.

I'm not sure "importPartially" is the best. "importIndirectly" is another name I can find.

AmaiKinono:
> I wonder whether you can do it with your account or not.

I can. In fact, I already tried it several times, but that didn't help. Also, the CI on the recent merge commit also ran into problem on macOS:

> GitHub Actions has encountered an internal error when running your job.

So maybe it's a problem on GitHub's side.

masatake:
I disabled macos-11.0 at `master`. Could you rebase your change on `master` and push --force here ?

AmaiKinono:
Done.

masatake:
I'm very sorry to be late.

I rethink the kinds and roles. How about these ones?
```Julia
# your proposal:
#   Lib: kind:module,roles:using
# my idea:
#   Lib: kind:module,roles:used
using Lib

# your proposal:
#   BigLib: kind:module,roles:using
#   thing1: kind:unknown,roles:using
#   thing2: kind:unknown,roles:using
# my idea:
#   BigLib: kind:module, roles:namespace 
#   thing1: kind:unknown,roles:used
#   thing2:> kind:unknown,roles:used
using BigLib: thing1, thing2

# your proposal:
#   Base: kind:module,roles:imported
#   thing1: kind:unknown,roles:imported
# my idea:
#   Base: kind:module, roles:namespace
#   thing1: kind:unknown,roles:imported
import Base.show
```

I think `used` is better than `using`. It is similar to `imported`.
I introduced a role not having much meaning: namespace.
Let's pick one as an example.
See `Base`.
For `import Base.show`, you assign `imported` as a role to `Base`. However, I think `Base` is not imported.
```
[jet@living]~/var/ctags-new% julia
julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.3 (2020-11-09)
 _/ |\__'_|_|_|\__'_|  |  Fedora 33 build
|__/                   |

┌ Warning: Terminal not fully functional
└ @ Base client.jl:390
julia> import Test.Pass
import Test.Pass

julia> Test
Test
ERROR: UndefVarError: Test not defined
Stacktrace:
 [1] top-level scope
 [2] run_repl(::REPL.AbstractREPL, ::Any) at /builddir/build/BUILD/julia/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288

julia> Pass
Pass
Pass
```

`Test` is not imported.  `Pass` is imported.
This is applicable to `Base.show`. `Base` is not imported. 
`Base` is just used for specifying show. So I introduced a meaningless role `namespace` and assigned it to `Base`.
I took this idea from python parser.

Acceptable?

If yes, it is better to have one page like man/ctags-lang-python.1.rst.in.

AmaiKinono:
Totally agree. Using a `namespace` role is also consistent with the Python parser (in the `from X import Y` case).

AmaiKinono:
@masatake I've finished it. I still want to do some refactor because the code parsing `import/using` expressions becomes a bit messy. You can read the added manpage first.

masatake:
@AmaiKinono, the man looks good to me.

masatake:
The unknown kind will be resolved in the future with #2741 :-P.

masatake (approved):
LGTM. Please, merge the commits with "Squash and merge" button.

PR #2728, opened 2020-11-25
Clojure: Fix skipMetadata not called for functions

Cnly:
Functions can also have metadata, but `skipMetadata` is not called for them. That causes functions like

```clj
(defn ^:private test-fn [])
```

to be parsed as `^:private` instead of `test-fn`. This PR fixes the problem.


masatake:
Thank you. Could you consider to add a unit test case?

https://docs.ctags.io/en/latest/units.html#how-to-write-a-test-case explains the way to add a test case.

Cnly:
@masatake Sure, should be done now.

masatake:
Thank you!

Issue #2697, opened 2020-11-14
Proposal: A possible way of incrementally updating tags file


AmaiKinono:
We had some discussion before, see https://github.com/universal-ctags/ctags/issues/2475#issuecomment-621366924. @masatake suggested the following scheme:

1. Drop tag entries for updated files (reported by git?) and write the remaining tag entries to a temporary tag file (tmp tag file),
2. Appende tag entries generated by running parsers for the updated files, and writing them to a new tag file,
3. Merge and Sort the two tags files.

2 and 3 can be done by the `-a` option.

I want to suggest a way to do 1 in this proposal. My thought is we use a group of pseudo tags to record the last modified time for each source file. It may look like this:

```
!_SOURCE_FILE_TIMESTAMP!/project/path/something.c	2020-11-11 23:33:28.908535595 +0800	/last modified time/
```

Then we offer an option like `--increment` to do incremental update. With this option, we compare this pseudo tag with the actual timestamp, and find files that are and are not updated (new files are counted as updated ones). This option should do nothing when generating a tags file for the first time.

For writing the temporary tags file, maybe we could use readtags like this:

``` console
$ readtags -t tags -Q '(not (#/(updated-file1)\|(updated-file2)\|.../ $input))' -Ene > temp_tags
```

Another problem is to ensure that the command line options are the same. A possible way we've discussed is to also put them in the pseudo tags, and @masatake said this is related to "multipass parsing". I'm not familiar with that idea, but for the scope of incremental tags file update, I think if people are doing this, then the command line is probably generated by a script/makefile/client tool/ctags configuration file, so it wouldn't become a problem if we ask for same options between multiple runs, rather than letting ctags ensure it.

masatake:
This includes multiple sub issues.
I made a Wiki page for this topic.
https://github.com/universal-ctags/ctags/wiki/Designing:-updating-tags-file-incrementally
@AmaiKinono, I assume you can edit the page.

I think about the initial goal.
"Adding -C action to readtags" is not so hard task but we need anyway.

```
$ readtags -t a.tags -C
--fields={...} --kinds-C=... ...
```

`-C` option reproduces the command line used for generating an existing tags file.
The option refers pseudo tags.

For achieving the goal, we have to do:
A. adding more pseudo tags
B. extending readtags command

masatake:
I found I implemented TAG_EXTRA_DESCRIPTION and TAG_FIELD_DESCRIPTION ago.

```
[...]

PR #2732, opened 2020-11-26
main: add timestamp field for file kind

masatake:
*The field name is changed from "tstamp" to 'epoch".* 

main,units.py: introduce "epoch:" field for ~f~F/file kind
    
    For implementing the feature "updating the tags file" in the future,
    this change introduces "epoch:", a field which records the last
    modified time of the input file. This field is enabled by default.
    
    Example output:
    
        $ ./ctags --extras=f --kinds-C= -o - main/main.c
        main.c      main/main.c     1;"     F       language:C      epoch:1078071702
    
    The resolution of time is second in the current implementation.
    For Using higher resolution timestamp in the future, ".000000000" is suffixed.
    
    NOTES:
    
    The "fileScope" member of tagEntryInfo structure was not used anywhere in the
    ctags source tree. So this change renames it to "timestamp".
    
    timestamp fields in tags output makes many test cases in Units not work.
    So this change puts "--fileds=-{epoch}" to units.py for turning off
    the field when running "make units".
    
### DIFF
--- a/Tmain/epoch-field.d/run.sh
+++ b/Tmain/epoch-field.d/run.sh
@@ -1,0 +1,34 @@
+# Copyright: 2020 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
+O=/tmp/ctags-tstamp-$$.c
+
+cat > $O <<EOF
+int main (void)
+{
+	return 0;
+}
+EOF
+
+is_json_avaiable()
+{
+	$1 --quiet --options=NONE --with-list-header=no --list-features | grep -q "json"
+}
+
+TZ=UTC+00:00 touch -t '200402291621.42' $O &&
+	$CTAGS --kinds-C= --extras=f --fields=T -o - $O \
+		| sed -e 's/.*\(epoch:.*\)/tags:\1/' &&
+	$CTAGS --kinds-C= --extras=f --fields=T -o - -x --_xformat="xref:epoch:%T" $O &&
+	{
+		if is_json_avaiable $CTAGS; then
+			$CTAGS --kinds-C= --extras=f --fields=T -o - --output-format=json $O \
+				| sed -e 's/.*"epoch": \([0-9]*\).*/json:epoch:\1/'
+		else
+			echo "json:epoch:1078071702"
+		fi
+	}
+s=$?
+rm $O
+exit $s
--- a/Tmain/epoch-field.d/stdout-expected.txt
+++ b/Tmain/epoch-field.d/stdout-expected.txt
@@ -1,0 +1,3 @@
+tags:epoch:1078071702
+xref:epoch:1078071702
+json:epoch:1078071702
--- a/Tmain/extras-field.d/run.sh
+++ b/Tmain/extras-field.d/run.sh
@@ -5,4 +5,3 @@ CTAGS=$1
 
 ${CTAGS} --quiet --options=NONE --fields=+Ere --extras=+qrf -o - input.cpp \
     | sed -e 's|[^	]*\(input.cpp\)|\1|'
-
--- a/Tmain/extras-field.d/run.sh
+++ b/Tmain/extras-field.d/run.sh
@@ -3,5 +3,5 @@
 
 CTAGS=$1
 
-${CTAGS} --quiet --options=NONE --fields=+Ere --extras=+qrf -o - input.cpp \
+${CTAGS} --quiet --options=NONE --fields=+Ere-T --extras=+qrf -o - input.cpp \
     | sed -e 's|[^	]*\(input.cpp\)|\1|'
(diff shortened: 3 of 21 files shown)


b4n:
Just a thought: mtimes are kind of fragile as they can change without the content changing and vice versa.  I'm not necessarily aware of the latest techniques to identify changes that recent build systems might use, but something like a hash of the file would be more robust, and as we'll read the file anyway generating it should be fairly cheap.

That's just a thought, and I'm actually a happy Make user so timestamps are fine with me :)  I'm just wondering if we shouldn't use a more robust/recent technique if we're introducing the feature.

masatake:
@b4n, thank you. I have needed your input.

I was also thinking about using content hash values.
I temporarily choose the timestamp because implementing it is so easy, as this pull request shows.
Switching to using the hash values later is not hard, and the timestamp field itself is useful.
Enabling the timestamp field by default is a bit questionable.

masatake:
https://stackoverflow.com/questions/12765650/mingw-msys-msvcrt-and-the-tz-environment-variable

masatake:
https://docs.microsoft.com/ja-jp/cpp/c-runtime-library/reference/tzset?view=msvc-160 (Japanese)

masatake:
On Windows, the environment variable TZ cannot be passed from a test case to the ctags process started from the script.
However, an environment variable TZX introduced just for testing can be passed.
Both TZ and TZX has the same value: Asia/Tokyo.
Why?
Should I call _tzset?

masatake:
Calling _tzset does nothing.

masatake:
`%z` in strftime on Windows doesn't print a numerical zone offset. I'm surprised at this.
I must add code for printing numerical zone offsets to my change by referring `_timezone`.

The question is whether strptime works well for the crafted output.
I wonder if just using an epoch is better than using a decoded string.

AmaiKinono:
I vote for (machine-friendly) epoch since the user is not likely to need to *read* this field. Even if they need the field, it's likely to be processed by a script/program.

masatake:
I have learned many things through this pull request.

Content-hash is nice but it requires reading the content of the file when COMPARING.
So having a timestamp field still has sense.

About timestamp, I would like to use the break-down format like "2020-12-06 02:51:27.926418871 +0900" because it is readable.
However, I found implementing it was too hard for me.

So I'm thinking about using epoch. However, the current implementation I have tried here is not good enough.
```
$ ./ctags --list-fields | grep epoch
T       epoch          yes     NONE             s--    no    the last...
```
It is represented in a string. It should be an integer.

For the future extension about the resolution of timestamp, I used floating-point format for the value of the epoch field.
```
        $ ./ctags --extras=f --kinds-C= -o - main/main.c
        main.c      main/main.c     1;"     F       language:C      epoch:1078071702.000000000
 ```
I wonder how to deal with the value in readtags.
I don't want to introduce double or float data types to the qualified language in readtags.
My idea is splitting the epoch field into two: epoch and epochLow.
With the idea the output can be:
```
        main.c      main/main.c     1;"     F       language:C      epoch:1078071702      epcohLow:000000000
```

What I don't want is increasing the sizeof tagEntryInfo. Eventually, I may have to use union.

AmaiKinono:
> For the future extension about the resolution of timestamp, I used floating-point format for the value of the epoch field
> I wonder how to deal with the value in readtags.

I don't know the specific meaning of "dealing with the value in readtags", but if all we want to do is comparing 2 timestamps, maybe splitting them at the dot and compare the 2 parts could work?

b4n:
> For the future extension about the resolution of timestamp, I used floating-point format for the value of the epoch field.
> 
> […]
> 
> I wonder how to deal with the value in readtags.
> I don't want to introduce double or float data types to the qualified language in readtags.

If you don't want to use a floating point value, what about just using a higher resolution for the timestamp?  Like instead of seconds, make it milliseconds or even nanoseconds.  Assuming a signed 64 bits value would still give a comfortable 292 years range in nanoseconds (`((1<<63)-1) / 1e9 / (365*24*60*60)`).
Using seconds is convenient for consumers because it's the "normal" value, and can be used with e.g. `stat %Y`, but I guess any other unit is gonna be fine so long as it's known.

I'm not sure it's a good idea, but it came to mind.

> My idea is splitting the epoch field into two: epoch and epochLow.

That seems a little convoluted to me, but would work, and a consumer would get a conventional seconds-based timestamp, and could fine-tune with the finer granularity if needed.  So maybe it's a good idea indeed, if we really want sub-second precision.

masatake:
@b4n, thank you for your comment.

`epoch:<seccons>` may be enough as the basis for exploring the way to solve #2697.
`stat -c %Y` is a really helpful tip to utilize `epoch:` field. We must write it to our man page.

I will update this pull request.

PR #2734, opened 2020-11-26
php: Add support for PHP 7.3 relaxed HereDoc/NowDoc syntax

### LINKED ISSUE #2717: PHP heredoc (<<<) syntax breaks tags creation
##### SUMMARY:

Tags generation stops when the PHP `heredoc` (`<<<`) syntax is encountered in a file. As the `nowdoc` PHP syntax is basically the same, that's another language element that breaks the file parsing.

##### The name of the parser: 

not sure about this. Assuming PHP

##### The command line you used to run ctags:

```
$ ctags --options=NONE foo.php
```

##### The content of input file:

```PHP
<?php

class LivingBeings {

    public function doSomething()
    {
        $foo = <<<FOO
        FOO;
    }

    public function doSomethingElse()
    {
    }
}
```

##### The tags output you are not satisfied with:

The `doSomethingElse` method is not listed in the file. As soon as I comment out the `heredoc` portion, the method is indexed normally, as you can see in the "expected output" section ahead.

```
!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combine/
!_TAG_PROC_CWD	/tmp/	//
!_TAG_PROGRAM_VERSION	5.9.0	/5a136315/
LivingBeings	foo.php	/^class LivingBeings {$/;"	c
doSomething	foo.php	/^    public function doSomething()$/;"	f	class:LivingBeings
```

##### The tags output you expect:

```
!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combine/
!_TAG_PROC_CWD	/tmp/	//
!_TAG_PROGRAM_VERSION	5.9.0	/5a136315/
LivingBeings	foo.php	/^class LivingBeings {$/;"	c
doSomething	foo.php	/^    public function doSomething()$/;"	f	class:LivingBeings
doSomethingElse	foo.php	/^    public function doSomethingElse()$/;"	f	class:LivingBeings
```

##### The version of ctags:

```
$ ctags --version
Universal Ctags 5.9.0(5a136315), Copyright (C) 2015 Universal Ctags Team
  Compiled: Nov 20 2020, 11:46:20
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +yaml, +packcc
```

##### How do you get ctags binary:

Building it locally:

```
$ cd ctags_source
$ make clean && make distclean
$ ./autogen.sh
$ ./configure --prefix=$HOME
$ make
$ make install
```

b4n:
See: https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes

Fixes #2717.


b4n:
I did *not* keep support for PHP < 7.3 syntax, because it seems a little too tricky to support both for not much gain, as I don't really expect much breakage anywhere (the syntax is very close, and the cases changing the behavior should really be extremely rare).  And if really some happen, I think we can reopen the subject then.

masatake:
Don't you need to update the comment for parseHeredoc()?

https://github.com/universal-ctags/ctags/blob/7f84a4438538080fd6c598df74f5d362881d222a/parsers/php.c#L613

```
3) closing ID must be immediately followed by a newline or a semicolon
 *     then a newline.
```

masatake:
If a commit includes only a test case, could you use "Units" or "Units(PHP)" as the prefix for the commit header?

b4n:
@masatake indeed, fixed both.  I also changed the additional test case to be CR-only, because I realized that main part normalizes CRLF into CR so there's nothing to test there specific to the PHP parser (which is the reason why coverage didn't increase with that commit).

I plan to update the code to remove specific handling of CRLF as it's redundant, but as there are other instances of it I though of doing it in a later commit.  If you prefer me not to include this redundant code in this PR, I can remove the parts I'm adding and only strip the other one separately.

b4n:
> I plan to update the code to remove specific handling of CRLF as it's redundant, but as there are other instances of it I though of doing it in a later commit. If you prefer me not to include this redundant code in this PR, I can remove the parts I'm adding and only strip the other one separately.

Actually there's no point to have this code just to remove it soon, so I dropped it now.

masatake (approved):

PR #2736, opened 2020-11-27
PHP: tests and corner cases

b4n:
This mostly adds tests to improve code coverage, plus 2 code changes:
1. Improve handling of `<?=` short echo tag so it doesn't risk generating spurious tags;
2. Remove redundant handling of CRLF line endings that the read.c already normalizes.

### DIFF
--- a/Units/parser-php.r/coverage.d/args.ctags
+++ b/Units/parser-php.r/coverage.d/args.ctags
@@ -1,0 +1,1 @@
+--fields=+im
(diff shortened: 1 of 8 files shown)


masatake (approved):

b4n:
I added an couple extra lines after seeing some code paths still weren't covered but could.
There won't be 100% coverage, because the parser includes some fallback code for invalid input, and there's no good test for that, but I'm trying to get full coverage of the normal code paths, which should be close to 100% anyway.

masatake (approved):

PR #2737, opened 2020-11-28
SystemVerilog: implement unsupported features in processDesignElement()

hirooih:
- processDesignElement() is divided into processDesignElementL() (for module, interface, and program) and processDesignElementS() (for others).
- tagging imported functions is implemented
- skipping package import declaration is implemented
- implement "interface class" support
- implement coverage_event handling
- units tests are heavily updated

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -432,24 +432,6 @@ static short isTempContext (tokenInfo const* token)
 	}
 }
 
-static short hasSimplePortList (verilogKind kind)
-{
-	switch (kind)
-	{
-		case K_TASK:
-		case K_FUNCTION:
-		case K_CHECKER:
-		case K_CLASS:
-		case K_INTERFACE:
-		case K_PROGRAM:
-		case K_PROPERTY:
-		case K_SEQUENCE:
-			return true;
-		default:
-			return false;
-	}
-}
-
 static void clearToken (tokenInfo *token)
 {
 	token->kind = K_UNDEFINED;	// to be set by updateKind()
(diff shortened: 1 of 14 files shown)


parsers/verilog.c
```diff
@@ -247,7 +252,8 @@ static const keywordAssoc KeywordTable [] = {
 	{ "typedef",       	K_TYPEDEF,   	{ 1, 0 } },
 	{ "union",         	K_STRUCT,    	{ 1, 0 } },
 	{ "var",           	K_REGISTER,  	{ 1, 0 } },
-	{ "void",          	K_REGISTER,  	{ 1, 0 } }
+	{ "void",          	K_REGISTER,  	{ 1, 0 } },
+	{ "with",          	K_WITH,			{ 1, 0 } }
```
masatake, line 256: (just comment) I recommend adding "," at the end. So we can make diff smaller when you add a new keyword next time.

masatake:
How should we handle "import", "using", "load", "require", "include", etc is one of the hard topic in ctags.
Of course, skipping it is the first step.  If you have time, please, watch #2697 and #2727.

parsers/verilog.c
hirooih: I know modern languages allow this, but I have been misunderstanding that C did not allowed this.

I've checked the K&R both the 1st and the 2nd edition, Most of examples do not use semicolon after the last item, but syntax rule says;

```
initializer:
	assignment-expression
	{ initializer-list }
	{ initializer-list , }


initializer-list:
	initializer
	initializer-list , initializer
```

Verilog/SystemVerilog does not allow this, it makes difficult scripting...

Thank!!!

hirooih:
> How should we handle "import", "using", "load", "require", "include", etc is one of the hard topic in ctags.

For import of System Verilog, it is not hard topic.  Adding a field should be enough.  I will add it on the TODO this.

PR #2740, opened 2020-11-28
R6class: avoid entering infinite loop

### LINKED ISSUE #2739: R: entring infinite loop
```
e <- R6Class(public = list({}}
```

masatake:
Close #2739.


Units/parser-r6class.r/infinite-loop.d/input.r
```diff
@@ -0,0 +1,2 @@
+# NOTHING HERE
+
```
masatake, line 2: The last empty line is not needed.

Issue #1960, opened 2018-12-19
go to definition of function within current class


baiwfg2:
Hi. I have a C++ member function `foo` in class A, `void A::foo();`. And I call it in a function `bar`.
```
void bar() {
  a.foo();
}
```
When I put cursor under `foo` and press `g + ]`, I expect to jump exactly to the `foo` definition of class A, not the function of same name in other classes. This will greatly reduce the number of entries in the selection list when there're many same function name across the project classes. Would ctags be able to do that ? Or I may need to find other plugin for vim to do that ?

(
Thank you for contacting us.

If you are reporting an issue with the parsing output, please fill
the following template.  As your custom CTags configuration can
affect results, please always use `--options=NONE` as the first
option when running `ctags`.

Otherwise, delete the template and write your issue from scratch.
Examples may help developers understanding your issue better.

Use GitHub web interface and markdown notation.
Using mail results broken text rendering that makes
the developers go crazy.
)

*****

The name of the parser:

The command line you used to run ctags:

```
$ ctags --options=NONE ...
```

The content of input file:

```C
/* THIS IS AN EXAMPLE */
int
main(void)
{
	...
```

The tags output you are not satisfied with:

```
!_THIS_IS_AN_EXAMPLE
mainVoid	foo.c	/^main(void)$/;"	kind:function	line:2	language:C	typeref:typename:int	signature:(void)	roles:def
...
```

The tags output you expect:

```
!_THIS_IS_AN_EXAMPLE
main	foo.c	/^main(void)$/;"	kind:function	line:2	language:C	typeref:typename:int	signature:(void)	roles:def
...
```

The version of ctags:

```
$ ctags --version
Universal Ctags 0.0.0(EXAMPLE), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: May 11 1018, 23:16:36
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml, +aspell
```

How do you get ctags binary:

(
Building it locally, via GNU/Linux distribution, as BSD's package,
win32 binary taken from Universal-ctags/ctags-win32 project, macosx
binary taken from Universal-ctags/homebrew-universal-ctags project,
etc.
)

masatake:
[...]

PR #2741, opened 2020-11-28
Experimental implementation of multi-pass parsing

masatake:
This pull request introduces `--_hint=<tag file>` option and internal APIs for utilizing the given tags file.
A parser can use the pre-existing tags file for improving the quality of parsing and tagging with the APIs.

This option is not for incremental updating.
Even you specify `--_hint=<tag file>`, ctags parsers all input files.

Python parser is the initial target for applying the APIs.
In the first pass, the Python parser attaches "unknown" kind to X in "from Y import X".
With the hint, the Python parser can resolve the real kind for X.


masatake:
* https://github.com/universal-ctags/libreadtags/pull/25 must be merged.
* https://github.com/universal-ctags/libreadtags/issues/22 must be fixed.

masatake:
Random ideas:

The option `--_hint` should be renamed like:

`--_hint-file=tagfile`:  error strict. If an error occurs in calling libreadtags APIs, ctags may stop.
`--_hint-file-maybe=tagfile`: error-tolerant version of `--_hint-file`. If an error occurs in calling libreadtags APIs, ctags just doesn't use the hint file.

ctags must reject specifying the same file for output and hint. Should I compare their inode numbers?

more API for parsers:
```
bool isHintFileAvaiable (void);
```

Just after opening a hint file, the main part of ctags should notify it to parsers that have a method 
```
void (*preprocessHintFile) (hintFile *file, hintFileInfo *info, langType lang);
```

This helps a parser build including/included (, require/provide, use/used, or import/package) relation graph before parsing.

How about introducing tags.c, a parser for tags.
```
./ctags a.tags
```
ctags parses a.tags. When ctags find `F` kind entry in a.tags, ctags parses the file tagged with the entry.

The code for realizing the multi-pass parsing and for updating a tags file are strongly related. But how?

Linking libreadtags proposed here is obviously needed for both features.
I guess we may need the "filesystem" language that deals with directories as the first-class objects.
tagEntryInfos for directories may be an important building block for dealing with "path" like include path, module pat, library path, etc.

For implementing "updating a tags file", I must revise the way of output.

When using a hint file, ctags must compare the options used for generating the hint file and the options just passed from the user.
This comparison is much more important when updating a tags file. Should we accept options in the second pass?
Saying NO is easy. However, we have to remember a user can have many lines in one's .ctags.

Comparing options is basic infrastructure for running ctags parsers parallel.

To support the other types of hint files, ctags must verify the type of the hint file with filename extension and pattern as ctags does for detecting a suitable parser for an input source file.

If querying hints is not done at hotspots, we can reuse the query engine used in the readtags command.

masatake:
I applied `--_hint-file=` to our C parser.

```
$ cat macro.h 
#define DEF(fn, rtype, signature, body)	\
	rtype fn signature BEGIN body END
$ u-ctags '--fields=+{language}{signature}' '--fields-C++=+{macrodef}' -o hint.tags macro.h 
$ cat hint.tags
!_TAG_OUTPUT_EXCMD	mixed	/number, pattern, mixed, or combineV2/
!_TAG_PROC_CWD	/home/jet/var/ctags-new/Units/parser-c.r/macrodef-hint-file.d/	//
!_TAG_PROGRAM_VERSION	5.9.0	/f35a3944/
DEF	macro.h	/^#define DEF(/;"	d	language:C++	signature:(fn,rtype,signature,body)	macrodef:rtype fn signature BEGIN body END
$ cat input.c 
#include "macro.h"

#define BEGIN {
#define END   }

DEF(add2, int, (int a, int b), a + b)
$ u-ctags --param-CPreProcessor:_expand=1 --_hint-file=hint.tags -o - input.c 
BEGIN	input.c	/^#define BEGIN /;"	d	file:
END	input.c	/^#define END /;"	d	file:
add2	input.c	/^DEF(add2, int, (int a, int b), a + b)$/;"	f	typeref:typename:int
$
```

I used the ctags command with my experimental patch to make tags for Qemu source code that I'm reading now.
```
$ time u-ctags '--fields=+{language}{signature}' '--fields-C++=+{macrodef}' -o hint.tags  -R
u-ctags '--fields=+{language}{signature}' '--fields-C++=+{macrodef}' -o  -R  2.03s user 0.13s system 98% cpu 2.196 total
$ time u-ctags --param-CPreProcessor:_expand=1 --_hint-file=hint.tags -R
u-ctags --param-CPreProcessor:_expand=1 --_hint-file=hint.tags -R  99.66s user 41.11s system 99% cpu 2:21.08 total
```

About 70 times slower.

masatake:
I implemented negative hint cache.
```
$ time ~/bin/u-ctags --fields='+{line}' --param-CPreProcessor:_expand=1 --_hint-file=hint.tags -R
time ~/bin/u-ctags --fields='+{line}' --param-CPreProcessor:_expand=1 --_hint-file=hint.tags -R

real	0m20.691s
user	0m15.114s
sys	0m5.488s
```

7 times faster than the version without the negative hint cache.
However still 10 times slower than running with no-hint.

masatake:
This is related to #1960.

masatake:
https://github.com/universal-ctags/ctags/pull/2741/commits/e5d7cedc9e41aa734dda2d8b60f49ce73bd02db5 must be included in ctags6.

Issue #2738, opened 2020-11-28
SystemVerilog: entering infinite loop


masatake:
input.sv:
```
interface static 0
```
with `ctags -o - input.sv`.

masatake:
How about this one?

input.sv:
```
module use!
```

hirooih:
> How about this one?
> 
> input.sv:
> 
> `module use!`

It should be OK.  I've confirmed it does not cause infinite-loop.
Note that `isWordToken ('!')` returns `false`.

***

make chop, slap, fuzz, and noise are too heavy for my current Windows PC.
The CPU fan keeps making a loud noise. I am worried that it might break.
I am trying to set up old Mac for the work.  OS update is done.  I will try `make noise`, etc.  this evening.

masatake:
"make noise" reports nothing after "git pull upstream master:master"!

hirooih:
> "make noise" reports nothing after "git pull upstream master:master"!

Thanks.

How long did it take?
I been running "make noise" since this morning.  It still running.

masatake:
I didn't measure the time.
About 2 or 3 days.
I have used tmpfs (on-memroy file system) when running "make noise".

Running "make noise" is never a must. I will run it when I find the time.

PR #2744, opened 2020-11-29
SystemVerilog: fix for processDesignElementL() not to go into infinite-loop (fix for #2738)

hirooih:
Fix for #2738.

I reviewed all loop and refactored some of them.

### DIFF
--- a/Units/parser-verilog.r/systemverilog-illegal.d/input.sv
+++ b/Units/parser-verilog.r/systemverilog-illegal.d/input.sv
@@ -1,7 +1,0 @@
-module m;
-  struct
-  {
-    s* ; // pushMembers() caused infinite loop : #2724
-  }
-  logic foo;
-endmodule
--- a/Units/parser-verilog.r/systemverilog-illegal.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-illegal.d/expected.tags
@@ -1,2 +1,0 @@
-m	input.sv	/^module m;$/;"	m
-foo	input.sv	/^  logic foo;$/;"	r	module:m
--- a/Units/parser-verilog.r/systemverilog-illegal.d/input-0-2724.sv
+++ b/Units/parser-verilog.r/systemverilog-illegal.d/input-0-2724.sv
@@ -1,0 +1,7 @@
+module m2724;
+  struct
+  {
+    s* ; // pushMembers() caused infinite loop : #2724
+  }
+  logic foo;
+endmodule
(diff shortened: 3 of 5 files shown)


Units/parser-verilog.r/systemverilog-illegal.d/input.sv
```diff
@@ -1,7 +1,10 @@
...
+module m2724;
   struct
   {
     s* ; // pushMembers() caused infinite loop : #2724
   }
   logic foo;
 endmodule
+
+// #2738: processDesignElementL() go into infinite loop
+interface static 0
```
masatake, line 10: It is better to introduce a new file for the broken input.
How do you think about putting these 3 lines into a new file, `input-0.sv`?
The "make units" processes "input.sv" and "input-[0-9].sv".

masatake (changes requested):

Units/parser-verilog.r/systemverilog-illegal.d/input.sv
hirooih: > The "make units" processes "input.sv" and "input-[0-9].sv".

I did not know this.
Do we require input.sv?  In other words,  input-0.sv (for original) and input-1.sv (for new one) are OK?

masatake: Yes, we require input.sv for the original file.
See also
https://docs.ctags.io/en/latest/units.html#how-to-write-a-test-case
.

hirooih: It works.  I've pushed again.

masatake: If you don't treat input.sv special, you can make input.sv empty like https://github.com/universal-ctags/ctags/pull/2740/files .

hirooih: > Units/TEST/input[-_][0-9].* Units/TEST/input[-_][0-9][-_]*.* optional

I made input.sv as an empty file.  And rename input-[01].sv as;
input-0-2724.sv
input-1-2738.sv

hirooih: > If you don't treat input.sv special, you can make input.sv empty like https://github.com/universal-ctags/ctags/pull/2740/files .

Sorry, I read this after I pushed again.

hirooih:
Uncovered lines are sanity checks for unexpected inputs.  They should be OK.

I am thinking that I will use systemverilog-illegal.d/input.sv to cover these uncovered lines.

masatake (approved):

PR #2746, opened 2020-12-01
appveyor: Revise packaging of debug symbols

k-takata:
Stop creating a separate debug package which includes whole documents.
Instead, create a package that includes only debug symbols.
Use `objcopy --only-keep-debug` and `objcopy --add-gnu-debuglink` to
extract debug symbols to a separate file.

How to use the debug symbols:
* Extract `*.zip` and `*.debuginfo.zip` in the same directory.
* Also put the ctags source code there.
* Use GDB to debug `ctags.exe`.
  GDB will automatically load the symbols from `ctags.exe.debug`.

### DIFF
--- a/win32/appveyor.bat
+++ b/win32/appveyor.bat
@@ -194,10 +194,17 @@ copy COPYING package\license > nul
 copy win32\mkstemp\COPYING.MinGW-w64-runtime.txt package\license > nul
 robocopy man package\man *.html > nul
 cd package
-7z a ..\ctags-%ver%-%ARCH%.debug.zip %filelist% %dirlist%
-strip *.exe
+call :strip ctags.exe
+call :strip readtags.exe
 7z a ..\ctags-%ver%-%ARCH%.zip %filelist% %dirlist%
+7z a ..\ctags-%ver%-%ARCH%.debuginfo.zip *.exe.debug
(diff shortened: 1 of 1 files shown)


win32/appveyor.bat
```diff
@@ -194,12 +194,19 @@ copy COPYING package\license > nul
 copy win32\mkstemp\COPYING.MinGW-w64-runtime.txt package\license > nul
 robocopy man package\man *.html > nul
 cd package
-7z a ..\ctags-%ver%-%ARCH%.debug.zip %filelist% %dirlist%
-strip *.exe
+call :strip ctags.exe
+call :strip readtags.exe
```
masatake, line 198: Doesn't "call :string *.exe" work?
With "call :string *.exe", we don't have to update win32/appveryor.bat when I add "edittags.exe" or some new commands.

masatake:
I added one comment but the change itself looks good for me.

win32/appveyor.bat
k-takata: I doesn't work, but we can use the "for" command to expand wildcards.
I will update.

k-takata: updated

masatake (approved):

Issue #2428, opened 2020-02-16
Making reference tags for external entities instead of definitions tags


masatake:
Some parsers capturing the tokens referring to external entities as definitions tags.
This should not. Instead, it should capture as reference tags.
The modifications for solving this issue may break compatibility.
However, implementing multi-pass parsing in the future, we must fix them now.

- [ ] \input and \include in Tex
- [] using in C++
...

masatake:
If external entities are captured proper kind and proper role, we can implement semantic recursion feature.

I have implemented such feature as a prototype (-S option):
```
[yamato@slave ctags-github]$ u-ctags --kinds-python=F --sort=no -o - --extras=+r -S /usr/lib/python3.6/site-packages/pygments/cmdline.py  
u-ctags --kinds-python=F --sort=no -o - --extras=+r -S /usr/lib/python3.6/site-packages/pygments/cmdline.py  
__main__	/usr/lib/python3.6/site-packages/pygments/__init__.py	/^if __name__ == '__main__':  # pragma: no cover$/;"	m	language:PythonMain
codetagify	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'codetagify':     CodeTagFilter,$/;"	f	language:ansibleFilter	pyexpr:CodeTagFilter
keywordcase	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'keywordcase':    KeywordCaseFilter,$/;"	f	language:ansibleFilter	pyexpr:KeywordCaseFilter
highlight	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'highlight':      NameHighlightFilter,$/;"	f	language:ansibleFilter	pyexpr:NameHighlightFilter
raiseonerror	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'raiseonerror':   RaiseOnErrorTokenFilter,$/;"	f	language:ansibleFilter	pyexpr:RaiseOnErrorTokenFilter
whitespace	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'whitespace':     VisibleWhitespaceFilter,$/;"	f	language:ansibleFilter	pyexpr:VisibleWhitespaceFilter
gobble	/usr/lib/python3.6/site-packages/pygments/filters/__init__.py	/^    'gobble':         GobbleFilter,$/;"	f	language:ansibleFilter	pyexpr:GobbleFilter
[...]

PR #2750, opened 2020-12-04
SystemVerilog: Misc Fixes

hirooih:
- implement life_time (static, automatic) on class
- new tags
  - constraint: 'O', 'constraint
  - nettype: 'N', 'nettype'
- support array manipulation methods
- skip class_type
- Units
  - add test for void function
  - fix comments for "virtual"

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -89,7 +89,7 @@ typedef enum {
 	K_MEMBER,
 	K_IFCLASS,	/* interface class */
 	K_CONSTRAINT,
-	K_IFCLASS	/* interface class */
+	K_NETTYPE,
 } verilogKind;
 
 typedef struct {
@@ -167,6 +167,7 @@ static kindDefinition SystemVerilogKinds [] = {
  { true, 'w', "member",    "struct and union members" },
  { true, 'l', "ifclass",   "interface class" },
  { true, 'O', "constraint","constraints" },
+ { true, 'N', "nettype",   "nettype declarations" },
 };
 
 static const keywordAssoc KeywordTable [] = {
@@ -258,6 +259,7 @@ static const keywordAssoc KeywordTable [] = {
 	{ "var",           	K_REGISTER,  	{ 1, 0 } },
 	{ "void",          	K_REGISTER,  	{ 1, 0 } },
 	{ "with",          	K_WITH,			{ 1, 0 } },
+	{ "nettype",       	K_NETTYPE,		{ 1, 0 } },
 };
 
 static tokenInfo *currentContext = NULL;
@@ -422,6 +424,7 @@ static short isContainer (verilogKind kind)
 		case K_PROPERTY:
 		case K_SEQUENCE:
 		case K_TYPEDEF:
+		case K_NETTYPE:
 		case K_ENUM:
 		case K_STRUCT:
 			return true;
@@ -435,6 +438,7 @@ static short isTempContext (tokenInfo const* token)
 	switch (token->kind)
 	{
 		case K_TYPEDEF:
+		case K_NETTYPE:
 		case K_ENUM:
 		case K_STRUCT:
 			return true;
@@ -1157,7 +1161,7 @@ static int processEnum (tokenInfo *const token, int c)
 // [ struct | union [ tagged ] ] [ packed [ signed | unsigned ] ] { struct_union_member { struct_union_member } } { [ ... ] }
 static int processStruct (tokenInfo *const token, int c)
 {
-	verilogKind kind = token->kind;	// K_STRUCT or K_TYPEDEF
+	verilogKind kind = token->kind;	// K_STRUCT, K_TYPEDEF, or K_NETTYPE
 
 	/* Skip packed, signed, and unsigned */
 	while (isWordToken (c))
@@ -1187,6 +1191,7 @@ static int processStruct (tokenInfo *const token, int c)
 //     | nettype [ class_type :: | package_identifier :: | $unit :: ] net_type_identifier net_type_identifier ;
 static int processTypedef (tokenInfo *const token, int c)
 {
+	verilogKind kindSave = token->kind;	// K_TYPEDEF or K_NETTYPE
 	verilogKind kind = K_UNDEFINED;
 	if (isWordToken (c))
 	{
@@ -1219,7 +1224,7 @@ static int processTypedef (tokenInfo *const token, int c)
 	}
 	c = processType (token, c, &kind);
 
-	createTag (token, K_TYPEDEF);
+	createTag (token, kindSave);
 
 	ptrArrayClear (tagContents);
 	return c;
@@ -1605,6 +1610,7 @@ static int pushMembers (tokenInfo* token, int c)
 static int processType (tokenInfo* token, int c, verilogKind* kind)
 {
 	verilogKind actualKind = K_UNDEFINED;
+	tokenInfo *tokenSaved;
 	do
 	{
 		// [ class_type :: | package_identifier :: | $unit :: ] type_identifier { [ ... ] }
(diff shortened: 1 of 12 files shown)


hirooih:
@masatake san,

I have the following valgrind errors on other parsers and others.
I have no idea.  How about you?

Thanks.

```
  #valgrind-error:                        30
	extra-total-lines
	parser-dts.r/dts-simple
	parser-dts.r/dts-fq-without-scope-field
	parser-cxx.r/rojas.h
	parser-cxx.r/typedefs.cpp
	parser-cxx.r/template-specializations
	parser-cxx.r/functions.cpp
	parser-cxx.r/bug-github-871.cpp
	parser-cxx.r/namespace-and-preprocessor.cpp
	parser-cxx.r/variable-declarations.cpp
	parser-cxx.r/operators.cpp
	review-needed.r/simple.vr
	parser-yacc.r/nested
	parser-ldscript.r/simple-ldscript
	readtags.r/backslash-at-the-end-of-pattern
	parser-cpreprocessor.r/if0-false-with-param.c
	parser-cpreprocessor.r/macros-specified-with-param.c
	parser-cpreprocessor.r/disable-cpp-cpp
	parser-cpreprocessor.r/if0-true-with-param.c
	parser-c.r/option-disable-undef.c
	parser-c.r/macrodef
	parser-c.r/c-digraphs
	parser-c.r/c-trigraphs
	parser-c.r/line_directives.c
	parser-c.r/macros.c
	parser-c.r/directives-2.c
	parser-c.r/bug1764143.h
	parser-c.r/bug1086609.c
	parser-c.r/directives.c
	parser-c.r/if0.c
make: *** [Makefile:5913: units] Error 1
Error: Process completed with exit code 2.

```

masatake:
The Valgrind errors may be nothing to do with this pull request. Maybe about parsers/cpreprocessor.c. I am inspecting them at #2751.

hirooih:
> The Valgrind errors may be nothing to do with this pull request. 

Thanks.

I've add a fix to improve code coverage.
I'll merge this after checks pass.

hirooih:
All uncovered lines are intended code for unexpected inputs.

parsers/verilog.c
```diff
@@ -1633,9 +1639,25 @@ static int processType (tokenInfo* token, int c, verilogKind* kind)
...
+		// break on "with"
+		if (token->kind == K_WITH)
+		{
+			// restore tokenSaved to token
+			//   tokenClear() cannot be used to free strings.
+			vStringDelete (token->name);
+			vStringDelete (token->blockName);
+			vStringDelete (token->inheritance);
+			*token = *tokenSaved;
+			eFree (tokenSaved);
```
masatake, line 1662: This is too tricky. 
I think you should introduce a helper function doing this dirty work.

```
static void swapToken (tokenInfo *t0, tokenInfo *t1)
{
	tokenInfo tmp = *t0;
	*t0 = *t1;
	*t1 = tmp;
}
...
+ 	if (token->kind == K_WITH)
+ 			swapToken (token, tokenSaved);
```

parsers/verilog.c
```diff
@@ -1633,9 +1639,25 @@ static int processType (tokenInfo* token, int c, verilogKind* kind)
...
+			break;
```
masatake, line 1664: If you introduce swapToken, you don't need this break.

parsers/verilog.c
hirooih: Really Cool!
I looked for any better way, but I could not think of this.
I already merged this request.  I will fix this in the next pull request.

Thanks!

masatake:
I got a question about constraint.
```
constraint word_align {addr[1:0] == 2'b0;}
```
It seems that `word_align` is a definition. So tagging this is appreciated.
How do you use (or refer) this name in the other area of the SystemVerilog input file?
```

class C;
  rand int x;
  constraint proto1;        // implicit form
  extern constraint proto2; // explicit form
endclass
```

Are proto1 and proto2 definitions?
If they refer to constraints defined in different places, they are not definitions.
In such a case, proto1 and proto2 should be tagged as reference tags.

hirooih:
By reading;

https://docs.ctags.io/en/latest/contributions.html?highlight=reference%20tag#what-should-be-tagged

> The secondary class is a reference tag. This is newly introduced in Universal-ctags and is not available in Exuberant-ctags. If a name is used (or referenced) in a file, it can be tagged as a reference tag.

I understand reference tags as follows.

```
int x, y;	// Both x and y are definition tags.
...
y = x + 1;	// Both x and y are reference tags.
```

In this mean the current Verilog/SystemVerilog parser does not support reference tags at all.
The parser uses skipExpressoin(), skipPastMatch("{}"), and etc. heavily.  Supporting reference tags requires whole redesign of it

On the other hand proto1 and proto2 in the following example may be tagged as K_PROTOTYPE.  They are declarations.  An implementation of class C may (implicit form) or must (explicit form) define them.

```
class C;
  rand int x;
  constraint proto1;        // implicit form
  extern constraint proto2; // explicit form
endclass
```

masatake:
> In this mean the current Verilog/SystemVerilog parser does not support reference tags at all.
The parser uses skipExpressoin(), skipPastMatch("{}"), and etc. heavily. Supporting reference tags requires whole redesign of it

This is o.k. Currently, there is no parser in ctags that supports reference tags perfectly. I'm introducing reference tags in a limited area (#2428). 

> On the other hand proto1 and proto2 in the following example may be tagged as K_PROTOTYPE. They are declarations. An implementation of class C may (implicit form) or must (explicit form) define them.

Could you show an example of the implementation of class C where proto1 and proto2 are defined?

hirooih:
> Could you show an example of the implementation of class C where proto1 and proto2 are defined?

Sure.

Here is an example from LRM 18.5.1 "External constraint blocks".

```
constraint C::proto1 { x inside {-4, 5, 7}; }
constraint C::proto2 { x >= 0; }
```

masatake:
@hirooih, thank you.

Is this valid valid SystemVerilog input?

```
class C;
   rand int x;
   rand int y;
   constraint proto0 {
      y <= 10;
   }
   constraint proto1;
   extern constraint proto2;
endclass

constraint C::proto1 { x inside {-4, 5, 7}; }
constraint C::proto2 { x >= 0; }
```

I need an input that can be used as the start point of discussion.

hirooih:
Here is;

```
class C;
  rand int x;
  constraint proto1;        // implicit form
  extern constraint proto2; // explicit form
endclass

class D extends C;
  constraint C::proto1 { x inside {-4, 5, 7}; } // optional
  constraint C::proto2 { x >= 0; }  // mandatory
endclass
```

I hope this explains I meant by;

> They are declarations. An implementation of class C may (implicit form) or must (explicit form) define them.

masatake:
Thank you. If we remove "C::" from the example, the example is still valid as SystemVerilog input?

hirooih:
>  If we remove "C::" from the example, the example is still valid as SystemVerilog input?

Yes, it is. Your example and my example are equivalent if I understand SystemVerilog syntax correctly.

masatake:
I understand your design. It is good. Sorry for making noise.
```
$ u-ctags --sort=no --fields=+KZ --kinds-SystemVerilog=+Q -o - /tmp/foo.sv
C	/tmp/foo.sv	/^class C;$/;"	class
x	/tmp/foo.sv	/^  rand int x;$/;"	register	scope:class:C
proto1	/tmp/foo.sv	/^  constraint proto1;        \/\/ implicit form$/;"	constraint	scope:class:C
proto2	/tmp/foo.sv	/^  extern constraint proto2; \/\/ explicit form$/;"	prototype	scope:class:C
D	/tmp/foo.sv	/^class D extends C;$/;"	class
proto1	/tmp/foo.sv	/^  constraint proto1 { x inside {-4, 5, 7}; } \/\/ optional$/;"	constraint	scope:class:D
proto2	/tmp/foo.sv	/^  constraint proto2 { x >= 0; }  \/\/ mandatory$/;"	constraint	scope:class:D
```

This is similar to the following input and tags
```
$ cat foo.cpp
class C {
  void proto1 (void);
  virtual void proto2 (void) = 0;
};

class D: C {
  void proto1 (void) {}
  void proto2 (void) {}
};
$ u-ctags --sort=no --fields=+KZ --kinds-C++='*' -o - /tmp/foo.cpp
C	/tmp/foo.cpp	/^class C {$/;"	class	file:
proto1	/tmp/foo.cpp	/^  void proto1 (void);$/;"	prototype	scope:class:C	typeref:typename:void	file:
proto2	/tmp/foo.cpp	/^  virtual void proto2 (void) = 0;$/;"	prototype	scope:class:C	typeref:typename:void	file:
D	/tmp/foo.cpp	/^class D: C {$/;"	class	file:
proto1	/tmp/foo.cpp	/^  void proto1 (void) {}$/;"	function	scope:class:D	typeref:typename:void	file:
proto2	/tmp/foo.cpp	/^  void proto2 (void) {}$/;"	function	scope:class:D	typeref:typename:void	file:
```

hirooih:
```
proto1	/tmp/foo.cpp	/^  void proto1 (void);$/;"	prototype	scope:class:C	typeref:typename:void	file:
proto2	/tmp/foo.cpp	/^  virtual void proto2 (void) = 0;$/;"	prototype	scope:class:C	typeref:typename:void	file:
```

This is what I tried to mean by;

> On the other hand proto1 and proto2 in the following example may be tagged as K_PROTOTYPE.

hirooih:
The issue advised on this ticket was fixed on 2508577 in #2760.

PR #2755, opened 2020-12-09
main: attach the last modified time to epcoh: field even for pre opened file

masatake:
There are cases that ctags pre-opens an input file for guessing a language for it. e.g. Detecting whether .h file is for for C++ or ObjectiveC.                                                                                 
                                                                                                                                                              
The original code didn't attach the last modified time of the file to "epcoh:" field.


masatake:
A test case for -G option is needed.

masatake:
Extended the test case to use -G option.

PR #2756, opened 2020-12-10
Fix Inko parser for string literals and variables, and add a man page

YorickPeterse:
This PR two bugs:

* String literals are now ignored, while previously ctags would tag any code in them
* Local variables with an explicit type name no longer produce tags

In addition, this PR adds a simple man page for Inko. Since the parser doesn't really have any options, the man page itself is pretty bare-bones; but it's a start :smile:


masatake:
Thank you very much. Looks good to me. Could you wait for a while till I fix the issue about "ci/circleci: centos_make" CI environment?

YorickPeterse:
@masatake 

> Could you wait for a while till I fix the issue about "ci/circleci: centos_make" CI environment?

Sure, no problem! :smile:

YorickPeterse:
@masatake Done! It seems `ci/circleci: fedora_distcheck` is failing, though I can't seem to view the output (apparently need to log in for that).

Units/parser-inko.r/let.d/input.inko
```diff
@@ -1,2 +1,3 @@
 let FOO = 10
 let BAR = 20
+let foo: Bla = 30
```
masatake, line 3: Don't you want to tag `foo`?
I wonder why you don't have to update expected.tags.

masatake:
> @masatake Done! It seems ci/circleci: fedora_distcheck is failing, though I can't seem to view the output (apparently need to log in for that).

...Please ignore this. I instructed Circleci web interface to rerun "fedora_distcheck". It ended successfully though Circleci doesn't report the result to this issue.

Units/parser-inko.r/let.d/input.inko
YorickPeterse: @masatake No, `foo` isn't supposed to be tagged. The bug was that in this case `Bla` _would_ be tagged, which is incorrect.

masatake: Oh, I see.

YorickPeterse: @masatake `foo` is a local variable, and they should not be tagged. They were not tagged either.

However, when defining a local variable with an explicit type (as is done in the above example), ctags would incorrectly generate a tag for `Bar`. The expected behaviour is no tags being generated for local variables ever.

masatake: I see.

PR #2757, opened 2020-12-10
circleci: update the name of repos enabling in centos_make job

masatake:
s/PowerTools/powertools/

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -110,7 +110,7 @@ jobs:
        - run:
            name: Install build tools
            command: |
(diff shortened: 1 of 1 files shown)


masatake:
Travis-ci is too slow.

k-takata:
> Travis-ci is too slow.

It might be better to consider using GitHub Actions.
Also, note that travis-ci.org will end this month. We should migrate to travis-ci.com if we still use Travis CI.
https://docs.travis-ci.com/user/migrate/open-source-repository-migration

PR #2762, opened 2020-12-13
misc/units: change option of stat on BSD-based OSes including macOS

hirooih:
This is a fix to run `make chop, slap, fuzz, or noise` on macOS.

It seems that `stat` command has some variants.  I looked for better way on the Web, and this was the best one I found.


masatake (approved):

masatake:
Thank you!

PR #2763, opened 2020-12-15
Julia: fill scope field for functions

AmaiKinono:
The idea is in the test `function_scope.d`. For a input file like:

``` julia
module MyModule
function func1() end

function func2()
    function func3() end
end

end
```

The `scope` field for the functions should be:

- `func1`: module:MyModule
- `func2`: module:MyModule
- `func3`: function:func2

For now there are some things not working well. I'll write them in code reviews.

### DIFF
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1046,17 +1046,16 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
         }
 
         addTag(name, NULL, vStringValue(arg_list), K_FUNCTION, line, pos, local_scope, local_parent_kind);
-        //addToScope(scope, name);
-        //parseExpr(lexer, true, K_FUNCTION, scope);
+        addToScope(scope, name);
+        parseExpr(lexer, true, K_FUNCTION, scope);
     }
     else if (lexer->cur_token == TOKEN_CLOSE_BLOCK)
     {
         /* Function without method */
         addTag(name, NULL, NULL, K_FUNCTION, line, pos, local_scope, local_parent_kind);
+        /* Go to the closing 'end' keyword */
+        skipUntilEnd(lexer);
     }
-
-    /* Go to the closing 'end' keyword */
-    skipUntilEnd(lexer);
 
     vStringDelete(name);
     vStringDelete(arg_list);
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1383,6 +1383,5 @@ extern parserDefinition* JuliaParser (void)
     def->parser     = findJuliaTags;
     def->keywordTable = JuliaKeywordTable;
     def->keywordCount = ARRAY_SIZE (JuliaKeywordTable);
-    def->useCork    = CORK_QUEUE;
     return def;
 }
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1016,7 +1016,7 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
         advanceChar(lexer);
         advanceToken(lexer, true);
     } else {
-        local_scope = scope;
+        local_scope = vStringNewCopy(scope);
         local_parent_kind = parent_kind;
     }
 
@@ -1059,6 +1059,7 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
 
     vStringDelete(name);
     vStringDelete(arg_list);
+    vStringDelete(local_scope);
 }
 
 /* Macro format:
@@ -1278,7 +1279,7 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
 {
     int level = 1;
     size_t old_scope_len;
-    vString *local_scope;
+    vString *local_scope = NULL;
 
     while (lexer->cur_token != TOKEN_EOF)
     {
@@ -1358,6 +1359,7 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
             break;
         }
     }
+    vStringDelete(local_scope);
 }
 
 static void findJuliaTags (void)
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1279,6 +1279,7 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
 {
     int level = 1;
     size_t old_scope_len;
+    vString *local_scope;
 
     while (lexer->cur_token != TOKEN_EOF)
     {
@@ -1316,14 +1317,28 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
                 parseImport(lexer, scope, kind);
                 break;
             case TOKEN_IDENTIFIER:
-                skipWhitespace(lexer, false);
-                if (lexer->first_token && lexer->cur_c == '(')
+                if (lexer->first_token && lexer->cur_c == '.')
                 {
-                    parseShortFunction(lexer, scope, kind);
+                    local_scope = vStringNewCopy(lexer->token_str);
+                    advanceChar(lexer);
+                    advanceToken(lexer, true);
+                    skipWhitespace(lexer, false);
+                    if (lexer->cur_c == '(')
+                    {
+                        parseShortFunction(lexer, local_scope, K_MODULE);
+                    }
                 }
                 else
                 {
-                    advanceToken(lexer, true);
+                    skipWhitespace(lexer, false);
+                    if (lexer->first_token && lexer->cur_c == '(')
+                    {
+                        parseShortFunction(lexer, scope, kind);
+                    }
+                    else
+                    {
+                        advanceToken(lexer, true);
+                    }
                 }
                 break;
             case TOKEN_OPEN_BLOCK:
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1319,7 +1319,11 @@ static void parseExpr (lexerState *lexer, bool delim, int kind, vString *scope)
             case TOKEN_IDENTIFIER:
                 if (lexer->first_token && lexer->cur_c == '.')
                 {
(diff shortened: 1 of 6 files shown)


parsers/julia.c
```diff
@@ -1037,14 +1045,14 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
             advanceToken(lexer, true);
         }
 
-        addTag(name, NULL, vStringValue(arg_list), K_FUNCTION, line, pos, scope, parent_kind);
+        addTag(name, NULL, vStringValue(arg_list), K_FUNCTION, line, pos, local_scope, local_parent_kind);
         //addToScope(scope, name);
         //parseExpr(lexer, true, K_FUNCTION, scope);
```
AmaiKinono, line 1050: For now, the test `function_scope.d` can't tag `func5`

If we uncomment these 2 lines, it's expected to tag `func5`, but the result is only `func1` and `func3` are tagged.

My guess is we have to move the lexer to the right place before `parseExpr`.

AmaiKinono:
I can't review places that's not changed, so I'll write it down here. See this piece of code in `parseExpr`:

```c
case TOKEN_IDENTIFIER:
    skipWhitespace(lexer, false);
    if (lexer->first_token && lexer->cur_c == '(')
    {
        parseShortFunction(lexer, scope, kind);
    }
```

Now the lexer doesn't jump over `.`, so this predicate is wrong. It looks like a heuristic, so I guess I could change it to `cur_c == '.'`, and do the jobs in `parseShortFunction`.

parsers/julia.c
```diff
@@ -1360,5 +1383,6 @@ extern parserDefinition* JuliaParser (void)
     def->parser     = findJuliaTags;
     def->keywordTable = JuliaKeywordTable;
     def->keywordCount = ARRAY_SIZE (JuliaKeywordTable);
+    def->useCork    = CORK_QUEUE;
```
AmaiKinono, line 1386: I don't think I understand cork correctly. From the doc and the clojure parser, it seems the way to use cork is make a tag for the scope/namespace first, then set a tag's `scopeIndex` to it. But we don't want to make a tag for `Base` in a block like

``` julia
function Base.func
    something...
end
```

The Ruby parser uses cork, but doesn't do this, and doesn't use `scopeIndex`.

Besides, setting `requestAutomaticFQTag` doesn't give me full qualified tags for Julia, so I must be using cork in a wrong way.

masatake: Julia parser, including your changes, manages scope information (both names and kinds manually) manually.
WIth the cork_queue, your parser becomes free from tracking them.
With the cork_queue, your parser can track only an integer (cork index) returned from makeTagEntry.
The cork index is a candidate for the scope of another tag.

As usual, I will show the way how to use the cork queue.
What you have to do first of all is:
```diff
diff --git a/parsers/julia.c b/parsers/julia.c
index ac7df5e5..3b17ecb7 100644
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -823,11 +823,11 @@ static void debugLexer (lexerState *lexer)
 }
 #endif
 
-static void addTag (vString* ident, const char* type, const char* arg_list, int kind, unsigned long line, MIOPos pos, vString *scope, int parent_kind)
+static int addTag (vString* ident, const char* type, const char* arg_list, int kind, unsigned long line, MIOPos pos, vString *scope, int parent_kind)
 {
     if (kind == K_NONE)
     {
-        return;
+        return CORK_NIL;
     }
     tagEntryInfo tag;
     initTagEntry(&tag, vStringValue(ident), kind);
@@ -843,7 +843,7 @@ static void addTag (vString* ident, const char* type, const char* arg_list, int
         tag.extensionFields.scopeKindIndex = parent_kind;
         tag.extensionFields.scopeName = vStringValue(scope);
     }
-    makeTagEntry(&tag);
+    return makeTagEntry(&tag);
 }
 
 /* Skip tokens until one of the goal tokens is hit. Escapes when level = 0 if there are no goal tokens.
@@ -1360,5 +1360,6 @@ extern parserDefinition* JuliaParser (void)
     def->parser     = findJuliaTags;
     def->keywordTable = JuliaKeywordTable;
     def->keywordCount = ARRAY_SIZE (JuliaKeywordTable);
+	def->useCork = CORK_QUEUE;
     return def;
 }
```

You can use the above change with code like `int parent_index = addTag ("PARENT"...);`.
The next step is for providing the way to fill a scope field for a new tag:
```diff
diff --git a/parsers/julia.c b/parsers/julia.c
index 3b17ecb7..048f6980 100644
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -823,7 +823,8 @@ static void debugLexer (lexerState *lexer)
 }
 #endif
 
-static int addTag (vString* ident, const char* type, const char* arg_list, int kind, unsigned long line, MIOPos pos, vString *scope, int parent_kind)
+static int addTag (vString* ident, const char* type, const char* arg_list, int kind, unsigned long line, MIOPos pos, vString *scope, int parent_kind,
+	int parent_represented_in_corkIndex)
 {
     if (kind == K_NONE)
     {
@@ -838,7 +839,9 @@ static int addTag (vString* ident, const char* type, const char* arg_list, int k
 
     tag.extensionFields.signature = arg_list;
     /* tag.extensionFields.varType = type; */  /* Needs a workaround */
-    if (parent_kind != K_NONE)
+	if (parent_represented_in_corkIndex != CORK_NIL)
+		tag.extensionFields.scopeIndex = parent_represented_in_corkIndex;
+	else if (parent_kind != K_NONE)
     {
         tag.extensionFields.scopeKindIndex = parent_kind;
         tag.extensionFields.scopeName = vStringValue(scope);
```

When you want to make a tag for "CHILD" having "PARENT" as the scope, call addTag like `addTag("CHILD", ..., parent_index, ...);`.  If you don't want to fill scope the original way, pass `CORK_NIL` instead of `parent_index`.

masatake: In some cases, you may want to make X as a scope without tagging X. Anonymous tag feature and reference tags may help you in that case.

AmaiKinono: Thanks!

> In some cases, you may want to make X as a scope without tagging X.

Yes, that's the main consideration. I do saw something like `extras:anonymous` in some tags files. I think I've got the idea.

I'll come back in 1 or 2 days.

masatake: If there is no name for the scope, the anonymous tag is for you.

input.c:
```c
struct {
  int x, y;
} origin;
```
u-ctags -o - /tmp/input.c
```
__anon6dc190750108	/tmp/input.c	/^struct {$/;"	s	language:C	file:
origin	/tmp/input.c	/^} origin;$/;"	v	language:C	typeref:struct:__anon6dc190750108
x	/tmp/input.c	/^  int x, y;$/;"	m	language:C	struct:__anon6dc190750108	typeref:typename:int	file:
y	/tmp/input.c	/^  int x, y;$/;"	m	language:C	struct:__anon6dc190750108	typeref:typename:int	file:
```

For making anonymous tags, you can use anon* functions declared in main/parse.h.
Don't forget to call `markTagExtraBit (&e, XTAG_ANONYMOUS);` after making an anonymous tag.

If the scope has a name but it is not a definition, the reference tag is for you.
input.py:
```python
from x import y
```
u-ctags --fields=+E --extras=+r -o - /tmp/input.py 
```
x	/tmp/input.py	/^from x import y$/;"	i	language:Python	extras:reference
y	/tmp/input.py	/^from x import y$/;"	x	language:Python	module:x	extras:reference
```

parsers/julia.c
AmaiKinono: Fixed.

parsers/julia.c
AmaiKinono: I gave up the idea of using cork, mainly because the julia parser handles scope information in its own way. The code becomes awkward if we mix them up, and it's beyond my ability to fully replace it using cork.

But thanks for your help anyway. Maybe I could use them to improve the documentation of cork.

masatake:
Just a question about Julia. This is not about your change but about Julia language itself.

```
function SomeModule.func2(a::Int)
    a
end
```
Does Julia accept the above code without defining SomeModule?
Is SomeModule defined when func2 is defined? or does Julia reject the above code because SomeModule is not defined?

AmaiKinono:
> Is `SomeModule` defined when `func2` is defined? or does Julia reject the above code because `SomeModule` is not defined?

It's the latter one. Julia will say `UndefVarError: SomeModule not defined`.

The syntax is only for extending/overriding functions in a imported module.

AmaiKinono:
@masatake I think this can be merged.

masatake:
@AmaiKinono, I see. Please, wait for a while. I would like to finish #2767.

masatake:
```
function SomeModule.func2(a::Int)
    a
end
```

I will merge this pull request. But before merging, I would like to think about how ideal ctags deals with the above input.

1. the parser reads "SomeModule".
2. the parser makes a reference tag for "SomeModule". the parser gets a corkIndex for the tag.
3. the parser reads "func2".
4. the parser makes a definition tag for "func2" with scope specified with the corkIndex.

The question is about the reference tag. The kind assigned to the tag is "module". 
I wonder what role we should assign to the tag. My idea is "indirectly referenced". 
 "indirectly referenced" is a bit too long to use in a tags file. So I would like to use "iref" in a tags file.
If someone finds a better name, I would like to hear it.

masatake:
Thank you.

AmaiKinono:
> "indirectly referenced" is a bit too long to use in a tags file

In Python we already have `indirectlyImported`. `indirectlyReferenced` is not much longer. For a shorter one, I think `indirectlyRef` is good. It's clear even without further explanation.

masatake:
@AmaiKinono, thank you very much. `indirectlyReferenced` is a bit log but using the same role names among parsers is a good idea. Let's use `indirectlyReferenced`.

PR #2774, opened 2020-12-28
Various minor fixes

masatake:


masatake:
travis-ci.com is unbelievably fast. However, we have to think about credits. 10000 credits are given per month. The testing for this pull request consumes 260 of 10000.

PR #2778, opened 2020-12-30
Inko: rename "object" keyword to "class"

YorickPeterse:
The "object" keyword has been renamed to "class". This commit updates
all Inko tests and the parser to match this change.


masatake:
Thank you.

PR #2780, opened 2020-12-31
Improve recursive macro expansion detection

pragmaware:
Don't allow the unget buffer to grow too large while expanding macros. This helps with
infinite recursive macro expansions. Fixes #2779 .

PR #2781, opened 2020-12-31
Cxx,CPreProcessor: limit to use a macro in a recursive macro expansion

masatake:
This change limits the usage of a macro definition in a recursive macro expansion up to CXX_PARSER_MAXIMUM_MACRO_USE_COUNT.

This change extends #2780.

### DIFF
--- a/parsers/cpreprocessor.c
+++ b/parsers/cpreprocessor.c
@@ -107,6 +107,7 @@ typedef struct sCppState {
 		conditionalInfo ifdef [MaxCppNestingLevel];
 	} directive;
 
+	cppMacroInfo * macroInUse;
 	hashTable * fileMacroTable;
 
 } cppState;
@@ -337,6 +338,7 @@ static void cppInitCommon(langType clientLang,
 
 	Cpp.directive.name = vStringNewOrClear (Cpp.directive.name);
 
+	Cpp.macroInUse = NULL;
 	Cpp.fileMacroTable =
 		doesExpandMacros && isFieldEnabled (FIELD_SIGNATURE) && isFieldEnabled (Cpp.macrodefFieldIndex)
 		? makeMacroTable ()
@@ -362,6 +364,16 @@ extern void cppInit (const bool state, const bool hasAtLiteralStrings,
 				   macrodefFieldIndex);
 }
 
+static void cppClearMacroInUse (cppMacroInfo **pM)
+{
+	for (cppMacroInfo *p = *pM; p; p = p->next)
+	{
+		CXX_DEBUG_PRINT("Macro <%p> clear useCount: %d -> 0", p, p->useCount);
+		p->useCount = 0;
+	}
+	*pM = NULL;
+}
+
 extern void cppTerminate (void)
 {
 	if (Cpp.directive.name != NULL)
@@ -383,6 +395,8 @@ extern void cppTerminate (void)
 	}
 
 	Cpp.clientLang = LANG_IGNORE;
+
+	cppClearMacroInUse (&Cpp.macroInUse);
 
 	if (Cpp.fileMacroTable)
 	{
@@ -505,6 +519,22 @@ void cppUngetString(const char * string,int len)
 	Cpp.ungetDataSize += len;
 }
 
+extern void cppUngetStringBuiltByMacro(const char * string,int len, cppMacroInfo *macro)
+{
+	if (macro->useCount == 0)
+	{
+		cppMacroInfo *m = Cpp.macroInUse;
+		Cpp.macroInUse = macro;
+		macro->next = m;
+	}
+	macro->useCount++;
+
+	CXX_DEBUG_PRINT("Macro <%p> increment useCount: %d->%d", macro,
+					(macro->useCount - 1), macro->useCount);
+
+	cppUngetString (string, len);
+}
+
 static int cppGetcFromUngetBufferOrFile(void)
 {
 	if(Cpp.ungetPointer)
@@ -522,6 +552,8 @@ static int cppGetcFromUngetBufferOrFile(void)
 		return c;
 	}
 
+	if (Cpp.macroInUse)
+		cppClearMacroInUse (&Cpp.macroInUse);
 	return getcFromInputFile();
 }
 
@@ -1693,7 +1725,7 @@ extern cppMacroInfo * cppFindMacroFromSymtab (const char *const name)
 
 /*  Determines whether or not "name" should be ignored, per the ignore list.
  */
-extern const cppMacroInfo * cppFindMacro (const char *const name)
+extern cppMacroInfo * cppFindMacro (const char *const name)
 {
 	cppMacroInfo *info;
 
@@ -1828,6 +1860,8 @@ static void saveIgnoreToken(const char * ignoreToken)
 	} else {
 		info->replacements = NULL;
 	}
+	info->useCount = 0;
+	info->next = NULL;
 
 	hashTablePutItem(cmdlineMacroTable,eStrndup(tokenBegin,tokenEnd - tokenBegin),info);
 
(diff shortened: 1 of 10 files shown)


pragmaware:
I like your solution much more.
We might be even able to get rid of my old tricks to gain some performance.

There are a couple of ideas that I'll add in the review.

(Happy New Year!)

parsers/cpreprocessor.c
```diff
@@ -518,6 +552,7 @@ static int cppGetcFromUngetBufferOrFile(void)
 		return c;
 	}
 
+	cppClearMacroInUse (&Cpp.macroInUse);
```
pragmaware, line 555: This call is in a very hot path.
I suggest to moving it to cppUngetStringBuiltByMacro() like this:

```
if(!Cpp.ungetPointer)
   cppClearMacroInUse(&Cpp.macroInUse)
```
I know that this is imperfect because it will retain macro usage for one substitution AFTER a sequence. However this should not hurt as long as we keep the maximum number of macro usages in a sequence high (See other comment below).

If you have a better idea for a place to move it we can use yours - the point is to move it out of the hot path.

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1202,6 +1202,14 @@ void cxxParserUngetCurrentToken(void)
 
 #define CXX_PARSER_MAXIMUM_TOKEN_CHAIN_SIZE 16384
 
+// We stop applying macro replacements if the unget buffer gets too big
+// as it is a sign of recursive macro expansion
+#define CXX_PARSER_MAXIMUM_UNGET_BUFFER_SIZE_FOR_MACRO_REPLACEMENTS 65536
+
+// We stop applying macro replacements if a macro is used so many
+// times in a recursive macro expansion.
+#define CXX_PARSER_MAXIMUM_MACRO_USE_COUNT 8
```
pragmaware, line 1211: I suggest to making this far higher. Like 1024 or something.
It will still pass the test but will allow for more complex macros to be used (like big tables buit by macros).

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1327,12 +1335,12 @@ bool cxxParserParseNextToken(void)
 			}
 		} else {
 
-			const cppMacroInfo * pMacro = cppFindMacro(vStringValue(t->pszWord));
+			cppMacroInfo * pMacro = cppFindMacro(vStringValue(t->pszWord));
 
-			if(pMacro)
+			CXX_DEBUG_PRINT("Macro %s <%p> useCount: %d", vStringValue(t->pszWord),
```
pragmaware, line 1340: I'd move this debug statement inside the if, so it isn't printed at every token.

parsers/cxx/cxx_parser_tokenizer.c
```diff
@@ -1359,7 +1367,11 @@ bool cxxParserParseNextToken(void)
 						// Exclude possible cases of recursive macro expansion that
 						// causes a single token chain to grow too big
 						//    -D'x=y.x'
-						(iInitialTokenChainSize < CXX_PARSER_MAXIMUM_TOKEN_CHAIN_SIZE)
+						(iInitialTokenChainSize < CXX_PARSER_MAXIMUM_TOKEN_CHAIN_SIZE) &&
+						// Detect other cases of nasty macro expansion that cause
+						// the unget buffer to grow fast (but the token chain to grow slowly)
+						//    -D'p=a' -D'a=p+p'
+						(cppUngetBufferSize() < CXX_PARSER_MAXIMUM_UNGET_BUFFER_SIZE_FOR_MACRO_REPLACEMENTS)
```
pragmaware, line 1374: This hack could probably be removed then. It doesn't hurt much... but well, as you wish.

pragmaware (commented):
Looks good to me!
I have added some comments to optimize it.

parsers/cpreprocessor.c
masatake: Cpp.ungetPointer? Didn't you want to write "Cpp.macroInUse"?

We can expect Cpp.ungetPointer is always NULL because (1) the if-condition before the line I changed is 
`if (Cpp.ungetPointer)` and (2) `return` statement exists at the end of the if-block.
So the change you suggested doesn't change the control flow. Am I missing something?

parsers/cxx/cxx_parser_tokenizer.c
masatake: This definition limits per-macro usage in a recursive expansion. It doesn't limit the usage of all macros in a recursive expansion. I think even 8 is large. At the early stage of development, I used 1 for the definition.

See 
https://github.com/universal-ctags/ctags/blob/7f7cb386009aed6738bb6aab3434671db0b16fe8/Units/parser-c.r/recursive-macros-2.d/expected.tags

1024 may make the expected.tags very large.

Do you still think the larger value is better?

parsers/cxx/cxx_parser_tokenizer.c
masatake: I see. I will reflect the comment to this pull request.

parsers/cxx/cxx_parser_tokenizer.c
masatake: If we use 1024 as the definition for CXX_PARSER_MAXIMUM_MACRO_USE_COUNT, your hack is still an important stopper of expansion.

masatake:
@pragmaware, thank you for reviewing. 
Happy new year.

parsers/cxx/cxx_parser_tokenizer.c
masatake: Though I didn't change CXX_PARSER_MAXIMUM_MACRO_USE_COUNT, I kept your code as a safe-stopper.

parsers/cxx/cxx_parser_tokenizer.c
masatake: I changed as following to report overly used macros:
```
+#ifdef DEBUG
+                       if(pMacro && (pMacro->useCount >= CXX_PARSER_MAXIMUM_MACRO_USE_COUNT))
+                       {
+                               /* If the macro is overly used, report it here. */
+                               CXX_DEBUG_PRINT("Overly uesd macro %s <%p> useCount: %d (> %d)",
+                                                               vStringValue(t->pszWord),
+                                                               pMacro, pMacro? pMacro->useCount: -1,
+                                                               CXX_PARSER_MAXIMUM_MACRO_USE_COUNT);
+                       }
+#endif
```

parsers/cxx/cxx_parser_tokenizer.c
masatake: I didn't change CXX_PARSER_MAXIMUM_MACRO_USE_COUNT in the update.

parsers/cpreprocessor.c
masatake: I used "Cpp.macroInUse" as the condition in the update.

pragmaware: The current version is quite better as it avoids the function call in most cases. So we can leave it like this.


Just for clarification of my previous comment, I suggested to moving the macro cleaning call to a different function: cppUngetStringBuiltByMacro().

So instead of 
```
static int cppGetcFromUngetBufferOrFile(void)
{
	...
	if (Cpp.macroInUse)
		cppClearMacroInUse (&Cpp.macroInUse);
	...
}
```
which is in the "hot path" because is called for every character read, do
```
extern void cppUngetStringBuiltByMacro(const char * string,int len, cppMacroInfo *macro)
{
	if (!Cpp.ungetPointer)
		// first macro replacement in sequence
		cppClearMacroInUse (&Cpp.macroInUse);
	...
}
```
which is not in the "hot path".

This is imperfect since the first replacement in a sequence retains the macro counts from the previous sequence. But as long as the limit is high, it won't hurt.

Though in fact with your additional check the difference might be negligible.

pragmaware:
LGTM!

parsers/cpreprocessor.c
masatake: Oh, I see. However, I would like to keep this pull request simple. So I would like you to create a new pull request stacked on my change (if you want).

PR #2786, opened 2021-01-03
Julia: parse attribute correctly with @kwdef macro

getzze:
I corrected a bug that was reported by a geany user.


masatake:
I will review this one after merging #2727.

masatake:
Thank you.

PR #2787, opened 2021-01-03
Julia - incorrect initialization at first line

getzze:
If the first line is empty, valgrind gives an error.
Short functions defined in the first line are also not parsed, try the following, where `fun2` is parsed but not `fun1`:

```julia
fun1(x) = 1
fun2(x) = 2
```

~~It is a work in progress, I am not sure where is the problem.~~
Solved by initializing the  `prev_c` member to `\0` and treating TOKEN_NONE as a new line TOKEN.


masatake:
Could you squash the two commits into one (when ready to merge)?
In addition, I would like you to use "Julia: " as the prefix of the commit header.

getzze:
I opened a new PR with clean commit #2790

PR #2788, opened 2021-01-03
Julia: correctly recognize Module.@macro as a macro call

getzze:
```julia
@eval fun1(x) = $a
Base.@eval fun2(x) = $a
```
`fun2` is not recognized as a short function, because `Base.@eval` is not recognized as a macro (with scope).

(currently `fun1` is also not parsed but it will be corrected by #2790).

~~This is a work in progress, it should be implemented using the new cork system introduced by #2763.~~
It is working now!

### DIFF
--- a/Units/parser-julia.r/scoped_macro.d/args.ctags
+++ b/Units/parser-julia.r/scoped_macro.d/args.ctags
@@ -1,0 +1,2 @@
+--fields=+Zs
+--sort=no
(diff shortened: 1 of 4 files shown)


AmaiKinono:
Actually in #2763 I didn't chose to use cork, so the mechanism dealing with scope is still the one you introduced.

getzze:
I don't understand what happens when I rebase, some other commits appear. Can you just cherry-pick the last commit, 7d04c71 ?

getzze:
ok, I managed to clean the commit tree, it can be merged.

masatake:
Please, wait for awhile.
I'm struggling the trouble in the build-system.

masatake:
I will review this one after merging #27272.

masatake:
@getzze, can I ask you to resolve the conflicts?

masatake:
Thank you.

PR #2790, opened 2021-01-04
Julia: correct valgrind error initialization

getzze:
... and treat TOKEN_NONE as a new line token.
rebase of #2787

If the first line is empty, valgrind gives an error.
Short functions defined in the first line are also not parsed, try the following, where fun2 is parsed but not fun1:

fun1(x) = 1
fun2(x) = 2

Solved by initializing the prev_c member to \0 and treating TOKEN_NONE as a new line TOKEN.


masatake:
Thakn you.

Issue #2795, opened 2021-01-08
Extra tags in multiline regex tables


bernardjosef:
Multiline regex tables do not support conditional tagging with extra. The following patch seems to solve the problem, but I do not know if there are drawbacks.

````
diff -urN ctags-orig/main/lregex.c ctags/main/lregex.c
--- ctags-orig/main/lregex.c	2021-01-06 18:31:30.000000000 +0100
+++ ctags/main/lregex.c	2021-01-08 02:45:40.000000000 +0100
@@ -2407,6 +2407,11 @@
 	{
 		regexTableEntry *entry = ptrArrayItem(table->entries, i);
 		regexPattern *ptrn = entry->pattern;
+
+		if ((entry->pattern->xtagType != XTAG_UNKNOWN)
+			&& (!isXtagEnabled (entry->pattern->xtagType)))
+			continue;
+
 		struct guestSpec  *guest = &ptrn->guest;
 
 		Assert (ptrn);
````

**Comment by masatake** [comment]
Thank you very much. You are correct. Can I ask you to make a pull request for this?
You can move the lines upward like:
```
 		regexTableEntry *entry = ptrArrayItem(table->entries, i);
+		if ((entry->pattern->xtagType != XTAG_UNKNOWN)
+			&& (!isXtagEnabled (entry->pattern->xtagType)))
+			continue;
+
 		regexPattern *ptrn = entry->pattern;
 		struct guestSpec  *guest = &ptrn->guest;
```

Could you add a test case I wrote to the pull request?
For adding, do "git am 0001-Tmain-add-a-case-for-testing-_extra-mtable-regex-fla.txt"
[0001-Tmain-add-a-case-for-testing-_extra-mtable-regex-fla.txt](https://github.com/universal-ctags/ctags/files/5790000/0001-Tmain-add-a-case-for-testing-_extra-mtable-regex-fla.txt)
.

masatake:
I wonder how do you find this bug because `_extra` flag is so minor.
You will be interested in my blog article. https://ctags.io/2021/01/05/optscript/.

bernardjosef:
I am currently working on a parser for Markdown wiki links and Pandoc citations in Markdown files for a Zettelkasten (slip box). This is why I have found the bug, because I have defined an extra flag there. My goal is rather unusual for Universal ctags, I believe. I want to follow links and citations inside a Zettelkasten with the help of tags.

masatake:
[...]

PR #2796, opened 2021-01-10
Extra mtable regex

bernardjosef:
Extra tags in multiline regex tables. See issue #2795.


masatake:
Thank you very much.
I'm waiting more contributions from the person like you who is not afraid of reading main/lregex.c!

PR #2799, opened 2021-01-11
build-sys: fix build script for cross compiling

masatake:


masatake:
#2813 solves the original issue.

PR #2802, opened 2021-01-14
build-sys: fix configure syntax error near unexpected token `dummy`

leleliu008:
add code to configure.ac to check if the `PKG_CHECK_EXISTS` macro is defined. If not defined, a fatal error thrown, tell the user to install pkg-config 0.18 or later.

Note: our project use the `PKG_CHECK_EXISTS` and `PKG_CHECK_MODULES` two macros, these two macros are provided by `pkg.m4` which is installed along with `pkg-config`. the `PKG_CHECK_EXISTS` since `0.18`, the `PKG_CHECK_MODULES` since `0.4`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -30,6 +30,8 @@ AM_SILENT_RULES([yes])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIR([m4])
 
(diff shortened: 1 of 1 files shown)


masatake (approved):

masatake:
Thank you! Plesae, merge this.

PR #2814, opened 2021-01-18
add unit test in centos:6.10 docker image container in CircleCI

leleliu008:


circle.yml
```diff
@@ -228,36 +228,38 @@ jobs:
...
+             yum install -y https://repo.ius.io/ius-release-el6.rpm
+             yum --enablerepo=ius-archive -y install python36
+             ln -s /usr/bin/python36 /usr/bin/python3
        - run:
            name: Get the source code of u-ctags
            command: |
              git clone https://github.com/universal-ctags/ctags
        - run:
            name: Build
            command: |
```
leleliu008, line 253: every step is running in a sub shell, so here no need use (commands)

circle.yml
```diff
@@ -228,36 +228,38 @@ jobs:
...
              sed -i 's|mirrorlist=|#mirrorlist=|g' /etc/yum.repos.d/CentOS-Base.repo
-             sed -i 's|mirror|vault|g' /etc/yum.repos.d/CentOS-Base.repo
-             sed -i 's|http:|https:|g' /etc/yum.repos.d/CentOS-Base.repo
+             sed -i 's|mirror|vault|g'             /etc/yum.repos.d/CentOS-Base.repo
+             sed -i 's|http:|https:|g'             /etc/yum.repos.d/CentOS-Base.repo
        - run:
            name: Install tools for building
            command: |
              yum install -y git gcc make autoconf automake pkgconfig
+             yum install -y https://repo.ius.io/ius-release-el6.rpm
```
leleliu008, line 244: use [ius-archive](https://repo.ius.io/) repository to install `python3.6`

circle.yml
```diff
@@ -228,36 +228,38 @@ jobs:
...
      working_directory: ~/universal-ctags
      docker:
-       - image: centos:6
+       - image: centos:6.10
      steps:
        - run:
            name: Rearrange CentOS-Base.repo file
            command: |
-             sed -i 's|#baseurl=|baseurl=|g' /etc/yum.repos.d/CentOS-Base.repo
+             sed -i 's|#baseurl=|baseurl=|g'       /etc/yum.repos.d/CentOS-Base.repo
```
masatake, line 236: The changes about whitespaces are nothing to do with the original issue that you wanted to solve.
So I woud like you to split the pull request into two if you want to change the whitespaces; one is for updating whitespaces and another is for fixing the original issue. In that case, I would like to use "circuleci,cosmetic: ..." as the prefix for the comment header for the former commit.
See 2a0a14105b2ac169f264f0a609883392223a4842 aa6af6384d758dfb55e47079dbece38275d0733cc nd as examples for the former commit.

masatake (approved):

PR #2815, opened 2021-01-18
Julia: correct bugs with struct constructor

getzze:
Correct three bugs with functions and constructors:
- `lenght(a) == 0` was recognized as a short function definition.
- Parametric constructors with `function` keyword were not recognized, like `function MyStruct{Int}(x) ... end`
- Inner constructors inside `struct` defined below a constructor with the `function` keyword had incorrect scope, it was considered to be below the constructor with `function`.
- [EDIT] Outer constructor, like parametric short function `func{Int}(x) = 0`, was not recognized as a short function.

### DIFF
--- a/parsers/julia.c
+++ b/parsers/julia.c
@@ -1086,6 +1086,14 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
         local_parent_kind = parent_kind;
     }
 
+    /* Scan for parametric type constructor */
+    skipWhitespace(lexer, false);
+    if (lexer->cur_c == '{')
+    {
+        scanCurlyBlock(lexer);
+        skipWhitespace(lexer, false);
+    }
+
     name = vStringNewCopy(lexer->token_str);
     arg_list = vStringNew();
     line = lexer->line;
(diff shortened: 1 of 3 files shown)


parsers/julia.c
```diff
@@ -1072,16 +1072,28 @@ static void parseFunction (lexerState *lexer, vString *scope, int parent_kind)
...
+    else if (lexer->cur_c == '.')
+    {
         local_scope = vStringNewCopy(lexer->token_str);
         local_parent_kind = K_MODULE;
         advanceChar(lexer);
         advanceToken(lexer, true, false);
-    } else {
+    }
+    else
+    {
```
masatake, line 1084: I'm sorry that I accepted this style of change. I should not.

This is nothing to do with the original issue you wanted to fix.
Could you make a separate commit for the change of style?
In that case, I would like to use "Julia,cosmetic" or "Julia,style" as the prefix for the header of the commit log.

INSTEAD, I would like you to invite this project. So you can control your parser especially when we get a pull request for your parser. We can avoid the mistake I took.

masatake: You can find "cosmetic" change with searching "cosmetic in git log output.

getzze:
Now I think it's working, but the code for function and short-function should be refactored to avoid these kinds of bugs, I will add it to the TODO list.

masatake (approved):

masatake:
Please, use "Create a merge commit".

PR #2836, opened 2021-02-01
readtags: add string-append/concat and regexp-quote operators

### LINKED ISSUE #2824: readtags: `concat` and `regexp-quote` operators
I think we need a `regexp-quote` operator.

A client tool may need to build regexp using literal strings, and it needs to make sure the generated regexp does match that string, by wrapping it in this operator.

Emacs and Vim both doesn't use POSIX extended regexp, so the built-in `regexp-quote` in editors is not reliable.

*Edit:* Just found that this may be more complicated than I thought:

- `concat` may be needed too, if we want to "build" a regexp inside readtags.
- We need to remove the limitation that the 1st arg of `regexp->string` must be a string literal (so we can write.

Maybe it's more realistic to figure out how to implement a `regexp-quote` for ERE in the client tool.

_Originally posted by @AmaiKinono in https://github.com/universal-ctags/ctags/issues/2475#issuecomment-765146595_

masatake:
Close  #2824.

May need more test cases.


masatake:
For debugging purposes, we need a print form(?) working at compile time.

dsl/dsl.c
```diff
@@ -157,11 +163,23 @@ static DSLProcBind pbinds [] = {
...
 	  .helpstr = "((string->regexp \"PATTERN\") $target) -> <boolean>; PATTERN must be string literal.",
-	  .macro = string2regex},
+	  .macro = macro_string2regexp },
+	{ "regexp-quote",    builtin_regexp_quote, NULL, DSL_PATTR_CHECK_ARITY, 1,
+	  .helpstr = "(regexp-quote <string>) -> <string>",
+	  .macro = macro_regexp_quote },
 	{ "print",   bulitin_debug_print, NULL, DSL_PATTR_CHECK_ARITY, 1,
 	  .helpstr = "(print OBJ) -> OBJ" },
+	{ "print-code", NULL, 0, 0,
+	  .helpstr = "(print-code CODE) -> COD; unlike `print', this works at compile time.",
```
AmaiKinono, line 181:
```suggestion
	  .helpstr = "(print-code CODE) -> CODE; unlike `print', this works at compile time.",
```

And maybe it's better to use `OBJ` rather than `CODE`? That's what `print` operator uses.

dsl/dsl.c
```diff
@@ -157,11 +163,23 @@ static DSLProcBind pbinds [] = {
...
 	  .helpstr = "(- <integer> <integer>) -> <integer>", },
-	{ "string->regexp",  NULL,                 NULL, 0, 0,
+	{ "concat",   builtin_string_append,NULL, 0, 0,
+	  .helpstr = "(concat any<string> ...) -> <string>; an alias for string-append",
+	  .macro = macro_string_append },
+	{ "string-append",   builtin_string_append,NULL, 0, 0,
+	  .helpstr = "(string-append any<string> ...) -> <string>",
+	  .macro = macro_string_append },
+	{ "string->regexp",  builtin_string2regexp,NULL, 0, 0,
 	  .helpstr = "((string->regexp \"PATTERN\") $target) -> <boolean>; PATTERN must be string literal.",
```
AmaiKinono, line 173:
```suggestion
	  .helpstr = "((string->regexp <string>) $target) -> <boolean>",
```

dsl/dsl.c
masatake: Thank you.

How about `EXPRESSION`?  I think `OBJ` implies something already evaluated. However, `print-code` works at compile-time, I thought using another work was better.
```
(print-expr EXPR) -> EXPR; ...
```

dsl/dsl.c
masatake: Thank you. I will update my change.

dsl/dsl.c
AmaiKinono: Yes, `EXPR` is better. In fact we have some help strings that uses `exp`:

```
eq?: (eq? exp0 exp1) -> <boolean>
```

maybe it's good to unify the term.

masatake: For runtime evaluation, I think we can use OBJ.
For compile-time action(?), we can use EXPR.

masatake:
`print-expr` is renamed to `printX`.
I updated the messages displayed in `--help-expression filter`:
```
Filter expression: 
     #/PATTERN/: (#/patter/ <string>) -> <boolean>; regular expression matching
    #/PATTERN/i: (#/patter/i <string>) -> <boolean>; in case insensitive way
          null?: (null? <any>) -> <boolean>
          begin: (begin <any:0> ... <any:n>) -> <any:n>
         begin0: (begin0 <any:0> ... <any:n>) -> <any:0>
            and: (and <any> ...) -> <boolean>
             or: (or <any> ...) -> <boolean>
             if: (if <any:cond> <expr:true> <expr:false>) -> <any:true>|<any:false>
            not: (not <any>) -> <boolean>
            eq?: (eq? <any> <any>) -> <boolean>
              <: (< <integer> <integer>) -> <boolean>
              >: (> <integer> <integer>) -> <boolean>
             <=: (<= <integer> <integer>) -> <boolean>
             >=: (>= <integer> <integer>) -> <boolean>
        prefix?: (prefix? <string:target> <string:prefix>) -> <boolean>
        suffix?: (suffix? <string:target> <string:suffix>) -> <boolean>
        substr?: (substr? <string:target> string:substr>) -> <boolean>
         member: (member <any> <list>) -> #f|<list>
       downcase: (downcase <string>|<list>) -> <string>|<list>
         upcase: (upcase <string>|<list>) -> <string>|<list>
         length: (length <string>) -> <integer>
              +: (+ <integer> <integer>) -> <integer>
              -: (- <integer> <integer>) -> <integer>
         concat: (concat <string> ...) -> <string>; an alias for string-append
  string-append: (string-append <string> ...) -> <string>
 string->regexp: ((string->regexp <string:pattern>) <string:target>) -> <boolean>
   regexp-quote: (regexp-quote <string>) -> <string>
          print: (print OBJ) -> OBJ
         printX: (printX EXPR) -> EXPR; do the same as `print' but this works before evaluating
           true: -> #t
          false: -> #f
            nil: -> ()
              $: ($ <string:field>) -> #f|<string>
          $name: -> <string>
         $input: -> <string>; input file name
       $pattern: -> #f|<string>
          $line: -> #f|<integer>
        $access: -> #f|<string>
           $end: -> #f|<integer>
        $extras: -> #f|<string>
          $file: -> <boolean>; whether the scope is limited in the file or not.
      $inherits: -> <list>
$implementation: -> #f|<string>
          $kind: -> #f|<string>
      $language: -> #f|<string>
         $scope: -> #f|<string>; $scope-kind:$scope-name
    $scope-kind: -> #f|<string>
    $scope-name: -> #f|<string>
     $signature: -> #f|<string>
       $typeref: -> #f|<string>
         $roles: -> <list>
         $xpath: -> #f|<string>
```

PR #2837, opened 2021-02-01
buildsys;units: fix make check failed on MSYS2 when MSYSTEM=MINGW32 and MSYSTEM=MINGW64

leleliu008:
use `cygpath` to convert `Windows path` to `Unix path` if `cygpath` exists.

### DIFF
--- a/Tmain/tag-relative-option.d/run.sh
+++ b/Tmain/tag-relative-option.d/run.sh
@@ -34,8 +34,18 @@ fi
     ${CTAGS} $O --_xformat="yes:     $F -> %F"  --tag-relative=yes    $F
     ${CTAGS} $O --_xformat="always:  $F -> %F"  --tag-relative=always $F
 } | {
-	# Normalize driver letter
-    sed -e 's|C:|/c|g'
+    # Normalize Windows driver letter
+    #
+    # comment time:   Sat Feb  6 13:11:44 UTC 2021
+    # comment author: leleliu008@gmail.com
+    #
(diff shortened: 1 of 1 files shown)


masatake:
The code looks useful.
How about putting the code to Tmain/utils.sh as a function?

leleliu008:
@masatake It is not commonly used although it looks useful. Because this implementation depends on the data structure of this test case. If change the implementation  I am afraid make it indeterminate.

masatake:
@leleliu008, I see.

Tmain/tag-relative-option.d/run.sh
```diff
@@ -34,8 +34,12 @@ fi
     ${CTAGS} $O --_xformat="yes:     $F -> %F"  --tag-relative=yes    $F
     ${CTAGS} $O --_xformat="always:  $F -> %F"  --tag-relative=always $F
 } | {
-	# Normalize driver letter
-    sed -e 's|C:|/c|g'
+    # Normalize driver letter
+    if command -v cygpath > /dev/null && command -v awk > /dev/null ; then
+        awk '{if ($NF ~ /^[A-Z]:/) { "cygpath "$NF | getline newpath; sub($NF,newpath) } print}'
+    else
+        sed 's|\([A-Z]\):|/\l\1|'
```
k-takata, line 41: I think that `\l` is a GNU extension.
I'm not sure we can use it here.

leleliu008: Yes, `\l` is a GNU extension. But only Windows's path match `[A-Z]:` pattern, `Cygwin` and `MSYS2` use `GNU sed`

masatake: You are correct but I feel the technique is a bit tricky.
So could you write ' \l is a GNU extension. But only Windows's path match [A-Z]: pattern, Cygwin and MSYS2 use GNU sed' as a comment at the line?

leleliu008: Yes, I can. In addition, If you worried about this may cause problem in some cases, I can use `python` or `perl` to implements `sed 's|\([A-Z]\):|/\l\1|'`

masatake (approved):

k-takata (approved):

PR #2838, opened 2021-02-01
perl: added support for method keyword

jimmygoogle:
I added Perl support for the method keyword used in the Function::Parameters class.

### DIFF
--- a/parsers/perl.c
+++ b/parsers/perl.c
@@ -419,6 +419,14 @@ static void findPerlTags (void)
 			spaceRequired = true;
 			qualified = true;
 		}
+		else if (strncmp((const char*) cp, "fun", (size_t) 3) == 0)
+		{
+			TRACE("this looks like a sub\n");
+			cp += 3;
+			kind = KIND_PERL_SUBROUTINE;
+			spaceRequired = true;
+			qualified = true;
+		}
 		else if (strncmp((const char*) cp, "use", (size_t) 3) == 0)
 		{
 			cp += 3;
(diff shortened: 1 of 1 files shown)


masatake:
Thank you for this pull request.

Don't you want to support "fun" keyword?

jimmygoogle:
I should add 'fun' as well. My use case what primarily for 'method'.

masatake:
I have two requests:

* supporting fun
* adding test cases
Could you look at https://docs.ctags.io/en/latest/testing.html#units-test-facility for adding a test case?

jimmygoogle:
Yes I can do this.

masatake:
I must write my plan.
I would like to convert your change to a sub parser like a parser for Moose instead of merging your code directly. See parsers/moose.c.

The test cases for "method" and "fun" is quite important to make the subparser work as you, the original author of the change, expected.

About the concept of sub parser,  see
https://docs.ctags.io/en/latest/running-multi-parsers.html?highlight=subparser#tagging-definitions-of-higher-upper-level-language-sub-base .

masatake:
See #2841.

jimmygoogle:
Looking at things, I am way out of my depth in making this a subparser. I
could do a unit test but thats about it.

masatake:
I would like you to see the test case in #2841.
fun and method are detected only in files where Function::Parameters is used.

masatake:
I found I didn't the test case to my #2841 though I merged the pull request.
I made a new pull request that includes only the test case I forgot to add to #2841.
Could you look at #2846?

jimmygoogle:
I am still working on the test case. I didn't have time to finish it last
night.

PR #2841, opened 2021-02-02
FunctionParameters: new sub parser running on Perl parser 

masatake:
This subparser extracts 'fun' and 'method' that can be defined with
Function::Parameeters[1] module in a perl script.

@jimmygoogle extended Perl parser for Function::Parameeters and made a
pull request (#2838). Conceptually this converts it to a subparser.

Close #2838.


masatake:
I forgot to add a test case.

PR #2825, opened 2021-01-23
R: support vector, list, data.frame and named attributes in them

masatake:
Close #2823.

With this change, ctags can extract in  https://github.com/kklot/eppasm/blob/1988/R/popClass.R.
```
$ u-ctags  --kinds-R=+n --fields=+ne  -o - popClass.R | grep ^infect_mix
infect_mix	popClass.R	/^infect_mix = function(hivpop, artpop, ii) {$/;"	n	line:381	vector:popFunc	end:402	constructor:function
```

DONE:
-  split the last commit into smaller ones (if possile), and
-  write a man page for R parser (and its subparsers).

PR #2844, opened 2021-02-03
R: use "function" kind for functions defined in a function

masatake:
<BUG>
The original code attaches "funcvar" kind to functions defined in a
function.

If a variable in a function gets a value returned from a well-known
constructor, R parser should attach a suitable kind for the
constructor to the variable.

In this case, function() is the well-known constructor.

<HOW IT SHOULD BE>
"function" kind is for a variable that has a value returned from
function(), or a value itself returned from function(). In the latter
case, ctags generates an anonymous name for tagging the value.

"function" kind should be used whether the variable is defined
in the top-level context or in a function.

<NOTE>

Let's call function() a well-known constructor.

If a variable in a function has a value returned from a function call
OTHER THAN the well-known constructor, the tag for the variable should
have "globalVar" or "functionVar" as its kind.  If the variable is
assigned in the top-level context, it should have "globalVar" kind. If
not, it should have "functionVar" kind.

In the future, R parser may support c(), list(), and data.frame() as
well-known constructors. We will introduce "vector",
"list" and "daraFrame" kinds for them

This change is not only a bug fix but also a preparation for #2825.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -224,6 +224,7 @@ man_MANS = \
 	man/ctags-lang-python.7 \
 	man/ctags-lang-verilog.7 \
 	man/ctags-lang-inko.7 \
+	man/ctags-lang-r.7 \
 	\
 	$(NULL)
 rst2man_verbose = $(rst2man_verbose_@AM_V@)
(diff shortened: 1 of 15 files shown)


masatake:
Withdraw this because these changes are part of #2825 now.

PR #2852, opened 2021-02-07
roundtrip: skip too large tag files

masatake:

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -58,7 +58,7 @@ jobs:
        - run:
            name: Test
            command: |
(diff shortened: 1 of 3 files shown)


masatake:
When setting the limit to 50 entries per tag file, timeouts occur.
When setting the limit to 25 entities per tag file, roundtrip tests are done in less than 5 minutes.
These are the results on circle ci.
On GitHub Actions, it takes 0s!
I wonder what I should do.

masatake:
On circle ci, I will put the limitation.
Not a good idea but we need a solution now.

PR #2853, opened 2021-02-07
man-test: new test harness

masatake:
`man-test` is a new test harness for testing examples in per-language man pages (man/ctags-lang-\*.7.rst.in).

If you have python3, you can run it with "make man-test."

TODO:
- ~documentation for writing a test case~
- ~running on CI/CD environment~

PR #2858, opened 2021-02-11
docs: update markup rules

hirooih:
- describe the markup rules in `docs/README.md`.
  - move most of contents of `man/README` to `docs/README.md` and `docs/contributions.rst` .
- update the contents in `docs/README.md`.
- delete unnecessary section markups from `docs/man/*.rst`.


docs/README.md
```diff
@@ -9,6 +9,38 @@ See [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.htm
...
+* put two single quotes with colon around field name like ``` ``field:`` ```.
+
+* put double quotes aground an example of command line, option usage like
+  "--langdef=MyLang", referring a section, and a file path.
+
+* use two asterisks for newly introduced conceptual `**word**` and reserved
+  words. The reserved words are
+
+  * "default" in --langmap
+
```
masatake, line 33: I found reserved words in ctags.h.

```
  * "all" in --list-kinds-full=all and ...
  * "auto" in --language-force=auto
  * "NONE" in --options=NONE
 ```

File: docs/README.md [hunk 2]
```diff
@@ -9,6 +9,38 @@ See [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.htm
...
+* use one asterisk for option parameter like `*LANG*` in --kind-<LANG> option.
```
masatake: I wonder which one we should use `*MATAVAR*`, `<METAVAR>`, or `*<METAVAR>*`.
In ``--help`` output, we have used  `<METAVAR>` because u-ctags did so.
Just using the uppercase string `METAVAR` is not enough in the help message. Let's think `C`. 
People may wonder whether it represents a name of a parser or a metavariable.
We cannot use a style like bold or italic in the help message.
So using `<` and `>` is understandable. 

On the man page, we can use a style for the purpose. So I wonder how we should do.
I'm familiar with `<...>` notation, so I would like to use `<METAVAR>` or `*<METAVAR>*`.

As far as I know, currently `<` and `>` are used only for `LANG`. I would like to use the notation for
other places like `<KIND>`.

This is just a comment.

masatake: It seems that we should use `*<METAVAR>*` in the man pages.

masatake (approved):

masatake:
Thank you. I need this to write a new document.

I want to introduce one more step, "making examples in the man page testable" in "writing a man page for a parser" procedures. 

Though I didn't get your review, I introduced `man-test` and its document a few days ago.
https://docs.ctags.io/en/latest/testing.html#testing-examples-in-language-specific-man-pages

docs/README.md
hirooih: Yes.  As you already found, on 7667692 I moved the contents from `man/README` to `docs/README.md` **as is**.
`*<METAVAR>*` in e37d287 is my proposal.

hirooih:
Thank you for your review an approval.  I've merged this PR.

> Though I didn't get your review, I introduced man-test and its document a few days ago.
> https://docs.ctags.io/en/latest/testing.html#testing-examples-in-language-specific-man-pages

I read, and have some questions about the description.  Let me ask you on #2853.

PR #2864, opened 2021-02-13
JavaScript: support fields in a class

### LINKED ISSUE #2849: JavaScript: support fields (Was: JavaScript stops parsing after property assigned to empty array)
JavaScript files stop parsing when encountering a class field that's assigned to an empty array, e.g.
```js
class EmptyArrayTest {
  field1 = [];
  ...
}
```
This does not seem to occur if the field is assigned to an array with elements (`['foo']`), numbers, strings, or object literals (empty or otherwise).  The presence or absence of the `static` modifier on the field doesn't make a difference.

In the example below, I have a class defined after the end of `EmptyArrayTest`, which doesn't show up at all in the ctags output.  If I move the `NumberTest` class before `EmptyArrayTest`, its tags are generated properly, but the method in `EmptyArrayTest` is not.

(Note: It looks like the default javascript parser doesn't generate tags for class fields.  This might be a bug, but it's not what this issue is about.)

*****

The name of the parser:
javascript

```
$ ctags --options=NONE -f - emptyarrayprop.js
```

The content of input file:

```js
class EmptyArrayTest { 
  field1 = [];

  method1() { 
    console.log('field1 has', this.field1.length, 'items');
  } 
}

class NumberTest { 
  fieldAnswer = 42;

  methodAnswer() { 
    console.log('the answer is', this.fieldAnswer);
  } 
}
```

```
EmptyArrayTest  emptyarrayprop.js       /^class EmptyArrayTest {$/;"    c
```

The tags output you expect:

```
EmptyArrayTest  emptyarrayprop.js       /^class EmptyArrayTest {$/;"    c
NumberTest      emptyarrayprop.js       /^class NumberTest {$/;"        c
method1 emptyarrayprop.js       /^  method1() {$/;"     m       class:EmptyArrayTest
methodAnswer    emptyarrayprop.js       /^  methodAnswer() {$/;"        m       class:NumberTest
```

The version of ctags:

```
$ ctags --version
  Compiled: Feb  6 2021, 12:49:59
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc
```

How do you get ctags binary:
MacPorts: `sudo port install universal-ctags`

*****

(@masatake added)
To solve this issue, we have to enhance ctags to support *fields*.

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations
* https://github.com/tc39/proposal-class-fields
* https://tc39.es/proposal-class-fields/

masatake:
Close #2849 as side effect of this change.

input0.js (quoted from #2849):

    class EmptyArrayTest {
      field1 = [];

      method1() {
	console.log('field1 has', this.field1.length, 'items');
      }
    }

output0.tags:

    EmptyArrayTest	input.js	/^class EmptyArrayTest { $/;"	c
    field1	input.js	/^  field1 = [];$/;"	M	class:EmptyArrayTest
    method1	input.js	/^  method1() { $/;"	m	class:EmptyArrayTest

"field1" is captured as a member/M kind tag.

References:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#field_declarations
* https://github.com/tc39/proposal-class-fields
* https://tc39.es/proposal-class-fields/

TODO:
* private methods
* class methods


masatake:
@b4n, can I ask you to review this?

In addition, I have a topic about that I would like to get a comment from you.

```JavaScript
v0 = function (x) { return x; }
class C0 {
    f0 = function (x) { return x; }
}
```

With the change I proposed in this pull request, ctags extracts `f0` as a tag with field kind.
On the other hand, v0 is extracted as a function, not a variable.
How do you think about this?

I call the technique used in assigning function kind to v0 "kind propagation".

We can extract `f0` as a function having C0 as its scope using the kind propagation.

However, I guess JavaScript people may want to call f0 "a field", not "a function".

During working on R, I met the same situation.
Instead of using the kind propagation, I introduced a parser specific field.

input.r:
```R
L <- list (
    first = function() { 1 },
    second = function() { 2 }
)
```

tags:
```console
$ ./ctags -o - --fields=+KZz input.r
L	input.r	/^L <- list ($/;"	kind:list
first	input.r	/^    first = function() { 1 },$/;"	kind:nameattr	scope:list:L	constructor:function
second	input.r	/^    second = function() { 2 }$/;"	kind:nameattr	scope:list:L	constructor:function
```

From the constructor field, a user or a tool used by the user can know `first` is a name-attribute in a list, AND it has a function as its value. "constructor:" field can be added to "field" kind of JavaScript parser. 
I am no sure whether Geany uses parser specific fields or not. But I would like to hear your comment.

masatake:
@tristanHessell, I would like you to see my last comment. If you have any comments, I would like to know them.

flwyd:
> ```js
> v0 = function (x) { return x; }
> class C0 {
>     f0 = function (x) { return x; }
> }
> ```
> 
> With the change I proposed in this pull request, ctags extracts `f0` as a tag with field kind.
> On the other hand, `v0` is extracted as a function, not a variable.
> How do you think about this?

I think it's good that `v0` is considered a function rather than a variable; this is a common JavaScript idiom for dynamically declaring functions.

I think the `f0` case is likely to be rare in the wild, since it would be simpler to define this as
```js
class C0 {
  f0(x) { return x; }
}
```
I searched through a few hundred JavaScript modules which use classes that are dependencies of an Ember project I've worked on.   I couldn't find any class members which were defined like `f0 = function(…)` or `f0 = x => …`.  When methods are dynamically assigned, it is customary to do so in the constructor:
```js
class C1 {
  constructor(y) {
    this.f0 = function(x) { return x + y; }
    this.f1 = x => x + y;
  }
}
```

So treating `f0` as a field in your example seems reasonable, and unlikely to affect most users.

masatake:
@flwyd, thank you very much. Your input is quite helpful for me because I don't know well about JavaScript.

parsers/jscript.c
```diff
@@ -1800,6 +1806,23 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 							makeJsTag (name, JSTAG_PROPERTY, NULL, NULL);
 				}
 			}
+			else if (isType (token, TOKEN_EQUAL_SIGN)
+					 || isType (token, TOKEN_SEMICOLON)
+					 || isType (token, TOKEN_CLOSE_CURLY)
+					 || (isType (token, TOKEN_IDENTIFIER)
+						 && name->lineNumber != token->lineNumber))
```
b4n, line 1813: I didn't review this carefully, but this strikes me as a huge problem, as it means indentation affects parsing.  AFAIK, there's no JS construct that require actual line breaks. I'm not familiar with the notation this PR handles, but I'd be surprised it introduced such a fundamental language change.

b4n: Either this would need specific handling in the prevtoken hack (speaking from memory), or record a line break has occurred so special handling can be done where needed.

In general implicit semicolons in JS is tricky to handle, because it can only be recognized on the parsing phase, not when lexing/tokenizing.  This means that to properly handle them, we should understand the *whole* JS syntax, and not the current "catchphrase-subset" we currently do (e.g. recognize specific interesting pieces and act upon that alone).  Up until now, we managed to get away with heuristics that worked in the vast majority (if not all) of the relevant cases, but maybe the field syntax breaks these, or we need to adjust it.

IIUC something like
```JS
class C0 {
    method1
    () {
    }
```
would be recognized as a `method1` field, although I'd bet JS actually thinks it's a method.

b4n:
> @b4n, can I ask you to review this?

I'll try, yes.  But not tonight, I'm not fresh enough -- and thus, my remarks above might lack some insight.

> In addition, I have a topic about that I would like to get a comment from you.
> 
> ```js
> v0 = function (x) { return x; }
> class C0 {
>     f0 = function (x) { return x; }
> }
> ```
> 
> With the change I proposed in this pull request, ctags extracts `f0` as a tag with field kind.
> On the other hand, v0 is extracted as a function, not a variable.
> How do you think about this?

I think f0 should be a function/method, because it's what it actually holds and can be used as.  The fact it uses one syntax or another to declare it doesn't really matter -- or *that* could be a parser-specific field if really somebody cares.  So, using your terms, IMO its kind should propagate.

> […]
> From the constructor field, a user or a tool used by the user can know `first` is a name-attribute in a list, AND it has a function as its value. "constructor:" field can be added to "field" kind of JavaScript parser.

I don't know R, but I don't really see how not having the kind `function` is useful.  A wild guess on my part would be that a name `function` inside a `list` ought to be a named element, so that adding that extra info is not really useful.  And more importantly, I would think it's more useful to know the values are functions than knowing they are values, so a tool that doesn't have specific handling for R should rather see them as functions.\
But all that is without actually knowing anything about R, maybe I'm missing something that makes your choice more appropriate.

> I am no sure whether Geany uses parser specific fields or not. But I would like to hear your comment.

We currently don't.  And supporting specific fields is a little harder when you're building a generic program that supports various languages, because at some level you'll have to be able to make them all fit in the same mold, would it only be to provide a consistent interface to interact with all of them.  Of course there can be speciific handling, but the more there are, the less useful it is to have a single tool like ctags to get the info from.

flwyd:
> > @b4n, can I ask you to review this?
> 
> I'll try, yes. But not tonight, I'm not fresh enough -- and thus, my remarks above might lack some insight.
> 
> > In addition, I have a topic about that I would like to get a comment from you.
> > ```js
> > v0 = function (x) { return x; }
> > class C0 {
> >     f0 = function (x) { return x; }
> > }
> > ```
> > 
> > 
> > With the change I proposed in this pull request, ctags extracts `f0` as a tag with field kind.
> > On the other hand, v0 is extracted as a function, not a variable.
> > How do you think about this?
> 
> I think f0 should be a function/method, because it's what it actually holds and can be used as. The fact it uses one syntax or another to declare it doesn't really matter -- or _that_ could be a parser-specific field if really somebody cares. So, using your terms, IMO its kind should propagate.

In the general case, whether a field contains a function can't be determined statically.  For example,
```js
let f2_function = true;
function logprefix(prefix) { return (...args) => console.log(prefix, ...args); }
function capitalize(str) { return str.toUpperCase(); }
class C0 {
  f0 = capitalize('hello');
  f1 = logprefix('hello');
  constructor() {
    this.f2 = f2_function ? this.f1 : this.f0;
}
```
creates a class where `new C0().f0` is the string `HELLO` while `new C0().f1('world')` calls a function that prints `hello world` to the console and `let x = new C0().f2` sets `x` to either a function or, if someone set `f2_function = false`, a string.  In TypeScript a dynamically-assigned "method field" would have a type, e.g. `class C0 { f0: (x) => string; }`, but since type inference is allowed for `class C0 { f0 = logprefix('hello'); }` the general case is still hard. 

The same is true of functions assigned to variables, so perhaps the best that can be hoped for is a heuristic on the syntactic form of the right-hand-side of an assignment, whether that's a variable, object property, or class field.  That said, function assignment to variables or object properties is much more common than function assignment as part of a class field.

While playing around with this example, I noticed that universal-ctags version 5.9.0 handles `function(x)` assignment differently than "fat arrow" assignment:
```js
let varfun = function(x) { console.log('varfun', x); }
let fatarrow = (x) => console.log('fatarrow', x);
let logprefix = (prefix) => { return (...args) => console.log(prefix, ...args); }
let logit = logprefix('it');

let myobj = {
  x: function(a) { return a + 2; },
  y: (b) => b - 2,
};
myobj.z = function(c) { return c * 2; };
myobj.w = (d) => d / 2;

const MyClass = function() {
  this.a = 'hello';
  this.b = function() { console.log(this.a, 'world'); }
  this.c = () => console.log(this.a, 'dolly');
}
MyClass.prototype.d = 'howdy';
MyClass.prototype.e = x => 'hello ' + x;
MyClass.prototype.f = function(x) { return this.d + ' ' + x; }
```
produces
```
MyClass funcs.js        /^const MyClass = function() {$/;"      c
b       funcs.js        /^      this.b = function() { console.log(this.a, 'world'); }$/;"       m  class:MyClass
d       funcs.js        /^MyClass.prototype.d = 'howdy';$/;"    m       class:MyClass
e       funcs.js        /^MyClass.prototype.e = x => 'hello ' + x;$/;"  m       class:MyClass
f       funcs.js        /^MyClass.prototype.f = function(x) { return this.d + ' ' + x; }$/;"    m  class:MyClass
fatarrow        funcs.js        /^let fatarrow = (x) => console.log('fatarrow', x);$/;" v
logit   funcs.js        /^let logit = logprefix('it');$/;"      v
logprefix       funcs.js        /^let logprefix = (prefix) => { return (...args) => console.log(prefix, ...args); }$/;"     v
myobj   funcs.js        /^let myobj = {$/;"     c
varfun  funcs.js        /^let varfun = function(x) { console.log('varfun', x); }$/;"    f
x       funcs.js        /^      x: function(a) { return a + 2; },$/;"   m       class:myobj
y       funcs.js        /^      y: (b) => b - 2,$/;"    p       class:myobj
z       funcs.js        /^myobj.z = function(c) { return c * 2; };$/;"  f       function:myobj
```
This identifies `varfun` as type `f` but `fatarrow` as type `v`.  For object properties, it treats `x` (an inline property defined with `function`) as type `m`, `y` (an inline property defined with a fat arrow) as `p`, and `z` (a non-inlined property defined with `function`) as type `f`, but `w` (a non-inlined property defined with fat arrow) is ignored.  The `MyClass` properties (using pre-ES2015 style) pick up `b` (`function` defined in the constructor), `d` (string defined on prototype), `e` (fat arrow defined on prototype), and `f` (`function` defined on prototype) as type `m` , while `a` (string defined in the constructor) and `c` (fat arrow defined in the constructor) are ignored.

It would be great if function expressions (`function(x) { ... }`) and fat arrow expressions (`x => ...`) were treated identically since they both result in functions, just with different semantics around variable scoping.  It would also be nice if old-style class fields (`this.foo` assignments in a function as well as `.prototype.` assignments) had the same ctags semantics as ES2015-style `class` fields (whether declared directly as fields or assigned like `constructor {this.foo = 'hello';}`).  It would also be great if JavaScript and TypeScript had the same ctags behavior.

b4n:
> It would be great if function expressions (`function(x) { ... }`) and fat arrow expressions (`x => ...`) were treated identically since they both result in functions, just with different semantics around variable scoping.

Yes.  Actually, if I'm up to date in this area, it's just that "fat arrow" is not recognized at all by the parser, so it just treats it as any other kind of value (but not function).  This could be improved to be recognized properly.

> It would also be nice if old-style class fields (`this.foo` assignments in a function as well as `.prototype.` assignments) had the same ctags semantics as ES2015-style `class` fields (whether declared directly as fields or assigned like `constructor {this.foo = 'hello';}`).

You mean having any `this.foo = …` assignation create a member in whichever calss `this` refers to?  I think we kind of try and do that, but that handling probably has several shortcomings.

> It would also be great if JavaScript and TypeScript had the same ctags behavior.

I'm not very familiar with TypeScrpit, but there's a specific parser for it in ctags.  I don't know if it's a good or bad thing, but reading your sentence suggests it is not.

At any rate, as you seem to be a whole lot more familiar with *using* JS and TS than I am (or any of the ctags devs are AFAIK), it would be stellar if you could propose test cases for missing support & al -- if not even providing patches directly :wink:.  I mean, I'd be happy to try and improve the parser, but it would help to have expected results coming from somebody that really uses it in a daily basis, instead of me that mostly know what the standard say when I read it, and do casual JS when I have to :slightly_smiling_face:

flwyd:
I'd be happy to provide some JS and TS test cases.  Should I create a separate issue for them, or add on to #2849?

b4n:
> I'd be happy to provide some JS and TS test cases. Should I create a separate issue for them, or add on to #2849?

I'd say if you could, even open a PR adding the test cases as "broken" ones (`Tests/parser-javascript.r/<something>.b`, see *docs/units.rst*).  The more targeted each test case the better, but don't hesitate to have more complex ones if needed as well.
If you can't PR, I'd say open separate issues.

parsers/jscript.c
masatake: Thank you.
After reading your review comment, I found comparing two `lineNumber` is not needed.

```
$ node --version
node --version
v14.15.4
$ cat /tmp/a.js 
cat /tmp/a.js 
class A {
    a0
    a1
}
$ cat /tmp/b.js 
cat /tmp/b.js 
class B {
    b0 b1
}

$ node /tmp/a.js 
node /tmp/a.js 
$ node /tmp/b.js 
node /tmp/b.js 
/tmp/b.js:2
    b0 b1
       ^^

SyntaxError: Unexpected identifier
    at wrapSafe (internal/modules/cjs/loader.js:979:16)
    at Module._compile (internal/modules/cjs/loader.js:1027:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

The code comparing two `lineNumber` is for rejecting `b1`.
However, the sequence of `b0 b1` is no valid JavaScript code. For such garbage input, ctags is allowed emitting garbage; JavaScript doesn't have to reject `b1` explicitly.

I will update the proposed changes.

Units/parser-javascript.r/js-fields.d/expected.tags
```diff
@@ -0,0 +1,32 @@
...
+field2	input.js	/^    field2;$/;"	M	class:Class0
+field3	input.js	/^    field3$/;"	M	class:Class0
+field4	input.js	/^    field4$/;"	M	class:Class0
+field5	input.js	/^    field5$/;"	M	class:Class0
+field6	input.js	/^    field6;$/;"	M	class:Class0
+field7	input.js	/^    field7$/;"	M	class:Class0
+field8	input.js	/^    field8 = 1;$/;"	M	class:Class0
+field9	input.js	/^    field9 = "abc"$/;"	M	class:Class0
+field10	input.js	/^    field10 = {}$/;"	M	class:Class0
+field11	input.js	/^    field11 = function(x, y) {$/;"	f	function:Class0
```
b4n, line 13: Should have scope `class:Class0` (see below)

parsers/jscript.c
```diff
@@ -1800,6 +1806,26 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
+			else if (isType (token, TOKEN_EQUAL_SIGN)
+					 || isType (token, TOKEN_SEMICOLON)
+					 || isType (token, TOKEN_CLOSE_CURLY)
+					 || (isType (token, TOKEN_IDENTIFIER)))
+			{
+				if (isType (token, TOKEN_EQUAL_SIGN))
+				{
+					readToken (token);
+					if (isKeyword (token, KEYWORD_function))
+						makeJsTag (name, JSTAG_FUNCTION, NULL, NULL);
```
b4n, line 1818: Inside a class it's gonna be a method, not a function (which is distinct in this parser, and will fix the scope kind for these fields)
```suggestion
						makeJsTag (name, JSTAG_METHOD NULL, NULL);
```

masatake: @b4n, thank you.
I fixed and pushed the change. If the changes should be squashed into one for your review, let me know.

Units/parser-javascript.r/js-fields.d/expected.tags
```diff
@@ -0,0 +1,31 @@
...
+field3	input.js	/^    field3$/;"	M	class:Class0
+field4	input.js	/^    field4$/;"	M	class:Class0
+field5	input.js	/^    field5$/;"	M	class:Class0
+field6	input.js	/^    field6;$/;"	M	class:Class0
+field7	input.js	/^    field7$/;"	M	class:Class0
+field8	input.js	/^    field8 = 1;$/;"	M	class:Class0
+field9	input.js	/^    field9 = "abc"$/;"	M	class:Class0
+field10	input.js	/^    field10 = {}$/;"	M	class:Class0
+field11	input.js	/^    field11 = function(x, y) {$/;"	m	class:Class0
+fun1	input.js	/^	var fun1 = function(x) { return x + 1 }$/;"	f	function:Class0.field11
```
masatake, line 14: The scope should be "method", not "function".

masatake: This one is hard to fix without cork.

PR #2875, opened 2021-02-17
Tmain(client): add test for vista.vim

liuchengxu:
The command used in the test is hardcoded as the command used in vista.vim is very stable now.

Ref https://github.com/liuchengxu/vista.vim/issues/386#issuecomment-780075143


Tmain/client-vista-vim-fields-expectation.d/run.sh
```diff
@@ -0,0 +1,11 @@
+# Copyright: 2021 <liuchengxu@github.com>
+# License: GPL-2
+#!/usr/bin/env bash
```
masatake, line 3: `#!...` line is not needed.

Tmain/client-vista-vim-fields-expectation.d/run.sh
```diff
@@ -0,0 +1,11 @@
...
+
+CTAGS=$1
+
+. ../utils.sh
+
```
masatake, line 8: Could you add
```
is_feature_available "${CTAGS}" json
```
This line is for skipping this test case if ctags has no json writer.

Tmain/client-vista-vim-fields-expectation.d/run.sh
```diff
@@ -0,0 +1,11 @@
...
+$CTAGS --format=2 --excmd=pattern --fields=+nksSaf --extras=+F --sort=no --append=no --extras=  --language-force=vim --vim-kinds=acfvmn --output-format=json --fields=-PF -f- $(pwd)/test.vim
```
masatake, line 9: Could you remove `$(pwd)/` ?

masatake (changes requested):
Could you change the commit log to
```
Tmain(client): add test for vista.vim

Ref liuchengxu/vista.vim#386 (comment)
```

masatake:
Thank you. Merged after squashing.

PR #2891, opened 2021-03-01
Tmain: using `shift_jis` instead of `cp932` coding.

leleliu008:
`musl-libc iconv API` dosn't support `cp932` coding but supports
`shift_jis` coding. `shift_jis` coding is also supportted by `GNU
libiconv`

although `shift_jis` coding is different from `cp932` coding, but in our
case, it can be replaced by each other.

References:
http://una.soragoto.net/topics/13.html
https://wiki.musl-libc.org/functional-differences-from-glibc.html
https://github.com/bminor/musl/blob/master/src/locale/iconv.c
https://www.gnu.org/software/libiconv

### DIFF
--- a/Tmain/input-encoding-option.d/run.sh
+++ b/Tmain/input-encoding-option.d/run.sh
@@ -11,7 +11,7 @@ BUILDDIR=$2
 if ${CTAGS} --quiet --options=NONE --list-features | grep -q iconv; then
   if ${CTAGS} --quiet --options=NONE \
 	      --pseudo-tags=-TAG_PROC_CWD \
-	      --input-encoding=utf-8 --input-encoding-java=cp932 --input-encoding-javascript=euc-jp \
+	      --input-encoding=utf-8 --input-encoding-java=shift_jis --input-encoding-javascript=euc-jp \
 	      -o ${BUILDDIR}/tags \
 	      input.js input.java ; then
       remove_commit_id ${BUILDDIR}/tags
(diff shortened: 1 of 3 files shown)


docs/news.rst
```diff
@@ -34,7 +34,7 @@ See :ref:`option_input_output_file` in :ref:`ctags(1) <ctags(1)>`.
 People may use their own native language in source code comments (or
 sometimes in identifiers) and in such cases encoding may become an issue.
 Nowadays UTF-8 is the most widely used encoding, but some source codes
-still use legacy encodings like latin1, cp932 and so on. These options
+still use legacy encodings like latin1, shift_jis and so on. These options
```
k-takata, line 37: I don't think this needs to be changed.

leleliu008: If this not changed, `musl-libc` user also will be confused when they find `cp932` is not work.

k-takata: As a Japanese user, many Windows source codes were written in cp932 not in shift_jis, using shift_jis as an example confuses many Japanese Windows users.

leleliu008: `cp932` coding is motioned in our docs, but it can not work when os use `musl-libc`, and our docs don't provide a solution, I think these people who really coming to find answers will be confused.

k-takata: The example `cp932` is for Japanese Windows users (as the iconv feature was originally created for them), and win32 u-ctags uses GNU iconv not musl.

leleliu008: OK. I will switch `docs/news.rst` back to Origin.

k-takata (approved):

masatake:
Could you add `Tmain: ` as the prefix for the commit header?
So we can know the change is about a test case in Tmain.
Other than this, it looks good.

leleliu008:
@masatake Done.

masatake (approved):

masatake:
Thank you.

PR #2903, opened 2021-03-07
man-test: specify encoding when opening a file

### LINKED ISSUE #2901: misc/man-test.py failes on UTF-8 charactors
From #2900

https://app.circleci.com/pipelines/github/universal-ctags/ctags/1131/workflows/9e34863f-bc8a-4c1e-bbf8-08fd26e2cb9e/jobs/11765

```
# Run test cases in ./man/ctags-lang-verilog.7.rst.in
```
Traceback (most recent call last):
  File "./misc/man-test.py", line 256, in <module>
    sys.exit(0 if man_tests (tmpdir, ctags, sys.argv[3:]) else 1)
  File "./misc/man-test.py", line 246, in man_tests
    if not man_test(tmpdir, ctags, m):
  File "./misc/man-test.py", line 224, in man_test
    for t in extract_test_cases (f):
  File "./misc/man-test.py", line 155, in extract_test_cases
    for line in  f.readlines():
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 7586: ordinal not in range(128)
make: *** [man-test] Error 1
```

This caused by the following input.

```
- Language Reference Manuals (LRM)
   - IEEE Standard for SystemVerilog — Unified Hardware Design, Specification, and
     Verification Language, IEEE Std 1800-2017,
```

Note that "—" is an UTF-8 character.

It is strange for me because I thought Python 3.x could handle UTF-8.

### PR DESCRIPTION
**Pull request by masatake**
Close #2901.

### DIFF
--- a/misc/man-test.py
+++ b/misc/man-test.py
@@ -68,14 +68,14 @@ def run_test_case(tmpdir, ctags, t):
     D = d + '/' + 'tags.diff'
     O0 = 'args.ctags'
     O = d + '/' + O0
-    with open(i, mode='w') as f:
+    with open(i, mode='w', encoding='utf-8') as f:
         f.write(t['code'])
 
-    with open(e, mode='w') as g:
+    with open(e, mode='w', encoding='utf-8') as g:
         g.write(t['tags'])
 
     inputf=None
-    with open(O, mode='w') as Of:
+    with open(O, mode='w', encoding='utf-8') as Of:
         for c in t['cmdline'].split():
             if c == '--options=NONE':
                 continue
@@ -84,12 +84,12 @@ def run_test_case(tmpdir, ctags, t):
                 continue
             print (c, file=Of)
 
-    with open(o, mode='w') as h:
+    with open(o, mode='w', encoding='utf-8') as h:
         cmdline = [ctags, '--quiet', '--options=NONE',
                    '--options=' + O0, inputf]
         subprocess.run(cmdline, cwd=d, stdout=h)
 
-    with open(D, mode='w') as diff:
+    with open(D, mode='w', encoding='utf-8') as diff:
         r = subprocess.run(['diff', '-uN', '--strip-trailing-cr', o0, e0],
                            cwd=d, stdout=diff).returncode
 
@@ -97,11 +97,11 @@ def run_test_case(tmpdir, ctags, t):
         t['result'] = True
         t['result_readable'] = 'passed'
     else:
-        with open(o) as f:
+        with open(o, encoding='utf-8') as f:
             t['actual_tags'] = f.read()
         t['result'] = False
         t['result_readable'] = 'failed'
-        with open(D) as diff:
+        with open(D, encoding='utf-8') as diff:
             t['tags_diff'] = diff.read()
     os.remove(O)
     os.remove(i)
@@ -220,7 +220,7 @@ def man_test (tmpdir, ctags, man_file):
     result = True
     print ('# Run test cases in ' + man_file)
     print ('```')
-    with open(man_file) as f:
+    with open(man_file, encoding='utf-8') as f:
         for t in extract_test_cases (f):
             t['man_file'] = man_file
             v = verify_test_case (t)

### CODE REVIEW

**Comment by masatake** [comment]
@hirooih, could you try this one?

**Comment by hirooih** [comment]
> @hirooih, could you try this one?

Sure...  But I cannot reproduce the fail.

man-test.py is not fixed yet (version in repository). 
ctags-lang-verilog.7.rst.in include UTF-8 characters.

````
$ python3 --version
Python 3.8.5
$ LANG=C ./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
# Run test cases in man/ctags-lang-verilog.7.rst.in
```
```
OK
$ 
````

Am I missing something?

**Comment by masatake** [comment]
How about removing LANG=C at the head of the command line?

**Comment by hirooih** [comment]
> How about removing LANG=C at the head of the command line?

Same.

````
$ grep -n -P "[^\x00-\x7F]" man/ctags-lang-verilog.7.rst.in
182:   - IEEE Standard for SystemVerilog — Unified Hardware Design, Specification, and
$ echo $LANG
C.UTF-8
$ ./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
# Run test cases in man/ctags-lang-verilog.7.rst.in
```
```
OK
$ LANG= ./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
# Run test cases in man/ctags-lang-verilog.7.rst.in
```
```
OK
$ 

````

I am using Ubuntu-20.04 on WSL2.

Can you reproduce the fail?

**Comment by masatake** [comment]
If things work fine, we don't need this change. You can close #2901. We can reopen this anytime we want.

**Comment by masatake** [comment]
Did you try Centos7?
Your original report is about `centos7_make` environment on circleci.

**Comment by masatake** [comment]
It may be centos7 specific.
With which branch of your repository the issue can be reproduced?
I will make a pull request mixing your branch and the change of this pull request.

**Comment by hirooih** [comment]
> Did you try Centos7?

No. I don't have Centos7 system, sorry.

```
  File "./misc/man-test.py", line 155, in extract_test_cases
    for line in  f.readlines():
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 7586: ordinal not in range(128)
```
It may be Python 3.6 specific.

**Comment by masatake** [comment]
> No. I don't have Centos7 system, sorry.
> 
> ```
>   File "./misc/man-test.py", line 155, in extract_test_cases
>     for line in  f.readlines():
>   File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
>     return codecs.ascii_decode(input, self.errors)[0]
> UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 7586: ordinal not in range(128)
> ```
> 
> It may be Python 3.6 specific.

How do you reproduce this?

**Comment by leleliu008** [comment]
@masatake I have confirmed  that this PR works fine on `CentOS 7.5.1804`. 

`man/ctags-lang-verilog.7.rst.in`'s origin encoding is `ISO-8859`, it should be convert to `UTF-8` encoding. one of the way is as following:
```
iconv -t utf-8 -o man/ctags-lang-verilog.7.rst.in.utf8 man/ctags-lang-verilog.7.rst.in
mv man/ctags-lang-verilog.7.rst.in.utf8 man/ctags-lang-verilog.7.rst.in
```

**Comment by masatake** [comment]
@leleliu008, thank you.

@hirooih, you may have changes that were not submitted as a pull request because of the encoding error.

Now you can make a pull request for your changes with the following steps:

1. make a breach for the changes.
2. cherry-pick my change of this pull request.
3. add your changes and commit.

The pull request derived from the branch must pass the man-test test cases.

**Comment by hirooih** [comment]
@masatake san,

> How do you reproduce this?

I referred the messages from https://app.circleci.com/pipelines/github/universal-ctags/ctags/1131/workflows/9e34863f-bc8a-4c1e-bbf8-08fd26e2cb9e/jobs/11765.

@leleliu008,

Thank you for your investigation.
It seems that the real problem was that I committed ISO-8895 file.

BTW How did you get original ISO-8895 encoded file?
I squashed the fix with ascii '-' and  "push --force".  The original file does not exist in my repository.

I converted man/ctags-lang-verilog.7.rst.in to ISO-8859-1.  But I could not reproduce the fail with LANG=, LANG=C, nor LANG=C.UTF-8.
Can you reproduce the fail?

**Comment by leleliu008** [comment]
After a careful reading of [open()](https://docs.python.org/3/library/functions.html?highlight=open#open) we can known that if text encoding is not specified when invoking `open(file)`, the default encoding will be used, the default encoding is `locale.getpreferredencoding(False)`, this value can be changed via set `LANG` environment variable to different value.

to reduce the whole file size, most(almost all) docker image's locale preset as following:
```
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```
You can see that the default locale is `POSIX` which will only work for displaying `ASCII` characters. that's why reported `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 7586: ordinal not in range(128)`

Non-English Language users must set `LANG` environment variable first to use docker.

But in our case, we don't modify the config file, because we usually avoid to change too much to fix a bug.

this PR is fix the issue, so you can't reproduce the issue.

about the `ISO-8895` encoding, I made a mistake, `man/ctags-lang-verilog.7.rst.in`'s origin encoding is `ASCII`, I add some `Unicode` characters, it changed to `ISO-8895`. this depends on what editor you are using. I'd like to remind you that this file must be 'UTF-8' encoding if you put some `Unicode` characters into it. otherwise you will face another problem.

**Comment by hirooih** [comment]
> After a careful reading of open() we can known that if text encoding is not specified when invoking open(file), the default encoding will be used, the default encoding is locale.getpreferredencoding(False), this value can be changed via set LANG environment variable to different value.

This is what I wrote in #2901 as;

> In general it is better for a program to follow the users' locale setting (as Python does). It is a POSIX way.

I am negative to set encoding (and locale) in a program explicitly.

> this PR is fix the issue, so you can't reproduce the issue.

I have been using the original misc/man-test.py which is not fixed yet.

Please replace "-" with "—" in your input file and try it under your environment.

I guess python 3.6.x fails and 3.8.x does not.
Even in ASICC mode, f.readlines() can read non-ASCII characters AS IS, as old standard C libraries does.

**Comment by leleliu008** [comment]
@hirooih 

image

you can run following command to get it's value:
```
python3 -c 'import locale; print(locale.getpreferredencoding());'
```
if the result is not `UTF-8`, it can reproduce.


**how to reproduce this issue:**

step1.  pull a docker image and run it:
```
docker run --tty --interactive --workdir=/root --hostname=centos-7.5.1804 --name=centos-7.5.1804 centos:7.5.1804 /bin/bash
```
step2.  install needed tools
```
yum -y install git python3
```
step3.  clone source code form github
```
```
step4.  change to ctags directory
```
cd ctags
```
step5.  edit `man/ctags-lang-verilog.7.rst.in`, add some Unicode characters and save it.
```
echo '—' >> man/ctags-lang-verilog.7.rst.in
```
step6.  run man-test
```
./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
```
report as following:
```
Traceback (most recent call last):
  File "./misc/man-test.py", line 256, in <module>
    sys.exit(0 if man_tests (tmpdir, ctags, sys.argv[3:]) else 1)
  File "./misc/man-test.py", line 246, in man_tests
    if not man_test(tmpdir, ctags, m):
  File "./misc/man-test.py", line 224, in man_test
    for t in extract_test_cases (f):
  File "./misc/man-test.py", line 155, in extract_test_cases
    for line in  f.readlines():
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 7862: ordinal not in range(128)
```

**Fix the issue:**

there exists two ways for fixing this issue.

**_WAY1:_**
step1:
```
export LANG=en_US.UTF-8
```
step2.  re-run man-test
```
./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
```
report as following:
```
OK
```

**_WAY2:_**
step1: edit `./misc/man-test.py`
change `with open(man_file) as f:` to `with open(man_file, encoding='utf-8') as f:`

step2.  re-run man-test
```
./misc/man-test.py /tmp/ ctags man/ctags-lang-verilog.7.rst.in
```
report as following:
```
OK
```

**Comment by masatake** [comment]
I think we don't have to reproduce the issue on our local PCs. If the issue can be reproduced on the CI/CD environment, it is enough. @hirooih, make a pull request that causes a failure on the CI/CD environment. I can take over the pull request for fixing the failure.

**Comment by hirooih** [comment]
>>  python3 -c 'import locale; print(locale.getpreferredencoding());'
> if the result is not UTF-8, it can reproduce.

On Ubuntu-20.04;

```
$ python3.8 -c 'import locale; print(locale.getpreferredencoding());'
UTF-8
$ LANG=C python3.8 -c 'import locale; print(locale.getpreferredencoding());'
UTF-8
$ LANG= python3.8 -c 'import locale; print(locale.getpreferredencoding());'
UTF-8
$ LANG=POSIX python3.8 -c 'import locale; print(locale.getpreferredencoding());'
UTF-8
$
```

This explains why I could not reproduce the issue,

From the document of locale.getpreferredencoding() you referred, this means python3.8 on Ubuntu-20.04 does not reproduce the fail because the UTF-8 mode is enabled.

Our issue is that readline() fails on non-ASCII characters with LANG=POSIX.
The issue is fixed on python3.8 on Ubuntu-20.04.

> Fix the issue:
> 
> there exists two ways for fixing this issue.
> 
> WAY1:
> ...
> WAY2;
> ...

We have one more.

WAY3; 
Use python3 with the UTF-8 mode is enabled (for example python3.8 on Ubuntu-20.04).
(For example do not use centos-7.5.1804 for misc/man-test.py test.)

I think WAY2 (this PR) is worse than WAY1 or WAY3.

I had been asking you if you could reproduce the issue to find out the real issue.

> I think we don't have to reproduce the issue on our local PCs. If the issue can be reproduced on the CI/CD environment, it is enough.

I disagree with you. I hope I could explain my intension to you.

> @hirooih, make a pull request that causes a failure on the CI/CD environment

Sure.  I've opened #2905.
I hope you chose WAY1 or WAY3.

leleliu008:
@hirooih 

the feature of `-X utf8` is added in `Python3.7`, some old linux distributions only support `Python3.6`

docs say `PYTHONUTF8: if set to 1, enable the UTF-8 mode.`, but `Ubuntu-20.04` is not set `PYTHONUTF8`

may be Python have been adopted by the Ubuntu repo maintainer, I just guess and don't confirmed.

image

Actually, the default encoding can be guessed as long as the UTF-8 locale files exist even if user just set `LANG=POSIX`

As it can be seen from the above, the `WAY3` doesn't work in some cases, that's why I didn't motion it.

masatake:
Merge via #2905.

PR #2911, opened 2021-03-13
docs: as a highlighter, specify CtagsOtplibLexer instance instead of class

masatake:


masatake:
See https://docs.ctags.io/en/latest/man/ctags-optlib.7.html#examples. .ctags is highlighted with colors!

PR #2924, opened 2021-03-26
build-sys,mk_mvc.mak: don't use ALL_SRCS to define ALL_OBJS

masatake:


mk_mvc.mak
```diff
@@ -23,7 +23,7 @@ WIN32_OBJS = $(WIN32_SRCS:.c=.obj)
 PEG_OBJS = $(PEG_SRCS:.c=.obj)
 PACKCC_OBJS = $(PACKCC_SRCS:.c=.obj)
 RES_OBJ = win32/ctags.res
-ALL_OBJS = $(ALL_SRCS:.c=.obj) $(REGEX_OBJS) $(FNMATCH_OBJS) $(WIN32_OBJS) $(PEG_OBJS) $(RES_OBJ)
+ALL_OBJS = $(ALL_OBJS) $(REGEX_OBJS) $(FNMATCH_OBJS) $(WIN32_OBJS) $(PEG_OBJS) $(RES_OBJ)
```
k-takata, line 26: nmake doesn't expand a variable inside a replacement. So, `$(ALL_SRCS:.c=.$(OBJEXT))` is not expanded as expected.
This is the reason why `$(ALL_SRCS:.c=.obj)` is used here.

masatake: Oh,  I see. Thank you.

Issue #1823, opened 2018-08-12
Segfault if base language flags given while new subparser uses `{dedicated}` mode


hadrielk:
`ctags` segfaults given the following options.ctags and a `.cpp` file to parse:
```
--kinds-C++=*

--langdef=Test{base=C++}{dedicated}
--map-C++=-.cpp
--map-Test=+.cpp

--_tabledef-Test=main
--_mtable-regex-Test=main/.//
```

If the `--kinds-C++=*` is removed, it works fine. If it's replaced with `--fields-C++=+{properties}` it segfaults. So long as no `C++`-specific options are given, it works fine.

Alternatively, if `--langdef=Test{base=C++}{dedicated}` is changed to `--langdef=Test{base=C++}`, it does not segfault. (of course it doesn't invoke the C++ parser either, but that's expected in that case)

The crash seems to occur due to `getNextSubparser()` in `parse.c` invoking `t = getSubparserLanguage(r);`, and then `getSubparserLanguage()` in `dependency.c` just doing `return s->slaveParser->id;`. Dereferencing the `s->slaveParser` is invalid, as it's `NULL` in this case.

I added an `Assert(s->slaveParser);` in `getSubparserLanguage()` and it asserted:
```
ctags: main/dependency.c:194: getSubparserLanguage: Assertion `s->slaveParser' failed.
```
I don't know this code well enough to know what *should* happen, so can't fix it myself.

```
$ ./ctags --version
Universal Ctags 0.0.0(9ffe430a), Copyright (C) 2015 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Aug 12 2018, 11:15:35
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +debug, +option-directory, +case-insensitive-filenames
```

masatake:
I knew it. I am surprised at this report because I expect none can find this bug.

This is related to the order of the initializations of parsers.

Original I expected initializations are done in following steps.

1. Links between parsers including relations base parsers and subparsers are established. (linking stage)
2. initialize parsers 

In your case, My code expects `--langdef=Test{base=C++}{dedicated}` is  evaluated before initializing  C++ parser. However, `--kinds-C++=*` triggers initializing C++ parser. This breaks my (wrong) expectation.

[...]

PR #2926, opened 2021-03-26
docs: restruct multiple parsers (and others)

hirooih:
Please review 5ff4861 and 7f81e6e carefully.


docs/running-multi-parsers.rst
```diff
@@ -216,6 +218,16 @@ higher level language, a subparser for the higher level language is
...
+	============================== ===================== ======================
+	input file                     tags of *<BASE_LANG>* tags of *<SUB_LANG>*
+	============================== ===================== ======================
+	*<BASE_LANG>* (``input.base``) top down              bottom up
+	*<SUB_LANG>*  (``input.sub``)  bottom up             top down
+	============================== ===================== ======================
+
+.. TODO: Q to @masatake san: Is this correct?
+	Especially "bottom up" in the middle of the bottom line is not sure.
+
```
masatake, line 241: I will answer to this question. Please, wait for awhile.

hirooih: Please take your time.  Thank you.

masatake: I will review this in this week end.

hirooih: Thank you.  But please respond at your convenience.

I've been busy with other works.
I have noticed after seeing my PR today there is a sentence in the first line of section "Top down parser choice and bottom up parser choice";

> There are two ways to run a subparser: top down or bottom up parser choices.

This means terms "top down" and "bottom up" are ones for a subparser.

Therefore the following table I proposed in this PR;

```
============================== ===================== ======================
input file                     tags of *<BASE_LANG>* tags of *<SUB_LANG>*
============================== ===================== ======================
*<BASE_LANG>* (``input.base``) top down              bottom up
*<SUB_LANG>*  (``input.sub``)  bottom up?            top down
============================== ===================== ======================
```

should be

```
============================== =================
input file                     subparser choices
============================== =================
*<SUB_LANG>*  (``input.sub``)  top down
*<BASE_LANG>* (``input.base``) bottom up
============================== =================
```

And the section "Top down parser choice and bottom up parser choice", which includes the table above, describes the implementation in C. Therefore, I reconsidered that it is better to move it to "API for subparser" in internal.rst.

Am I correct?

masatake: I found the quality of implementation and *concept* of the "direction" is very low.
I must refine the concept itself.

```
============================== =================
input file                     subparser choices
============================== =================
*<SUB_LANG>*  (``input.sub``)  top down
*<BASE_LANG>* (``input.base``) bottom up
============================== =================
```

This looks correct.

> And the section "Top down parser choice and bottom up parser choice", which includes the table above, describes the implementation in C. Therefore, I reconsidered that it is better to move it to "API for subparser" in internal.rst.

Developers of optlib parsers should know this to use the direction flags. So I don't think the section should be move.

hirooih: > This looks correct.

Thanks.  I will update the commit.

> Developers of optlib parsers should know this to use the direction flags. So I don't think the section should be move.

Let me explain some more why I proposed moving the section.

- "Defining a subparser" section in optlib.rst is well-described without using words "top down" nor "bottom up". 
- I think a developer who needs subparser should use optlib, because is much more easy and simple.
- Only maintainers of C-API for subparser should know the concept of "top down" and "bottom up".

How do you think?

masatake: I see. Please, move the section.

A developer of an optlib parser should know "top down" and "bottom up" choosing. I wanted to show you an example. However, it doesn't work well because of a bug.

docs/optlib.rst
```diff
@@ -1425,14 +1412,13 @@ You can implement such roles by extending the built-in C parser:
 
 .. _guest-regex-flag:
 
-Running a guest parser with ``_guest`` regex flag
+Defining a guest parser with ``_guest`` regex flag
```
masatake, line 1409: How about `Scheduling running a parser as a guest parser` ?
`Defining` implies writing a new parser for me.

masatake:
I'm reviewing
> docs: updates for multiple parsers 7f81e6e

masatake:
I'll review 7f81e6e today.

masatake:
It seems my review comments are disassociated with the line changed.
But you can see the original line I commented by visiting https://github.com/universal-ctags/ctags/commit/7f81e6e78c6549c46fc75f4329a9163b9945ab3a .


I'm very sorry for taking much time for reviewing.

I made some minor comments to 7f81e6e78c6549c46fc75f4329a9163b9945ab3a.
Other than them, the changes look good to me.

The code  for direction I wrote is broken.  The evaluation for direction flag is done by the code initialing parsers. It is broken.
 In addition to https://github.com/universal-ctags/ctags/issues/1823, I must revise the direction flag.

masatake:
About multiple parsers, I would like you or @AmaiKinono to update ctags-client-tools(7) and/or ctags-incompatiblities(7) about following aspect:

Exuberant Ctags runs a parser for an input file. However, Universal ctags runs multiple parsers.
That means tags for an input file can have different language: fields like:

```
[yamato@dev64 ctags]$ cat /tmp/foo.html 
<html>
  <script>var x = 1</script>
  <h1>title</h1>
</html>
[yamato@dev64 ctags]$ ./ctags -o - --extras=+g /tmp/foo.html 
title	/tmp/foo.html	/^  <h1>title<\/h1>$/;"	h
x	/tmp/foo.html	/var x = 1/;"	v
[yamato@dev64 ctags]$ ./ctags -o - --extras=+g --fields=+l /tmp/foo.html 
title	/tmp/foo.html	/^  <h1>title<\/h1>$/;"	h	language:HTML
x	/tmp/foo.html	/var x = 1/;"	v	language:JavaScript
```

I guest some client tools may assume "tags for an input file have the same language: fields".
The behavior of u-ctags may break such clients.

I wonder whether we should call this is an incompatibly or not.

docs/optlib.rst
hirooih: I agree with you.  I've fixed to "Defining".

hirooih: I've fixed.

docs/running-multi-parsers.rst
hirooih: I've update the table as we agreed.

> I see. Please, move the section.
> 
> A developer of an optlib parser should know "top down" and "bottom up" choosing. I wanted to show you an example. However, it doesn't work well because of a bug.

I left the section as is, because it will be required when the bug is fixed.

hirooih:
> It seems my review comments are disassociated with the line changed.

No problem.

> Could you comment out this section?
> 
> The 'Q' is understandable.
> I found I must explain "placeholder tag". The ghost kind is useful for fill the find field of the placeholder tag. I will write about the Ghost kind when I write about the placeholder tag. I will write about the placeholder tag when I write about Optscript.

I've commented out the section.  I also left your comment.

> Good question.
> 
> Unlike subparser, guest/host relationships are very dynamic. So we cannot list guest parsers till running.

I see.
I left the following comments.

> .. NOTE: We don't have --list-guest-parser option as --list-subparser for
> 	a subparser, because a guest parser does not define a new parser and it just
> 	uses parsers which are already defined.

> About multiple parsers, I would like you or @AmaiKinono to update ctags-client-tools(7) and/or ctags-incompatiblities(7) about following aspect:

I'll take this.  I will make another PR.

I made fixes you suggested.  Thank you for your advices.
Let me know if you think that I can merge this PR.

masatake (approved):

hirooih:
Thank you for your review and approval.

PR #2928, opened 2021-03-26
Ada: adjust parsing mode at the end of generic

masatake:
Close #2925.

The original code reset the parsing mode to ROOT
at the end of generic package.

This setting causes a bug if the generic package is
not declared at the top level.

Issue #2723, opened 2020-11-23
Julia parser  TODO list


masatake:
See #2654 

- [ ] [tagging imported object] (https://github.com/universal-ctags/ctags/pull/2654#discussion_r499113865)
- [ ] [short function](https://github.com/universal-ctags/ctags/pull/2654#issuecomment-707817011)

@getzze is the original author.

masatake:
@getzze, feel free to edit this dashboard (if you need).

PR #2942, opened 2021-04-05
Ada : Improve formal parameter tags in generic function/procedure

koenmeersman:
Improving generic procedure/function handling.
When testing https://github.com/universal-ctags/ctags/pull/2928 
I noticed formal parameters of generic procedure/functions are not  tagged.

### DIFF
--- a/Units/parser-ada.r/ada-generic-in-package.d/args.ctags
+++ b/Units/parser-ada.r/ada-generic-in-package.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--kinds-Ada=+{formal}
--- a/Units/parser-ada.r/ada-generic-in-package.d/expected.tags-e
+++ b/Units/parser-ada.r/ada-generic-in-package.d/expected.tags-e
@@ -6,3 +6,39 @@ package My_Package isMy_Package/s2,48
     type Num is digits <>;Num5,81
   type Missing_Tag is recordMissing_Tag/t10,208
     Num : Integer;Num11,237
+
+input_1.adb,1666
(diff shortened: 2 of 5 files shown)


masatake:
@koenmeersman, thank you for receiving the invitation.
Please, let me explain some rules I used in maintaining ctags.

1. if possible, keep a commit smaller. So reviewing becomes easier.
2. use prefixes like "Ada: ", "main: ", "build-sys: ", etc., for the headers of commits. 
3. use the secondary prefies like "cosmetic", "style", "refactor", etc. for the headers of commits if the commits doesn't change the behavior of program.  See 698c4a3ad35c65ecf1ab66cd0c8c1e2fe135a308 as an example.
4. include the final results of your efforts for improving ctags to your pull request; try-and-error commits should not be recorded to the ctags history. See #2951.
5. don't include merge commit. You have two choices; don't merge "master" after making your topic branch or rebase on "master".
6. you can make a "TEMP" pull request for running test cases on CI. I do this frequently to understand Windows and Mac because I have only GNU/Linux at hand.
7. You can make and maintain the "Ada" specific meta issue like https://github.com/universal-ctags/ctags/issues/2723 .
8. Use cherry-pick instead of git merge when taking commits from my branch. So your pull request becomes simple.

Sorry for the many rules unexplained well.

I will review this pull request. Could you wait for a while?

masatake:
This also explains the rules. https://docs.ctags.io/en/latest/contributions.html#committing-and-submitting-a-pull-request

masatake:
Merged with some modifications. Thank you.

PR #2491, opened 2020-04-04
SystemVerilog: parse a multi dimensional array in typedef correctly

masatake:
Close #2489.

Signed-off-by: Masatake YAMATO <yamato@redhat.com>

PR #2947, opened 2021-04-07
build-sys: use upstream packcc (take 2)

masatake:
This is based on #2491.
Instead of updating packcc.c, this change updates the upstream of 'git subtree'.


masatake:
We cannot merge this till the Kotlin parser maintainer says "go".

misc/pull-packcc.sh
```diff
@@ -6,4 +6,4 @@
 # License: GPL 2 or later
 
 cd $(git rev-parse --show-toplevel)
-git subtree pull --prefix misc/packcc https://github.com/universal-ctags/packcc.git master --squash
+git subtree pull --prefix misc/packcc https://github.com/arithy/packcc.git master --squash
```
dolik-rce, line 9: Are you sure you want to pull in complete packcc repository? Most of it (test/, build/) won't be really useful for anything here?

Unfortunately, I think it is not possible to pull only some files using `git subtree`. On the other hand, you're probably not using `git subtree` with `push` or `merge` and you `--squash` the commits, so it's not really necessary to use it at all. You could simply pull the desired files using `curl`. It could look like this:
```sh
cd $(git rev-parse --show-toplevel)
SHA="$(curl 'https://api.github.com/repos/arithy/packcc/branches/master' | jq -r .commit.sha)"
for FILE in src/packcc.c LICENSE README.md; do
    mkdir -p "$(dirname "misc/packcc/$FILE")"
    curl "https://raw.githubusercontent.com/arithy/packcc/$SHA/$FILE" -o "misc/packcc/$FILE"
done
git commit -m "Updated packcc to $SHA" -- misc/packcc
```

peg/kotlin.peg
```diff
@@ -22,6 +22,10 @@
```
dolik-rce, line 22: I've made few more adjustments to the grammar. Could you please use the file from 
https://github.com/dolik-rce/ctags/blob/35ea29c889b345946b3cd11bf3593b82de2de4eb/peg/kotlin.peg?

dolik-rce (approved):
I have only two minor comments. The update looks good, even if you decide to not act on them.

misc/pull-packcc.sh
masatake: I think getting the whole source tree is o.k.
We can avoid installing jq and curl for building ctags source tree if we use just git subtree.
We can expect the pull-packcc script works even if the directory layout of packcc is changed.
Of course, ctags' building script like Makefile.am must be revised when the layout is changed. However, we can reduce the items we have to think about when updating.

peg/kotlin.peg
masatake: I put the new file to "build-sys: use upstream packcc" commit.
Is this o.k.?

We should separate the changes about kotlin.peg to another commit. However, when I removed the change about kotlin.peg from "build-sys: use upstream packcc" commit, I cannot build ctags. So we may have to put everything into the commit.

masatake:
I added 7334d8c to recover the mistake I took in c1e350d.

peg/kotlin.peg
masatake: packcc used in u-ctags was derived from enechaev/packcc.
I wonder the fixes in https://github.com/enechaev/packcc/commits/master are in the new upstream.

misc/pull-packcc.sh
dolik-rce: You don't need curl and jq to build ctags, only if you want to update packcc. Or is the script called automatically somewhere in the build/release process? I thought you're running it manually only when you want to update.

But as I said, I don't really mind having more files here, you can leave this as is.

peg/kotlin.peg
dolik-rce: > I put the new file to "build-sys: use upstream packcc" commit.
Is this o.k.?

I don't mind. If you wan't it in two separate commits and both of them passing tests, then you'd  have to first change the grammar and then add `%earlysource` later in the second commit.

> I wonder the fixes in https://github.com/enechaev/packcc/commits/master are in the new upstream.

If you mean those 6 changes you listed in arithy/packcc#25, then yes, those are present in the new upstream. I'm not sure about other changes, but I think it was mostly features added by enechaev. As far as I can tell, everything works just fine with current state of this branch (I just checked again now).

masatake: I made a separate commit for changing about kotlin.peg. Thank you.

masatake:
@dolik-rce, thank you very much.

k-takata:
@masatake Did you leave the switch-packcc-repo branch intentionally? Or can we delete the branch?
We also have some stale branches. Some of them might be deleted?
https://github.com/universal-ctags/ctags/branches/stale

We can also make our packcc repository archived.

masatake:
@k-takata, thank you. I deleted some of the obviously unnecessary branches. I converted universal-ctags/packcc repo to an archive.

PR #2955, opened 2021-04-12
docs(man): minor improvements for ctags(1)

masatake:
- Use `fileScope``/``F`` instead of ``F/fileScope``.
- Introduce   ``--map-<LANG>`` in addition to ``--langmap``.


hirooih:
Thanks.

PR #2956, opened 2021-04-14
build-sys,visual studio: Add x64 build support

ArcsinX:
This patch adds ability to build ctags using Visual Studio project from x64 native shell


k-takata:
I haven't tested this on my local environment, but the changes look good.
Let's merge this and see what's happen.

PR #2969, opened 2021-04-23
GitHub Actions(netbsd): add autoconf dependency in .github/workflows/testing-netbsd.yml

leleliu008:
It seems that the `NetBSD pkgin repository dependency tree` changed, `autoconf` is not installed along with `automake` now.


masatake:
LGTM. Could you add 'GitHub Actions (netbsd):' prefix for the commit header like: 
```
GitHub Actions(netbsd): add autoconf dependency in .github/workflows/testing-netbsd.yml 
```

masatake (approved):

PR #2971, opened 2021-04-24
SystemVerilog: Avoid using non-ASCII characters

### LINKED ISSUE #2970: warning C4819 and error C2220
Hello,

I have already git clone the newest version of universal-ctags. But the warning C4819 and error C2220 happen for some files as below. It can be solved by modify the file on localy but I suggest it can be solved from here and everyone can have it.
image

k-takata:
Fix #2970
See also #2710

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1672,7 +1672,7 @@ static int processType (tokenInfo* token, int c, verilogKind* kind, bool* with)
 }
 
 // class_type ::=
-//       ps_class_identifier [ # ( … ) ] { :: class_identifier [ # ( … ) ] }
+//       ps_class_identifier [ # ( ... ) ] { :: class_identifier [ # ( ... ) ] }
 // "interface_identifier ." is also handled
 static int skipClassType (tokenInfo* token, int c)
 {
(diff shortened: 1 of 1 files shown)


hirooih (approved):
Thanks.

I considered adding a check for non-ASCII characters.

https://stackoverflow.com/questions/3001177/how-do-i-grep-for-all-non-ascii-characters

I don't like either of them.

I thought there must be a way to fix the root cause, but ...

https://stackoverflow.com/questions/31028931/reset-default-file-encoding-to-utf-8-in-visual-studio-2013

It seems that Visual Studio 2013 does not support UTF-8 really.  But Visual Studio 2015 does.

k-takata:
I think VS2013 accepts UTF-8 characters if the file has UTF-8 BOM. (I don't like it, though.)
VS2015 has the `/utf-8` command-line option, but it causes an error on VS2013.

hirooih:
> I think VS2013 accepts UTF-8 characters if the file has UTF-8 BOM. 

This is described on the bottom of the link I referred above. But if I understand correctly, UTF-8 BOM is not C standard compliant. It is not a solution.

My questions is "Do we really have to support VS2013?".

Issue #2964, opened 2021-04-19
JSON output format


dciug:
Command:
```
$ ctags --fields=* --output-format=json -f ctags.json file.c
```

The output isn't valid json. It needs to be a JSON array.
So, the output needs enclosing "[ ]" and commas to separate the objects.

masatake:
You are correct. The format is jsonl (https://jsonlines.org/).
We must fix the document and help output.
However, changing the option to `--output-format=jsonl` may be a bad idea because the original option is already used.

hirooih:
Shall I send the following PR?

```diff
--- a/docs/output-json.rst
+++ b/docs/output-json.rst
@@ -4,7 +4,8 @@
 JSON output
 ======================================================================

-Experimental JSON output has been added. ``--output-format=json`` can be
+Experimental `JSON <https://www.json.org/>`_  (strictly speaking `JSON Lines
+<https://jsonlines.org/>`_) output has been added. ``--output-format=json`` can be
 used to enable it.

 .. code-block:: console
```

masatake:
@hirooih, yes, please. In addition, could you do the same thing to ctags(1)?

hirooih:
> In addition, could you do the same thing to ctags(1)?

Sure.

But it is a little bit annoying... Here is the description about JSON format in ctags(1).

```
``--output-format=(u-ctags|e-ctags|etags|xref|json)``
	Specify the output format. The default is ``u-ctags``.
	See tags(5) for ``u-ctags`` and ``e-ctags``.
	See ``-e`` for ``etags``, and ``-x`` for ``xref``.
	``json`` is experimental format, and available only if
	the ctags executable is built with ``libjansson``.

.. TODO: convert output-json.rst to ctags-json-output.1.rst (ctags-json-output(1)).
   and add a link to it here.
```

JSON is the only experimental format here. Other formats have no description here.  It is not smart to describe only about JSON here.

How about changing it as follows?

```
	...
	``json`` is `an experimental format <https://docs.ctags.io/en/latest/output-json.html>`_, and available only if
	...
```

masatake:
@hirooih, thank you.

Let's remove the "experimental" label from the json-output. It is one of the new features that have real users: 
https://github.com/google/zoekt/blob/d92b3b80e582e735b2459413ee7d9dbbf294d629/ctags/json.go#L46

The JSON output includes the version information of the file (stream) format. So I think we can add it to
the public interface of u-ctags.

```
[...]

PR #2979, opened 2021-04-30
docs: update the description of JSON output  (#2964)

hirooih:
- add link to JSON and JSON Lines.
- Now it is not experimental.


masatake (approved):
Thank you.

masatake:
I will inspect the reason of the failure on fedoradist at circleci .
Generally, we should stop our development when a failure is reported.
However, this change is obviously nothing to do with the faulire. So, please merge this at the timing as you want.

Issue #2976, opened 2021-04-26
About multiple parsers, I would like you or @AmaiKinono to update ctags-client-tools(7) and/or ctags-incompatiblities(7) about following aspect:


hirooih:
About multiple parsers, I would like you or @AmaiKinono to update ctags-client-tools(7) and/or ctags-incompatiblities(7) about following aspect:

Exuberant Ctags runs a parser for an input file. However, Universal ctags runs multiple parsers.
That means tags for an input file can have different language: fields like:

```
[yamato@dev64 ctags]$ cat /tmp/foo.html 
<html>
  <script>var x = 1</script>
  <h1>title</h1>
</html>
[yamato@dev64 ctags]$ ./ctags -o - --extras=+g /tmp/foo.html 
title	/tmp/foo.html	/^  <h1>title<\/h1>$/;"	h
x	/tmp/foo.html	/var x = 1/;"	v
[yamato@dev64 ctags]$ ./ctags -o - --extras=+g --fields=+l /tmp/foo.html 
title	/tmp/foo.html	/^  <h1>title<\/h1>$/;"	h	language:HTML
x	/tmp/foo.html	/var x = 1/;"	v	language:JavaScript
```

I guest some client tools may assume "tags for an input file have the same language: fields".
The behavior of u-ctags may break such clients.

I wonder whether we should call this is an incompatibly or not.

_Originally posted by @masatake in https://github.com/universal-ctags/ctags/issues/2926#issuecomment-826275491_

hirooih:
@masatake san,

I realized that I had missed an important thing after I opened this issue.

The multiple parsers are still experimental features.  So they are out of scope of man pages, ctags-client-tools(7) nor ctags-incompatibilities(7).

I can add descriptions on news.rst.

The question is "Should we describe all incompatibilities of experimental features in news.rst?".
If answer is "yes", I also have to add descriptions about option files (option-files.rst), new output formats (output-*.rst) and optlibs (optlib.rst).  They are not described in news.rst yet.

It is not hard works because they can be simple descriptions as one for [interactive mode](https://docs.ctags.io/en/latest/news.html#interactive-mode).

How do you think?

masatake:
> The multiple parsers are still experimental features. So they are out of scope of man pages, ctags-client-tools(7) nor ctags-incompatibilities(7).

As you wrote, using options for defining "Multiple parsers" for an input file in a user's .ctags.d is experimental features.
[...]

PR #2980, opened 2021-04-30
docs: ctags-client-tools(7): describe MULTIPLE-LANGUAGES FOR AN INPUT FILE (#2976)

hirooih:
There was a section for this already.


masatake (approved):
Thank you.

PR #2983, opened 2021-04-30
Tmain: add missing --options=NONE to run.sh for making the cases environment independent

masatake:
Signed-off-by: Masatake YAMATO <yamato@redhat.com>

PR #2986, opened 2021-04-30
circleci: target fedora 34 explicitly (x86_64) 

masatake:
Derived from  #2983.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ jobs:
   #
   # This is for the latest RELEASED Fedora.
   # Update the version of Fedora when new version is released.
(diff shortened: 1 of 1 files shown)


masatake:
Reproduced with podman + fedora:34 locally.
I wonder  this, https://github.com/actions/virtual-environments/issues/2658, is something to do with what I see.

masatake:
I used Fedora28 for the reproduction. Maybe the OS is too old. When I used Fedora33 with the same container image based on fedora34, it is not reproduced. So I was convinced that what @leleliu008 wrote in #2983 was correct.
I withdraw this.

PR #3000, opened 2021-05-04
IPythonCell: a new subparsre stacked on Python parser

### LINKED ISSUE #2978: Adding support for Python code cells
Cells are useful for structuring code and navigating through them in Vim using [Tagbar](https://github.com/preservim/tagbar) would be very useful. Tagbar and other outline plugins depend on `ctags` to get the tags they show. I was wandering if there is a way to enable cell tagging atm, and if not is it planned for a future release?

I originally posted this issue on Tagbar here it is for reference https://github.com/preservim/tagbar/issues/759

Cells can be defined in several ways, e.g. `# %%`:

```python
import numpy as np
from matplotlib import pyplot as plt
from scipy.io import wavfile
import os

# %% generate sound
f = 12000
fs = 44100
t = np.arange(0, 1, 1/fs)
sound = np.sin(2*np.pi*f * t)

# %% plot sound
plt.plot(t, sound)

# %% play sound
wavfile.write('sound.wav', fs, np.int16(sound * 2**15))
os.system('play sound.wav')
```

Atm this code only shows variables in the tagbar. 

image

masatake:
See https://github.com/hanschen/vim-ipython-cell.

Close #2978.


masatake:
man-test.py rejects 7935a2c. This is the time to rewrite the script as @hirooih suggested in https://github.com/universal-ctags/ctags/pull/2853 .

masatake:
I introduced a quick fix to man-test.py.

gerazov:
@masatake - it seems that the `c` tag was already used for classes in Python. Can you update the merge to use capitalized `C` for code cells instead?

masatake:
> @masatake - it seems that the c tag was already used for classes in Python. Can you update the merge to use capitalized C for code cells instead?

See https://github.com/universal-ctags/ctags/issues/2978#issuecomment-829451352 .
See also https://docs.ctags.io/en/latest/man/ctags-client-tools.7.html#multiple-languages-for-an-input-file.

Use the `language` field to distinguish whether a tag having c as its kind means `class` in the Python or` cell` in the IPythonCell.
`--fields=+l` or `--fields=+{language}'.

I will not update. Even if I will update as you wrote, your issue will occur again when `C` kind is introduced to the Python parser.

I recommend you to enable `--fields=+Kzl`.

The rest of the work is up to the developers of your client tool.

I must add this to ctags-faq(7).

gerazov:
You're right - arigato for the pointer! :bow:

PR #3013, opened 2021-05-07
test;misc/roundtrip: fix failed on FreeBSD-11.4

### LINKED ISSUE #3011: misc/roundtrip failed on NetBSD8
relavent code: [misc/roundtrip#L139-L140](https://github.com/universal-ctags/ctags/blob/master/misc/roundtrip#L139-L140)
```
t="${t%
_}"
```

here why split into 2 line? can it be change to `t="${t%_}"` ?

leleliu008:
close #3011

### DIFF
--- a/misc/roundtrip
+++ b/misc/roundtrip
@@ -135,9 +135,13 @@ for tags in $tagfiles; do
 	# remove it, plus the extra newline, using variable substitutions.
 	# Note: we use "printf '%s\n'" instead of "echo" because Dash's "echo"
 	# unconditionally expands some sequences, like "\t" and alike.
-	t="$(printf '%s\n' "$name" | expandEscapeSequences; printf _)"
-	t="${t%
-_}"
+    if [ "$name" = "\r" ] ; then
+        t="$(printf "\r")"
+    elif [ "$name" = "\n" ] ; then
+        t="$IFS"
+    else
+        t="$(printf '%s\n' "$name" | expandEscapeSequences)"
+    fi
 	if [ 1 -gt $("${READTAGS}" -t "$tags" - "$t" | wc -l) ]; then
 	    printf 'FAILED: "%s" -t "%s" - "%s"\n' "${READTAGS}" "$tags" "$t"
 	    printf '	The raw tag name was "%s"\n' "$name"


masatake:
As the original comment shows, this area is quite complicated.
Could you write much more commit logs?

leleliu008:
To be honest, I didn't understand the origin code and comments:
```
for tags in $tagfiles; do
    tagnames=$(sed -e 's/^\([^  ]*\)    .*/\1/' "$tags" | limit_entries)
    for name in $tagnames; do
    # Yes, there is a reason for this craziness.  We need to properly
    # handle embedded newlines (expanded from "\n"), including trailing
    # ones the shell would strip automatically.  To work around this, we
    # add a dummy character at the end to inhibit stripping, and then
    # remove it, plus the extra newline, using variable substitutions.
    # Note: we use "printf '%s\n'" instead of "echo" because Dash's "echo"
    # unconditionally expands some sequences, like "\t" and alike.
    t="$(printf '%s\n' "$name" | expandEscapeSequences; printf _)"
    t="${t%
_}"
    printf "t=%s\n" "$t"
    if [ 1 -gt $("${READTAGS}" -t "$tags" - "$t" | wc -l) ]; then
        printf 'FAILED: "%s" -t "%s" - "%s"\n' "${READTAGS}" "$tags" "$t"
        printf '    The raw tag name was "%s"\n' "$name"
        s=1
    fi
    done
done
```
I have tried many times and finally found this workaround.

misc/roundtrip
```diff
@@ -43,9 +43,9 @@ fi
 limit_entries()
 {
     if [ -n "${MAX_ENTRIES}" ]; then
-	head -${MAX_ENTRIES}
+	    head -${MAX_ENTRIES}
```
masatake, line 46: Cold you remove this change from the commit?
I believe FreeBSD is not so whtespace sensitive.

misc/roundtrip
```diff
@@ -43,9 +43,9 @@ fi
...
     else
-	cat
+	    cat
```
masatake, line 48: Cold you remove this change from the commit?

If you find/think the indentation style of this line really wrong, you can make a separate commit and add to this pull request.

misc/roundtrip
```diff
@@ -125,24 +125,22 @@ IFS='
...
-	t="${t%
-_}"
-	if [ 1 -gt $("${READTAGS}" -t "$tags" - "$t" | wc -l) ]; then
-	    printf 'FAILED: "%s" -t "%s" - "%s"\n' "${READTAGS}" "$tags" "$t"
-	    printf '	The raw tag name was "%s"\n' "$name"
-	    s=1
-	fi
+for tagfile in $tagfiles; do
+    for tagname in $(cut -f1 < "$tagfile" | limit_entries)
+    do
```
masatake, line 130: Instead of 
```
for ...
do
```
, could you keep
```
for ...; do
```

The area we are dealing with is quite complicated.  So let's minimize the diff.

misc/roundtrip
```diff
@@ -126,23 +126,20 @@ IFS='
...
+            t="$IFS"
+        else
+            t="$(printf '%s\n' "$name" | expandEscapeSequences)"
+        fi
+
+        if [ 1 -gt $("${READTAGS}" -t "$tags" - "$t" | wc -l) ]; then
+            printf 'FAILED: "%s" -t "%s" - "%s"\n' "${READTAGS}" "$tags" "$t"
+            printf '	The raw tag name was "%s"\n' "$name"
+            s=1
+        fi
```
leleliu008, line 142: L138-142: nothing to do except for indentation

masatake: I think it is possible that you can remove L138-142 from the change by adjusting indentation (or whitespaces).
I think you can use tagnames, the original variables.
Does FreeBSD require removing the comment?
Updating the comment for FreeBSD is welcome. However, removing the whole the comment is not needed, I think.

The minimized version may be:
```
diff --git a/misc/roundtrip b/misc/roundtrip
index 536da3add..04006776b 100644
--- a/misc/roundtrip
+++ b/misc/roundtrip
@@ -126,7 +126,7 @@ IFS='
 
 tagfiles=$(find "$UNITS" -name expected.tags)
 for tags in $tagfiles; do
-    tagnames=$(sed -e 's/^\([^	]*\)	.*/\1/' "$tags" | limit_entries)
+    tagnames=$(cut -f1 < "$tags" | limit_entries)
     for name in $tagnames; do
 	# Yes, there is a reason for this craziness.  We need to properly
 	# handle embedded newlines (expanded from "\n"), including trailing
@@ -135,9 +135,13 @@ for tags in $tagfiles; do
 	# remove it, plus the extra newline, using variable substitutions.
 	# Note: we use "printf '%s\n'" instead of "echo" because Dash's "echo"
 	# unconditionally expands some sequences, like "\t" and alike.
-	t="$(printf '%s\n' "$name" | expandEscapeSequences; printf _)"
-	t="${t%
-_}"
+	if [ "$name" = "\r" ] ; then
+	    t="$(printf "\r")"
+	elif [ "$name" = "\n" ] ; then
+	    t="$IFS"
+	else
+	    t="$(printf '%s\n' "$name" | expandEscapeSequences)"
+	fi
 	if [ 1 -gt $("${READTAGS}" -t "$tags" - "$t" | wc -l) ]; then
 	    printf 'FAILED: "%s" -t "%s" - "%s"\n' "${READTAGS}" "$tags" "$t"
 	    printf '	The raw tag name was "%s"\n' "$name"

```
Could you make my differ smaller?

If you think we should remove the comment or remove the variable, of course, it is o.k. Improving the code are welcome.
However, they should be done in another commit with another reason.

leleliu008: I don' known why should be done in another commit with another reason.

When I fixed this problem as you expected, the code will be very ugly and no any readability. 
ugly: no indentation and redundant variables
no any readability: the comments not match the code

In my idea, the code must be written for people to read, the readability is very important.

masatake: > I don' known why should be done in another commit with another reason.

Keeping a commit for single purpose reduces the load of reviewers.
In this case, I am the reviewer.

The original commit does more than one thing:

* improving readability,
* making the script work on FreeBSD

Updating g the comment for the change making the script work on FreeBSD is acceptable.
I don't talk about the number of lines.

I found a better explanation what I would like to say in 
https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#separate-your-changes .

I know "keeping a commit for single purpose" requires extra work to the author of the changes.

leleliu008: Thank you for explaining.

masatake:
Please, wait for a while. I cannot find enough time now.

misc/roundtrip
```diff
@@ -126,7 +126,7 @@ IFS='
 
 tagfiles=$(find "$UNITS" -name expected.tags)
 for tags in $tagfiles; do
-    tagnames=$(sed -e 's/^\([^	]*\)	.*/\1/' "$tags" | limit_entries)
+    tagnames=$(cut -f1 < "$tags" | limit_entries)
```
masatake, line 129: Doesn't the sed command line work on FreeBSD-11.4?
If it works, keep the commad line as is in this commit.
If you think using cut is better, you can make another commit for it (or another pull request).
As far as reading your commit log, you focus on t=... .

masatake: I guess this change works with the bellow changes. Am I correct?

masatake:
I expect your change adapts the script to FreeBSD without changing the script's original behavior (or semantics?).
I can find a case breaking the original behavior with your change:
```
[yamato@dev64]~/var/ctags% git diff
diff --git a/Units/roundtrip-escapes.d/expected.tags b/Units/roundtrip-escapes.d/expected.tags
index 927ba14c..03fc0479 100644
--- a/Units/roundtrip-escapes.d/expected.tags
+++ b/Units/roundtrip-escapes.d/expected.tags
@@ -1,3 +1,4 @@
+a\n    input   //
 _\\_   input   //
 _\a_   input   //
 _\b_   input   //
[yamato@dev64]~/var/ctags% git branch
  leleliu008-fix-issue-3011
* master
[yamato@dev64]~/var/ctags% git checkout leleliu008-fix-issue-3011 
M       Units/roundtrip-escapes.d/expected.tags
Switched to branch 'leleliu008-fix-issue-3011'
[yamato@dev64]~/var/ctags% make roundtrip
  RUN      roundtrip                     
FAILED: "./readtags" -t "/home/yamato/var/ctags/Units/roundtrip-escapes.d/expected.tags" - "a"
        The raw tag name was "a\n"
make: *** [Makefile:6374: roundtrip] Error 1
[yamato@dev64]~/var/ctags% git checkout master                   
M       Units/roundtrip-escapes.d/expected.tags
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
[yamato@dev64]~/var/ctags% make
REPOINFO   main/repoinfo.h
make  all-am
make[1]: Entering directory '/home/yamato/var/ctags'
REPOINFO   main/repoinfo.h
  CCLD     ctags
make[1]: Leaving directory '/home/yamato/var/ctags'
[yamato@dev64]~/var/ctags% make roundtrip     
  RUN      roundtrip
[yamato@dev64]~/var/ctags% 
```

I read the original comment. Though I'm bad at English, I can understand the comment.
I think 4 base pieces of knowledge are needed to understand the comment.

* The format of tags file, especially about the escape sequence introduced in Universal Ctags. The document of the format is at https://docs.ctags.io/en/latest/man/tags.5.html.
* readtags command, a command for reading tags file. The document of the command is at https://docs.ctags.io/en/latest/man/readtags.1.html.
* The roundtrip test harness itself. There is no documentation for it. However, the commit log illustrates the basic idea. See https://github.com/universal-ctags/libreadtags/commit/6f0c79d112c90e25f8009b7ca7f4a31b6c744d3f . The page includes the original version of the script. 
* Units/roundtrip-escapes.d/expected.tags file. When you run `make roundtrip`, readtags scans all the lines of all expected.tags.
  Most of all expeced.tags files are for testing ctags. However, roundtrip-escapes.d/expected.tags is special. It is for testing readtags.

After reading the above documents and files, you will know what you can/should do for the miserious `t="${t%
}`.  

Thank you for your efforts, and I'm sorry to be late giving the above input to you.

leleliu008:
@masatake Thank you  for your guidance. I don't wanna continue this PR, because this fix is very little thing, most people don't  run roundtrip test on FreeBSD-11.4, our CI not support FreeBSD-11.4 too. I wanna focus on other important things such as #3030

masatake:
> I don't wanna continue this PR.

Feel free to close this pull request. However, the change for replacing sed with cut may make sence.
Incremental small changes are always welcome.

PR #3012, opened 2021-05-07
SQL: introduce a heuristic to distinguish PL/SQL inquiry directives and PostgreSQL dollar quoted strings

masatake:
Partially close #3006.

    [1] explains the inquiry directive ($$name).
    
    Quoted from the issue reported by @bagl:
    
        The PL/SQL directive is a token of two dollar signs followed by alpha
        and possibly more alphanum or #_$ characters (regex:
        \$\$[A-Za-z][A-Za-z0-9$_#]*). This definition will probably collide
        with PostgreSQL dollar quoted strings[2] such as $$Dianne's horse$$ or
        $SomeTag$Dianne's horse$SomeTag$ implemented here.
    
    [1] https://docs.oracle.com/en/database/oracle/oracle-database/18/lnpls/plsql-language-fundamentals.html#GUID-E918087C-D5A8-4CEE-841B-5333DE6D4C15
    [2] https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING
    
    This change focus on predefined directives.
    
    Signed-off-by: Masatake YAMATO <yamato@redhat.com>

PR #3024, opened 2021-05-13
docs: as a highlighter, specify CtagsOtplibLexer class in sphinx-4

masatake:
Close #3012.

Tested-by: fpliu <leleliu008@gmail.com>


k-takata:
I think that the class name `CtagsOtplibLexer` should be `CtagsOptlibLexer`.

PR #3027, opened 2021-05-15
main: introduce --_makeTagEntryReflection-<LANG> option to filter tags

masatake:
See #3020.

The new option `--_makeTagEntryReflection-<LANG>` allows users to add optscript code which ctags runs when making a tag.
In the optscript code, you can modify the tag, make an extra tag derived from the tag, and/or make the tag not being printed.

Let's see examples.

input.md:
```
# TITLE LEVEL 1

## TITLE LEVEL 2

## ANOTHERTITLE LEVEL 2

# ANOTHER TITLE LEVEL 1
```

mymd.ctags
```postscript
--_extradef-Markdown=withfname,appending input filename
--extras-Markdown=+{withfname}
--_makeTagEntryReflection-Markdown={{
    /Markdown.withfname _extraenabled {
        . :extras {
            /Markdown.withfname _amember not
        } {
            true
        } ifelse
        {
            % Make the original tag invisible
            . _markplaceholder

            mark . :name (@) . :input _buildstring
            . :kind
            . _tagloc _tag dup /Markdown.withfname _markextra
            _commit pop
        } if
    } if
}}
```

the command session:
```
$  ./ctags --options=mymd.ctags  --fields=+'{extras}' -o - input.md
ANOTHER TITLE LEVEL 1@input.md	input.md	/^# ANOTHER TITLE LEVEL 1$/;"	c	extras:withfname
ANOTHERTITLE LEVEL 2@input.md	input.md	/^## ANOTHERTITLE LEVEL 2$/;"	s	extras:withfname
TITLE LEVEL 1@input.md	input.md	/^# TITLE LEVEL 1$/;"	c	extras:withfname
TITLE LEVEL 2@input.md	input.md	/^## TITLE LEVEL 2$/;"	s	extras:withfname
```

The input file name is appended to the names of original tags.
The script hides the original tags.
If you don't want to hide them, remove the line `. _markplaceholder` in mymd.ctags.

TODO:
- [x]  initForeignRefTagEntry used in rpmspec parser crashes with this pull request.
- [ ] performance evaluation and optimization. This pull request increases sub parser calls.
- [x] test cases 
- [ ] reconsider the name of the option
- [ ] introduce `--_makeTagEntryNotification-<LANG>` option and support it in optlib2c (?, should I do this when we really need it?).
- [ ] documentation for ctags.io

PR #3028, opened 2021-05-16
optscript: bug fix, new operators, refactoring

masatake:
This is a preparation pull request for mering #3027.

The last commit of #3027 includes a critical bug. So the running the test cases fails.
This pull requests includes all commits other than the last one.
All changes are useful even if we don't merge #3027.


masatake:
I fixed the critical bug in  #3027. So I withdraw this and I continue to improve #3027 for merging.

masatake:
In #3027, there are still areas that I have to improve before merging.
So I reopen this request for merging the acceptable subset of #3027.

PR #3031, opened 2021-05-18
Markdown:  skip frontmatter area

masatake:
Ideally, we should run a guest parser on the area.
This time, I make ctags just skip the areas.


masatake:
@leleliu008, do you have an idea to recover the broken image source of centos6?

leleliu008:
@masatake 

use my latest version of pre-build Python:

```
curl -LO https://github.com/leleliu008/git-prebuild/releases/download/2.30.2/git-2.30.2-x86_64-linux-glibc.tar.xz
tar xf git-2.30.2-x86_64-linux-glibc.tar.xz -C /opt
ln -sf /opt/git-2.30.2-x86_64-linux-glibc/bin/git /usr/bin/git
```
Note: installation dir must be `/opt/git-2.30.2-x86_64-linux-glibc`.

PR #3033, opened 2021-05-19
circleci (centos6*): install the pre-build Python

masatake:
https://repo.ius.io/ius-release-el6.rpm is removed.
We install the pre-build python instead.
@leleliu008 provides this solution.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -394,10 +394,10 @@ jobs:
        - run:
            name: Install tools for building
            command: |
-             yum install -y git gcc make autoconf automake pkgconfig
-             yum install -y https://repo.ius.io/ius-release-el6.rpm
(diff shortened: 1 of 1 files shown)


masatake:
```
  File "./misc/units.py", line 115
    print(msg, file=sys.stderr)
```

I wonder why.

leleliu008:
because every step run in different bash process. 

`export PATH=/opt/python-3.9.5-x86_64-linux-glibc/bin:$PATH`
change to
`echo 'export PATH=/opt/python-3.9.5-x86_64-linux-glibc/bin:$PATH' >> $BASH_ENV`

masatake:
@leleliu008, thank you!

PR #3035, opened 2021-05-22
Units: add test using GNU extension regex pattern, i.e. \s, \w

hirooih:
dd177c3 just copy regex tests.
fc71cc9 uses GNU extension regex patterns.

Some changes are not equivalent with original to try more use of extensions.


masatake:
If you rebase your changes on the latest HEAD of the master branch, the failures on centos6 will be fixed.

masatake:
These test cases can be used for testing p/pcre2 regex flags.

PR #3051, opened 2021-06-02
build-sys: remove AX_PROG_CC_FOR_BUILD, we don't need to use AX_PROG_CC_FOR_BUILD now. because the packcc is written in C89 now. we don't need to check C99 now.

leleliu008:


masatake:
@leleliu008, you edited the header of this pull request. However, what I wanted you to update was the header of the commit log of git:https://github.com/universal-ctags/ctags/pull/3051/commits/94f50a6fdb8e7cd3d072953241f90b7acb64b996 .

leleliu008:
@masatake Yes, I known, I will push a new commit, and the commit log and the title will be changed too.

masatake:
Oh, I see.

masatake (approved):

k-takata:
Does the cross-compilation section in autotools.rst need to be updated?

leleliu008:
the old variables keep to use, so `autotools.rst` do not need to be updated.

Issue #3034, opened 2021-05-21
write the specification of regex implemented 


hirooih:
I've been expecting to update regex engine as #1861. It was because I just wanted to use \s,  \S, \w, and \W which I was familiar with via Perl.
Today I found they were already supported the GNU regex which is being used by ctags.

I've been referring man page of e?grep(1), regex(7), and manual of the GNU libc.  But they don't describe about the feature.
I searched on the Web, https://www.gnu.org/software/grep/manual/grep.html#Regular-Expressions is the only document I could find.

@masatake -san, if you agree with me, I'd like to update examples in ctags-optlib.7.rst and optlib.rst to use them. It will improve readability of patterns. Some of them do not handle white-spaces correctly, i.e. `...private) +)?func (..`.

masatake:
Before starting this task, could you add a case testing  \s, \S, \w, and \W to Tmain or Units?
@leleliu008 has added many platforms where our test runs. We can know whether we can expect  \s, \S, \w, and \W are available on all the platforms or not.

hirooih:
> Before starting this task, could you add a case testing \s, \S, \w, and \W to Tmain or Units?

PR #3035.  Failed on Cygwin, MSYS2, and MacOS 10.15.
Interesting... Is is possible for them to use a different regex engine from `gnu_regex/` in the ctags distribution?

masatake:
If a platform has an implementation of regex library, ctags (build-system) may use it.
The code in `gnu_regex/` directory is for the platform where the library is not available.

hirooih:
> If a platform has an implementation of regex library, ctags (build-system) may use it.
> The code in gnu_regex/ directory is for the platform where the library is not available.

That makes sense and explains this situation.
Let me think some more.

hirooih:
@masatake -san,

> If a platform has an implementation of regex library, ctags (build-system) may use it.
> The code in gnu_regex/ directory is for the platform where the library is not available.

Why don't we use gnu_regex always?

If ctags can use a same regex engine on all platforms, we can ensure the compatibility of optlib parsers among the platforms.
And regex engines on some platform has less functionalities than the GNU regex.  `\s` and `\w` are very useful for optlib parsers as we see on c18057e.

I tired to use gnu_regex on my MacOS machine and confirmed it passed the new tests.

Here is my fix of `configure.ac*. I'm not familiar with Autoconf.  There must be better fixes.

```diff
--- a/configure.ac
+++ b/configure.ac
[...]

PR #3054, opened 2021-06-05
Import regex from gnulib

hirooih:
Currently ctags use a regex library of the system on which ctags is build.
As the result, there is no specification of the regex for `optlib`, and an `optlib` script cannot use full features of a regex library considering compatibility with other system using a regex library whose features are limited.
ctags distribution includes regex library from glibc, but it is too old (from glibc-2.10.1 released in 2009).
See also #3034.

This PR imported the latest gnulib regex module.


hirooih:
There are 3 types of errors remain.

### [run units target on MSYS2 / testing (MINGW64)](https://github.com/universal-ctags/ctags/pull/3054/checks?check_run_id=2755420092)

  - Autoconf, configure, and make runs without error.
  - Some of Units tests fails.
  - I have no idea how this can be happen.  Any hints are welcome.

### build_and_test : [fedora33_cross_aarch64](https://app.circleci.com/pipelines/github/universal-ctags/ctags/1420/workflows/7f20cb62-1514-46a4-809c-ba48753d1646/jobs/15741) 

configure finds pthread,

```
checking pthread.h usability... yes
checking pthread.h presence... yes
checking for pthread.h... yes
```

But ld cannot find the pthread functions.

```
  CCLD     readtags
/usr/lib/gcc/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: gnulib/libgnu.a(regex.o): in function `re_compile_internal':
/root/universal-ctags/gnulib/regcomp.c:759: undefined reference to `pthread_mutex_init'
/usr/lib/gcc/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: /root/universal-ctags/gnulib/regcomp.c:814: undefined reference to `pthread_mutex_destroy'
/usr/lib/gcc/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: gnulib/libgnu.a(regex.o): in function `re_search_stub':
/root/universal-ctags/gnulib/regexec.c:391: undefined reference to `pthread_mutex_lock'
...
```

Can I get config.log for this test?

### others

Others are configurations which does not use Autoconf.  This should be fixed easily.

masatake:
> Can I get config.log for this test?

How about putting `cat config.log` somewhere in the script running the test on the CI/CD environment?

leleliu008:
you should check `-lpthread` if needed in our build script.

masatake:
I will limit the input for the roundtrip harness. So we can ignore `run units target on MSYS2 / testing (MSYS) (pull_request) `.
On the other hand, `run units target on MSYS2 / testing (MINGW64) (pull_request)  ` reported "Segmentation fault"
I wonder why. The reports are not only for ctags executions but also readtags executions.

hirooih:
@masatake and @leleliu008, Thank you for your advices.

> How about putting cat config.log somewhere in the script running the test on the CI/CD environment?

This works and I could find the place where test for pthread library is done.

> you should check -lpthread if needed in our build script.

The configure script test with `-pthread` and set it `$LIBPTHREAD`.

FYI here is the comment in configure script.

```
      # If -pthread works, prefer it to -lpthread, since Ubuntu 14.04
      # needs -pthread for some reason.  See:
      # https://lists.gnu.org/r/bug-gnulib/2014-09/msg00023.html
```
fedora33_cross_aarch64 now works.

***

> Others are configurations which does not use Autoconf. This should be fixed easily.

This was more difficult than I thought. To compile gnulib without Autoconf we have to generate config.h and some other header files. It is not practical to debug this via CI/CD environment.
I gave up and for now I decided to use gnu_regex/* for configurations which does not use Autoconf as we do now.

***

> On the other hand, run units target on MSYS2 / testing (MINGW64) (pull_request) reported "Segmentation fault"
> I wonder why. The reports are not only for ctags executions but also readtags executions.

This is the only error left. 

I missed segmentation faults yesterday. How can we debug this?

masatake:
> I missed segmentation faults yesterday. How can we debug this?

See #3057.

hirooih:
>> I missed segmentation faults yesterday. How can we debug this?
 
> See #3057.

I understand that we cannot use a debugger interactively....

Failed on the following line;

```
lregex.c:1423 	errcode = regcomp (result, regexp, cflags);
```

I guess that one of arguments may be broken.

Why gdb was not invoked by `Tmain/getter-extras-field.d/run.sh`.

masatake:
I'm very surprised at we don' have to modify ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters.
The changes for .mak files are also small.

hirooih:
> I'm very surprised at we don' have to modify ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters.
> The changes for .mak files are also small.

That is because I cheated.

> > Others are configurations which does not use Autoconf. This should be fixed easily.
> 
> This was more difficult than I thought. To compile gnulib without Autoconf we have to generate config.h and some other header files. It is not practical to debug this via CI/CD environment.
> I gave up and for now I decided to use gnu_regex/* for configurations which does not use Autoconf as we do now.

masatake:
> That is because I cheated.
>> ...

Oh, I see.

mk_mvc.mak
```diff
@@ -9,15 +9,20 @@
 
 include source.mak
 
+OBJEXT = obj
+
+# to be replaced by gnulib/regex.{ch}
+REGEX_HEADS = gnu_regex/regex.h
+REGEX_SRCS = gnu_regex/regex.c
+REGEX_OBJS = $(REGEX_SRCS:.c=.$(OBJEXT))
```
k-takata, line 17: Using a variable in replacement doesn't work on MSVC.

mk_mvc.mak
```diff
@@ -9,15 +9,20 @@
...
+
 REGEX_DEFINES = -DHAVE_REGCOMP -D__USE_GNU -DHAVE_STDBOOL_H -Dstrcasecmp=stricmp
 
-OBJEXT = obj
 COMMON_DEFINES =
 DEFINES = -DWIN32 $(REGEX_DEFINES) -DHAVE_PACKCC $(COMMON_DEFINES) -DHAVE_REPOINFO_H -DREADTAGS_DSL
 INCLUDES = -I. -Imain -Ignu_regex -Ifnmatch -Iparsers -Ilibreadtags -Idsl
 OPT = /O2 /WX
 PACKCC = packcc.exe
-REGEX_OBJS = $(REGEX_SRCS:.c=.obj)
```
k-takata, line 20: So, this line cannot be removed.

hirooih: Thank you.

It passed in
https://ci.appveyor.com/project/universalctags/ctags/builds/39626407/job/kklumhnaft1n0i7u?fullLog=true.

This is the only environment I can use.
I've fix it.

hirooih:
@k-takata san,

> This was more difficult than I thought. To compile gnulib without Autoconf we have to generate config.h and some other header files. It is not practical to debug this via CI/CD environment.
> I gave up and for now I decided to use gnu_regex/* for configurations which does not use Autoconf as we do now.

For the configurations which does not use Autoconf, in other words which use `mk_msc.mak` and `mk_mingw.mak`, the difficult part is generating header files.

I think we can use header files you generated on #3069 with `mk_mingw.mak`, and we may use them with `mk_msc.mak`.
I'd like to try this.  Could you send me the following files you generated on #3069?

```
gnulib/inttypes.h
gnulib/langinfo.h
gnulib/limits.h
gnulib/locale.h
gnulib/stdlib.h
gnulib/unistd.h
gnulib/wchar.h
gnulib/wctype.h
gnulib/sys/types.h
```

k-takata:
[gnulib-generated-header.zip](https://github.com/universal-ctags/ctags/files/6673162/gnulib-generated-header.zip)
[config.h.zip](https://github.com/universal-ctags/ctags/files/6673185/config.h.zip)

These are the generated files.

hirooih:
Thanks.
I will try them this weekend.

hirooih:
> For the configurations which does not use Autoconf, in other words which use mk_msc.mak and mk_mingw.mak, the difficult part is generating header files.

I misunderstood.  Header files generated in the gnulib directory are architecture/OS independent. 
Only `config.h` depends on arhitecture/OS.
By fixing `config.h` I could build on Win32 env, too.

Now we are ready to migrate from old glibc based regex to the lastest gnulib based regex.

hirooih:
I've updated the commits of this PR.

- import the latest gnulib as of today
- commit `config.h` provided by @k-takata san as is , then update it.  This shows us which lines were changed for mingw and mvc.
- separate commits for mingw and mvc.
- add gnulib/README.md
- update commit logs

As I commented on #3034, this PR just gives us the up-to-date implementation of GNU regex and better compatibility checks.

masatake:
I will review this pul in this weekend. Please, wait for a while.

mk_mvc.mak
```diff
@@ -9,9 +9,14 @@
...
 
+OBJEXT = obj
+
+# to be replaced by gnulib/regex.{ch}
+REGEX_HEADS = gnu_regex/regex.h
+REGEX_SRCS = gnu_regex/regex.c
+
 REGEX_DEFINES = -DHAVE_REGCOMP -D__USE_GNU -DHAVE_STDBOOL_H -Dstrcasecmp=stricmp
 
-OBJEXT = obj
```
masatake, line 14: Do we need this change?

mk_mvc.mak
```diff
@@ -9,9 +9,14 @@
 
 include source.mak
 
+OBJEXT = obj
```
masatake, line 12: Do we need this change?

mk_mingw.mak
```diff
@@ -12,7 +19,6 @@ CC = gcc
 WINDRES = windres
 OPTLIB2C = ./misc/optlib2c
 PACKCC   = ./packcc.exe
-OBJEXT = o
```
masatake, line 15: Do we need this change?

mk_mingw.mak
```diff
@@ -2,6 +2,13 @@
 
 include source.mak
 
+OBJEXT = o
```
masatake, line 5: Do we need this change?

configure.ac
```diff
@@ -6,7 +6,7 @@
 
 #	Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.59])
+AC_PREREQ([2.64])
```
masatake, line 9: NOTE for myself: the version of autoconf shipped as part of RHEL6 is 2.63.
The time for thinking about preparing weekly tarball with generated configure comes.

Makefile.am
```diff
@@ -1,3 +1,7 @@
+ACLOCAL_AMFLAGS = -I m4
+#AM_CPPFLAGS = -I$(top_builddir)/gnulib -I$(top_srcdir)/gnulib
+DEFAULT_INCLUDES = -I$(top_srcdir)/gnulib
+SUBDIRS = gnulib
```
masatake, line 5: I don't think libgnu.a is our prodcuts though we will put libgnu.a in our source tree.
To represent libgnu.a is a separated product, I would like to introduce a variable like:

```
LIBGNU_LIBS = gnulib/libgnu.a $(LIBPTHREAD)
```

So <A>...

Makefile.am
```diff
@@ -66,23 +68,16 @@ readtags_CPPFLAGS = -I. -I$(srcdir) -I$(srcdir)/main -I$(srcdir)/libreadtags
 readtags_CFLAGS = $(COVERAGE_CFLAGS) $(WARNING_CFLAGS)
 dist_readtags_SOURCES = $(READTAGS_SRCS) $(READTAGS_HEADS)
 readtags_CPPFLAGS += -DREADTAGS_DSL -I$(srcdir)/dsl
+readtags_LDADD = gnulib/libgnu.a $(LIBPTHREAD)
```
masatake, line 71: ...*A* you can rewrite this line like:
```
readtags_LDADD = $(LIBGNU_LIBS)
```
and ... *A*

Makefile.am
```diff
@@ -177,6 +171,7 @@ dist_libctags_a_SOURCES = $(ALL_LIB_HEADS) $(ALL_LIB_SRCS)
 ctags_CPPFLAGS = $(libctags_a_CPPFLAGS)
 ctags_CFLAGS = $(libctags_a_CFLAGS)
 ctags_LDADD  = libctags.a
+ctags_LDADD += gnulib/libgnu.a $(LIBPTHREAD)
```
masatake, line 174: ...*A* you can rewrite this line like:
```
ctags_LDADD = $(LIBGNU_LIBS)
```

Makefile.am
```diff
@@ -199,6 +194,7 @@ endif
 mini_geany_CPPFLAGS = $(libctags_a_CPPFLAGS)
 mini_geany_CFLAGS = $(libctags_a_CFLAGS)
 mini_geany_LDADD  = libctags.a
+mini_geany_LDADD += gnulib/libgnu.a $(LIBPTHREAD)
```
masatake, line 197: ...*A* you can rewrite this line like:
```
mini_geany_LDADD = $(LIBGNU_LIBS)
```
and ... *A*

Makefile.am
```diff
@@ -211,6 +207,7 @@ optscript_CPPFLAGS = $(libctags_a_CPPFLAGS)
 optscript_CFLAGS = $(libctags_a_CFLAGS)
 optscript_LDADD  = libctags.a
 # TODO: remove these dependencies from optscript command.
+optscript_LDADD += gnulib/libgnu.a $(LIBPTHREAD)
```
masatake, line 210: ...*A* you can rewrite this line like:
```
optscript_LDADD = $(LIBGNU_LIBS)
```
.

Makefile.am
```diff
@@ -1,3 +1,7 @@
+ACLOCAL_AMFLAGS = -I m4
+#AM_CPPFLAGS = -I$(top_builddir)/gnulib -I$(top_srcdir)/gnulib
+DEFAULT_INCLUDES = -I$(top_srcdir)/gnulib
```
masatake, line 3: I don't like this change. I would like to keep the code representing "WE ARE USING GNULIB" explicitly.
How about using following code?
```
GNULIB_CPPFLAGS = I$(top_srcdir)/gnulib
```

So ... *B*

Makefile.am
masatake, line 71: ...B we add a line like:
```
readtags_CPPFLAGS += $(GNULIB_CPPFLAGS)
```

and ... *B*

Makefile.am
```diff
@@ -66,23 +68,16 @@ readtags_CPPFLAGS = -I. -I$(srcdir) -I$(srcdir)/main -I$(srcdir)/libreadtags
 readtags_CFLAGS = $(COVERAGE_CFLAGS) $(WARNING_CFLAGS)
 dist_readtags_SOURCES = $(READTAGS_SRCS) $(READTAGS_HEADS)
 readtags_CPPFLAGS += -DREADTAGS_DSL -I$(srcdir)/dsl
+readtags_LDADD = gnulib/libgnu.a $(LIBPTHREAD)
 dist_readtags_SOURCES += $(READTAGS_DSL_SRCS) $(READTAGS_DSL_HEADS)
-if !HAVE_REGCOMP
-readtags_CPPFLAGS += $(REGCOMP_CPPFLAGS)
```
masatake, line 71: ...B we add a line like:
```
readtags_CPPFLAGS += $(GNULIB_CPPFLAGS)
```

and ... *B*

Makefile.am
```diff
@@ -108,7 +103,6 @@ if HAVE_STRNLEN
 libctags_a_CPPFLAGS += -DUSE_SYSTEM_STRNLEN
```
masatake, line 103: ...B we add a line like:
```
libctags_a_CPPFLAGS += $(GNULIB_CPPFLAGS)
```

and ... *B*

Makefile.am
```diff
@@ -108,7 +103,6 @@ if HAVE_STRNLEN
...
 endif
 libctags_a_CPPFLAGS+= $(FNMATCH_CPPFLAGS)
-libctags_a_CPPFLAGS+= $(REGCOMP_CPPFLAGS)
```
masatake, line 111: ...B we add a line like:
```
readtags_CPPFLAGS += $(GNULIB_CPPFLAGS)
```

and ... *B*

Makefile.am
masatake: Using DEFAULT_INCLUDES reduces the lines of the build script, however, it breaks the rhythm (or pattern) of lines.

mk_mingw.mak
```diff
@@ -89,7 +86,6 @@ V_WINDRES   = $(V_WINDRES_$(V))
 V_WINDRES_0 = @echo [WINDRES] $@;
 V_WINDRES_1 =
 
-
```
masatake, line 86: Very very very minor thing. The line deletion may not be needed.

mk_mingw.mak
```diff
@@ -109,7 +105,7 @@ $(PACKCC_OBJS): $(PACKCC_SRCS)
 $(PACKCC): $(PACKCC_OBJS)
 	$(V_CC) $(CC_FOR_PACKCC) $(OPT) -o $@ $^
 
-ctags.exe: $(ALL_OBJS) $(ALL_HEADS) $(PEG_HEADS) $(PEG_EXTRA_HEADS) $(REGEX_HEADS) $(FNMATCH_HEADS) $(WIN32_HEADS)
+ctags.exe: $(ALL_OBJS) $(ALL_HEADS) $(PEG_HEADS) $(PEG_EXTRA_HEADS) $(GNULIB_HEADS) $(FNMATCH_HEADS) $(WIN32_HEADS) $(LIBGNU)
```
masatake, line 108: Just a question. Where is LIBGNU defined?

masatake:
I put comments. There is no critical one. Thank you.

When I tried the change on Fedora34. Though the libc of Fedora34 provides regcomp, libgnu.a is linked to ctags.
```
[yamato@dev64]~/var/ctags-github% nm -D /lib64/libc.so.6  | grep regcomp
nm -D /lib64/libc.so.6  | grep regcomp
00000000000e4600 W regcomp@@GLIBC_2.2.5
[yamato@dev64]~/var/ctags-github% nm ./ctags  | grep regcomp
nm ./ctags  | grep regcomp
00000000004e1b80 T rpl_regcomp
[yamato@dev64]~/var/ctags-github% nm -D ./ctags  | grep regcomp
```

Do you expect this?

mk_mvc.mak
hirooih: No. These `OBJEXT` related changes are not mandatory.

I think a Makefile variable should be defined before it is referred for readability, even if it is a _recursively expanded variable_ (cf. https://www.gnu.org/software/make/manual/make.html#Flavors).

So I moved the line before the following line I added;

```
GNULIB_OBJS = $(GNULIB_SRCS:.c=.$(OBJEXT))
```

(I did this on both mk_mingw.mak and mk_mvc.mak.  But `nmake` does not support `$(GNULIB_SRCS:.c=.$(OBJEXT))`. So there is no need for mk_mvc.mak.)

But I found `$(OBJEXT)` is also referred in `include source.mak`. I had to move the line `OBJEXT = obj` before the line `include source.mak`.

Which do you prefer?

1. do not change. (revert the change)
2. Move the line `OBJEXT = obj` before the line `include source.mak`.
3. Move the line `OBJEXT = obj` before the line `include source.mak` in a separate commit or PR.

Makefile.am
hirooih: I used `DEFAULT_INCLUDES` because I thought `-I$(top_srcdir)/gnulib` had to be defined earlier than other include paths.
It is because  `-I$(top_srcdir)/gnulib` may override system include files. But other include paths do not override system include files.

I will fix them as you suggested.

hirooih:
> When I tried the change on Fedora34. Though the libc of Fedora34 provides regcomp, libgnu.a is linked to ctags.
> ...
> Do you expect this?

I've checked on my environment (Ubuntu 20.04), and the regex check fails, too.

from: config.log

```
...
configure:16737: checking for working re_compile_pattern
configure:17014: gcc -o conftest -g -O2   conftest.c  >&5
configure:17014: $? = 0
configure:17014: ./conftest
configure:17014: $? = 64
configure: program exited with status 64
...
```

Here is the test in `configure` failed.

```c
            ...
            /* Matching with the compiled form of this regexp would provoke
               an assertion failure prior to glibc-2.28:
                 regexec.c:1375: pop_fail_stack: Assertion 'num >= 0' failed
               With glibc-2.28, compilation fails and reports the invalid
               back reference.  */
            re_set_syntax (RE_SYNTAX_POSIX_EGREP);
            memset (&regex, 0, sizeof regex);
            s = re_compile_pattern ("0|()0|\\1|0", 10, &regex);
            if (!s)
              result |= 64;     // !!! FAILED HERE (added by hirooih) !!!
            else
              {
                if (strcmp (s, "Invalid back reference"))
                  result |= 64;
                regfree (&regex);
              }
            ...
```

The comment says it expects glibc-2.29 or later.
glibc-2.29 was released on Feb. 1st, 2019.  https://sourceware.org/legacy-ml/libc-announce/2019/msg00000.html
It is pretty new.

mk_mvc.mak
masatake: > No. These OBJEXT related changes are not mandatory.
> 
> I think a Makefile variable should be defined before it is referred for readability, even if it is a recursively expanded variable (cf. https://www.gnu.org/software/make/manual/make.html#Flavors).
> 

The change for improving readability is really appreciated. However, please do it in a separate commit.
So the change I have to review becomes smaller.
See  also
https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#separate-your-changes .

> 1. ...
> 2. ...
> 3. ...

I have no preference. Please, do as you think better. What I want is only the separation.

masatake:
The version of glibc on Fedora34 is
```
[yamato@dev64]~/var/ctags-github% rpm -q glibc
glibc-2.33-18.fc34.x86_64
```
So it is not old. I wonder why libgnu.a is linked. I will inspect the reason.

BTW I think it is better to add "gnulib_regex" or something to the output of `--list-features` output.

hirooih:
> So it is not old. I wonder why libgnu.a is linked. I will inspect the reason.

Does the same regex test fail on Fedora34?

> BTW I think it is better to add "gnulib_regex" or something to the output of --list-features output.

I need a condition.  By readling `config.h` generated;

```
#ifdef _REGEX_INCLUDE_LIMITS_H
```

or

```
#ifdef regexec
```

may work, but I'm not sure.

mk_mvc.mak
hirooih: I move the changes for OBJEXT in a separate commit, "build: move OBJEXT above `include source.mak'".

Makefile.am
hirooih: I've fixed, but I chose a name `GNULIB_LIBS`.

Makefile.am
hirooih: I've fixed.

mk_mingw.mak
hirooih: I've reverted the change.

masatake:
I put the log I got when building the latest code of the import-regex-from-gnulib branch o Fedora34.

[LOG.txt](https://github.com/universal-ctags/ctags/files/6809137/LOG.txt)

Makefile.am
```diff
@@ -65,24 +68,18 @@ bin_PROGRAMS+= readtags
...
 dist_readtags_SOURCES = $(READTAGS_SRCS) $(READTAGS_HEADS)
+readtags_CPPFLAGS += $(GNULIB_CPPFLAGS)
 readtags_CPPFLAGS += -DREADTAGS_DSL -I$(srcdir)/dsl
+readtags_LDADD = $(GNULIB_LIBS)
 dist_readtags_SOURCES += $(READTAGS_DSL_SRCS) $(READTAGS_DSL_HEADS)
-if !HAVE_REGCOMP
-readtags_CPPFLAGS += $(REGCOMP_CPPFLAGS)
-dist_readtags_SOURCES += $(REGEX_SRCS) $(REGEX_HEADS)
-endif
+
```
masatake, line 75: Very very very minor item. The empty line is no needed.

masatake:
What kind of output will configure show when it decides regex in gnulib should be used?

hirooih:
> BTW I think it is better to add "gnulib_regex" or something to the output of --list-features output.

It was easier than I thought. I've added this feature.
I chose `#ifdef _REGEX_INCLUDE_LIMITS_H` by finding [this article](https://lists.gnu.org/archive/html/bug-gnulib/2011-07/msg00435.html).
I tested this change by forcing `result` to 0 in configure.

****

The only issue remains is a regex test fails with glibc-2.31 and glibc-2.33.

My env is using glibc-2.31.

```
$ /lib/x86_64-linux-gnu/libc.so.6
GNU C Library (Ubuntu GLIBC 2.31-0ubuntu9.2) stable release version 2.31.
...
```

The test was added by the following commit of gnulib.

> commit 55a4abd92a0a8fa0a9d9aff3892505f7b0c6d73c
> Author: Jim Meyering <meyering@fb.com>
> Date:   Sat Dec 15 15:24:21 2018 -0800
> 
>     regex: work around a bug in glibc-2.27 and prior
> 
>     * m4/regex.m4 (gl_REGEX): Reject any system regexp that gets a failed
>     assertion for /0|()0|\1|0/.
>     * tests/test-regex.c (main): Add the same test here.
> ...

What I can think of is asking to Gnulib ML.

masatake:
Now I understand what you wrote.

masatake:
> The only issue remains is a regex test fails with glibc-2.31 and glibc-2.33.

YES.

hirooih:
I've sent a bug report to bug-gnulib@gnu.org.

@k-takata san,
Could you review especially the fixes for Win32 environments?

hirooih:
>> The only issue remains is a regex test fails with glibc-2.31 and glibc-2.33.

> I've sent a bug report to bug-gnulib@gnu.org.

https://lists.gnu.org/archive/html/bug-gnulib/2021-07/msg00025.html

The fix was committed to the Gnulib git repository.  I pulled and merge it to this PR.

https://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=commit;h=e707dbe7c6da9dd8300cb3d60141f144a7a5d5b1

> However, glibc 2.33 (the current version) still fails to pass the configure-time test, due to glibc bug 11053 which is fixed in Gnulib; see:
> 
> https://sourceware.org/bugzilla/show_bug.cgi?id=11053

This is OK.

But the test fails causes a segmentation fault during `configure`.  It seems that a `conftest` which should be removed automatically remains, it causes fail on fedora33_distcheck.

How can I fix this...?

hirooih:
`make distcheck` failed on my env (Ubuntu-20.04, Windows 10 WSL2).  They are fixed by #3093 and #3096, and now `make distcheck` on my env.

I cannot not reproduce the following error occurred during `fedora33_distcheck` test.

```
...
ERROR: files left in build directory after distclean:
./core.6821.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest
make[1]: *** [Makefile:6151: distcleancheck] Error 1
make[1]: Leaving directory '/root/universal-ctags/universal-ctags-5.9.0/_build/sub'
make: *** [Makefile:6080: distcheck] Error 1
```

hirooih:
During debugging the fails on `make tmain` and `make distcheck` I found I had to add a make rule for `libgnu.a`.  I added the following fix on `Makefile.am`.

```diff
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,8 +1,11 @@
 ACLOCAL_AMFLAGS = -I m4
 #AM_CPPFLAGS = -I$(top_builddir)/gnulib -I$(top_srcdir)/gnulib
-GNULIB_CPPFLAGS = -I$(top_srcdir)/gnulib
-GNULIB_LIBS = gnulib/libgnu.a $(LIBPTHREAD)
-SUBDIRS = gnulib
+GNULIB_DIR = gnulib
+GNULIB_CPPFLAGS = -I$(top_builddir)/$(GNULIB_DIR) -I$(top_srcdir)/$(GNULIB_DIR)
+GNULIB_LIBS = $(GNULIB_DIR)/libgnu.a $(LIBPTHREAD)
+$(GNULIB_DIR)/libgnu.a:
+	$(MAKE) -C $(GNULIB_DIR) libgnu.a
+SUBDIRS = $(GNULIB_DIR)

 include $(top_srcdir)/source.mak
```

masatake:
> I cannot not reproduce the following error occurred during fedora33_distcheck test.

I will inspect this after you merge #3093 and #3096.

hirooih:
> I will inspect this after you merge #3093 and #3096.

Thank you for your approvals.  I've merged them.

masatake:
@hirooih, could you rebase this pull request on the latest version of the "master" branch?
I wonder how the fedora33_distcheck says.

hirooih:
> @hirooih, could you rebase this pull request on the latest version of the "master" branch?
> I wonder how the fedora33_distcheck says.

I did and have the same error at the end of `make distcheck`.

```
...
ERROR: files left in build directory after distclean:
./core.6844.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest
make[1]: *** [Makefile:6178: distcleancheck] Error 1
make[1]: Leaving directory '/root/universal-ctags/universal-ctags-5.9.0/_build/sub'
make: *** [Makefile:6107: distcheck] Error 1
```

The `configure` generated on my env has the following lines;

```
...
_ACEOF
if ac_fn_c_try_run "$LINENO"; then :
  gl_cv_func_re_compile_pattern_working=yes
else
  gl_cv_func_re_compile_pattern_working=no
fi
rm -f core *.core core.conftest.* gmon.out bb.out conftest$ac_exeext \
  conftest.$ac_objext conftest.beam conftest.$ac_ext
fi
```

`conftest` should be removed the `rm` command line.

masatake:
I expected `./core.6844.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest` is a core file.
So limiting linux kernel to create a core file should solve the issue...

masatake:
Something wrong in your branch management.

I wrote:
> @hirooih, could you rebase this pull request on the latest version of the "master" branch?
> I wonder how the fedora33_distcheck says.

Did you rebase your branch on the *LATEST* version of the "master" branch?

As far as seeing https://github.com/hirooih/ctags/commits/import-regex-from-gnulib, you rebaed your branch on a bit older commit.

You may have to do:
```
$ git checkout master
$ git pull upstream master:master
$ git checkout import-regex-from-gnulib
$ git rebase master
```

I guess you didn't `git pull upstream master:master`, the step for updating local "master".

The latest master is a branch having a commit https://github.com/universal-ctags/ctags/commit/0a3699def2c633dbb9a46a3ef49c9e4f50653cf6 . This commit limits kernel to generate the core file.

hirooih:
> I thought you wanted to the effect of #3093 and #3096.  I did not know you added commit after them.

I've pull the latest and pushed again.

> I expected ./core.6844.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest is a core file.
> So limiting linux kernel to create a core file should solve the issue...

I don't expect the fix, because conftest is an execution binary and a core file also removed the following line;

```
rm -f core *.core core.conftest.* gmon.out bb.out conftest$ac_exeext \
  conftest.$ac_objext conftest.beam conftest.$ac_ext
```

We will see the result.

BTW can you reproduce the fail on your local machine?

hirooih:
> We will see the result.

It passed!!!
I don't undertand why...

masatake:
The core pattern doesn't match the name of the core file actually generated:
`./core.6844.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest`.

The pattern should be 'core.*conftest'. 
See core(5) man page:
```

   Naming of core dump files
       By default, a core dump file is named core, but the /proc/sys/kernel/core_pattern file (since Linux 2.6 and 2.4.21) can be set to define a template
       that is used to name core dump files.  The template can contain % specifiers which are substituted by the following values when a core file is cre‐
       ated:

...
           %E  Pathname of executable, with slashes ('/') replaced by exclamation marks ('!') (since Linux 3.0).
```

masatake (approved):

hirooih:
@masatake san,

Now I understand that `./core.6844.!root!universal-ctags!universal-ctags-5.9.0!_build!sub!conftest` is the name of core file.
Thank you for your approval.

@k-takata san, could you review 

- gnulib: make use of gnulib regex module for mingw (a5d10cd)

and

- gnulib: make use of gnulib regex module for mvc (0342ff4)

masatake:
@hirooih, let's merge this.

hirooih:
> @hirooih, let's merge this.

Sure.

@masatake san, @k-takata san,
Thank you for your valuable advices.

masatake:
The test is passed on RHEL6. I wonder why.

PR #3062, opened 2021-06-10
C++: fix scope kind of tparam in a prototype

masatake:
Close #3060.

input.h:

	template <class T> int f (T t);

The scope kind for T was function.
It should be prototype.

Issue #3043, opened 2021-06-01
Can I strip the C parser out?


IngwiePhoenix:
Hello!

After experimenting with this for a while, I think I have found what I was looking for... phew, finally.

However, I only need to parse C headers - and C headers only - to retrive the definitions from it.

Is it possible to strip the C parser out to repurpose it for this? Here is what I want to do, basically:

```c
// input in C:
typedef struct MyStruct {
  int aThing;
} MyStruct_t;
MyStruct_t add(MyStruct_t, MyStruct_t);
```
```v
// Output in V:
[typedef] // Needed, so V knows that it's `MyStruct_t` instead of `struct MyStruct_t`
struct C.MyStruct_t {
  aThing int
}
[unsafe] // This tells V that this call "might" be unsafe.
fn C.add(C.MyStruct_t, C.MyStruct_t) C.MyStruct_t
```

So, I just want to save myself from copytyping a gazillion headers just to achieve something small. Hence, I would love to take definitions from a C header and convert them.

Would it be possible to strip the C parser out, and re-use it for this?

Thank you in advance!

masatake:
C: report parameters in a prototype with anonymous tags

input.h
```
MyStruct_t add(MyStruct_t, MyStruct_t);
```
expected.tags
```
anon1 ... typeref:typename:MyStruct_t ...
anon2 ... typeref:typename:MyStruct_t ...
```
This feature helps using tags for code generation as reported in #3043.

masatake:
>  Can I strip the C parser out?

I don't understand the sentence. However, I guess what you want to use ctags as part of a c-heaer-to-vlang translator.

I hope ctags and readtags are suitable for this kind of hacking. I wrote a shell script. Surprisingly, I found a critical limitation in C parser. See #3045. Signature handling requires sed. I would like to remove this limitation in the future.  

c2v.sh:
```bash
#!/bin/bash

CTAGS=u-ctags
READTAGS=readtags
TAGS=$$.tags

# TAB CHARACTER MUST BE HERE.
# TAB CHARACTER MUST BE HERE.
TAB='	' # TAB CHARACTER MUST BE HERE.
# TAB CHARACTER MUST BE HERE.
# TAB CHARACTER MUST BE HERE.

emit_member()
{
	printf "  %s %s\n" "$1" "$2"
}

handle_member()
{
	local name=$1
	local rest=$2
	local tmp
	local typename

	case "$rest" in
		(*typeref:typename:*)
[...]

Issue #2471, opened 2020-03-25
Some problems of readtags options


AmaiKinono:
## About `-l`

The usage given by `readtags -h` is :

```
readtags [-ilp] [-n] [-Q EXP] [-s[0|1]] [-t file] [-] [name(s)]
```

But if I do:

``` console
$ readtags -l -Q '(eq? "citre" $name)' -t ./tags
```

then all the tags are listed. However, if I follow the examples given in the [doc](https://docs.ctags.io/en/latest/news.html#examples-of-filter-expressions), and put `-l` at last:

``` console
$ readtags -Q '(eq? "citre" $name)' -t ./tags -l
```

It gives the wanted lines.

Also, `readtags -h` says `-l` list all tags, and `-Q EXP` filters the result, so my understanding is we put `-l` first to list the tags, then followed by `-Q` to filter it. Why this doesn't work?

## About `-e` and `-n`

If I do:

``` console
$ readtags -Q '(prefix? "Concepts" $name)' -t ./tags -le
```

the output is:

```
Concepts        README.md       /^# Concepts$/
```

But with:

```
$ readtags -Q '(prefix? "Concepts" $name)' -t ./tags -el
```

I get the extension fields:

```
Concepts        README.md       /^# Concepts$/;"        kind:chapter
```

Similar things happen to `-n`. My understanding is they are like "switches" and affect `-l` that comes later. But I feel the help message is not clear about this.

## About `-i`

`-i` "performs case-insensitive matching", but it only works when name is given directly, like

``` console
$ readtags -i ./tags Citre
```

This is not enough. It should make `eq?`, `prefix?`, `suffix?` and `substr?`case insensitive, or at least (maybe better) we should have case-insensitive versions of these functions. It's needed for client tools.

## About other arguments

- `-p` "performs partial matching". But from my experiment I think it should be "prefix matching".

masatake:
What we have to do are

- revise  the code,
- improve -h output,
- add readtags(1) man page, and
- add more test cases to Tmain.

When improving readtags, I wonder how seriously we have to think about compatibility.
On Fedora, ctags package doesn't include readtags command.
In Exuberant-ctags, install target of Makefile installs only ctags command and its man page.
So I guess readtags command is not so popular.

masatake:
I didn't check the actual behaviour of readtags. I just revise the help message:
```
Find tag file entries matching specified names.

Usage: 
    ./readtags -h
        Print this help message.
[...]

PR #3063, opened 2021-06-10
Cxx: extract parameters in a prototype

### LINKED ISSUE #3045: C: report parameters in a prototype with anonymous tags
C: report parameters in a prototype with anonymous tags

input.h
```
MyStruct_t add(MyStruct_t, MyStruct_t);
```
expected.tags
```
anon1 ... typeref:typename:MyStruct_t ...scope:protoype:add...
anon2 ... typeref:typename:MyStruct_t ...scope:protoype:add...
```
This feature helps using tags for code generation as reported in #3043.

_Originally posted by @masatake in https://github.com/universal-ctags/ctags/issues/3043#issuecomment-852057589_

masatake:
Close #3045.

Parameters in functions were tagged.
This change extends it.
With this change, parameters in prototypes are tagged.

It is usual that a parameter in a prototype has no name. In such a case, this change generates an anonymous name.

### DIFF
--- a/Tmain/list-kinds-full.d/stdout-expected.txt
+++ b/Tmain/list-kinds-full.d/stdout-expected.txt
@@ -14,7 +14,7 @@ t	typedef	yes	no	0	C	typedefs
 u	union	yes	no	0	C	union names
 v	variable	yes	no	0	C	variable definitions
 x	externvar	no	no	0	C	external and forward variable declarations
-z	parameter	no	no	0	C	function parameters inside function definitions
+z	parameter	no	no	0	C	function parameters inside function or prototype definitions
 #LETTER	NAME	ENABLED	REFONLY	NROLES	MASTER	DESCRIPTION
 A	alias	no	no	0	NONE	namespace aliases
 D	macroparam	no	no	0	C	parameters inside macro definitions
@@ -37,4 +37,4 @@ t	typedef	yes	no	0	C	typedefs
 u	union	yes	no	0	C	union names
 v	variable	yes	no	0	C	variable definitions
 x	externvar	no	no	0	C	external and forward variable declarations
-z	parameter	no	no	0	C	function parameters inside function definitions
+z	parameter	no	no	0	C	function parameters inside function or prototype definitions
--- a/Tmain/nested-subparsers.d/stdout-expected.txt
+++ b/Tmain/nested-subparsers.d/stdout-expected.txt
@@ -14,7 +14,7 @@ t  typedefs
 u  union names
 v  variable definitions
 x  external and forward variable declarations [off]
-z  function parameters inside function definitions [off]
+z  function parameters inside function or prototype definitions [off]
 L  goto labels [off]
 D  parameters inside macro definitions [off]
 #
@@ -49,7 +49,7 @@ t       typedef    yes     no      0      C      typedefs
 u       union      yes     no      0      C      union names
 v       variable   yes     no      0      C      variable definitions
 x       externvar  no      no      0      C      external and forward variable declarations
-z       parameter  no      no      0      C      function parameters inside function definitions
+z       parameter  no      no      0      C      function parameters inside function or prototype definitions
 #
 # list kinds-full Event
 #
--- a/Units/parser-cxx.r/attribute.cpp.d/expected.tags
+++ b/Units/parser-cxx.r/attribute.cpp.d/expected.tags
@@ -18,6 +18,7 @@ e1e1	input.cpp	/^	e1e1 __attribute__((deprecated)),$/;"	e	enum:e1	file:
 e1e2	input.cpp	/^	e1e2$/;"	e	enum:e1	file:
 p1	input.cpp	/^__attribute__((noreturn)) void p1(void),$/;"	p	typeref:typename:void	file:	signature:(void)
 p2	input.cpp	/^	__attribute__((format(printf, 1, 2))) p2 (const char *, ...),$/;"	p	typeref:typename:void	file:	signature:(const char *,...)
+__anon6f96ad27010d	input.cpp	/^	__attribute__((format(printf, 1, 2))) p2 (const char *, ...),$/;"	z	prototype:p2	typeref:typename:const char *	file:
 p3	input.cpp	/^	p3 (void);$/;"	p	typeref:typename:void	file:	signature:(void)
 p4	input.cpp	/^void (__attribute__((noreturn)) ****p4) (void) __attribute__((deprecated));$/;"	v	typeref:typename:void (****)(void)	properties:deprecated
 p5	input.cpp	/^void p5() __attribute__ ((weak, alias ("p4")));$/;"	p	typeref:typename:void	file:	signature:()
--- a/Units/parser-cxx.r/complex-macros.d/args.ctags
+++ b/Units/parser-cxx.r/complex-macros.d/args.ctags
@@ -11,6 +11,6 @@
 -D STRINGIFY(token)=#token
 -D IMPLEMENT_FUNCTIONS(_prefix)=void _prefix ## a(){ }; void _prefix ## b(){ };
 -D DECLARE_VARS(_prefix)=int _prefix ## a; int _prefix ## b;
--D DECLARE_FUNCTION_4(Ret,Class,Method,...)=Ret Class##__##Method(Class *this, ##__VA_ARGS__)
+-D DECLARE_FUNCTION_4(Ret,Class,Method,...)=Ret Class##__##Method(Class *that, ##__VA_ARGS__)
 -D DECLARE_FUNCTION_4_BEGIN(...)={
 -D DECLARE_FUNCTION_4_END(...)=}
--- a/Units/parser-cxx.r/complex-macros.d/expected.tags
+++ b/Units/parser-cxx.r/complex-macros.d/expected.tags
@@ -11,6 +11,8 @@ _ret	input.cpp	/^#define DECLARE_FUNCTION_3(_ret,_name,...) _ret _name(__VA_ARGS
 _name	input.cpp	/^#define DECLARE_FUNCTION_3(_ret,_name,...) _ret _name(__VA_ARGS__);$/;"	D	macro:DECLARE_FUNCTION_3
 DECLARE_FUNCTION_3	input.cpp	/^#define DECLARE_FUNCTION_3(/;"	d	file:	signature:(_ret,_name,...)
 p3	input.cpp	/^DECLARE_FUNCTION_3(int,p3,int a,int b);$/;"	p	typeref:typename:int	file:	signature:(int a,int b)
+a	input.cpp	/^DECLARE_FUNCTION_3(int,p3,int a,int b);$/;"	z	prototype:p3	typeref:typename:int	file:
+b	input.cpp	/^DECLARE_FUNCTION_3(int,p3,int a,int b);$/;"	z	prototype:p3	typeref:typename:int	file:
 DEPRECATED	input.cpp	/^#define DEPRECATED(/;"	d	file:	signature:(...)
 p4	input.cpp	/^DEPRECATED(int p4());$/;"	p	typeref:typename:int	file:	signature:()	properties:deprecated
 _prefix1	input.cpp	/^#define DECLARE_TWO_VERSIONS_OF_FUNCTIONS(_prefix1,_prefix2) \\$/;"	D	macro:DECLARE_TWO_VERSIONS_OF_FUNCTIONS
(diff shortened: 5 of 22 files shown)


pragmaware:
Currently the function parameters are enabled via +z which is "function parameters inside function definitions".
We have to change the help string to include prototypes too.... but I wonder if it makes sense to hide these behind a different "switch" at all so one could enable prototypes and function parameters but not the prototype parameters.
The same consideration could extend also to the template arguments related to prototypes...

masatake:
> Currently the function parameters are enabled via +z which is "function parameters inside function definitions".
We have to change the help string to include prototypes too.... but I wonder if it makes sense to hide these behind a different "switch" at all so one could enable prototypes and function parameters but not the prototype parameters.

I think using the same switch 'z' for both function parameters and prototype parameters is better.
Other than scope, I think they are the same.
About template parameter, using different switch 'Z' is good for me.  For me, template parameter looks very different concept from function and prototype parameter. Maybe because I'm not familier with C++. Anyway we already introduced Z. So we should not change it.  

The following shows how the pull request works:
```console
$ cat input.c
cat input.c
void p(int x);

void f(int y)
{
}
$ ./ctags -o - input.c
./ctags -o - input.c
f	input.c	/^void f(int y)$/;"	f	language:C	typeref:typename:void
$ ./ctags -o - --kinds-C=+p input.c
./ctags -o - --kinds-C=+p input.c
f	input.c	/^void f(int y)$/;"	f	language:C	typeref:typename:void
p	input.c	/^void p(int x);$/;"	p	language:C	typeref:typename:void	file:
$ ./ctags -o - --kinds-C=+z input.c
./ctags -o - --kinds-C=+z input.c
f	input.c	/^void f(int y)$/;"	f	language:C	typeref:typename:void
y	input.c	/^void f(int y)$/;"	z	language:C	function:f	typeref:typename:int	file:
$ ./ctags -o - --kinds-C=+zp input.c
./ctags -o - --kinds-C=+zp input.c
f	input.c	/^void f(int y)$/;"	f	language:C	typeref:typename:void
p	input.c	/^void p(int x);$/;"	p	language:C	typeref:typename:void	file:
x	input.c	/^void p(int x);$/;"	z	language:C	prototype:p	typeref:typename:int	file:
y	input.c	/^void f(int y)$/;"	z	language:C	function:f	typeref:typename:int	file:
$ ./ctags -o - --kinds-C=+zp-f input.c
./ctags -o - --kinds-C=+zp-f input.c
p	input.c	/^void p(int x);$/;"	p	language:C	typeref:typename:void	file:
x	input.c	/^void p(int x);$/;"	z	language:C	prototype:p	typeref:typename:int	file:
y	input.c	/^void f(int y)$/;"	z	language:C	function:f	typeref:typename:int	file:
```

In this pull request, ctags doesn't emit tags for parameters of a prototype if the prototype kind is disabled <1>.
On the other hands, ctags emits tags for parameters of a function even if the function kind is disabled <2>.

In the most of all parsers I wrote, I implemented the behavior <2>: tagging children even if the kind for their parent object is disabled. However, in this pull request I chose <1> because emitting tags for parameters of prototype when disabling prototype kind looks too noisy.

@pragmaware, how I should do? 
See https://github.com/universal-ctags/ctags/issues/3043 about the background of this pull request.

pragmaware:
> [...]
> Other than scope, I think they are the same.

Well, they obviously refer to the same "physical" parameters, so in this sense they are the same. However, the usage by the ctags clients may be very different. Take the very common operation of "completion": most editors that use ctags do that. The parameters of the function are essentially local variables: one wants to complete them in the scope of the function and thus they are very useful. The parameters of the prototype have no such role: one never wants to complete them or even to look them up by name. The only use of prototype parameters is to help analyzing the signature of the function "from outside", for example when filling the arguments of the call. But that's a rather advanced feature, few editors really have it and now that I look at it ctags doesn't even have enough information to implement it that way (index of the parameter is missing).

My editor, for instance, has the first feature (completion) but has no use for the parameters of the prototype: it would discard them. That's why I'd like to keep function parameters on but prototype parameters off. But in the end, it's not a big deal: they are easy to filter out, now that they are marked with the "prototype" scope.

> About template parameter, using different switch 'Z' is good for me. For me, template parameter looks very different concept from function and prototype parameter. Maybe because I'm not familier with C++.

From an editor-using-ctags perspective they are kind-of similar. Inside a function definition the template parameters are placeholders for types. One completes and resolves them like types in the function body. The prototype ones, again, are used as map/translation but never as completion. Editors fully handling template parameters are even more exotic beasts...

> Anyway we already introduced Z. So we should not change it.

I agree... though for the prototype template parameters nobody would notice if we added another kind :)

> In the most of all parsers I wrote, I implemented the behavior <2>: tagging children even if the kind for their parent object is disabled. However, in this pull request I chose <1> because emitting tags for parameters of prototype when disabling prototype kind looks too noisy.

Makes sense. I can't think of a good reason one would want to see the prototype parameters but not the prototypes...

masatake:
Thank you for the ​suggestive comment.

> ... (index of the parameter is missing).

Yes. I have been thinking about adding "nth:" as a COMMON field.

input.c:
```
struct point *make_point(int x0, int y0);
struct point {
  struct object base;
  int x;
  int y;
};
void abort(void);
struct char *format(char *fmt,...);
```
expected.tags:
```
make_point	f.c	/^struct point *make_point(int x0, int y0);$/;"	p	typeref:struct:point *	file:	arity:2
x0	f.c	/^struct point *make_point(int x0, int y0);$/;"	z	prototype:make_point	typeref:typename:int	file:	nth:0
y0	f.c	/^struct point *make_point(int x0, int y0);$/;"	z	prototype:make_point	typeref:typename:int	file:	nth:1
point	f.c	/^struct point {$/;"	s	file:
base	f.c	/^  struct object base;$/;"	m	struct:point	typeref:struct:object	file:	nth:0
x	f.c	/^  int x;$/;"	m	struct:point	typeref:typename:int	file:	nth:1
y	f.c	/^  int y;$/;"	m	struct:point	typeref:typename:int	file:	nth:2
abort	f.c	/^void abort(void);$/;"	p	typeref:typename:void	file:	arity:0
format	f.c	/^struct char *format(char *fmt,...);$/;"	p	typeref:struct:char *	file:	arity:1+
fmt	f.c	/^struct char *format(char *fmt,...);$/;"	z	prototype:format	typeref:typename:char *	file:	nth:0
```
(signature fields are omitted in this example.)
pParamInfo keeps parameters in arrays. So filling nth fields for parameters may be easy.
I would like to implement nth after merging this.
I think nth: field for struct members is not so bad idea. It may be less useful than that for parameters.
In C language, for representing inheritance, developers put an object of the parent class to nth:0 member of a structure.
In other words, from the nth:0, we can know class hierarchy.

I'm also thinking about arity: field. I'm not sure how useful it is.


About template parameter(tparam), whether a tparam is for template prototype or template function can be known from its scope field. See f281087 included in this pull request or #3062.

masatake:
I found nth: field is not enough if the target language, like C++, supports overloading.

input.cc:
```
void f(int x) {}
void f(float y, int x) {}
void f(int x, float y) {}
```

u-ctags --sort=no --fields=+oS --kinds-C++=+zp -o - input.cc  | grep ^x
```
x	input.cc	/^void f(int x) {}$/;"	z	function:f	typeref:typename:int	file:	nth:0
x	input.cc	/^void f(float y, int x) {}$/;"	z	function:f	typeref:typename:int	file:	nth:1
x	input.cc	/^void f(int x, float y) {}$/;"	z	function:f	typeref:typename:int	file:	nth:0
```

I wonder how about introducing `inSignature:` field like:
```
x	input.cc	/^void f(int x) {}$/;"	z	function:f	typeref:typename:int	file:	nth:0	inSignature:(int x)
x	input.cc	/^void f(float y, int x) {}$/;"	z	function:f	typeref:typename:int	file:	nth:1	inSignature:(float y, int x)
x	input.cc	/^void f(int x, float y) {}$/;"	z	function:f	typeref:typename:int	file:	nth:0	inSignature:(int x, float y)
```

pragmaware:
> I found nth: field is not enough if the target language, like C++, supports overloading.

Yes, this is a problem. In my editor I solve it by looking at line numbers.
When looking for the containing scope of a symbol I preferentially pick the one that is defined in the same file, with a line number that is lower or equal and the closest one. This mostly works.

However, ctags internally *has* the hierarchy information in this case so it could be output in some way. For example, we could assign a synthetic unique identifier to the parent symbol (function definition) and reference it in the child symbol (parameter)... that would be unambiguous.

pragmaware:
Now that I think of it, it's not only a problem with overloads. It's potentially with any symbol. For example, when taking into consideration a large C/C++ code base, possibly including libraries, it's not uncommon to have multiple classes with the same name in the global namespace. Or maybe the same static helper function repeated in many cpp files... It's just enough to make sure the compiler can see only one of these symbols at a time. I believe this can happen also in most other languages.

So a tree-building ctags client has always to solve the child-parent matching problem in some way. File+line-number matching is one way but it's not 100% reliable. Having unique ids for symbols would be a nice addition.

masatake:
>  Having unique ids for symbols would be a nice addition.

Do you think cork-indexes are appropriate as the ids?

To improve ctags in the view of this topic, I thought what we needed.

A. tags.d. Managing whole the source tree in a tags file is hard. I'm thinking about per-source file tags files. tags.d is a name of directory where the tags files are stored.  This allows incremental updating.

B. filesystem langauge. ctags is a file oriented program. It doesn't consider relation between two input files. `filesystem language` provides the way to record more information about input files to tags file in extensible way. ctags has `file` kind already. However, it is hard to attach more information on it. 

C. multi-pass parsing. An experimental implementation is proposed in #2741. #2471 reads a tags file as hint. I must read whole the tags file. If the tags file is larger, loading takes more time. Instead of dealing a tags file, tags files under tags.d may allow on-demand loading.

I'm not sure these are the way to go.
Much more studies are needed.
I wonder how code3 uses a tags file.

masatake:
@pragmaware, can I merge this pull request?

pragmaware:
> Do you think cork-indexes are appropriate as the ids?

Anything unique in the context of a tags file will do. If the cork indexes are unique and don't change then they are OK.

> I wonder how code3 uses a tags file.

I must say that a tags file by itself has limited application. When it's properly sorted it can be reasonably used to lookup a single symbol at a time without consideration of scope. However this is only the most basic form of "developer assistance". Most IDEs and code editors do far more than that and use derived structures that are more complex (and efficient) than a tag file.

In fact, in code3 case the tags file is never really written to disk. ctags is launched on subtrees of code (whole source tree, an include directory, a single file as it is being edited and so on) and its standard output is processed directly into memory.

The memory structure is called "symbol store" and is composed of different pieces:
- a list of "symbol files", which are lists of symbols associated to a single source file sorted by line number plus a few additional maps used to speed-up the context lookups
- a symbol tree for each language that is being manipulated: a real tree, with all the container -> member relationships
- a sorted list of symbols for each language (similar to a tag file but with in-memory objects interconnected with the other structures)

If you want to take a look, [the symbol store is here](https://github.com/pragmaware/code3/blob/f96191a506fec0590a28985c5bbf7b870a22fdb5/code3/src/lib/symbol/tools/C3SymbolStore.h#L32), while the symbol tree parts [are here](https://github.com/pragmaware/code3/tree/master/code3/src/lib/symbol/symbol/).

This whole structure allows the editor to quickly do complex and accurate completions. Say that the user is editing the following file with the cursor placed at position ``<cursor>``:
```
...
Class2 c;
...
void Class1::method1()
{
   c.mem<cursor>
}
...
```

If the user now presses tab, the editor will:
- extract the text before the cursor -> ``c.mem``
- figure out that ``mem`` is the token to complete and it is a member of an aggregate that is in the ``c`` variable
- find the current symbol file in the symbol store
- lookup the current context in the symbol file, that is, find the symbol (function) that is containing the current line number (that would be function ``method1`` - note how start and end line numbers are useful here)
- recursively look for the symbol ``c``: the algorithm would first look for variables or parameters inside the context symbol ``method1``, then look for member variables in its container ``Class1``... and finally climb up to the global scope to find ``c`` as global variable.
- resolve the type of ``c``, that is ``Class2`` and look it up in the global scope (the scope of definition of ``c``)
- find all the members (variables or functions) of ``Class2`` that start with the string ``mem``
- if there is only one then just complete ``c.mem`` to the name of the symbol
- if there are multiple then complete the common part and present a choice dialog to the user

Quite complicated process. In more involved cases may take a noticeable time with optimized in-memory structures. Totally unfeasible with only a tag file.

I don't think ctags should try to do this process internally. The reasons are:
- it's really too complicated
- it requires several complicated structures in memory that must be updated on-the-fly
- it assumes a global knowledge of the whole source tree the user is manipulating, something that a single ctags run may not have
- would almost certainly slow ctags down quite a lot.. but this process **must** be fast.
- would possibly require compatibility-breaking changes

What is mean is simply: the tags file **is good as it is**. It's just a plain symbol storage format.

I think ctags should:

- provide the maximum amount of information attached to each symbol 
- allow the user to choose what is needed and what not (because the useless information is wasted space/time).
- possibly provide a more reliable way to re-build the tree: as I have noted in previous comments there is considerable amount of "guessing"
- eventually provide a **separate library** to [implement the process I've described](https://github.com/pragmaware/code3/tree/f96191a506fec0590a28985c5bbf7b870a22fdb5/code3/src/lib/symbol), all in memory

pragmaware (approved):
Go for it!

masatake:
> >  Do you think cork-indexes are appropriate as the ids?
> Anything unique in the context of a tags file will do. If the cork indexes are unique and don't change then they are OK.

A cork index is unique in a input file. However, it can be duplicated. So unique id for each input file is needed in addition to the cork index.

pragmaware:
> A cork index is unique in a input file. However, it can be duplicated. So unique id for each input file is needed in addition to the cork index.

Just use a global integer incremented at each new symbol?

PR #3066, opened 2021-06-13
minitrip: limited version of the roundtrip target for CI/CD environments

### LINKED ISSUE #2851: Select and limit the input for roundtrip target
It takes a too long time.
```
[jet@living]~/var/ctags-new% find -name expected.tags | xargs wc -l | sort -n -r
find -name expected.tags | xargs wc -l | sort -n -r
  15684 total
   1781 ./Units/review-needed.r/test.vhd.t/expected.tags
    444 ./Units/parser-emacsLisp.r/simple-emacsLisp.d/expected.tags
    356 ./Units/parser-cxx.r/cxx-keywords-as-c-identifiers.b/expected.tags
 ...
```

./Units/review-needed.r/test.vhd.t/expected.tags is too large.
Removing this is one of the solutions.

masatake:
Close #2851.

### DIFF
--- a/misc/roundtrip
+++ b/misc/roundtrip
@@ -22,6 +22,22 @@ export LC_ALL=C
 
 READTAGS=${1:-./readtags}
 UNITS=${2:-./Units}
+shift 2
+
+MINITRIP=
+while [ $# -gt 0 ]; do
+    case $1 in
+	--minitrip)
+	    MINITRIP=minitrip
+	    ;;
+	*)
+	    echo "Unexpected argument: $1" 1>&2
+	    exit 1
+	    ;;
+    esac
+    shift
+done
+
 s=0
 
 if ! [ -f "${READTAGS}" ]; then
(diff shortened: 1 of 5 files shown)


masatake:
@hirooih, could you look at e6409f9 ?

misc/roundtrip
```diff
@@ -22,6 +22,22 @@ export LC_ALL=C
 
 READTAGS=${1:-./readtags}
 UNITS=${2:-./Units}
+shift 2
+
+MINITRIP=
+while [ $# -gt 0 ]; do
+    case $1 in
+	--minitrip)
```
masatake, line 30: The option is implemented here.

hirooih: Oh, I see.

I did not understand that e6409f9 was one of commits in this PR.  I went to e6409f9 directly last night.
I checked https://github.com/universal-ctags/ctags/commits/master/misc/roundtrip.

`minitrip` is really an internal option.  It is not an important information for readers.

> *minitrip* internal option was introduced because roundtrip target
> took too long time on CI/CD environments. The option is automatically
> enabled when running the target on the environments.
> 
> When the option is enabled, the target uses only marked `expected.tags`
> files instead of using all the `expected.tags` files. An empty file

This could be;

> On CI/CD environments, the target uses only marked `expected.tags`
> files instead of using all the `expected.tags` files because the roundtrip target
> took too long time. An empty file ...

How about this?

masatake: I agree with you. Thank you. I will update the commit.

PR #3067, opened 2021-06-13
main: introduce nth field

masatake:
For input.h,
```
inline void f(int x, int y) {}
#define F(X,Y) (X+Y)
```

wiith ``--fields=+oS --kinds-C++=+zD``, ctags emits:
```
f	input.h	/^inline void f(int x, int y) {}$/;"	f	typeref:typename:void	signature:(int x,int y)
x	input.h	/^inline void f(int x, int y) {}$/;"	z	function:f	typeref:typename:int	file:	nth:0
y	input.h	/^inline void f(int x, int y) {}$/;"	z	function:f	typeref:typename:int	file:	nth:1
X	input.h	/^#define F(X,Y) (X+Y)$/;"	D	macro:F	nth:0
Y	input.h	/^#define F(X,Y) (X+Y)$/;"	D	macro:F	nth:1
F	input.h	/^#define F(/;"	d	signature:(X,Y)
```


masatake:
We must get an ack at https://github.com/liuchengxu/vista.vim/issues/404 for merging this pull request.

masatake:
We got an ack from vista.vim.

main/field.h
```diff
@@ -65,8 +65,9 @@ typedef enum eFieldType { /* extension field content control */
 	FIELD_XPATH,
 	FIELD_END_LINE,
 	FIELD_EPOCH,
+	FIELD_NTH,
 
-	FIELD_BUILTIN_LAST = FIELD_EPOCH,
+	FIELD_BUILTIN_LAST = FIELD_NTH,
```
masatake, line 70: Requiring updating FIELD_BUILTIN_LAST when we add a field can cause a bug.

PR #3070, opened 2021-06-16
C++: recognize operator= even if the statement starts from struct, class or union

### LINKED ISSUE #2840: C++: assignment operator overloading in struct will cause parse error
I find assignment operator overloading in struct will cause parse error.

The test code here.
```
typedef struct Currency
{
    int Dollar;
    int Cents;

    struct Currency &operator=(Currency& value)
    {
        Dollar = value.Dollar;
        Cents = value.Cents;
        return *this;
    }
};

void test()
{
}
```

I use following command to parse:
```
ctags.exe --output-format=json --langmap=c:+.inc --kinds-c=+plz --fields=+nie -o - D:\tmp\test\testIfMacro.c
```

Result is :
```
{"_type": "tag", "name": "Cents", "path": "D:\\tmp\\test\\testIfMacro.c", "pattern": "/^    int Cents;$/", "file": true, "line": 4, "typeref": "typename:int", "kind": "member", "scope": "Currency", "scopeKind": "struct", "end": 4}
{"_type": "tag", "name": "Currency", "path": "D:\\tmp\\test\\testIfMacro.c", "pattern": "/^typedef struct Currency$/", "file": true, "line": 1, "kind": "struct"}
{"_type": "tag", "name": "Dollar", "path": "D:\\tmp\\test\\testIfMacro.c", "pattern": "/^    int Dollar;$/", "file": true, "line": 3, "typeref": "typename:int", "kind": "member", "scope": "Currency", "scopeKind": "struct", "end": 3}
```

As you can see from the result, the function `test` after the struct definition is missing.

masatake:
Close #2840.

The original code assumed a statement startnig from struct as a
definition of a struct. This caused that the parser failed in
recognizing an operator= definition (or prototype) returning a struct typed data.

### DIFF
--- a/Units/parser-cxx.r/struct-keyword-not-for-defining-struct.d/args.ctags
+++ b/Units/parser-cxx.r/struct-keyword-not-for-defining-struct.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--kinds-C++=+plz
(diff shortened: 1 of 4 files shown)


pragmaware (approved):
Looks good!

masatake:
Thank you for reviewing.

Issue #3014, opened 2021-05-07
Failed to open file with unicode name longer than 4


TroyDanielFZ:
```
ctags --options=NONE 比较长的测试文件.cpp
```
The file is encoded in cp936, and when the filename is short as `文件.cpp` or long ascii name, it works well. 
However, when the filename is in unicode, it fails. Sometimes for3,4 or longer names.
I can figure out why, and this maybe a bug?

k-takata:
Please use the issue template.
What is the ctags version? What is your OS? How do you get the executable?

TroyDanielFZ:
Sorry for duplicate submission due to network issues.

The name of the parser:


The command line you used to run ctags:

```
ctags --options=NONE 比较长的测试文件.cpp
ctags --options=NONE 比较长的测试文.cpp
ctags --options=NONE 比较长的测试.cpp
ctags --options=NONE 比较长的测.cpp
ctags --options=NONE 比较长的.cpp
ctags --options=NONE 比较长.cpp
ctags --options=NONE 比较.cpp
ctags --options=NONE 比.cpp
```

All the  input files are empty.

No output content, e.g. no file `tags` generated.

Outputs in the command prompt are:

```
e:\temp\>ctags --options=NONE 比较长的测试文件.cpp
ctags: Notice: No options will be read from files or environment
ctags: Warning: cannot open input file "姣旇緝闀跨殑娴嬭瘯鏂囦欢.cpp" : No such file or directory

e:\temp\>ctags --options=NONE 比较长的测试文.cpp
ctags: Notice: No options will be read from files or environment
ctags: Warning: cannot open input file "姣旇緝闀跨殑娴嬭瘯鏂?cpp" : No space left on device

e:\temp\>ctags --options=NONE 比较长的测试.cpp
ctags: Notice: No options will be read from files or environment
ctags: Warning: cannot open input file "姣旇緝闀跨殑娴嬭瘯.cpp" : No such file or directory

e:\temp\>ctags --options=NONE 比较长的测.cpp
ctags: Notice: No options will be read from files or environment
ctags: Warning: cannot open input file "姣旇緝闀跨殑娴?cpp" : No space left on device

e:\temp\>ctags --options=NONE 比较长的.cpp
ctags: Notice: No options will be read from files or environment

e:\temp\>ctags --options=NONE 比较长.cpp
ctags: Notice: No options will be read from files or environment

e:\temp\>ctags --options=NONE 比较.cpp
[...]

Issue #2996, opened 2021-05-03
main/main.c has logical mistakes


leleliu008:
1. [main/main.c#L125-L164](https://github.com/universal-ctags/ctags/blob/master/main/main.c#L125-L164)
```
#elif defined (HAVE__FINDFIRST)

#if defined (HAVE__FINDFIRST)
#endif

#endif
```

2. [main/main.c#L91-L164](https://github.com/universal-ctags/ctags/blob/master/main/main.c#L91-L164)
```
#if defined (HAVE_OPENDIR) && (defined (HAVE_DIRENT_H) || defined (_MSC_VER))
        static bool recurseUsingOpendir (const char *const dirName){}
#elif defined (HAVE__FINDFIRST)
        static bool createTagsForWildcardEntry (){}
        static bool createTagsForWildcardUsingFindfirst (const char *const pattern){}
#endif
```
here should change to following:
```
#if defined (HAVE_OPENDIR) && (defined (HAVE_DIRENT_H) || defined (_MSC_VER))
        static bool recurseUsingOpendir (const char *const dirName){}
#endif

#if defined (HAVE__FINDFIRST)
        static bool createTagsForWildcardEntry (){}
        static bool createTagsForWildcardUsingFindfirst (const char *const pattern){}
#endif
```
I find on `MSYS2/MINGW32` and `MSYS2/MINGW64` have `opendir()` function, but don't define `HAVE_OPENDIR` in `./main/e_msoft.h`, so `#if defined (HAVE_OPENDIR) && (defined (HAVE_DIRENT_H) || defined (_MSC_VER))` always `false`, 

When I use CMake build system, I checked `opendir()` function in build script, and defined `HAVE_OPENDIR` macro, so the logic error.

masatake:
Please, make a pull request for the solution that you found:-)

masatake:
Run 'git blame' if you find a code questionable?
I have changed many areas in main/. However, some areas have not been changed from Exuberant Ctags.
If a questionable area has not been changed since the Exuberant Ctags, and if you know the way to improve,
in many cases, opening a pull request is better.

If a questionable area has been changed by @masatake, opening an issue makes sense.

leleliu008:
@masatake OK.

k-takata:
This will be fixed in #3072.

PR #3072, opened 2021-06-18
build-sys,main,mingw: Enable MANUAL_GLOBBING

k-takata:
On Windows, the shell doesn't expand wildcards.
Ctags should expand them by ourselves.

See also: https://github.com/universal-ctags/ctags/issues/3014#issuecomment-863628889


masatake:
I tried this change in my MINGW64 environment.
```
$ ./ctags -o - main/main.c
```
prints nothing. 
ctags rebuilt after `git checkout HEAD^` works fine.

I wonder this is something to do with https://github.com/universal-ctags/ctags/issues/2996 .

k-takata:
Codecov reports nothing about the change. It might be better to use codecov also in the MinGW environment.

masatake (approved):

masatake:
LGTM. Sorry to be late for the review.

PR #3078, opened 2021-06-18
TEMP: win32 gnulib regex

hirooih:
Test PR for win32 gnulib regex.


hirooih:
Now works on Win32 env.

PR #3080, opened 2021-06-24
GitHub Actions: pass GITHUBACTIONS=1 when running make-targets for testing

masatake:

### DIFF
--- a/.github/workflows/testing-alpine.yml
+++ b/.github/workflows/testing-alpine.yml
@@ -30,5 +30,5 @@ jobs:
       - run: make install
       - run: file /usr/bin/ctags
       - run: ctags --version
-      - run: make check V=1 APPVEYOR=1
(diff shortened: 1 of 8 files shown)


k-takata:
Most of the CI environments set the `CI` environment variable to `true` (or `True`).
* https://docs.github.com/actions/reference/environment-variables
* https://www.appveyor.com/docs/environment-variables/
* https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

If we don't need to distinguish each CI environment and just want to know whether we are running the tests on any CI environment, we might be able to use the `CI` variable.

masatake:
@k-takata, thank you! I'm thinking about introducing unified environment variable.
I will make a pull request for using the `CI` variable after merging this.

masatake:
I merged this for reducing the time taking for roundtrip target.

Issue #3093, opened 2021-07-17
"lang-sequel" test fails on "make distcheck" 


hirooih:
`make distcheck` fails on `lang-sequel`.
FYI `make tmain` passes.

The version of ctags: Current HEAD
commit 23ea3765d15a4cc361a1dfef497617c91c704fbe (HEAD -> master, upstream/master)
Merge: 88cd4981 8d3a6bc8
Author: Masatake YAMATO <yamato@redhat.com>
Date:   Thu Jul 15 07:41:49 2021 +0900

```
$ make distclean
...
$ ./autogen.sh
...
$ ./configure
...
$ make distcheck
...
Testing lang-sequel
------------------------------------------------------------
stdout                                                      passed
stderr                                                      failed (diff: /home/hiroo/GitHub/ctags/universal-ctags-5.9.0/_build/sub/Tmain/lang-sequel.d/stderr-diff.txt)
exit                                                        failed (diff: /home/hiroo/GitHub/ctags/universal-ctags-5.9.0/_build/sub/Tmain/lang-sequel.d/exit-diff.txt)
...
Failed tests
============================================================
lang-sequel/stderr-compare (not committed/cached yet)
lang-sequel/exit-compare (not committed/cached yet)

make[1]: *** [Makefile:6315: tmain] Error 1
make[1]: Leaving directory '/home/hiroo/GitHub/ctags/universal-ctags-5.9.0/_build/sub'
make: *** [Makefile:5477: distcheck] Error 1
$ 
```

hirooih:
I found this during debugging #3054.
`make distcheck* falis with the error I reported on this PR.
I tested with 'master' branch and have the same fail, as I reported on this PR.

I'm using Ubuntu 20.04, Windows 10 WSL2.

#3094 is a similar issue. `make distcheck` passed on `fedora33_distcheck`.  `make distcheck` also fails on the branch on my env.

hirooih:
I've sent fixes for this issue on PR #3096.

masatake:
You can close this:-)

hirooih:
Fixed by #3096. Closed

PR #3096, opened 2021-07-17
Tmain: fix fails on make distcheck (fix for #3093)

hirooih:
`make distcheck` fails on Ubuntu 20.04 on Windows 10 WSL2.
These are fixes for them.

I also don't understand why `Tmain/epoch-field.d/run.sh` and  `Tmain/readtags-qualifier-begin.d/run.sh` passed on `make tmain`.

The fix of `Tmain/lang-sequel.d/args.ctags` is only for `make distcheck`.  But I don't understand why it passes on `fedora33_distcheck` CI test.

I found these issues during debugging #3054.

### DIFF
--- a/Tmain/lang-sequel.d/run.sh
+++ b/Tmain/lang-sequel.d/run.sh
@@ -7,4 +7,4 @@ CTAGS=$1
 
 . ../utils.sh
 
-${CTAGS} --quiet --options=NONE --options=./args.ctags input.unknown
+${CTAGS} --quiet --options=NONE --options=./args.ctags -o - input.unknown


Tmain/readtags-qualifier-begin.d/run.sh
```diff
@@ -20,8 +20,8 @@ fi
 
 run_readtags()
 {
-	local fexpr=$1
-	local action=$2
+	local fexpr="$1"
+	local action="$2"
```
masatake, line 24: Is this for fixing a bug or just for prevention?

```
[jet@living]/tmp/ctags/universal-ctags-5.9.0/_build/sub/Tmain/lang-sequel.d% dash
$ x='a b'
$ y=$x         
$ echo $y
a b
```

I think this change is for prevention.

Tmain/lang-sequel.d/args.ctags
```diff
@@ -3,3 +3,4 @@
 --_sequel-UnknownX={{
     (bye) ==
 }}
+-o -
```
masatake, line 6: Good catch!

Could you write this in run.sh, not here.

Tmain/readtags-qualifier-begin.d/run.sh
hirooih: It was for a bug fix.  I saw errors on this test (and epoch-field.d, too). But I cannot reproduce them.
It seems that something was wrong with me.

I misunderstood that shell expansions rule.  Shell expansions are done after variable assignments.

I've revert the change.

Tmain/lang-sequel.d/args.ctags
hirooih: > Could you write this in run.sh, not here.

Sure.  I've done.

masatake (approved):

PR #3101, opened 2021-07-23
Import fnmatch from gnulib

hirooih:
ctags distribution includes fnmatch library from glibc, but it is too old.
This fnmatch version of #3054 (for regex).


hirooih:
@masatake san,

Fails on `build_and_test` are issues I reported on #3102.
Please review this PR.

Thanks.

hirooih:
I forgot to remove fnmatch/*.  I've commented 035afa5.

masatake:
I disabled the testing on centos6. Could you rebase your changes on the latest "master" branch?

hirooih:
> Could you rebase your changes on the latest "master" branch?

Done.  All checks have passed.

masatake:
Could you add gnulib_fnmatch feature as you did in gnulib_regex ?
It will be useful for debugging in the future.

hirooih:
> Could you add gnulib_fnmatch feature as you did in gnulib_regex ?

For regex `configure` may choose regex in the system library.  On the other hand `gnulib/fnmatch.c` is always used for fnmatch.
And I cannot find any clues (for example #define) to know GNU fnmatch is used.

Only I can do is adding a feature item unconditionally.

```diff
--- a/main/options.c
+++ b/main/options.c
@@ -554,6 +554,7 @@ static struct Feature {
        /* Following two features are always available on universal ctags */
        {"wildcards", "can use glob matching"},
        {"regex", "can use regular expression based pattern matching"},
+ {"gnulib_fnmatch", "linked with the Gnulib fnmatch library"},

 /* https://lists.gnu.org/archive/html/bug-gnulib/2011-07/msg00435.html */
 #ifdef _REGEX_INCLUDE_LIMITS_H
	{"gnulib_regex", "linked with the Gnulib regular expression library"},
```

Is this what you expect?

masatake:
I found `REPLACE_FNMATCH = 0` in Makefile. I expect we can use it.
I wonder why there is no `REPLACE_REGCOMP`.

hirooih:
@masatake san,

> I found REPLACE_FNMATCH = 0 in Makefile. I expect we can use it.

REPLACE_FNMATCH just used as follows in fnmatch.in.h;

```
# if @REPLACE_FNMATCH@   // --> #if 0
...
# else
#  if !@HAVE_FNMATCH@   // #if !1
...
#  endif
_GL_CXXALIAS_SYS (fnmatch, int,
                  (const char *pattern, const char *name, int flags));
# endif
```

This results;

> Only I can do is adding a feature item unconditionally.

masatake:
```
# if @REPLACE_FNMATCH@   // --> #if 0
```

I don't understand this.

Do you mean `@REPLACE_FNMATCH@` is ALWAYS replaced with 0?

hirooih:
> Do you mean @REPLACE_FNMATCH@ is ALWAYS replaced with 0?

No.  I just showed how `@REPLACE_FNMATCH@` is used.
It does not define anything we (`main/options.c`) can use.

masatake:
> No. I just showed how @REPLACE_FNMATCH@ is used.
> It does not define anything we (main/options.c) can use.

In that case, I think we can define something we can use if REPLACE_FNMATCH is not 0 in Makefile.
I will show you how to do it. Give me time.

masatake:
How about this one?
```
index 1d049980..c089f5c1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,8 @@ AH_TEMPLATE([MSDOS_STYLE_PATH],
 	[Define to 1 if your system uses MS-DOS style path.])
 AH_TEMPLATE([ENABLE_GCOV],
 	[Define to 1 if gcov is instrumented.])
+AH_TEMPLATE([USE_GNULIB_FNMATCH],
+	[Define to 1 when using fnmatch implementation in bundled gnulib.])
 
 
 # Define convenience macros
@@ -536,6 +538,10 @@ PRETTY_ARG_VAR([WARNING_CFLAGS], [C compiler warning flags],
 #   header files or functions.
 gl_INIT
 
+if test "$REPLACE_FNMATCH" != 0; then
+	AC_DEFINE([USE_GNULIB_FNMATCH])
+fi
+
 AC_CHECK_FUNCS(asprintf)
 AC_CHECK_FUNCS(strstr)
 AC_CHECK_FUNCS(strcasecmp stricmp, break)
diff --git a/main/options.c b/main/options.c
index 369ba05c..9eb249ef 100644
--- a/main/options.c
+++ b/main/options.c
@@ -553,6 +553,9 @@ static struct Feature {
 #endif
 	/* Following two features are always available on universal ctags */
 	{"wildcards", "can use glob matching"},
+#ifdef USE_GNULIB_FNMATCH
+	{"gnulib_fnmatch", "linked with the Gnulib fnmatch library"},
+#endif
 	{"regex", "can use regular expression based pattern matching"},
 
 /* https://lists.gnu.org/archive/html/bug-gnulib/2011-07/msg00435.html */
```

hirooih:
> How about this one?

Now I see. `configure` script can access the variables in m4 macros after `gl_INIT`.
I've merged your fix and push it.

masatake:
My change works only for the platforms using ./configure; the script sets USE_GNULIB_FNMATCH.
The other platforms, we must set USE_GNULIB_FNMATCH manually. Can I ask you to set in...`*.mak` files or somewhere.

masatake:
BTW, REPLACE_FNMATCH is a shell variable, not m4 variable.

hirooih:
> The other platforms, we must set USE_GNULIB_FNMATCH manually. Can I ask you to set in...*.mak files or somewhere.

Good catch!  I've added it on `win32/config_mingw.h` and `win32/config_mvc.h`.

> BTW, REPLACE_FNMATCH is a shell variable, not m4 variable.

You are right.

masatake (approved):
Thank you very much. Sorry for taking much time for reviewing.

hirooih:
You are so welcome. Thank you for your precious advices.

Issue #1249, opened 2016-12-23
No rst2man = no man page


dtikhonov:
I think that requiring that `rst2man` be installed in order to generate and install a man page is too much.  Here is my story:

When I recently installed a new system, I git-cloned Universal Ctags and installed it (as I always do).  Some time later, I wanted to look up a ctags option:

- `man ctags`
- `locate ctags.1`
- `find /usr -name ctags.1`
- WTF...

I went back to the cloned git repo and found ctags.1.rst.  A few minutes later I found 15413bacefa25b8ceb3767f774726cbbd89a9641 and a few more minutes later I installed `rst2man` and had my man page.

On Ubuntu 16.04.1 LTS, installing `docutils-common` (for which I had to run a web search, as `apt search rst2man` produces no hits) installs ten packages.  They are:

- docutils-common:amd64 (0.12+dfsg-1)
- docutils-doc:amd64 (0.12+dfsg-1, automatic)
- python-chardet:amd64 (2.3.0-2, automatic)
- python-docutils:amd64 (0.12+dfsg-1, automatic)
- libpaper1:amd64 (1.1.24+nmu4ubuntu1, automatic)
- python-roman:amd64 (2.0.0-2, automatic)
- python-pygments:amd64 (2.1+dfsg-1, automatic)
- libpaper-utils:amd64 (1.1.24+nmu4ubuntu1, automatic)
- libwebpmux1:amd64 (0.4.4-1, automatic)
- python-pil:amd64 (3.1.2-0ubuntu1, automatic)

That's a lot of packages.  Also, what if the user does not have root privileges -- what is he to do for a man page?

I think it would be better to simply check in ctags.1 as changes to ctags.1.rst are made.  This way, people who want to use Universal Ctags do not have to jump through hoops just to get a fine man page.

b4n:
IMO it's not a good idea.  IMO, the solution here is for those people to use release tarballs that contain those generated files, and to leave in Git only the actual sources.  Having generated files checked in creates noise at some point or another, e.g. if 2 contributors don't have the same version of *rst2man*, etc.
Also, where to stop?  Should *configure* be checked in, so people don't need Autoconf, Automake, Aclocal, Perl and whatnot?  Should the object files be checked in so people don't need a C compiler?  Should the executable be checked in so people don't need a linker?  Yes, I'm shamelessly using *reductio ad absurdum*, sorry :)

But well, that's my opinion :)

dtikhonov:
[...]

PR #3111, opened 2021-08-07
bulid-sys: use Automake on man directory

hirooih:
Make use of Automake on man directory.

When we add a man file, now we have to add it on 3 files, `configure.ac`, `Makefile.am`, and `man/Makefile`.
Now adding to only `man/Makefile.am` is enough.

Some more additional related fixes are included.  See commit log for details.


hirooih:
Finaly CI/CD tests passed.

@masatake san,
Please review.

masatake:
(reviewing now)

man/Makefile.am
```diff
@@ -0,0 +1,123 @@
...
+else
+reppat := $(foreach m,$(cites),-e 's/[[:<:]]$(m)/:ref:`& <&>`/g')
+endif
+
+update-docs: $(doc_files)
+
+# Delete the line "------" in the first 10 lines of ../docs/man/*.rst to
+# suppress unnecessary section indices.
+../docs/man/%.rst: %.rst
+	$(MKDIR_P) ../docs/man
```
masatake, line 93: A minor issue. Remove a whitespace character at the end of the line.

man/Makefile.am
```diff
@@ -0,0 +1,123 @@
+#
+# This is a stand-alone Makefile for Universal-ctags maintainers.
```
masatake, line 2: This was?

man/Makefile.am
```diff
@@ -0,0 +1,123 @@
...
+# GNU Make is needed.
+#
+# This doesn't work with Makefile.am at the top of source tree.
```
masatake, line 5: Now this file works well with Makefile.am at the top of source tree:-)

man/Makefile.am
```diff
@@ -54,7 +54,7 @@ doc_files  = $(addprefix ../docs/man/,$(rst_files))
 # for automake
 # generate man pages only when rst2man is installed
 if HAVE_RST2MAN
-man_MANS = $(man_pages)
+dist_man_MANS = $(man_pages)
```
masatake, line 57: I don't object to this change. However, I would like to point out a minor issue of this change.

configure takes an option, --program-prefix.
You can use the option like:
```
./configure --program-prefix=u-
make
make install 
```
The above command sequences install 'u-ctags'. `u-` is prepended to the command name.

The original man page generator runs after running configure.
The original man page generator reflects the option to the generated man pages.
e.g. in man/ctags-lang-sql.7.rst.in:
```
SYNOPSIS
--------
|       **@CTAGS_NAME_EXECUTABLE@** ... [--extras={guest}] --languages=+SQL ...
```

@CTAGS_NAME_EXECUTABLE@ was replaced.

The replacement done in the original man page generator can be done because it runs after running configure.

`dist_man_MANS` means generated man pages to our tarball.
The man page may be generated after running ./configure by a person who will do "make dist" before running ./configure by another person who does download "universal-ctags.tar.gz" and will do "make".
This is a degradation of the new man page generator with `dist_`.

I think this is acceptable degradation, though.

man/Makefile.am
```diff
@@ -0,0 +1,123 @@
...
+update-docs: $(doc_files)
+
+# Delete the line "------" in the first 10 lines of ../docs/man/*.rst to
+# suppress unnecessary section indices.
+../docs/man/%.rst: %.rst
+	$(MKDIR_P) ../docs/man 
+	sed $(reppat) -e '1,10s/^-*$$//' < $< > $@
+
+# use `[@]` not to be replaced by automake
+%.rst: %.rst.in
```
masatake, line 97: Now we will use configure, so I wonder why we have to replace @...@ manually.

masatake: ... I understand. With this, we don't have to add an entry to configure.in when adding a new man page.
However, as I wrote, `--program-prefix` doesn't work with this change.
I will look for a better solution.

masatake:
I would like to clear the goal.

We would like to reduce the number of lists when adding a new man page.

However, we don't break the following things:
- --program-perfix configure option works well, and
- man pages at docs.ctags.io use normal names like `ctags`.

masatake:
We can externalize the output file list in configure.ac like:
```
[yamato@dev64]~/var/ctags-github% git diff --cached
index 6f2fc603..84b302e1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -797,7 +797,7 @@ fi
 
 AC_CHECK_FUNCS(mblen)
 
-AC_CONFIG_FILES([Makefile gnulib/Makefile man/Makefile])
+AC_CONFIG_FILES(m4_include(default.conflist))
 
 AC_OUTPUT
 
diff --git a/default.conflist b/default.conflist
new file mode 100644
index 00000000..1d8ca807
--- /dev/null
+++ b/default.conflist
@@ -0,0 +1 @@
+Makefile gnulib/Makefile man/Makefile
```

The externalized list can be share between configure.ac and Makefile.am.
The command or scriptlet generating docs/man can use the list, too.
I will study this approach more.

man/Makefile.am
hirooih: Fixed.

man/Makefile.am
hirooih: I've changed as follows;

> # This is a Makefile.am read by the Autotools.

man/Makefile.am
hirooih: I've removed the last sentence.

man/Makefile.am
hirooih: I had to write my intension of this change in the commit log.
I've added the following to it.

>     This allows users who have not installed rst2man to install man pages.

But I am OK to withdraw this change.

***

#1249 was in my mind. 
The prefixed man pages are installed if the users run `configure` with `--prefix` option.

```
$ ./autogen.sh && ./configure --program-prefix=u_ --prefix=/tmp/ctags
...
$ make install
...
make[3]: Nothing to be done for 'install-exec-am'.
 /usr/bin/mkdir -p '/tmp/ctags/share/man/man1'
 /usr/bin/install -c -m 644 'ctags.1' '/tmp/ctags/share/man/man1/u_ctags.1'
 /usr/bin/install -c -m 644 'readtags.1' '/tmp/ctags/share/man/man1/u_readtags.1'
...
```

> I think this is acceptable degradation, though.

I also think so, although I overlooked that the replacement of `@CTAGS_NAME_EXECUTABLE@` did not occur as you pointed.
A user has to type `man u_ctags`, and he/she sees the page starting "ctags - Generate tag files for source code".

man/Makefile.am
hirooih: The change here is not to change the behavior.

hirooih:
> However, we don't break the following things:
> 
> - --program-perfix configure option works well, and
> - man pages at docs.ctags.io use normal names like ctags.

I agree with you.

> The externalized list can be share between configure.ac and Makefile.am.

I don't understand this. Makefile.am does not use the list.  It just added a file to be maintained.

masatake:
> >    The externalized list can be share between configure.ac and Makefile.am.

> I don't understand this. Makefile.am does not use the list. It just added a file to be maintained.

The comment area is too small to illustrate the idea.

#3114 implements the idea.
See https://github.com/universal-ctags/ctags/pull/3114/commits/32e9fbe84e3d7afe5687d598b3eac0f1d5f35209#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR242.

The technique used in this pull request throws away the improvement introduced in 1eace6c605a429e34fab7d389c8e7abec124c260.

The technique used in #3114 keeps the improvement. However, #3114 is complicated; it runs two code generators in autogen.sh.

The complexity comes from the requirements for
A. making contents for docs.ctags.io in this source tree, and
B. putting generated man pages to tar.gz.

About A, I expect we can find another solution if we study sphinx and readthedoc more.

About B.
Currently we maintain files .rst.in. I wonder what happens if we maintain files .in.rst instead.
Putting .in file generated from .in.rst by rst2man to tar.gz may help people build ctags on the platform where rst2man is unavailable.
However, I wonder whether we should take time to satisfy B. I want to ask people to install rst2man if they want to use the latest u-ctags.

hirooih:
> > > The externalized list can be share between configure.ac and Makefile.am.
> > 
> > I don't understand this. Makefile.am does not use the list. It just added a file to be maintained.
> > 
> 
> The comment area is too small to illustrate the idea.
> 
> #3114 implements the idea.

Now I understand what you meant.

I thought "the externalized list" meant the following list which is not shared.

> -AC_CONFIG_FILES([Makefile gnulib/Makefile man/Makefile])
> +AC_CONFIG_FILES(m4_include(default.conflist))

But "the externalized list" means shared lists like the list of man pages.  If so, I'm convinced.

> A. making contents for docs.ctags.io in this source tree, and
> ...
> About A, I expect we can find another solution if we study sphinx and readthedoc more.

These days I also read the the document of read the doc, but I concluded that it did not allow making contents.

> B. putting generated man pages to tar.gz.

The the following commit make this.

> - man_MANS = $(man_pages)
> + dist_man_MANS = $(man_pages)

As I wrote in #3114, this is only for users who does not install rst2man.
It has no effect on users who install rst2man.
The only issue is;

> A user has to type man u_ctags, and he/she sees the page starting "ctags - Generate tag files for source code".

as I wrote above.

> Putting .in file generated from .in.rst by rst2man to tar.gz may help people build ctags on the platform where rst2man is unavailable.

I think this can be implemented easily. Problem is that about 10 .in files must be committed.  It should be in another PR.
Shall I revert the `dist_man_MANS` change?

masatake:
> I think this can be implemented easily. Problem is that about 10 .in files must be committed. It should be in another PR.
> Shall I revert the dist_man_MANS change?

Could you explore the idea of .in.rst?
I guess rst2man must be run in autogen.sh in this approach.

TEMPORARILY, please revert the dist_man_MANS change?

hirooih:
> TEMPORARILY, please revert the dist_man_MANS change?

Done.

PR #3114, opened 2021-08-10
build-sys: unify "generated man page" lists into one

masatake:
Quoted from #3111 opened by @hirooih:

    When we add a man file, now we have to add it on 3 files,
    configure.ac, Makefile.am, and man/Makefile.  Now adding to only
    man/Makefile.am is enough.

This change introduces man/generated-man.list as the unified list.
misc/generate-man-m4.sh called from autogen.sh converts the unified
list to m4/generated-man.m4. configure.ac includes m4/generated-man.m4.

misc/generate-man-am.sh called from autogen.sh converts the unified
list to makefiles/generated-man.am. Makefile.am includes
makefiles/generated-man.am.


masatake:
@hirooih, how about this approach?

hirooih:
> @hirooih, how about this approach?

Thanks.

It reduces the number of files to be edited when a man page is added.
But it introduces 3 new files and makes build-process more complex a little.

I think my #3111 is better, because it just simplies the build process (reduces the number of lines in the files) using common Automake features.

masatake:
I withdraw this.

PR #3115, opened 2021-08-12
build-sys: generage man/*.in and add them to make dist

hirooih:
This PR is a continuation of "bulid-sys: use Automake on man directory" #3111.


hirooih:
We have 2 fails. Both are failed on `dist` target.

> Makefile:1374: *** Cannot make ctags.1.in: rst2man is not installed.  Stop.

This PR requires rst2man on dist and distcheck targets.  So this is an expected error.
I add docutils-python3 on the list of installed tools.

masatake (approved):

PR #3119, opened 2021-08-14
build-sys: misc updates

hirooih:
Several updates for build-sys

- Makefile.am
  - Fix compile error on `make man-test` following `./configure`.
  - remove more files by `make clean`
  - make better use of make-variables for packcc
- man/Makefile.am
  - implement silencing variable
- ctags-client-tools.7.rst.in
  - fix a warning error
- makefiles/testing.mak, etc.
  - define checkgen target
- source.mak, etc.
  - merge  makefiles/*_input.mak
  - reorder for consistency
  - define GNULIB_HEADS, GNULIB_SRCS in source.mak
- makefiles/help.mak
  - update `make help`
- etc.


Makefile.am
```diff
@@ -43,32 +43,37 @@ MAINTAINERCLEANFILES = autom4te.cache
...
 
 PACKCC = $(top_builddir)/packcc$(BUILD_EXEEXT)
+CLEANFILES += $(PACKCC)
 
 cc4b_verbose = $(cc4b_verbose_@AM_V@)
 cc4b_verbose_ = $(cc4b_verbose_@AM_DEFAULT_V@)
 cc4b_verbose_0 = @echo CC4BUILD "  $@";
-$(PACKCC): $(top_srcdir)/misc/packcc/src/packcc.c
-	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $(top_srcdir)/misc/packcc/src/packcc.c
+$(PACKCC): $(top_srcdir)/$(PACKCC_SRCS)
```
masatake, line 74: I strongly expect the number of the source file(s) of packcc is one even in the future.
However, the name `PACKCC_SRCS` implies it can hold more than one item though it holds only one item actually.

So, how about doing like:
```Make
$(PACKCC): $(foreach let,$(PACKCC_SRCS),$(top_srcdir)/$(let))
```
Ideally, we should not use GNU Make extension.

Makefile.am
```diff
@@ -43,32 +43,37 @@ MAINTAINERCLEANFILES = autom4te.cache
...
+	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $(top_srcdir)/$(PACKCC_SRCS)
```
masatake, line 75: You can use `$^` instead of `$(top_srcdir)/$(PACKCC_SRCS)`.
(But I don't know the portability of the automatic variable.)

source.mak
```diff
@@ -430,6 +430,42 @@ WIN32_HEADS = main/e_msoft.h
 WIN32_SRCS = win32/mkstemp/mkstemp.c
 WIN32_OBJS = $(WIN32_SRCS:.c=.$(OBJEXT))
 
+# common to MVC and MINGW
+CMN_GNULIB_HEADS = \
```
masatake, line 434: Could you use `COMMON_` as a prefix instead of `CMN_`?

masatake:
Excellent work!

Makefile.am
masatake: Another choice is defining top_PACKCC_SRCS or something manually in sources.mak.

Makefile.am
hirooih: > However, the name PACKCC_SRCS implies it can hold more than one item though it holds only one item actually.

I made this change to support the case where PACKCC_SRCS hold more than one item.  But you are right, I overlooked the concatenation of `$(top_srddir)/`.

> Ideally, we should not use GNU Make extension. but I cannot find a way without the extension.

If we can use GNU Make extension, we can simply fix as follows.  (Use VPATH and `$<`)

```diff
- $(PACKCC): $(top_srcdir)/misc/packcc/src/packcc.c
- 	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $(top_srcdir)/misc/packcc/src/packcc.c
+ $(PACKCC): $(PACKCC_SRCS)
+ 	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $<
```

Actually I used `$<' when I pushed first. But it failed on some platforms, because use of `$<` does not work on some platforms. (See [12.1 $< in Ordinary Make Rules](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.70/autoconf.html#g_t_0024_003c-in-Ordinary-Make-Rules)),

How about the following fix? (Add `$(top_srcdir)/` at the definition of `$(PACKCC_SRCS)`)

```diff
diff --git a/Makefile.am b/Makefile.am
index 22b3a4bb..5cffa241 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -71,8 +71,8 @@ CLEANFILES += $(PACKCC)
 cc4b_verbose = $(cc4b_verbose_@AM_V@)
 cc4b_verbose_ = $(cc4b_verbose_@AM_DEFAULT_V@)
 cc4b_verbose_0 = @echo CC4BUILD "  $@";
-$(PACKCC): $(top_srcdir)/$(PACKCC_SRCS)
-   $(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $(top_srcdir)/$(PACKCC_SRCS)
+$(PACKCC): $(PACKCC_SRCS)
+ $(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(PACKCC_CPPFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) -o $@ $(PACKCC_SRCS)

 if USE_READCMD
 bin_PROGRAMS+= readtags
diff --git a/source.mak b/source.mak
index b4b1e677..48977d85 100644
--- a/source.mak
+++ b/source.mak
@@ -420,8 +420,9 @@ READTAGS_HEADS = \
        $(NULL)
 READTAGS_OBJS  = $(READTAGS_SRCS:.c=.$(OBJEXT))

+# "$(top_srcdir)/" is required since VPATH does not work on this.
 PACKCC_SRCS = \
-   misc/packcc/src/packcc.c \
+ $(top_srcdir)/misc/packcc/src/packcc.c \
        \
        $(NULL)
 PACKCC_OBJS = $(PACKCC_SRCS:.c=.$(OBJEXT))
```

Makefile.am
hirooih: I did not read this review comment when I wrote the comment above.
What I proposed is similar to yours.

source.mak
hirooih: Sure.

hirooih:
Code coverage decreased packcc was not built correctly on the first push.  It is fixed.
The test does not rerun after forced-push.

Makefile.am
masatake: I wonder whether mk_mingw.mak works well or not with this change.

hirooih: > I wonder whether mk_mingw.mak works well or not with this change.

With and without this change, both `mk_mingw.mak` and `mk_mvc.mak`  do not work on the case where PACKCC_SRCS holds more than one item. They assume a single file input.

mk_mingw.mak

```make
$(PACKCC_OBJS): $(PACKCC_SRCS)
	$(V_CC) $(CC_FOR_PACKCC) -c $(OPT) $(CFLAGS) $(COMMON_DEFINES) -o $@ $<

$(PACKCC): $(PACKCC_OBJS)
	$(V_CC) $(CC_FOR_PACKCC) $(OPT) -o $@ $^
```

The following should work.

```make
$(PACKCC): $(PACKCC_SRCS)
	$(V_CC) $(CC_FOR_PACKCC) $(OPT) $(CFLAGS) $(COMMON_DEFINES) -o $@ $^
```

`mk_mvc.mak` has the same issue.

BTW how about renaming `PACKCC_SRCS` to `PACK_SRC` and commenting that it must be a single file?

```diff
--- a/source.mak
+++ b/source.mak
@@ -420,12 +420,9 @@ READTAGS_HEADS = \
        $(NULL)
 READTAGS_OBJS  = $(READTAGS_SRCS:.c=.$(OBJEXT))

-PACKCC_SRCS = \
-   misc/packcc/src/packcc.c \
-   \
-   $(NULL)
-PACKCC_OBJS = $(PACKCC_SRCS:.c=.$(OBJEXT))
+# only one file can be specified.
+PACKCC_SRC = misc/packcc/src/packcc.c
+PACKCC_OBJ = $(PACKCC_SRC:.c=.$(OBJEXT))

 WIN32_HEADS = main/e_msoft.h
 WIN32_SRCS = win32/mkstemp/mkstemp.c
```

P.S.
I found I understood `$^` (all the prerequisite) and `$<` (the first prerequisite) oppositely.

masatake: > BTW how about renaming PACKCC_SRCS to PACK_SRC and commenting that it must be a single file?

It is o.k. No comment is needed. The name tells everything developers should know.
Could you make a separate commit for renaming the macro names?

hirooih:
> Could you use COMMON_ as a prefix instead of CMN_?

Squash into the original commit.

> Could you make a separate commit for renaming the macro names?

  build-sys: rename PACKCC_SRCS and PACKCC_OBJS to PACKCC_SRCS and PACKCC_OBJS, respectively. (bbdad51)

I also added one more new commit.

  build-sys: Makefile.am: remove unnecessary $(top_builddir) (0059ecd)

masatake:
Removing `$(top_builddir)` is really o.k.? 
Though any CI environment doesn't report a failure, when cross-compiling ctags,`$(top_builddir)` may be used. 

@hirooih, if you are sure what you did, make a pull request only for 0059ecd, and ask @leleliu008 to review. I am not a good person for reviewing 0059ecd.

leleliu008:
I don't encourage you to do so although removing `$(top_builddir)` is ok. I think`$(top_builddir)` is more readable.

masatake:
@leleliu008, I would like to know the background of "more readable."

My understanding is that `$(top_builddir)` can be a hint for readers that it is something to do with cross-compiling.
Am I correct?

leleliu008:
@masatake No, `$(top_builddir)` is nothing to do with `cross-compiling`. Actually `$(top_builddir)` is top level Makefile's dir. In our case it can be replace by `./` but I don't encourage someone to do so.

masatake:
Is `./` meaningful when building ctags out of the source tree like:
```
$ cd /tmp
$ mkdir build
$ cd build
$ ~/devel/ctags/configure
$ make
```
?

leleliu008:
> I don't encourage you to do so although removing `$(top_builddir)` is ok. I think`$(top_builddir)` is more readable.

It is my subjective opinion, maybe someone think replacing `$(top_builddir)` with `./` is better. I just give my suggestion.

leleliu008:
> Is `./` meaningful when building ctags out of the source tree like:
> 
> ```
> $ cd /tmp
> $ mkdir build
> $ cd build
> $ ~/devel/ctags/configure
> $ make
> ```
> 
> ?

Sorry, I don't understand what do you wanna say?

leleliu008:
according to logical, [0059ecd](https://github.com/universal-ctags/ctags/commit/0059ecd4f27994606b3599fc7d7b233879013ebc) is ok.

masatake:
> Actually $(top_builddir) is top level Makefile's dir. In our case it can be replace by ./ but I don't encourage someone to do so.

When or how the `./` is used or referenced in our building process?

k-takata:
> build-sys: rename PACKCC_SRCS and PACKCC_OBJS to PACKCC_SRCS and PACKCC_OBJS, respectively. ([bbdad51](https://github.com/universal-ctags/ctags/commit/bbdad5111086321724fce8865f05fcf62babccf0))

It looks that the commit message is wrong. It should be:
build-sys: rename PACKCC_SRCS and PACKCC_OBJS to PACKCC_SRC and PACKCC_OBJ, respectively.

leleliu008:
In top level Makefile, './' is relative to itself. `$(top_builddir)` is top level Makefile's dir. they are same thing in some cases. `$(top_builddir)` is absolute path and './' is relative path

masatake:
@leleliu008, I see. Thank you.

k-takata:
I thought `$(top_srcdir)` and `$(top_builddir)` were used to support out-of-source build, but I'm not sure.
It would be nice if the CI also checks out-of-source build (if currently it is not checked).

leleliu008:
Although our CI don't check out-of-source build, my `ndk-pkg` CI have checked. [reference](https://github.com/leleliu008/ndk-pkg-formula-repository/runs/3307914735?check_suite_focus=true)

leleliu008:
@masatake how about add a `out-of-source-build.yml` GitHub Actions workflow?

hirooih:
All star casts are here!
Thank you for your comments.

Sorry for my following long comments.  But I hope it helps you.

@k-takata san,

> It looks that the commit message is wrong. It should be:
> build-sys: rename PACKCC_SRCS and PACKCC_OBJS to PACKCC_SRC and PACKCC_OBJ, respectively.

Thanks.  I've fixed.

@k-takata san,

> It would be nice if the CI also checks out-of-source build (if currently it is not checked).

@leleliu008  san,

> Although our CI don't check out-of-source build, my ndk-pkg CI have checked. reference

Our CI already includes `make distcheck` which checks "the out-of-source build".

Here is the out-line of `make distcheck`.

```shell
create universal-ctags-5.9.0.tar.gz
tar -xzf universal-ctags-5.9.0.tar.gz
cd universal-ctags-5.9.0
mkdir _build/sub
cd _build/sub
../../configure
make all
make check
make distclean && check all files are removed.
```

***

@masatake san,

> @hirooih, if you are sure what you did, make a pull request only for 0059ecd, and ask @leleliu008 to review. I am not a good person for reviewing 0059ecd.

The day has finally come when we can teach you (my mentor) something :-)

From [Autoconf manual, 4.8.1 Preset Output Variables](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.70/autoconf.html#Preset-Output-Variables);

- Variable: `builddir`
  - Rigorously equal to ‘.’. Added for symmetry only.
- Variable: `top_builddir`
  - The relative name of the top level of the current build tree. In the top-level directory, this is the same as `builddir`.
- Variable: `srcdir`
  - The name of the directory that contains the source code for that makefile.
- Variable: `top_srcdir`
  - The name of the top-level source code directory for the package. In the top-level directory, this is the same as `srcdir`.

The `Makefile.am` we are discussing is in the top-level directory. `top_builddir` is equal to `.`.  (The main topic of this thread)
`top_builddir` and `top_srcdir` are for Makefiles in subdirectories.
We don't have to use `builddir`.  It is _"Added for symmetry only"_.

If we `make` at the top-level directory (`./configure`), `top_srcdir` is equal to `top_builddir`. 

For the `make distcheck` case above,
- `pwd`: `./universal-ctags-5.9.0/_build/sub`
- `top_srcdir`: `../..`

Note that $(top_builddir) may not be an absolute path.  If we need an absolute path, we have $(abs_top_builddir).

> $(top_builddir) is absolute path and './' is relative path

***

We sometimes have to use `srcdir`.  For example;

```makefile
$(PACKCC): $(srcdir)/$(PACKCC_SRC)
	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD)  ... -o $@ $(srcdir)/$(PACKCC_SRC)
```

If we only use GNU Make, this can be;

```makefile
$(PACKCC): $(PACKCC_SRC)
	$(cc4b_verbose)$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD)  ... -o $@ $^
```

`VPATH` is set to `top_srcdir` (`../..`).
make cannot find `$(PACKCC_SRC)` (`misc/packcc/src/packcc.c`), it adds paths in `VPATH`, then  `../../misc/packcc/src/packcc.c` is found. `$^` is replaced by `../../misc/packcc/src/packcc.c`.

***

@leleliu008 san,

If we start using `builddir` or `top_builddir` for the cases in 0059ecd, we can use them everywhere.  I think we need any reasoning when we use them.

For example the following line in `Makefile.am` tests if it is the out-of-source build or not.

```
clean-local:
	@if test "$(top_srcdir)" != "$(top_builddir)"; then \
```

> > I don't encourage you to do so although removing $(top_builddir) is ok. I think$(top_builddir) is more readable.
> It is my subjective opinion, maybe someone think replacing $(top_builddir) with ./ is better. I just give my suggestion.

Thanks you.  An I hope I could let you think replacing `$(top_builddir)` with `./` is better.

hirooih:
> Removing `$(top_builddir)` is really o.k.?
> Though any CI environment doesn't report a failure, when cross-compiling ctags,`$(top_builddir)` may be used.
> 
> @hirooih, if you are sure what you did, make a pull request only for [0059ecd](https://github.com/universal-ctags/ctags/commit/0059ecd4f27994606b3599fc7d7b233879013ebc), and ask @leleliu008 to review. I am not a good person for reviewing [0059ecd](https://github.com/universal-ctags/ctags/commit/0059ecd4f27994606b3599fc7d7b233879013ebc).

@masatake -san,
I think @leleliu008 already reviewed the change, and agreed *almost*. However I'd like to make a pull request for 0059ecd and some more related fixes.

I've pushed this PR without 0059ecd. Please approve this if you see any other issues.

hirooih:
FYI
fedora33_gmake_roundtrip sometimes fails by timeout. After rerun it passes.

masatake:
> FYI
> fedora33_gmake_roundtrip sometimes fails by timeout. After rerun it passes.

I found the reason. I will fix this.

masatake (approved):

PR #3125, opened 2021-08-19
Add statx to seccomp rules

athos-ribeiro:
Add statx to seccomp rules as a fallback for newfstatat. This is
specially important for the case where the newfstatat system call is not
available on a platform using glibc >= 2.33

This patch was created as an effort to fix FTBFS builds on Ubuntu after the glibc 2.33 migration. See https://bugs.launchpad.net/ubuntu/+source/universal-ctags/+bug/1934829 for further reference.


masatake (approved):

masatake:
Thank you.

PR #3126, opened 2021-08-19
TEMP,ci: enable fedora34 as a CI target

masatake:


masatake:
We can switch from fedora 33 to fedora 34!

PR #3127, opened 2021-08-19
circleci: use fedora34

masatake:
But `fedora33_cross_aarch64` is kept as is for now.

### DIFF
--- a/.github/workflows/testing-gnulinux.yml
+++ b/.github/workflows/testing-gnulinux.yml
@@ -28,6 +28,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: update package information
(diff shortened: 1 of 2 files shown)


masatake:
The coverage decreased.

The reason may be related to the output: 
```
Scanning . for .gcda files ...
Found 199 data files in .
Processing optlib/libctags_a-RSpec.gcda
/home/runner/work/ctags/ctags/ubuntu-20.04-gcc/optlib/libctags_a-RSpec.gcda:stamp mismatch with notes file
...
```

https://github.com/gcovr/gcovr/issues/314


The coverage for C files translated by optlib2c is unimportant because the source files just define the regular expression and register to the main part.

masatake:
After rebasing on the HEAD of the master branch, the issue about coverage is gone.
I wonder what happened...

PR #3128, opened 2021-08-19
build-sys: generate win32/ctags_vs2013.vcxproj* from source.mak

hirooih:
temporary push to test on CI env.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,8 @@ EXTRA_DIST   = README.md autogen.sh \
 	       win32/ctags.rc win32/ctags.exe.manifest win32/resource.h \
 	       win32/ctags_vs2013.sln \
 	       win32/ctags_vs2013.vcxproj win32/ctags_vs2013.vcxproj.filters \
+	       win32/ctags_vs2013.vcxproj.in win32/ctags_vs2013.vcxproj.filters.in \
+	       win32/Makefile \
 	       win32/gen-repoinfo.bat \
 	       misc/txt2cstr \
 	       $(PEG_INPUT) $(OPTLIB2C_INPUT) $(TXT2CSTR_INPUT) \
--- a/win32/Makefile
+++ b/win32/Makefile
@@ -1,0 +1,87 @@
+#
+# Copyright (c) 2021, Hiroo Hayashi
+#
+# This source code is released for free distribution under the terms of the
+# GNU General Public License version 2 or (at your option) any later version.
+#
+# Makefile to generate ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters
+#
+#   usage: make [-B]
+#
+# Restrictions:
+#   - Input Files: ctags_vs2013.vcxproj.in and ctags_vs2013.vcxproj.filters.in
+#     - The last charactor of the files must be '>'.
+#       cf. check_eof_chars_in_vcxproj() in misc/src-check.
+#     - Other lines must be end with LF.
+#   - GNU make is required.
+
+VCXPROJ = ctags_vs2013.vcxproj
+VCXPROJ_FILTERS = ctags_vs2013.vcxproj.filters
+SOURCE_MAK = ../source.mak
+
+all: $(VCXPROJ) $(VCXPROJ_FILTERS)
+
+include $(SOURCE_MAK)
+
+# exclude some files for Win32 and replace a slash (/) to a backslash (\)
+MVC_SRCS = $(MVC_GNULIB_SRCS) $(CMDLINE_SRCS) $(LIB_SRCS) $(OPTLIB2C_SRCS) $(PARSER_SRCS) $(OPTSCRIPT_DSL_SRCS) $(DEBUG_SRCS) $(WIN32_SRCS)
+MVC_SRCS_EXCLUDE = main/mbcs.c main/seccomp.c main/trace.c
+MVC_SRCS_CONV = $(sort $(subst /,\\,$(filter-out $(MVC_SRCS_EXCLUDE),$(MVC_SRCS))))
+
+MVC_HEADS = $(MVC_GNULIB_HEADS) $(CMDLINE_HEADS) $(LIB_HEADS) $(OPTLIB2C_HEADS) $(PARSER_HEADS) $(OPTSCRIPT_DSL_HEADS) $(DEBUG_HEADS)  $(WIN32_HEADS)
+MVC_HEADS_EXCLUDE = main/interactive_p.h main/mbcs.h main/mbcs_p.h main/trace.h
+MVC_HEADS_CONV = $(sort $(subst /,\\,$(filter-out $(MVC_HEADS_EXCLUDE),$(MVC_HEADS))))
+
+MVC_INC_DIRS1 = ..;../main;../gnulib;../parsers;../parsers/cxx;../dsl;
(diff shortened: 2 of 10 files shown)


hirooih:
@k-takata san,

Could you review the before Masatake-san, since this is related the win32 build which you developed.

I committed ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters in 4 separate commits as follows to see the differences easier. I will marge them after your reviews.

- build-sys: ctags_vs2013.vcxproj*: sort 4165772
  - just sort
- build-sys: ctags_vs2013.vcxproj*: fix typos and missing files a8383c1
  - fix typos in file names and add missing files
- build-sys: ctags_vs2013.vcxproj*: use lower-case canonically 8b706e6
  - only directory name `main` and `parser` were capitalized. Use lower-case for them, too
- build-sys: ctags_vs2013.vcxproj*: generated version 9dc3d48
  - Delete the warning level condition for regex.C
  - mkstemp.c uses the path name format same as others.

***

I have two questions about ctags_vs2013.vcxproj*.

Q1. Why the mistakes in the second change above does not cause error on CI-test, appveyor.yml:msbuild?

Q2. The function check_eof_chars_in_vcxproj() in misc/src-check checks if ctags_vs2013.vcxproj* do not end with newline.
The CI-test passes even when it end with newline on this check, too.
Do we really need this restriction and the test?

I'm afraid that the CI-test might not check the ctags_vs2013.vcxproj*.

k-takata:
> Q1. Why the mistakes in the second change above does not cause error on CI-test, appveyor.yml:msbuild?

Hmm, not sure. Maybe an error is shown when someone tries to open a misspelled header file in Visual Studio, but it may not cause an error when compiling because the correct header file actually exists.

> The CI-test passes even when it end with newline on this check, too.

That might be a bug of the script.
The reason why we wanted to check that the vcxproj files don't end with CRLF is that when someone modifies the vcxproj files with Visual Studio, VS automatically removes the last CRLF.
Now, the vcxproj files are not intended to be edited by VS, we can remove the check of the last CRLF.
The `.editorconfig` file should be also updated accordingly.

hirooih:
> Maybe an error is shown when someone tries to open a misspelled header file in Visual Studio, but it may not cause an error when compiling because the correct header file actually exists.

That's makes sense. But it means that the CI-test does not check the ctags_vs2013.vcxproj*.

> That might be a bug of the script.

Does "the script" mean `misc/src-check`?
I don't think it is a bug of `misc/src-check`.

> The reason why we wanted to check that the vcxproj files don't end with CRLF is that when someone modifies the vcxproj files with Visual Studio, VS automatically removes the last CRLF.
> Now, the vcxproj files are not intended to be edited by VS, we can remove the check of the last CRLF.

As I added in contributions.rst as follows, using the Makefile proposed by this PR is still optional.

> If you have GNU make, ``make -C win32`` updates the win32 files described above from ``makefile.mak``.

I think that real users of the vcxproj files cannot use GNU Make nor sed. They still can edit sources.mak and vcxproj files by hand.
The test in `misc/src-check` still valid.

> The .editorconfig file should be also updated accordingly.

Now I understand why my VS code changes the end-of-line chars of vcxproj files. I've been looking for the cause in Git documents, etc.
I've added the following lines to .editorconfig.

```
[win32/*.vcxproj*.in]
end_of_line = lf
```

win32/Makefile
```diff
@@ -0,0 +1,80 @@
+#
```
masatake, line 1: Could you put your copyright notice here?

masatake:
run units target on Cygwin / testing (pull_request):

> ERROR: The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: The process cannot access the file because it is being used by another process.

I think the error is nothing to do with the commits in this pull request; the error is reported in a very early stage of testing.

win32/Makefile
hirooih: I've added the copyright notice.
I also update the comment in this header.

hirooih:
Sorry, I did not update this PR.

> I think the error is nothing to do with the commits in this pull request; the error is reported in a very early stage of testing.

Yes, the error was fixed by #3137.
This push adding the copyright notice should be clean.

@k-takata san,
Do you see any issues after reading my comment, https://github.com/universal-ctags/ctags/pull/3128#issuecomment-903875843?

hirooih:
>> I think the error is nothing to do with the commits in this pull request; the error is reported in a very early stage of testing.

> Yes, the error was fixed by #3137.

It seems that we have other issue.  I agree that "the error is nothing to do with the commits in this pull request".

```
...
ranlib libctags.a
rm: cannot remove 'ctags.exe': Device or resource busy
make[2]: *** [Makefile:2648: ctags.exe] Error 1
make[2]: Leaving directory '/cygdrive/d/a/ctags/ctags'
make[1]: *** [Makefile:5771: check-recursive] Error 1
make[1]: Leaving directory '/cygdrive/d/a/ctags/ctags'
make: *** [Makefile:6065: check] Error 2
```

parsers/verilog.c
```diff
@@ -420,25 +420,49 @@ static short hasSimplePortList (verilogKind kind)
...
 }
 
+static tokenInfo *dupToken (tokenInfo *token)
+{
+	tokenInfo *dup = newToken ();
+	vStringDelete (dup->name);
+	vStringDelete (dup->blockName);
+	vStringDelete (dup->inheritance);
+	*dup = *token;
+	dup->name = vStringNew ();
```
masatake: The three fields are deleted, then allocated. Smaller overhead but how about using the following code instead?
```
tokenInfo backup = *dup
*dup = *token;
dup->name = backup.name;
dup->blockName = backup.blockName;
dup->inheritance = backup.inheritance;
// ...copy the contents.
```

masatake: ??? I'm confused. Is this change is needed for generating a vcxproj file?

masatake:
> It seems that we have other issue. I agree that "the error is nothing to do with the commits in this pull request".

But, this is not reproduced pull requests I'm working on these days.
How about remaking the branch based on the latest master branch?

parsers/verilog.c
hirooih: Sorry for your confusion.

Last week I setup a native Linux box instead of Windows 10 WSL Ubuntu I have been using. I pushed the PR from it.
It seems that I made a mistake.

hirooih:
I've force-pushed from the original machine.
Everything looks good.

masatake (approved):

Makefile.am
```diff
@@ -32,6 +32,8 @@ EXTRA_DIST   = README.md autogen.sh \
 	       win32/ctags.rc win32/ctags.exe.manifest win32/resource.h \
 	       win32/ctags_vs2013.sln \
 	       win32/ctags_vs2013.vcxproj win32/ctags_vs2013.vcxproj.filters \
+	       win32/ctags_vs2013.vcxproj.in win32/ctags_vs2013.vcxproj.filters.in \
+		   win32/Makefile \
```
k-takata, line 36: It looks the indent is different from the other lines.

hirooih: I've fixed it.
Thank you.

k-takata (approved):

hirooih:
Thank you for your views.

PR #3130, opened 2021-08-20
build-sys: builddir fixes

hirooih:
This PR came from #3119.  See https://github.com/universal-ctags/ctags/pull/3119#issuecomment-899565667 for more details.

`misc/testing.mak` has a related issue.  A shell variable `builddir` holds the output of `pwd` command.
But in this case the full pathnames are not required.  The relative paths are enough.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,7 +65,7 @@ if HAVE_STRNLEN_FOR_BUILD
 PACKCC_CPPFLAGS_FOR_BUILD += -DPCC_USE_SYSTEM_STRNLEN
 endif
 
-PACKCC = $(top_builddir)/packcc$(BUILD_EXEEXT)
+PACKCC = packcc$(BUILD_EXEEXT)
 CLEANFILES += $(PACKCC)
 
 cc4b_verbose = $(cc4b_verbose_@AM_V@)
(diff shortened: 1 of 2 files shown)


masatake:
Is this a bug fix?
I think keeping the redundant variables can be hints for understanding our build process.

masatake:
If `misc/units` does `cd`, the directory passed as the last argument (e.g. `./Tmain`) can be solved wrongly.
Passing the directory as an absolute path on the caller side like `$$(pwd)/Tmain` can avoid the risk.

I agree that my code, the makefile, and the `units` script are written in a redundant way.
I cannot remember the detail. However, the redundant code like `$$(pwd)/Tmain` makes me remember what kind of potential bugs I was afraid of.

leleliu008:
I suggest that you spend your time doing real contributions such as fixing issues or improving performance, instead of wasting time on a "refactoring witch hunt" with AFAICT no clear goals in mind.

Don't change code unless absolutely necessary. Which means code refactoring is necessary only when potential bugs found or performance, readability are improved obviously.

masatake:
> But for command name of a shell program, foo and ./foo is not equivalent. I did not know the detail implementation of units.py and unit script, it might have similar "search path" feature.

In that case, I wonder changing from 
```
builddir=$$(pwd)
...
$${PROG} $${c} $(srcdir)/Units $${builddir}/Units
```
to
```
$${PROG} $${c} $(srcdir)/Units ./Units
```
is safe enough or not.

Anyway,  I will close this. As you wrote the original code has redundancy but the redundancy has helped me understand the code.

If the changes to this pull request are needed in your other pull requests, please, reopen this.

Issue #3131, opened 2021-08-21
Question: how to update files under docs/man?


masatake:
@hirooih, when I update foo.rst.in file under man directory, what I should do for updating docs/man/foo.rst ?

As far as reading docs/README.md, just `make` is enough.
However, it doesn't work in my environment.

For getting what I want, I have to do `make -C man update-docs`.

hirooih:
I've sent PR #3133 .

***

> For getting what I want, I have to do make -C man update-docs.

Files `docs/man/*`  are used only by Sphinx documents.  `make -C docs html`, etc. generates them.
And `make check-genfile` invoked by `make check`, etc. checks if any of  `docs/man/*` must be committed.

I also thought about modifying the `docs/README.md`, but I chose to fix `Makefile.am`.  I think it is better to show developer that any of  `docs/man/*` is updated.

masatake:
> I've sent PR #3133 .

I see. Thank you.

PR #3133, opened 2021-08-21
build-sys: man: add update-docs to default target

hirooih:
clean-docs is also invoked by make clean
fix for #3131


masatake (approved):

masatake:
I need this now.
So I will merge this.

PR #3134, opened 2021-08-26
build-sys: .editerconfig update

hirooih:
update rule for the GNU source codes and Makefiles.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,7 +19,7 @@ indent_size = 4
 indent_style = space
 trim_trailing_whitespace = true
 
-[*.mak]
+[{Makefile*,*.mak}]
 indent_size = 8
 indent_style = tab
 trim_trailing_whitespace = true


masatake:
@hirooih, thank you.

Could you try the following patch?
If everything works well, GDB reports the backtrace for the crash.

```
diff --git a/.github/workflows/testing-cygwin.yml b/.github/workflows/testing-cygwin.yml
index bd06ae0f..b8b4735b 100644
--- a/.github/workflows/testing-cygwin.yml
+++ b/.github/workflows/testing-cygwin.yml
@@ -15,7 +15,7 @@ jobs:
         shell: C:\tools\cygwin\bin\bash.exe -l -i {0}
 
     steps:
-    - run: choco install -y --source cygwin gcc-g++ make automake autoconf pkg-config dos2unix libiconv-devel libjansson-devel libxml2-devel libyaml-devel
+    - run: choco install -y --source cygwin gcc-g++ make automake autoconf pkg-config dos2unix libiconv-devel libjansson-devel libxml2-devel libyaml-devel gdb
       shell: pwsh
     
     - run: git config --global core.autocrlf input
@@ -26,8 +26,8 @@ jobs:
     - run: printf 'cd %s' "$(cygpath '${{ github.workspace }}')" >> ~/.bashrc
 
     - run: ./autogen.sh
-    - run: ./configure --prefix=/usr
-    - run: make V=1
+    - run: ./configure --prefix=/usr --enable-debugging
+    - run: make V=1 AM_CFLAGS='-O0 -g'
     - run: make install
     - run: file /usr/bin/ctags
     - run: /usr/bin/ctags --version
```

hirooih:
@masatake san,  Thank you.

I opened #3137.

I updated my local Cygwin environment which I have not been using for a few years, and I could reproduce the fail on it.
The backtrace command does not work because stack is broken, but I find it causes a segmentation fault in a pthread function called by regcomp.

I will continue on #3137.

masatake:
> The backtrace command does not work because stack is broken, but I find it causes a segmentation fault in a pthread function called by regcomp.

Oh, I see. It is understandable.

Issue #3137, opened 2021-08-28
CI test on Cygwin fails


hirooih:
During working on #3128 CI test on Cygwin started failing causing segmentation faults.
It also fails on #3134 which changed only `.editorconfig`.

I've updated my local Cygwin environment and reproduce the fail.

masatake:
Is Valgrind available on cygwin?

masatake:
What I really don't want to know is why regcomp needs pthread:-O.

hirooih:
I guess you meant;

> What I really want to know is why regcomp needs pthread:-O.

Me, too!

masatake:
I don't want to know it. It is o.k. for me as far as regcomp works fine.
If regcomp doesn't work, we have to know it. So I don't want to know if possible.
However, the most of all computer programs are broken. So eventually we have to know it whether we want to know it or not.

hirooih:
I see.
Your comment reminds me the similar fail https://github.com/universal-ctags/ctags/pull/3069#issuecomment-861916353.
Similar fix works.

I guess a change related to weak-reference was made in Cygwin upstream.

I've sent PR #3138.

Thanks.

hirooih:
fixed by #3138.

PR #3138, opened 2021-08-28
cygwin: disable weak reference for pthread library

hirooih:
Fix for #3137 preventing glthread_lock_init() from causing a segmentation fault.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -510,6 +510,9 @@ case "$host" in
 	AC_DEFINE(MANUAL_GLOBBING)
 	AC_DEFINE(HAVE__FINDFIRST)
 	;;
+  *-cygwin*)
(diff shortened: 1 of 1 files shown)


masatake (approved):

masatake:
Thank you for taking time for the issue.
I'm not sure this is an ideal solution. However, we need a solution now. Let's merge this.

hirooih:
> I'm not sure this is an ideal solution.

I should mention to @k-takata san. He taught us the original fix. on https://github.com/universal-ctags/ctags/pull/3069#issuecomment-861987951.

>However, we need a solution now. Let's merge this.

Sure.

k-takata:
I think this is correct.

hirooih:
> I think this is correct.

Thanks!

PR #3141, opened 2021-08-28
ttcn: Skip comments correctly

karolba:
Right now comments don't get skipped all the way, leading to code like

    type enumerated Enum {
        /* a comment */
        Value
    }

or

    type enumerated Enum {
        // a comment
        Value
    }

resulting in "Value" not being present in the output, due to `getNonWhiteSpaceChar()` returning `'*'` (for `/* */`comments) or `'\n'` (for `//` comments)


masatake:
@karolba, thank you.

I have a comment about your usage of git.
You used your `master` branch for making a pull request.
However, I recommend you make a topic branch for each pull request. So you can keep your master branch clean.
See https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows for more details.

I wrote this because I expect more contributions from you:-).
Thank you again.

PR #3145, opened 2021-09-02
lua: don't include module in function names

### LINKED ISSUE #3144: Lua tags always include qualifier
The name of the parser:

Lua

```
$ ctags --options=NONE --languages=lua -o - test.lua
```

The content of input file:

```lua
local M = {}

function M.say_hello()
  print("Hello world")
end

return M

```

```
M.say_hello	test.lua	/^function M.say_hello()$/;"	f
```

The tags output you expect:

```
say_hello	test.lua	/^function M.say_hello()$/;"	f
```

The version of ctags:

```
$ ctags --version
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +json, +interactive, +case-insensitive-filenames, +packcc
```

How do you get ctags binary:

From nixpkgs.

---

The problem with the above is that when I require my test module in another Lua file:

```lua
local test = require('test')

test.say_hello()
```

The `say_hello` function in `test.lua` is not found. Adding functions to a table in Lua and then returning the table at the end of the module is a common pattern in Lua, so the table name itself should not be included (unless `--extras=+q` is used).

Note this is a different issue than https://github.com/universal-ctags/ctags/issues/1798, which is kind of asking for the opposite (ironically).

gpanders:
Closes #3144.

### DIFF
--- a/parsers/lua.c
+++ b/parsers/lua.c
@@ -60,7 +60,7 @@ static bool is_a_code_line (const unsigned char *line)
 
 static bool isLuaIdentifier (char c)
 {
-	return (bool) !(isspace(c)  || c == '(' || c == ')' || c == '=');
+	return (bool) !(isspace(c)  || c == '(' || c == ')' || c == '=' || c == '.' || c == ':');
 }
 
 static void extract_next_token (const char *begin, const char *end_sentinel, vString *name)
@@ -92,7 +92,12 @@ static void extract_next_token (const char *begin, const char *end_sentinel, vSt
 
 	for (const char *c = begin; c <= end; ++c)
 	{
-		if (isLuaIdentifier (*c))
+		if (*c == '.' || *c == ':')
+		{
+			/* Do not include module names in function name */
+			vStringClear (name);
+		}
+		else if (isLuaIdentifier (*c))
 			vStringPut (name, (int) *c);
 		else
 		{
--- a/Units/parser-lua.r/simple.lua.d/expected.tags
+++ b/Units/parser-lua.r/simple.lua.d/expected.tags
@@ -1,3 +1,3 @@
-test.i_123	input.lua	/^ test.i_123 = function  (x)$/;"	f
-test.me_12a	input.lua	/^function test.me_12a(one, two)$/;"	f
-test.me_12b	input.lua	/^function test.me_12b (one, two)$/;"	f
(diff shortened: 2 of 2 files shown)


parsers/lua.c
```diff
@@ -60,7 +60,7 @@ static bool is_a_code_line (const unsigned char *line)
 
 static bool isLuaIdentifier (char c)
 {
-	return (bool) !(isspace(c)  || c == '(' || c == ')' || c == '=');
+	return (bool) !(isspace(c)  || c == '(' || c == ')' || c == '=' || c == '.' || c == ':');
```
masatake, line 63: Do we need this?

isLuaIdentifier() is used not only in extract_next_token() but also in extract_prev_toke().
I wonder whether the change for isLuaIdentifier() has an impact on extract_prev_toke() or not.

If you are not sure, keep isLuaIdentifier() as is.
If you are sure, could you write your conviction to the commit log.

parsers/lua.c
```diff
@@ -92,7 +92,10 @@ static void extract_next_token (const char *begin, const char *end_sentinel, vSt
 
 	for (const char *c = begin; c <= end; ++c)
 	{
-		if (isLuaIdentifier (*c))
+		if (*c == '.' || *c == ':') {
```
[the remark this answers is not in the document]
gpanders: My mistake, I'll update this.

parsers/lua.c
gpanders: It is needed, since `extract_prev_token()` moves backward as long as `isLuaIdentifier()` returns true. So we add the cases for `'.'` and `':'` to have it stop as soon as it reaches one of those characters.

masatake: I see. Thank you for your explanation.

gpanders: Updated the commit message.

masatake (approved):

masatake:
@gpanders, please allow me to take over this pull request.

I extend your pull request to support the old use case.
See #3146 and https://github.com/universal-ctags/ctags/pull/3146/commits/ea9b5e2624bae39af5b5dcd035396227e409714c .

masatake:
I will merge your change via #3146.

gpanders:
Sounds good, thanks!

PR #3146, opened 2021-09-02
Lua: don't include module in function names but attach it as scope fields

### LINKED ISSUE #3144: Lua tags always include qualifier

masatake:
This extends #3145, @gpanders's pull request.
Closes #3144.


masatake:
After reviewing the newly added parsing code by myself, I will merge this.
I have to give more broken output for testing the code.

PR #3149, opened 2021-09-03
ITcl,TclOO: introduce forceuse parameters

### LINKED ISSUE #3148: Itcl,TclOO: support none-self-contained script
Both Itcl and TclOO parsers expect the namespace is imported explicitly for the packages.
However, there are cases the namespace is imported externally.

----

>  If you want to provide that functionality as a command line (forcing Itcl parsing when parsing Tcl), I'm open to that too.

I wonder when we want to provide the extended command line interface.

My guessing:

file1.tcl:
```
package require Itcl
namespace import itcl::*
source file2.tcl
```

file2.tcl:
```
class MyClass { .... }
```

In this case, do you expect MyClass is tagged by the ITcl parser?
In the other words, file2.tcl is a valid ITcl program file?

_Originally posted by @masatake in https://github.com/universal-ctags/ctags/discussions/3136#discussioncomment-1248103_

masatake:
Close #3148.

### DIFF
--- a/Tmain/list-params.d/stdout-expected.txt
+++ b/Tmain/list-params.d/stdout-expected.txt
@@ -1,10 +1,11 @@
 # ALL
-#LANGUAGE      NAME    DESCRIPTION
-CPreProcessor  _expand expand macros if their definitions are in the current C/C++/CUDA input file (true or [false])
-CPreProcessor  define  define replacement for an identifier (name(params,...)=definition)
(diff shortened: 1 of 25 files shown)


masatake:
@bigfaceworm, if you are good at English (I'm not), could you review 
ff17464  in this pull request?

masatake:
https://github.com/universal-ctags/ctags/blob/master/docs/parser-tcl.rst should be merged to the new man page.

PR #3155, opened 2021-09-10
Small fixes

rootkea:
Found by: [scan-build](https://clang-analyzer.llvm.org/scan-build)

### DIFF
--- a/main/field.c
+++ b/main/field.c
@@ -824,30 +824,20 @@ static bool defaultDoesContainAnyChar (const tagEntryInfo *const tag CTAGS_ATTR_
 extern bool  doesFieldHaveTabOrNewlineChar (fieldType type, const tagEntryInfo *tag, int index)
 {
 	fieldObject *fobj = fieldObjects + type;
-	const char *value;
 	bool (* doesContainAnyChar) (const tagEntryInfo *const, const char*, const char*) = fobj->def->doesContainAnyChar;
 
 	Assert (tag);
-	Assert (index == NO_PARSER_FIELD || ((unsigned int)index) < tag->usedParserFields);
+	Assert (index == NO_PARSER_FIELD || (index >= 0 && ((unsigned int)index) < tag->usedParserFields));
+
+	if (index == NO_PARSER_FIELD)
+		return false;
 
 	if (doesContainAnyChar == NULL)
-	{
-		if (index == NO_PARSER_FIELD)
-			return false;
-		else
-			doesContainAnyChar = defaultDoesContainAnyChar;
-	}
+		doesContainAnyChar = defaultDoesContainAnyChar;
 
-	if (index >= 0)
-	{
-		const tagField *f = getParserFieldForIndex (tag, index);
+	const tagField *f = getParserFieldForIndex (tag, index);
 
-		value = f->value;
-	}
-	else
-		value = NULL;
-
-	return (* doesContainAnyChar) (tag, value, "\t\n");
+	return (* doesContainAnyChar) (tag, f->value, "\t\n");
 }
 
 /*  Writes "line", stripping leading and duplicate white space.
--- a/configure.ac
+++ b/configure.ac
@@ -350,7 +350,7 @@ AC_PROG_LN_S
 AC_CHECK_PROG(STRIP, strip, strip, :)
 AC_SYS_LARGEFILE
 
-AC_CHECK_PROGS([perl_found], [perl], [yes], [no])
+AC_CHECK_PROG([perl_found], [perl], [yes], [no])
 AM_CONDITIONAL([RUN_OPTLIB2C], [test "${perl_found}" = "yes"])
 AM_CONDITIONAL([RUN_TXT2CSTR], [test "${perl_found}" = "yes"])
 
--- a/main/args.c
+++ b/main/args.c
@@ -145,8 +145,8 @@ static char* nextFileLine (FILE* const fp)
 					ungetc (c, fp);
 			}
 			vStringStripTrailing (vs);
-			result = xMalloc (vStringLength (vs) + 1, char);
 			vStringStripLeading (vs);
+			result = xMalloc (vStringLength (vs) + 1, char);
 			strcpy (result, vStringValue (vs));
 		}
 		vStringDelete (vs);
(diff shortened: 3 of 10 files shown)


masatake:
Thank you.
Could you make a pull request only for 9e65c42?
In that case, could you use 'Tex: ' as prefix the commit header ? 

```
Tex: add missing break statement
```

I would like to merge 9e65c42 first.

rootkea:
I just opened #3156 fixing missing `break.`

masatake:
Could you make a pull request for 27ec7c8  and send it to https://github.com/universal-ctags/libreadtags/pulls ?
ctags just uses the library developed at the repository.

rootkea:
Opened https://github.com/universal-ctags/libreadtags/pull/34 for `NULL` check on `file` before accessing `file->err`

masatake:
I will cherry-pick the commits from this pull request incrementally.

rootkea:
> I will cherry-pick the commits from this pull request incrementally.

Or if you want me to open separate PRs for each commit then I can also do that. :)

rootkea:
I have a build question:

Why don't I see `#define DEBUG` in config.h even when I supply `--enable-debugging` to ./configure?

Looks like we're missing `AC_DEFINE(ENABLE_DEBUG)` in configure.ac?

masatake:
> Why don't I see #define DEBUG 1 in config.h even when I supply --enable-debugging to ./configure?
> Looks like we're missing AC_DEFINE(DEBUG) in configure.ac?

I don't know or I cannot remember.
Maybe there is a historical reason.

Universal Ctags (u-ctags)  was derived from Exuberant Ctags (e-ctags).

e-ctags has its own debug infrastructure.
When I started working on u-ctags, I didn't know and study the debug-infra.
The debug infra was related to C language parser of e-ctags.
u-ctags introduced new C/C++ parsers. So I didn't find a strong reason to study the debug-infra.

On the other hand, u-ctags introduced a new debug-infra called "tracing".
It is a spin-off of the new  C/C++ parsers.
`./configure --enable-debuggin` enables the tracing feature.

And now we are here. I often want to revise the debug-infra. However, I didn't.

rootkea:
@masatake Thanks for the reply!

> Why don't I see `#define DEBUG 1` in `config.h` even when I supply `--enable-debugging` to `./configure`?

Answering my own question: We pass `-DDEBUG` instead of defining DEBUG in config.h for `--enable-debugging`.

rootkea:
I'll look into tests failures tomorrow.

masatake:
The failure is reproduced even on my note pc (make tmain).
After removing `main: Refactor defaultDoesContainAnyChar() e08c926 ` from the branch, the failure is gone.

Could you remove the pull request from the this pull request?
So I can merge the "good" changes first.

Then make a separate pull request for  `main: Refactor defaultDoesContainAnyChar() e08c926 ` ?
I will also inspect the reason of the failure. I guess your change exposes a bug.

masatake:
*  main: Refactor defaultDoesContainAnyChar() e08c926

  I have not reviewed yet because it make the test on the CD
  enbvironment failure. I will take time to inspect the failure
  because I think the change exposures pre-existing bug.

* Remove redundant code 752d982

  I wonder how should I deal with this.
  The same/similar change is already proposed in #2280 and #2278.
  I have not taken an action for them because I would like to keep
  dead assignments to make code understandable when extending parsers.

  It is not easy for me to explain why keeping dead assignments helps 
  people understanding code. So I didn't take action.

  I wonder how I should do wit this change, #2280, and #2278.

* main: Allocate the exact amount of memory we need edf5194

  This is obvious improvement. I will cherry-pick this.

* main: Use safer vsnprintf() instead of vsprintf() 5d1fa0f 

  The size is already ensured by calling mem_try_ensure_space in line 842.
  So just using vsprintf is enough.
  I will not merge this change.

* buildsys: Fix check for `perl` 80c50eb

  Using AC_CHECK_PROGS is bug?
  I don't think so. I will not merge this change.

rootkea:
> Using AC_CHECK_PROGS is bug?

Check the [`AC_CHECK_PROGS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html) syntax.

Currently, if you have `perl` on your system the configure.ac fails to detect it.

masatake:
```
[yamato@dev64]~/var/ctags-github% which perl
which perl
/usr/bin/perl
[yamato@dev64]~/var/ctags-github% ./configure 2>&1 | grep perl
./configure 2>&1 | grep perl
checking for perl... no
```
You are correct. I will cherry pick 80c50eb, too. Thank you.

masatake:
edf5194 and 80c50eb are merged.

main/field.c
```diff
@@ -824,30 +824,20 @@ static bool defaultDoesContainAnyChar (const tagEntryInfo *const tag CTAGS_ATTR_
 extern bool  doesFieldHaveTabOrNewlineChar (fieldType type, const tagEntryInfo *tag, int index)
 {
 	fieldObject *fobj = fieldObjects + type;
-	const char *value;
 	bool (* doesContainAnyChar) (const tagEntryInfo *const, const char*, const char*) = fobj->def->doesContainAnyChar;
 
 	Assert (tag);
-	Assert (index == NO_PARSER_FIELD || ((unsigned int)index) < tag->usedParserFields);
+	Assert (index == NO_PARSER_FIELD || (index >= 0 && ((unsigned int)index) < tag->usedParserFields));
```
masatake, line 830: This line makes sense.

main/field.c
```diff
@@ -824,30 +824,20 @@ static bool defaultDoesContainAnyChar (const tagEntryInfo *const tag CTAGS_ATTR_
...
 
-	if (doesContainAnyChar == NULL)
-	{
-		if (index == NO_PARSER_FIELD)
-			return false;
-		else
-			doesContainAnyChar = defaultDoesContainAnyChar;
-	}
+	if (index == NO_PARSER_FIELD)
+		return false;
```
masatake, line 833: Your change doesn't consider the combination of `index == NO_PARSER_FIELD` and `doesContainAnyChar != NULL`.

masatake:
Acceptable changes are already merged to the main tree.
Thank you. So I will close this one in this time.

PR #3156, opened 2021-09-10
Tex: add missing `break` statement

rootkea:
Found by: [scan-build](https://clang-analyzer.llvm.org/scan-build)


masatake:
Thank you.

PR #3162, opened 2021-09-17
GitHub Actions: remove the ubuntu-16.04 environment due to it will be removed 3 days later

leleliu008:
The `ubuntu-16.04` github actions environtment will
 be removed on September 20, 2021. Migrate to `ubuntu18.04` or `ubuntu-latest` instead. For more details see https://github.com/actions/virtual-environments/issues/3287


masatake (approved):
Thank you for taking care about these topics.

PR #3164, opened 2021-09-20
C++ fix: tags generated by --extras=+q now contain 'end' field

fwmechanic:
--extras=+q is supposed to cause an identical copy of a tag to be created
bearing a fully-qualified name, when the fully-qualified name exists (is
different from the standard/default/non-qualified tag name).

Prior to this fix, fully-qualified-name tags generated by --extras=+q
would be missing the end field (generated by --fields=+e) which their
non-qualified counterpart tag would contain.

The end (line number) value is (AFAICT the only tag field value that
is) not known when the non-qualified tag entry is created (which is the
time that the fully-qualified tag entry is created if --extras=+q by
renaming and cloning the non-qualified tag).

Existing code patches the end line number into the non-qualified tag
entry once that value has been ascertained, but would not patch the
corresponding fully-qualified tag entry.  This PR/commit adds patching
of the corresponding fully-qualified tag entry when the non-qualified tag
entry is being patched.


parsers/cxx/cxx_parser.c
```diff
@@ -1505,12 +1515,14 @@ void cxxParserAnalyzeOtherStatement(void)
 			// out its proper scope. Better avoid emitting this one.
 			CXX_DEBUG_PRINT("But it has been preceded by the 'friend' keyword: this is not a real prototype");
 		} else {
-			int piCorkQueueIndex;
-			int iScopesPushed = cxxParserEmitFunctionTags(&oInfo,CXXTagKindPROTOTYPE,CXXEmitFunctionTagsPushScopes, &piCorkQueueIndex);
-			if (piCorkQueueIndex != CORK_NIL)
+			int iCorkQueueIndex, iCorkQueueIndexFQ;
```
masatake, line 1518: I'm not sure whether @pragmaware cares or not but I have a comment about this change.

This change, renaming piCorkQueueIndex to iCorkQueueIndex, is correct but this change is not needed for filling end fields of FQ tags.
So before introducing changes for filling end fields of FQ tags, I would like you to make a small separated commit only for fixing the wrong variable name, and add the commit as "preparation commit" for this pull request.

masatake (commented):
Thank you.
Looks good for me other than one minor topic.

fwmechanic:
BTW I considered adding helper wrapper functions for `cxxParserMarkEndLineForTagInCorkQueue(iCorkQueueIndex)` and `cxxParserSetEndLineForTagInCorkQueue(iCorkQueueIndex,lnum)` which would contain the `if(iCorkQueueIndex > CORK_NIL)` test, so most of those tests could be removed from the point of calls of these functions (making the calling code less noisy and eliminating the possibility that the variables tested vs passed to the call might inadvertently diverge), but hypothesized performance-related objections.  

If this refinement would be acceptable, I would be happy to modify this PR to implement it (perhaps using inline function implementations to mitigate any performance concern?).

masatake:
Could you add "Cxx: " as the prefix for the headers of your commit logs like: `Cxx: correct naming of variable` ?

masatake:
Thank you. Let's see how the Cxx parser maintainer says.

pragmaware:
Sorry for lagging in reviewing this one: I'm currently travelling. I'll take a deeper look once I find the environment to do so.

PR #3175, opened 2021-10-16
Lex parser

masatake:


hirooih:
> @hirooih, your win32 target works perfectly when adding this parser. Thank you very much.

I'm glad to hear that!

PR #2654, opened 2020-10-02
Add julia parser

getzze:
Tested with Geany (https://github.com/geany/geany/pull/2584)

PR #3194, opened 2021-11-20
Add GDScript support

Davidy22:
Tested in geany, heavily borrowed from python

Used #2654 as reference somewhat, although I do notice there's some additional files that may be requested but the test case format looks fairly dense and I'm hoping the output files are generated and someone'll point to how to generate them.


parsers/gdscript.c
```diff
@@ -0,0 +1,784 @@
+/*
+*   Copyright (c) 2000-2003, Darren Hiebert
```
masatake, line 2: You are not Darren Hiebert.

masatake: This year is 2021.

parsers/gdscript.c
```diff
@@ -0,0 +1,784 @@
...
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   This module contains functions for generating tags for GDScript language
+*   files.
```
masatake, line 8: If you know, could add a URL for the GDScript language reference here?

parsers/gdscript.c
```diff
@@ -0,0 +1,784 @@
...
+		if (*cp == '\0')  /* skip blank line */
+			continue;
+
+		/* Skip comment if we are not inside a multi-line string. */
+		if (*cp == '#' && !longStringLiteral)
+			continue;
+
+		/* Deal with line continuation. */
+		if (!line_skip) vStringClear(continuation);
+		vStringPut(continuation, line);
```
masatake, line 661: As the second argument, VStringPut expects a character, not a C string.
Your compiler may warn something.

masatake:
See https://docs.ctags.io/en/latest/testing-parser.html about writing a test case.
Could you add a GDScript to docs/news.rst as a newly introduced parser?

parsers/gdscript.c
masatake: If you really want to put the C string `line` at the end o continuation, `vStringCatS` may be the function you may want to use.

masatake:
```
parsers/gdscript.c: In function 'constructParentString':
parsers/gdscript.c:426:23: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
    vStringPut(result, (int) e->name);
                       ^
parsers/gdscript.c: In function 'findVariable':
parsers/gdscript.c:541:5: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
     if (strncmp (ln, "var", 3) != 0)
     ^~
parsers/gdscript.c:544:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  cp = strstr(ln, "=");
  ^~
parsers/gdscript.c: In function 'findGDScriptTags':
parsers/gdscript.c:661:28: warning: passing argument 2 of 'vStringPut' makes integer from pointer without a cast [-Wint-conversion]
   vStringPut(continuation, line);
                            ^~~~
In file included from ./main/field.h:22,
                 from ./main/entry.h:21,
                 from parsers/gdscript.c:17:
./main/vstring.h:105:64: note: expected 'int' but argument is of type 'const char *'
 CTAGS_INLINE void vStringPut (vString *const string, const int c)
                                                      ~~~~~~~~~~^
parsers/gdscript.c:645:27: warning: unused variable 'candidate' [-Wunused-variable]
   const char *cp = line, *candidate;
                           ^~~~~~~~~
```

parsers/gdscript.c
Davidy22: I was somewhat conflicted about whether to change it because this used so much of his code from the python one. I figured that I'd leave it since if I had made the same amount of changes in a patch to the python file I'd surely have left that line alone. Do I just replace it anyway, or do I add myself after it as an addendum?

parsers/gdscript.c
Davidy22: Whoops, they were all vStringCatS before and the CI warning recommended put, wasn't as familiar with the c offerings

parsers/gdscript.c
Davidy22: Yeah sure

parsers/gdscript.c
masatake: Oh, I see. 
I'm not sure how it should be. 
How about this one?

```
/*
*   Copyright (c) 2000-2003, Darren Hiebert
*   Copyright (c) 2014-2016, Colomban Wendling <ban@herbesfolles.org>
*   Copyright (c) 2021, David Yang
*
*
*   This module contains functions for generating tags for GDScript language
*   files. This module is derived from the Python module.
*
* 	GDScript language reference:
* 	https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html
*
*/
```

masatake:
This is nothing to do with whether I will merge your changes or not. I will merge your changes anyway. 
However, I have some words because I expect you will contribute to this project more in the future:-P

It seems that you use your `master` branch for developing your parser. However, I think you should make a topic branch for the purpose and use it. https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows may help you.

parsers/gdscript.c
```diff
@@ -0,0 +1,789 @@
...
+	}
+}
+
+// TODO: Add typing and annotations
+static const char *findVariable(const char *line)
+{
+	/* Parse global and class variable names (C.x) from assignment statements.
+	 * Object attributes (obj.x) are ignored. */
+	const char *cp, *sp, *eq, *start, *ln;
+
```
masatake, line 536: Though it is invisible on the GitHub web interface, there are unneeded whitespaces. Could you remove them?

parsers/gdscript.c
```diff
@@ -0,0 +1,789 @@
...
+		++ln;
+		while (isIdentifierCharacter ((int) *ln))
+			++ln;
+		while (isspace ((int) *ln))
+			++ln;
+	}
+
+	if (strncmp (ln, "var", 3) != 0)
+		return NULL;
+
```
masatake, line 548: Though it is invisible on the GitHub web interface, there are unneeded whitespaces. Could you remove them?

parsers/gdscript.c
masatake: Updated because the GDScript parser is based on the line-oriented older Python parser.

masatake:
It seems that you have developed GDScript parser based on a bit old Python parser.
The old one is a line-oriented parser.
The new one is a token-oriented parser.
See https://github.com/universal-ctags/ctags/blob/master/parsers/python.c.

Anyway, we can replace the implementation later.
What we need is a test case.

It seems that the first example in https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_basics.html is good for starting.

Davidy22:
Oh yeah, going to get to the test case, I just pushed a few times to get compile to pass before I had to go out for a bit, I'll take a run at test cases some time tonight when I'm back

parsers/gdscript.c
Davidy22: Yeah sure, I had modified before my name but I can switch to fancy copyright. The first line'll probably change I think, if I'm switching to using the new python parser as base

Davidy22:
Took a minute, but redone from the newer version. The instructions for running test cases locally might be outdated, make doesn't have a units target on what I pulled, but I think it'll run in CI so can probably see if it passes new cases here.

I set enumerated values to K_VARIABLE because I couldn't for the life of me figure out how to get them to register on symbols list in geany, but that might be external to ctags. I'll take another run at it if the distinction is critical.

masatake:
> Took a minute, but redone from the newer version.

Thank you.

> The instructions for running test cases locally might be outdated, make doesn't have a units target on what I pulled, 

Could you tell me the detail of the trouble?
The "units" test facility is the heart of u-ctags development. I would like to fix it.

> but I think it'll run in CI so can probably see if it passes new cases here.
> 
> I set enumerated values to K_VARIABLE because I couldn't for the life of me figure out how to get them to register on symbols list in geany, but that might be external to ctags. I'll take another run at it if the distinction is critical.

As far as I heard, geany has mappings from u-ctags's kinds to geany's kind.
So, please define and use kinds that are natural and/or comfortable for GDScript developpers. The design of kinds is the most important part of a parser development. Rewritting a parser is always allowed. However, redesigning kinds is not allowed onece the code is merged to our repo.

Davidy22:
The command from [this section](https://docs.ctags.io/en/latest/testing-parser.html#how-to-run-unit-tests) of the unit test guide says to run `make units`, ran make units from the top level and it turned up nothing. I did a quick search in the repo and found a units target in makefiles/testing.mak, so I guess I just need to use something other than GNU make to run this locally.

But while I'm still looking at unit tests through CI, looks like there's some existing files to append to, I'll get to trying to get these to pass

Since all the other languages with enums use the enumerator kind in their enums I probably should be using that instead of variable, I'll take another run at it.

masatake:
> The command from this section of the unit test guide says to run make units, ran make units from the top level and it turned up nothing.

Can I see the terminal output of "make units" ?

Davidy22:
'''
make: *** No rule to make target 'units'.  Stop.
'''

It's not much to go on, but at least github's not charging us for their CI so I can see the test results here.

Tests mostly fixed up I think, except for the one in list-roles-with-kind-names.d where it seems like something I've changed has wiped the whole output of that file so I'll need to figure that out

masatake:
```
$ make: *** No rule to make target 'units'.  Stop.
```
Reproduced:

```
$ cd /tmp
cd /tmp
$ git clone https://github.com/universal-ctags/ctags
Cloning into 'ctags'...
remote: Enumerating objects: 56326, done.        
remote: Counting objects: 100% (3757/3757), done.        
remote: Compressing objects: 100% (2008/2008), done.        
remote: Total 56326 (delta 1997), reused 3008 (delta 1645), pack-reused 52569        
Receiving objects: 100% (56326/56326), 18.77 MiB | 10.63 MiB/s, done.
Resolving deltas: 100% (35458/35458), done.
$ cd ctags
cd ctags
$ make units
make units
make: *** No rule to make target 'units'.  Stop.
$ 
```

How about doing following:
```
$ bash ./autogen.sh 
bash ./autogen.sh 
##################################################################
#                The paths and versions for tools                #
##################################################################
autoreconf is /usr/bin/autoreconf
------------------------------------------------------------------
...
$ ./configure
./configure
Universal Ctags, version 5.9.0
Linux 5.14.17-301.fc35.x86_64 #1 SMP Mon Nov 8 13:57:43 UTC 2021 x86_64
...
$ make
make
REPOINFO   main/repoinfo.h
make  all-recursive
make[1]: Entering directory '/tmp/ctags'
Making all in gnulib
...
$ make units
make units
REPOINFO   main/repoinfo.h
  CCLD     ctags
  RUN      units

Category: ROOT
Testing vera-interface as Vera                              passed
Testing zephir-simple as Zephir                             passed
...
```

masatake:
I'm inspecting the reason why 'make tmain' fails.
With your change, I got
```
$ ./ctags --quiet --options=NONE --list-roles=all.'{header}'
Segmentation fault (core dumped)
```
.

masatake:
The entries in GDScriptKinds and gdscriptKind are associated 1 to 1.
You defined K_METHOD in gdscriptKind, however no "method" in GDScriptKinds.
You defined K_MODULE in gdscriptKind, however no "module" in GDScriptKinds.
You defined K_PARAMETER in gdscriptKind, however no "parameter" in GDScriptKinds.
You defined K_LOCAL_VARIABLE in gdscriptKind, however no "localvar" in GDScriptKinds.

On the other hand, you defined  "namespace" in GDScriptKinds. However, no K_NAMESPACE in GDScriptKinds.

COUNT_KIND is a special. You don't need an associated entry for it in GDScriptKinds.

Could you revise the kind definition for GDScript?

I have two more requests.
1. Could you rename "Units/parser-gdscript.r/sample" to "Units/parser-gdscript.r/sample.d"?
2. Could you add  "args.ctags" file under  "Units/parser-gdscript.r/". The new file must include a line "--sort=no".

Units/parser-gdscript.r/sample/input.gd
```diff
@@ -0,0 +1,77 @@
+# A file is a class!
```
masatake, line 1: Could you add the origin of the test input?
```
# Taken from https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html
```
or
```
# Derrived from https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_basics.html
```

Davidy22: Went with taken from, I did change one line slightly so that it would test typed constants but it's still basically taken from the docs.

Davidy22:
I've even built stuff that's used the autogen, configure steps and it didn't come to mind here. Well, got to run the tests, changed code till the test output looked right and then just used the test output as expected instead of trying to handwrite it.

Renamed, file added, and switched enumerated values back to K_ENUMERATOR, got them to show up in the geany editor too. Tests pass locally, I'm assuming they should pass in CI so now we should be on semantics.

Units/parser-gdscript.r/sample.d/expected.tags
```diff
@@ -0,0 +1,17 @@
...
+THE_NAME	input.gd	/^const THE_NAME:String = "Charly"$/;"	v	typeref:typename:const String
+anon_enum	input.gd	/^enum {UNIT_NEUTRAL, UNIT_ENEMY, UNIT_ALL}$/;"	g
+Named	input.gd	/^enum Named {THING_1, THING_2, ANOTHER_THING=1}$/;"	g
+v2	input.gd	/^var v2 = Vector2(1, 2)$/;"	v
+v3	input.gd	/^var v3 = Vector3(1, 2, 3)$/;"	v
+some_function	input.gd	/^func some_function(param1, param2):$/;"	f
+something	input.gd	/^func something(p1, p2):$/;"	f
+Something	input.gd	/^class Something:$/;"	c
+a	input.gd	/^    var a = 10$/;"	v	class:Something
+_init	input.gd	/^func _init():$/;"	f
```
masatake, line 17: Could you add a test case for method, module, enumerator, parameter, and local?

Units/parser-gdscript.r/sample.d/input.gd
```diff
@@ -0,0 +1,79 @@
...
+
+# A file is a class!
+
+# Inheritance
+
+extends BaseClass
+
+# (optional) class definition with a custom icon
+
+class_name MyClass, "res://path/to/optional/icon.svg"
```
masatake, line 11: I wonder MyClass should be tagged with "class" kind?
How do you think?

parsers/gdscript.c
```diff
@@ -0,0 +1,1343 @@
...
+			if (lvEntry && lvEntry->kindIndex != K_CLASS)
+				kind = K_LOCAL_VARIABLE;
+
+			readNext = parseVariable (token, kind, decorators, token->keyword);
+		}
+		else if (token->keyword == KEYWORD_enum)
+		{
+			readNext = parseEnum (token);
+		}
+		else if (token->type == '@' && atStatementStart &&
```
masatake, line 1272: https://godotengine.org/article/gdscript-progress-report-new-gdscript-now-merged
It seems that GDScript has the "annotation" feature. Could you add a test case for testing "@..."?
So the test coverage increases.

parsers/gdscript.c
```diff
@@ -0,0 +1,1343 @@
...
+	KEYWORD_enum,
+};
+typedef int keywordId; /* to allow KEYWORD_NONE */
+
+typedef enum {
+	ACCESS_PRIVATE,
+	ACCESS_PROTECTED,
+	ACCESS_PUBLIC,
+	COUNT_ACCESS
+} accessType;
```
masatake, line 55: Does GDScript the concept "access" ?

If yes, could you add a test case?

parsers/gdscript.c
```diff
@@ -0,0 +1,1343 @@
...
+	unsigned int parameterCount = 0;
+	int corkIndex;
+
+	readToken (token);
+
+	if (token->type == '{') {
+		name = newToken ();
+		copyToken (name, token);
+		name->string = vStringNew();
+		vStringCatS(name->string, "anon_enum");
```
masatake, line 986: You can use
```
extern vString *anonGenerateNew (const char *prefix, int kind);
```
declared in main/parse.h.

Units/parser-gdscript.r/sample.d/input.gd
```diff
@@ -0,0 +1,79 @@
...
+
+# Constants
+
+const ANSWER = 42
+const THE_NAME:String = "Charly"
+
+# Enums
+
+enum {UNIT_NEUTRAL, UNIT_ENEMY, UNIT_ALL}
+enum Named {THING_1, THING_2, ANOTHER_THING=1}
```
masatake, line 31: It seems that your parser has the ability to tag THING_1, THING_2,...
I wonder why they are not in expected.tags.

parsers/gdscript.c
```diff
@@ -0,0 +1,1343 @@
...
+	vString   *parameterTypes [ARRAY_SIZE(parameterTokens)] = { NULL };
+	unsigned int parameterCount = 0;
+	int corkIndex;
+
+	readToken (token);
+
+	if (token->type == '{') {
+		name = newToken ();
+		copyToken (name, token);
+		name->string = vStringNew();
```
masatake, line 985: The original memory object pointed by name->string is lost.
You don't have to allocate vStringNew() here. A string is already assigned when you did newToken().

parsers/gdscript.c
masatake: What I wrote above is incorrect.

You can use 
```
extern void anonGenerate (vString *buffer, const char *prefix, int kind);
```
that use the existing vstring.

masatake:
Valgrind reports some memory-leaks. You can run it locally with the command: "make units LANGUAGES=GDScript VG=1" if you installed valgrind locally.

The test coverage is still low. I think I can help you increase the coverage. Give me time.

Units/parser-gdscript.r/sample.d/input.gd
Davidy22: The class_name keyword in gdscript is a little bit weird. Actually, all files in GDScript are by default sort of anonymous classes. `extends` is class inheritance for files, and class_name sets a name for the class that the file defines. The identifier, which in this case is MyClass, is actually undefined inside the file where class_name is used. Instead, class_name automatically includes the file as a class in all other files in the project, in this case using `MyClass` as the identifier, without the need for any explicit `extends` or `preload`. If we actually try to use the MyClass identifier within this file though, Godot will throw an error and you probably don't ever need to try reference MyClass from inside MyClass anyways because you can just directly refer to other variables and functions in here.

This actually means that the parser I've submitted is slightly incomplete, because it also needs to crawl the rest of the directory for .gd files with `class_name` declarations to list out every class that will be available in a file because GDScript can include files without you saying that you wanted to import anything, but I thought this might be a little out of scope.

parsers/gdscript.c
Davidy22: Oh whoops, I had some annotations in the first test I was using, slipped the mind when I grabbed the docs sample. I can add some annotations sure

parsers/gdscript.c
Davidy22: GDScript borrowed python's logic of making anything with a name starting with an underscore_ protected, there's an _init() already in the test file. Are we expecting more information in the output for this?

parsers/gdscript.c
Davidy22: Oh thanks, should be some amount cleaner

Units/parser-gdscript.r/sample.d/expected.tags
Davidy22: I have lines that would be related to module, enumerator, parameter and local, not showing up in the test output though. Module is extends (kind of), enumerator is by the two enums, there's some functions with parameters and a `var local_var = 5` in `some_function`.  Can add a method to the class.

Units/parser-gdscript.r/sample.d/input.gd
Davidy22: My first handwritten test case had lines for the enumerated values because I thought they'd show up. Thought this was how it was supposed to be, they do show up in the geany symbols list, back to checking code then. Local variables in functions also don't show up outside

Davidy22:
Satisfied valgrind, and the test case file is mutating. Thought anonGenerateNew was something different when it was recommended, it does make the enum names a little ugly but this ensures uniqueness I guess. I suspect this isn't going to push the test coverage over the threshold, maybe it does though.

Some files named thrift generated in peg/ when running tests could stand to get added to the gitignore maybe or just cleaned up post run, deleted them a good few times.

parsers/gdscript.c
```diff
@@ -0,0 +1,1346 @@
...
+			escaped++;
+	}
+}
+
+static void readIdentifier (vString *const string, const int firstChar)
+{
+	int c = firstChar;
+	do
+	{
+		vStringPut (string, (char) c);
```
masatake, line 411: The type of the second parameter of vStringPut is `int`. So you don't have to convert the type of `c` to `char` here.

masatake: Please, ignore above comment. I found it came from the python parser. I will fix both python and gdscript parsers together after merging gdscript.

Davidy22: I think I haven't touched this particular function, checked the python parser and it also has the little extra cast. Defies expectation a little. Almost done with going through the code coverage moans, this should put it over the threshold.

Davidy22: Oh I just realised you can click through to see more code coverage complaints, I have a lot more lines to look at

Davidy22: Alright, got a little hard to match line numbers near the end as code changed but this should surely get us above the threshold

Davidy22:
Oh word just looked, it's so close, just need to find a little more to add

parsers/gdscript.c
masatake: > Are we expecting more information in the output for this?

Yes. Let's the access information to expected.tags.
You can do it with putting
```
--fields=+{access}
```
to the args.ctags file.

parsers/gdscript.c
```diff
@@ -0,0 +1,1275 @@
...
+
+	readToken (token);
+
+	if (token->type == '{') {
+		copyToken (name, token);
+		vString *enumName = anonGenerateNew ("anon_enum_", K_ENUM);
+		vStringClear(name->string);
+		vStringCatS(name->string, vStringValue(enumName));
+		vStringDelete(enumName);
+		name->type = TOKEN_IDENTIFIER;
```
masatake, line 925: You don't have to make a new vstring.
Instead, let's reuse name->string area.
```
diff --git a/parsers/gdscript.c b/parsers/gdscript.c
index f9f83f206..e912ea28f 100644
--- a/parsers/gdscript.c
+++ b/parsers/gdscript.c
@@ -918,10 +918,8 @@ static bool parseEnum (tokenInfo *const token)
 
        if (token->type == '{') {
                copyToken (name, token);
-               vString *enumName = anonGenerateNew ("anon_enum_", K_ENUM);
-               vStringClear(name->string);
-               vStringCatS(name->string, vStringValue(enumName));
-               vStringDelete(enumName);
+               vStringClear (name->string);
+               anonGenerate (name->string, "anon_enum_", K_ENUM);
                name->type = TOKEN_IDENTIFIER;
        } else if (token->type == TOKEN_IDENTIFIER) {
                copyToken (name, token);
```

parsers/gdscript.c
masatake: I confirmed that the annotations are attached to tags in the test cases.

Units/parser-gdscript.r/sample.d/expected.tags
masatake: Put
```
--kinds-GDScript=+{parameter}
--kinds-GDScript=+{local}
```
to the args.ctags file.

parsers/gdscript.c
Davidy22: Oh, anonGenerate doing that makes it a good bit nicer

Davidy22:
Alright I flipped everything to true in GDScriptKinds and just changed some things on the geany symbols list side to make undesired things not show up, and made the whitespace in the test input file worse. `parameter` and `local` are still flipped to false in the python parser, so by default the python parser still should skip those. Output file now has parameters and local variables. I tried to type a carriage return, strangely difficult on linux, and it doesn't show up on git but I'm pretty sure enough more is being added to clear the test case threshold.

Units/parser-gdscript.r/sample.d/expected.tags
```diff
@@ -0,0 +1,26 @@
+a	input.gd	/^@export_range(start=0, end=100, step=1) var a = 5$/;"	v	annotations:export_range(start=0, end=100, step=1)
+s	input.gd	/^var s = "Hello"$/;"	v	annotations:export
+arr	input.gd	/^@onready var arr = [1, 2, 3]$/;"	v	annotations:onready
+dict	input.gd	/^var dict = {"key": "value", 2: 3}$/;"	v
+typed_var	input.gd	/^var typed_var: int$/;"	v	typeref:typename:int
+inferred_type	input.gd	/^inferred_type\\$/;"	v	annotations:onready,export_multiline
+ANSWER	input.gd	/^const ANSWER = 42$/;"	v	typeref:typename:const
```
masatake, line 7: This one, `ANSWER` is tagged as a `v`, a variable.
However, I think it should be tagged as a `const` or `constant`.
How do you think?

Davidy22: I just followed the example of the first file with a constant variable that I saw in a grep, I went by `parser-cxx.r/cxx11-raw-strings.d/expected.tags`, which has:

```
12:str4	input.cpp	/^static const char* str4 = LR"blah(";int bug4;)blah";$/;"	v	typeref:typename:const char *	file:
13:str5	input.cpp	/^static const char* str5 = u8R"blah(";int bug5;)blah";$/;"	v	typeref:typename:const char *	file:
14:str6	input.cpp	/^static const char* str6 = uR"blah(";int bug6;)blah";$/;"	v	typeref:typename:const char *	file:
15:str7	input.cpp	/^static const char* str7 = UR"blah(";int bug7;)blah";$/;"	v	typeref:typename:const char *	file:
```

Which classifies them with v and puts const in the type, which I followed. GDScript does do a thing where you substitute the keyword `var` entirely with the word `const` to declare a constant, but it's functionally the same as a constant variable I'd think

Davidy22: Actually I could drop into godot dev chat real quick and get a thought from there on which to prefer real quick

masatake:
Current TODO items.

* extends ad class_name keywords
  I wonder what we should do. I must learn GDScript to decide how we should do it.
  Once the decision is made, @masatake can implement anything we need about `extends` ad `class_name` keywords.
  This will be the most interesting area of this parser.

* e/enumerator
  ```
  enum x {a, b}
  ```
  `x` is tagged. However, it seems that `a` and `b` are not tagged.
  @masatake may be able to implement what we need.

* const
 constants are tagged as variables.
 Once the decision is made, @masatake can implement anything we need.

masatake:
I'm studying GDScript slowly.
@Davidy22, how do you think about signal? I think developers using GDScript may be happy if ctags can extract signals.
https://gdscript.com/solutions/signals-godot/

I don't say extracting signals is not a must item for merging your parser to the ctags repo.
I'm talking about the design of kinds. If kinds (and fields) are designed well, someone (including myself) can implement it even the one doesn't know GDScript. However, only a person who knows GDScript like you can design kinds and fields for the parser.

masatake:
The coverage is high enough. However, we can increase it more.
Could you add 
```
--extras=+r
--fields=+r
```
to the args.ctags?

Davidy22:
Oh, signals probably would be pretty useful. I kind of passed over them because signals are referenced with strings as parameters to a lookup function as opposed to actual identifiers, variables/functions/etc, that usually make up the symbols list. It would be pretty useful to have though, if we're willing to overlook how signals aren't identifiers in the same way everything else is. Some things to deal with if we want to show all signals available in a file:

- Signals can only travel up the node tree hierarchy, we'll need to parse the [tscn](https://docs.godotengine.org/en/stable/development/file_formats/tscn.html) file to find valid signals.
- Signal names can be in the autoloaded files defined in project.godot
- The above two points mean we're also crawling other .gd files to read their defined signals

Added the extra bit to args, I can take another run at enumerators sometime

Davidy22:
Pulled back because [something on ci test.](https://app.circleci.com/pipelines/github/universal-ctags/ctags/1880/workflows/8f222ada-a39c-40d9-bc6c-b26e3a429b49/jobs/20403) I'll be back in a while to see what's going on.

masatake:
To enable the assertion, you must rebuild ctags:
```
./configure --enable-debugging
make clean
make
```

masatake:
You defined the function method and the method kinds. I guess only the method kind is needed. I would like to hear your comment.b

```
func something(p1, p2):
    .something(p1, p2)


# Inner class

class Something:
    var a = 10
    const _private_var:String = "hi\n\\escape"
	func foooooooo() -> String:
		print("""
		test\\
		
		test""")
        return ""
```

So the class has no name, a class is being defined behind the seen.
Let's call it `anonClass`.

I guess `something` is a method of `anonClass`. Am I correct?
`Something` is a class defined in `anonClass`.

If with the `class_name` keyword, a developer can give a name for the class.

Units/parser-gdscript.r/sample.d/expected.tags
masatake: There are developers think differently.

`constant` kind is used in https://github.com/syskrank/vim-gdscript-ctags   .
`C` kind is used for constants in https://github.com/PrestonKnopp/gdtags .

C/C++ parser takes a different approach.  Base on the target language's sematic, the parser recognizes `const` is a part of a type.
```
$ cat /tmp/foo.c
const int i = 1;
$ ./ctags -o - /tmp/foo.c
i	/tmp/foo.c	/^const int i = 1;$/;"	v	typeref:typename:const int
```

ctags is a low-level tool. If possible, ctags should represents different things in different ways. This is one of the mottos of u-ctags.
If you want you can map "const" kind items extracted in ctags to "variable" kind items in geany.

A questionable thing is "const" defined in a method. What kind we should use for it? My idea is just using "const".
Instead, we can delete the "local" kind.  You may think deleting "local" kind conflicts with the motto.
However, it is acceptable because the scope: field of a tag for a local variable can represent the difference.

Let's think about the followig input:
```
$ cat /tmp/foo.gd
var v
const c = 1
func _init():
    var lv = Something.new()
	const lc = 2
```

The current parser output:
```
$ ./ctags --sort=no --kinds-GDScript='*' --fields=+K -o - /tmp/foo.gd
v	/tmp/foo.gd	/^var v$/;"	variable
c	/tmp/foo.gd	/^const c = 1$/;"	variable	typeref:typename:const
_init	/tmp/foo.gd	/^func _init():$/;"	function
lv	/tmp/foo.gd	/^    var lv = Something.new()$/;"	local	function:_init	file:
lc	/tmp/foo.gd	/^	const lc = 2$/;"	local	function:_init	typeref:typename:const	file:
```

My idea:
```
$ ./new-ctags --sort=no --kinds-GDScript='*' --fields=+K -o - /tmp/foo.gd
v	/tmp/foo.gd	/^var v$/;"	variable
c	/tmp/foo.gd	/^const c = 1$/;"	constant
_init	/tmp/foo.gd	/^func _init():$/;"	function
lv	/tmp/foo.gd	/^    var lv = Something.new()$/;"	variable	function:_init	file:
lc	/tmp/foo.gd	/^	const lc = 2$/;"	constant	function:_init	file:
```

masatake:
I convet my comments to C language:-). See #3199.

Units/parser-gdscript.r/sample.d/expected.tags
Davidy22: If the local type gets wiped in favor of metadata I'll have to figure out how to look at metadata to exclude locals from the symbol table in geany, otherwise this is probably fine I guess. Still feel like the relationship between const and var is more akin to C behavior despite the keywording though

masatake: Here ctags is built from the source code I proposed in #3194.

Consider the following as input:
```
$ cat -n /tmp/foo.gd
     1	var x
     2	func f(x):
     3		var x
     4		func g(x):
     5			var x
     6			
```

ctags-#3194's output:
```
$ u-ctags --sort=no --fields=+n -o - /tmp/foo.gd 
anon_class_49f7b8910100	/tmp/foo.gd	/^var x$/;"	c	line:1
x	/tmp/foo.gd	/^var x$/;"	v	line:1	class:anon_class_49f7b8910100
f	/tmp/foo.gd	/^func f(x):$/;"	m	line:2	class:anon_class_49f7b8910100
x	/tmp/foo.gd	/^func f(x):$/;"	z	line:2	method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	var x$/;"	l	line:3	method:anon_class_49f7b8910100.f	file:
g	/tmp/foo.gd	/^	func g(x):$/;"	f	line:4	method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	func g(x):$/;"	z	line:4	function:anon_class_49f7b8910100.f.g	file:
x	/tmp/foo.gd	/^		var x$/;"	l	line:5	function:anon_class_49f7b8910100.f.g	file:
```
`--fields=+n` option is for attaching `line:` fields:

If we add `--fields=+Z` option to the command line, `scope:` field markers are added. It helps us understand tags file:
```
$ u-ctags --sort=no --fields=+nZ -o - /tmp/foo.gd 
anon_class_49f7b8910100	/tmp/foo.gd	/^var x$/;"	c	line:1
x	/tmp/foo.gd	/^var x$/;"	v	line:1	scope:class:anon_class_49f7b8910100
f	/tmp/foo.gd	/^func f(x):$/;"	m	line:2	scope:class:anon_class_49f7b8910100
x	/tmp/foo.gd	/^func f(x):$/;"	z	line:2	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	var x$/;"	l	line:3	scope:method:anon_class_49f7b8910100.f	file:
g	/tmp/foo.gd	/^	func g(x):$/;"	f	line:4	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	func g(x):$/;"	z	line:4	scope:function:anon_class_49f7b8910100.f.g	file:
x	/tmp/foo.gd	/^		var x$/;"	l	line:5	scope:function:anon_class_49f7b8910100.f.g	file:
```

If we add `--fields=+ZK` option to the command line, the long-name of kinds are used. It helps us understand tags file:
```
$ u-ctags --sort=no --fields=+nZK -o - /tmp/foo.gd 
anon_class_49f7b8910100	/tmp/foo.gd	/^var x$/;"	class	line:1
x	/tmp/foo.gd	/^var x$/;"	variable	line:1	scope:class:anon_class_49f7b8910100
f	/tmp/foo.gd	/^func f(x):$/;"	method	line:2	scope:class:anon_class_49f7b8910100
x	/tmp/foo.gd	/^func f(x):$/;"	parameter	line:2	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	var x$/;"	local	line:3	scope:method:anon_class_49f7b8910100.f	file:
g	/tmp/foo.gd	/^	func g(x):$/;"	function	line:4	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	func g(x):$/;"	parameter	line:4	scope:function:anon_class_49f7b8910100.f.g	file:
x	/tmp/foo.gd	/^		var x$/;"	local	line:5	scope:function:anon_class_49f7b8910100.f.g	file:
```

The question is how it will be if we use `variable` kind instead of `local`. Let's replace "local" with "variable" with a text editor:
```
anon_class_49f7b8910100	/tmp/foo.gd	/^var x$/;"	class	line:1
x	/tmp/foo.gd	/^var x$/;"	variable	line:1	scope:class:anon_class_49f7b8910100
f	/tmp/foo.gd	/^func f(x):$/;"	method	line:2	scope:class:anon_class_49f7b8910100
x	/tmp/foo.gd	/^func f(x):$/;"	parameter	line:2	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	var x$/;"	variable	line:3	scope:method:anon_class_49f7b8910100.f	file:
g	/tmp/foo.gd	/^	func g(x):$/;"	function	line:4	scope:method:anon_class_49f7b8910100.f	file:
x	/tmp/foo.gd	/^	func g(x):$/;"	parameter	line:4	scope:function:anon_class_49f7b8910100.f.g	file:
x	/tmp/foo.gd	/^		var x$/;"	variable	line:5	scope:function:anon_class_49f7b8910100.f.g	file:
```

Both using the kind information (parameter or variable) and the scope fields, you can distinguish all `x`.

So I think deleting the `local` kind is acceptable in your use case.
However, keeping is o.k. My original suggestion was introducing "const" kind.

Davidy22:
Yeah, it is technically true that every function is a method because files are anonymous classes. GDScript docs use the words function and method in the typical sense though, and the distinction in calling functions and methods is like in other languages, `function()` as opposed to `object.function()`. The `.something()` in the example above calls `something` slightly differently because it's demonstrating syntax for calling the original function when overriding, it's actually not recursion. `something(1,2)` would be a legal call, and it would call the function defined above like you'd expect from a function, and `foooooooo()` would be illegal because it needs to be called from an object defined from class `Something`, like so:

```
var obj = Something.new()
print(obj.foooooooo())
```

`Something` is a class  defined in the larger class file, yeah. We can keep nesting class definitions if we wanted to:

```
class Something:
    class Something1:
        class Something2:
            class Something3:
                class Something4:
                    class Something5:
                        var a
```
And because files are classes, even `Something` is already a class definition nested within the `anonClass`.

`class_name` gives the file a class name and automatically makes the class available in all other project files, so in order to get a complete list of all classes that could be instantiated in a file, we need to parse all the other .gd files present in the directory for `class_name` calls. Or maybe this is the job of the text editor to track open files with a `class_name` call.

masatake:
> class_name gives the file a class name and automatically makes the class available in all other project files, so in order to get a complete list of all classes that could be instantiated in a file, we need to parse all the other .gd files present in the directory for class_name calls. Or maybe this is the job of the text editor to track open files with a class_name call.

Try #3199 (https://docs.ctags.io/en/latest/contributions.html#testing-a-pr-locally-before-being-merged).
ctags-#3199 recognizes `class_name` and gives a proper name to  `anonClass` tag.

Davidy22:
Oh, actually, do we even need to worry about the cross-file thing? The main issue with `class_name` and `signal` is that they're only really relevant in other files in the project, but maybe sharing those symbols between files is just the editor's job and collecting the symbols per file is good enough

masatake:
> Oh, actually, do we even need to worry about the cross-file thing? The main issue with class_name and signal is that they're only really relevant in other files in the project, but maybe sharing those symbols between files is just the editor's job and collecting the symbols per file is good enough

At least ctags should extract names newly defined/introduced in a source file.

An editor can ignore the names extracted by ctags.

Davidy22:
Grabbing the class name and signals isn't too tall an order, it's the bit about them needing to be available to other files to provide a complete view of available symbols that I was worried about, but that might just be the editor's job now that I think about it, to aggregate the files it has open

PR #3202, opened 2021-11-28
RSpec: rewrite in C language

masatake:


masatake:
I make a pull request only for ecd30f1 first of all.

PR #3205, opened 2021-11-29
Pascal: Add parsing of function/proc signatures

eht16:
Parse the signatures and return type for functions and procedures in Pascal.

This features is ported from the Geany Pascal parser where we use this code already for many years and it might be useful for others too.

### DIFF
--- a/Units/parser-pascal.r/simple-pascal.d/args.ctags
+++ b/Units/parser-pascal.r/simple-pascal.d/args.ctags
@@ -1,0 +1,2 @@
+--fields=+tS
+--sort=no
--- a/parsers/pascal.c
+++ b/parsers/pascal.c
@@ -38,10 +38,22 @@ static kindDefinition PascalKinds [] = {
 */
 
 static void createPascalTag (
-		tagEntryInfo* const tag, const vString* const name, const int kind)
+		tagEntryInfo* const tag, const vString* const name, const int kind,
+		const vString *arglist, const vString *vartype)
 {
 	if (PascalKinds [kind].enabled  &&  name != NULL  &&  vStringLength (name) > 0)
+	{
 		initTagEntry (tag, vStringValue (name), kind);
+		if (arglist != NULL && !vStringIsEmpty(arglist))
+		{
+			tag->extensionFields.signature = vStringValue(arglist);
+		}
+		if (vartype && !vStringIsEmpty(vartype))
+		{
+			tag->extensionFields.typeRef[0] = "typename";
+			tag->extensionFields.typeRef[1] = vStringValue(vartype);
+		}
+	}
 	else
 		/* TODO: Passing NULL as name makes an assertion behind initTagEntry failure */
 		initTagEntry (tag, NULL, KIND_GHOST_INDEX);
@@ -74,6 +86,64 @@ static bool tail (const char *cp)
 	return result;
 }
 
+static void parseArglist(const char *buf, vString *arglist, vString *vartype)
+{
+	const char *start, *end;
+	int level;
+
+	if (NULL == buf || arglist == NULL)
+		return;
+
+	/* parse argument list which can be missing like in "function ginit:integer;" */
+	if (NULL != (start = strchr(buf, '(')))
+	{
+		for (level = 1, end = start + 1; level > 0; ++end)
+		{
+			if ('\0' == *end)
+				break;
+			else if ('(' == *end)
+				++ level;
+			else if (')' == *end)
+				-- level;
+		}
+	}
+	else /* if no argument list was found, continue looking for a return value */
+	{
+		start = NULL;
+		end = buf;
+	}
+
+	/* parse return type if requested by passing a non-NULL vartype argument */
+	if (NULL != vartype)
+	{
+		char *var, *var_start;
+
+		if (NULL != (var = strchr(end, ':')))
+		{
+			var++; /* skip ':' */
+			while (isspace((int) *var))
+				++var;
+
+			if (starttoken(*var))
+			{
+				var_start = var;
+				var++;
+				while (intoken(*var))
+					var++;
+				if (endtoken(*var))
+				{
+					vStringNCatS(vartype, var_start, var - var_start);
+				}
+			}
+		}
+	}
+
+	if (NULL == start) /* no argument list */
+		vStringCatS(arglist, "()");
+	else
+		vStringNCatS(arglist, start, end - start);
+}
+
 /* Algorithm adapted from from GNU etags.
  * Locates tags for procedures & functions.  Doesn't do any type- or
  * var-definitions.  It does look for the keyword "extern" or "forward"
@@ -83,6 +153,8 @@ static bool tail (const char *cp)
 static void findPascalTags (void)
 {
 	vString *name = vStringNew ();
+	vString *arglist = vStringNew();
+	vString *vartype = vStringNew();
 	tagEntryInfo tag;
 	pascalKind kind = K_FUNCTION;
 		/* each of these flags is true iff: */
@@ -205,7 +277,12 @@ static void findPascalTags (void)
 			for (cp = dbp  ;  *cp != '\0' && !endtoken (*cp)  ;  cp++)
 				continue;
 			vStringNCopyS (name, (const char*) dbp,  cp - dbp);
-			createPascalTag (&tag, name, kind);
+
+			vStringClear (arglist);
+			vStringClear (vartype);
+			parseArglist((const char*) cp, arglist, (kind == K_FUNCTION) ? vartype : NULL);
+
+			createPascalTag (&tag, name, kind, arglist, (kind == K_FUNCTION) ? vartype : NULL);
 			dbp = cp;  /* set dbp to e-o-token */
 			get_tagname = false;
 			found_tag = true;
@@ -246,6 +323,8 @@ static void findPascalTags (void)
 			}
 		}  /* while not eof */
 	}
+	vStringDelete (arglist);
+	vStringDelete (vartype);
 	vStringDelete (name);
 }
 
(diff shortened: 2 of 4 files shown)


parsers/pascal.c
```diff
@@ -38,10 +38,16 @@ static kindDefinition PascalKinds [] = {
...
 static void createPascalTag (
-		tagEntryInfo* const tag, const vString* const name, const int kind)
+		tagEntryInfo* const tag, const vString* const name, const int kind,
+		const char *arglist, const char *vartype)
 {
 	if (PascalKinds [kind].enabled  &&  name != NULL  &&  vStringLength (name) > 0)
+	{
 		initTagEntry (tag, vStringValue (name), kind);
+		tag->extensionFields.signature = arglist;
+		tag->extensionFields.typeRef[0] = eStrdup ("typename");
```
eht16, line 48: Is it correct to dup the string here?
I found varying uses in other parsers and so I am unsure whether it should be done or not. It seems to work without as well but I might have just been lucky.

Same for the assignment of `typeRef[1]` below.

Units/parser-pascal.r/simple-pascal.d/args.ctags
```diff
@@ -0,0 +1 @@
+--fields=+tS
```
masatake, line 1: Could you add
```
--sort=no
```
?

So reviewing the test case becomes easier a bit.

eht16: Done (I will squash the commits before merge).

masatake: I see. Thank you.

parsers/pascal.c
masatake: In this case, you should not do "dup".
"dup" is needed once a tag entry is copied to the cork queue.
See https://docs.ctags.io/en/latest/internal.html?highlight=cork#output-tag-stream-1.

If the cork queue is enabled in a parser, a tag passed makeTagEntry is copied to a memory buffer named "cork queue". makeTagEntry returns an `id` on the queue. You can get back the tag entry with passing the `id` to getEntryInCorkQueue(id). After getting back, you can update a member of the tag entry. When updating the member, you must do "dup".

The pascal parser doesn't use the cork queue. So you don't have to do "dup".

If you can use valgrind, you can test the mem leak with "make units LANGUAGES=Pascal VG=1".

eht16: Thanks for detailed explanation.
I checked with Valgrind locally in the meantime and noticed there also that the string duplication was not necessary.
Squashed and pushed the fix.

masatake:
If you are thinking about your hoidays:-), I would like you to see https://github.com/universal-ctags/ctags/pull/2241, too.
I need helps from a person who knows pascal.

eht16:
> If you are thinking about your hoidays:-), I would like you to see #2241, too. I need helps from a person who knows pascal.

Ha, nice try.
Actually, the "type" desclaration parsing is, apart from this PR, the only difference between the ctags and Geany Pascal parser. I was about to submit a PR for this too (not realizing #2241 is already there). But I noticed the current implementation is incomplete and buggy and so rather needs to be rewritten or removed. I personally do not know Pascal any good, learned it a bit in school but this was a few decades ago... :).
I don't want and can't spend time on implementing this properly. Sorry.

masatake:
@eht16, I see.
After merging GDScript, I will take more time for this pull request.

parsers/pascal.c
```diff
@@ -74,6 +80,68 @@ static bool tail (const char *cp)
...
+				{
+					*var = '\0';
+					*vartype = strdup(var_start);
+				}
+			}
+		}
+	}
+
+	*end = '\0';
+	*arglist = strdup(start);
```
masatake, line 140: My understanding:

To do `*end = '\0'`, you have to do `c = strdup(buf)`.
To do `strdup(start)`, you have to do `*end = '\0'`, terminating the C string.

I think you can reduce `strdup()` with `eStrndup(const char* str, size_t len);` declared in `main/routines.h`.
With it, you can make a new C string without doing `*end = '\0'`.

parsers/pascal.c
```diff
@@ -74,6 +80,68 @@ static bool tail (const char *cp)
...
+
+	eFree(c);
```
masatake, line 142: If `eStrndup` helps you expectedly, you can remove this `eFree()`.

parsers/pascal.c
```diff
@@ -74,6 +80,68 @@ static bool tail (const char *cp)
...
+			if (starttoken(*var))
+			{
+				var_start = var;
+				var++;
+				while (intoken(*var))
+					var++;
+				if (endtoken(*var))
+				{
+					*var = '\0';
+					*vartype = strdup(var_start);
```
masatake, line 133: Could you eStrdup() declared n `main/routines.h` instead o `strdup()`.
`eStrdup()` terminates the ctags process if allocating memory for the new string is failed.

parsers/pascal.c
```diff
@@ -74,6 +80,68 @@ static bool tail (const char *cp)
 	return result;
 }
 
+static void parseArglist(const char *buf, char **arglist, char **vartype)
```
masatake, line 83: Using vString() for storing arglist and vartype is an alternative approach.

In a micro view, an operation on vString tasks higher const than using `char *` directly.
However, in a macro view, vString is not so bad. Using char *, you must call strdup twice in every calling parseArglist.
vString can be reused once it is new'ed in findPascalTags. 
A vString can clear its internal buffer keeping char * with vStringClear.
Though, it is cleared, the buffer is not deallocated. 

Writing English is hard for me. Let's show me in pseudo-code:
```
/* Your code using raw C string */

static void parseArglist(const char *buf, char **arglist, char **vartype)
{
  *arglist = strdup();
  *vartype = strdup();  
}

static void findPascalTags (void)
{
  char *arglist = NULL;
  char *vartype = NULL;
  // ...
  loop() {
    createPascalTag();
    // ...
    free (arglist);
    free (vartype);    
    parseArglist(, &arglist, &vartype);
  }
  free (arglist);
  free (vartype);    
  
}

/* What we can do with vString */
static void parseArglist(const char *buf, vString *arglist, vString *vartype)
{
  char c, *str;
  /* ... */
  vStringPut(arglist, c);
  /* or */ vStringCatS(arglist, str);
  /* or */ vStringNCatS(arglist, str, len);
  vStringPut(vartype, c);
  /* or */ vStringCatS(vartype, str);
  /* or */ vStringNCatS(vartype, str, len);

}

static void findPascalTags (void)
{
  vString *arglist = vStringNew();
  vString *vartype = vStringNew();
  // ...
  loop() {
    createPascalTag();
    // ...
    vStringClear (arglist);
    vStringClear (vartype);    
    parseArglist(, arglist, vartype);
  }
  vStringDelete (arglist);
  vStringDelete (vartype);    
  
}
```

Of course, I don't care about the performance of this level.
However, vString helps me understand the logic.

This comment is not for "must-fix". Your code is already works fine. So this is a just comment.

masatake:
NOTE for myself: following semi-fuzz tests are passed.
```
make chop LANGUAGES=Pascal
make slap LANGUAGES=Pascal
make noise LANGUAGES=Pascal
```

parsers/pascal.c
eht16: Thanks for the hint, did it and then rewrote it with your `vString` suggestion below :D.

parsers/pascal.c
eht16: Fortunately, that code vanished by using `vString`.

parsers/pascal.c
eht16: Fortunately, that code vanished by using `vString`.

parsers/pascal.c
eht16: Great idea and thanks for the detailed explanation.
I rewrote the code to use `vString` (back in 2009 when I wrote the code initially I wasn't so much aware of the `vString` helper and didn't touch the since then).

It's simpler now and we don't need any more string copying than what is done implicitly by `vStringCat*()`.

eht16:
The `chop`, `noise` and `slap` make targets do not output anything for me and quit pretty fast. I'm not sure if they actually did test anything.

parsers/pascal.c
```diff
@@ -276,11 +278,9 @@ static void findPascalTags (void)
...
 			vStringNCopyS (name, (const char*) dbp,  cp - dbp);
 
-			if (arglist != NULL)
-				eFree(arglist);
-			if (kind == K_FUNCTION && vartype != NULL)
-				eFree(vartype);
-			parseArglist((const char*) cp, &arglist, (kind == K_FUNCTION) ? &vartype : NULL);
+			vStringClear (arglist);
+			vStringClear (vartype);
+   			parseArglist((const char*) cp, arglist, (kind == K_FUNCTION) ? vartype : NULL);
```
masatake, line 283: A minor issue. The line 282 and the line 283 are not aligned.

masatake: Line 282 uses only tag characters.
However, the line 283 uses tabs and spaces mixed way.

eht16: Oops, will fix.

Another thing I noted (and probably was introduced by me in 2009), there are some inconsistencies in using a space between function name and the opening brace of the argument list. Should I fix them as well, to consistently use a space between?

masatake: > Should I fix them as well, to consistently use a space between?

Thank you for the offering. If possible, could you fix them in ANOTHER commit?
In a commit, let's do one thing.

eht16: Ok, the wrong indentation is fixed.

Did you mean another commit in this PR to fix the argument list space problem or a separate PR?
I don't mind.

masatake: I also don't mind. However, you already submitted a good commit that aligns with the goal of this pull request.
So I will merge this.

masatake:
Thank you.

PR #3206, opened 2021-12-01
Add gdscript entry in new parser docs

Davidy22:
Quick page written up for #3199.


masatake:
Thank you.
I'll merge this after merging the parser.
I have a plan to write a man page. I would like you to review it in the future.

PR #3212, opened 2021-12-02
Pascal: extract type

masatake:
This is mostly based on #2241 submitted by @ntrel.
I adjusted the changes to fit the latest codebase.


parsers/pascal.c
```diff
@@ -256,6 +264,14 @@ static void findPascalTags (void)
 					verify_tag = false;
 				}
 			}
+			else if (tolower ((int) *dbp) == 't')
+			{
+				if (tail ("type"))      /*  check for type declaration */
+				{
+					found_tag = false;
```
masatake, line 271: These lines are not covered by the test cases.

masatake:
If more than two types are defined under the `type` keyword, the change doesn't work.

https://www.gnu-pascal.de/gpc/Type-Declaration.html
```
     type
       { This side is the }     { That side is the }
       { type declaration }     { type definition  }
     
       Arrayfoo            = array [0 .. 9] of Integer;  { array definition }
       Recordfoo           = record                      { record definition }
                               Bar: Integer;
                             end;
     
            { schema def with discriminants ``x, y: Integer'' }
       SchemaFoo (x, y: Integer) = array [x .. y] of Integer;
       CharSetFoo          = set of Char;              { Def of a set }
       ObjectFoo           = object                    { Def of an object }
                               procedure DoAction;
                               constructor Init;
                               destructor Done;
                             end;
       SubrangeFoo         = -123..456;                { subrange def }
     
       EnumeratedFoo       = (Pope,John,the,Second);   { enum type def }
            { Def of a pointer to another type identifier }
       PInteger            = ^arrayfoo;
            { Def of an alias name for another type identifier }
       IdentityFoo         = Integer;
            { Def of an integer which was initialized by 123 }
       InitializedFoo      = Integer value 123;
```

PR #3217, opened 2021-12-05
Python: improve the way to parser arglist of cdef functions

### LINKED ISSUE #3216: Incorrect cython signature when 'z' kind enabled
For the following cython code
```
cdef int int_identity(int i) : return i  
```
processing with `ctags --fields=S --python-kinds=+z` produces
```
int_identity	test.py	/^cdef int int_identity(int i) : return i  $/;"	signature:(inti)
```
(notice the missing space between `int` and `i` in signature) while `ctags --fields=S` produces the correct result with the space:
```
int_identity	test.py	/^cdef int int_identity(int i) : return i  $/;"	signature:(int i)
```

masatake:
Close #3216.

Much more effort may be needed to support Cython.
In sagemath, I found cdef func lines which cannot be handled this pull request:
```
# Based on the code from sagemath-9.4-2.fc35                                                                                                                                                              
                                                                                                                                                                                              
cdef inline void* sig "sig"(size_t n) nogil:                                                                                                                                                  
    pass                                                                                                                                                                                      

cdef int squares(uint n, uint res[2]):                                                                                                                                                        
    pass                                                                                                                                                                                      `

cdef int singular(poly **ret, poly *p, ring *r, list args, poly *(*get_element)(object)):
    pass

```


masatake:
It will be better to introduce Cython parser separated from Python parser.

Instead, we need routines supporting Python-like languages.
We can (re)write GDScript, Python, Cython, and maybe Nim based on the routines.

masatake:
Better than nothing.

PR #3226, opened 2021-12-17
DON't MERGE: fine spotter locating the isssue on "run units target on MSYS2"/MSYS

masatake:
These changes try to locate the cause of the issue I showed in #3225.

Calling `lock_init` causes the issue.

PR #3227, opened 2021-12-17
DON'T MERGE: use the regex module in gnulib in single thread mode

masatake:
This pull request is for demonstrating the last change of this pull request can detour the issue reported as #3225 and #3226.

The word, "HUNTER", in the branch name is not correct. This change hunts no bug: just escaping.


masatake:
The segmentation faults are gone.

Unexpectedly "Unexpectedly I survived.\n" is not printed because `lock_init()` returns 0 with this change.

PR #3228, opened 2021-12-17
build-sys: use the regex module in gnulib in single thread mode

masatake:
On "run units target on MSYS2"/MSYS CI environment, many
Tmain cases are failed with segmentation faults as show in

in regcomp.c of gnulib.

change.

This change is a work-around. The root cause is unknown.

Close #3225, #3226, and #3227.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -560,6 +560,10 @@ PRETTY_ARG_VAR([WARNING_CFLAGS], [C compiler warning flags],
 #   header files or functions.
 gl_INIT
 
+AC_DEFINE([GNULIB_REGEX_SINGLE_THREAD], [1],
+  [Define if all programs in this package call functions of the Gnulib
+   'regex' module only from a single thread.])
+
 if test "$REPLACE_FNMATCH" != 0; then
 	AC_DEFINE([USE_GNULIB_FNMATCH])
 fi


masatake:
@hirooih, last week, I have inspected segmentation faults on "run units target on MSYS2"/MSYS CD environment.
#3225 is for reporting the segmentation faults.
I tried to locate the code triggering the segmentation faults in #3226. I found `lock_init()` causes the segmentation faults.

I found #3227 can be used as a work-around.

This pull request cherry-picks the work-around. All the changes for demonstrating the issue are removed.

The segmentation faults have been stopped the development of ctags. So I will merge this workaround anyway.

However, if you have any comments or advice, I would like to hear them.

Using gdb in batch mode can be used for inspecting the fault more. However, I cannot find any more... energy in my mind.
Using `objdump -d` to read the generated code for `lock_init()` is another thing I should do.

masatake:
I merge this to merge other changes.
However, comments are welcome even after merging.

hirooih:
@masatake san,

> However, if you have any comments or advice, I would like to hear them.

I think your fix is fair.

By searching `GNULIB_REGEX_SINGLE_THREAD` which you added in this PR, I found the following description the Gnulib manual.
https://www.gnu.org/software/gnulib/manual/html_node/Multithreading-Optimizations.html

> You may define the C macro GNULIB_REGEX_SINGLE_THREAD, if all the programs in your package invoke the functions of the regex module only from a single thread.

As far as I grep the source files of u-ctags, u-ctags does not call any thread functions.  I think defining GNULIB_REGEX_SINGLE_THREAD is correct.

We had another similar segmentation fault on a msys2 system on #3069. But this PR fixes the root cause of it, too.

masatake:
> We had another similar segmentation fault on a msys2 system on #3069. But this PR fixes the root cause of it, too.

Oh, I forgot this one.
I must read the discussion.

Thank you.

hirooih:
Sorry, I found my English was ambiguous.

> But this PR fixes the root cause of it, too.

"this PR" meant this #3228, i.e. your fix.

k-takata:
Sorry for late.
I think that the `lock_init` issue is exactly the same as https://github.com/universal-ctags/ctags/pull/3069#issuecomment-861916353.

masatake:
@k-takata, thank you for the confirmation.

PR #3229, opened 2021-12-19
Cxx: accept bit-width for enum-typed structure member

### LINKED ISSUE #3223: C: enum bit field
```C
enum token_type {
  ID, KEYWORD,
};
struct token {
	enum token_type	token_type : 8;
};
```

`token_type` is not extracted.
I found this in linux kernel.
I'm not sure whether the input is valid as standard C.

masatake:
The original code could not extract enum-typed
structure members having bit-width like "mber"
in the input:

    enum token_type {
      ID, KEYWORD,
    };
    struct token {
	    enum token_type	mber : 8;
    };

Close #3223.

### DIFF
--- a/Units/parser-c.r/enum-bit-fields.d/args.ctags
+++ b/Units/parser-c.r/enum-bit-fields.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=no
+--fields-C=+{macrodef}
+--param-CPreProcessor._expand=1
+--fields=+{signature}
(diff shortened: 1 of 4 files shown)


masatake:
The parser maintainer may be still busy.

- One month has passed. 
- The change has small.
- The change has no impact on the parser core.

For these reason, I will merge this.

@pragmaware, your review is welcome even after merging the change.

PR #3230, opened 2021-12-21
Man: extract  subsections

masatake:


masatake:
TODO

- [ ] `end:` fields for subsections should be filled.
- [ ] add a test case for the new operator

masatake:
ea090f6 must be reviewed.

masatake:
@leleliu008, it seems that the build environments related to MacOS are broken. Could you fix them?

PR #3237, opened 2021-12-25
Fortran parser improvements from Geany

techee:
These are various improvements we had in Geany for the Fortran parser:

- reporting of anonymous tags
- reporting scopes for interfaces and enums
- allowing keyword names to be used as names of various Fortran types

Note: I don't know what I'm doing much, I don't know Fortran and it's possible I'm doing something wrong - the patches are just based on what is present in Geany to allow us to use the uctags Fortran parser with correct scope information.

### DIFF
--- a/parsers/fortran.c
+++ b/parsers/fortran.c
@@ -469,22 +469,26 @@ static tokenInfo *newToken (void)
 	return token;
 }
 
-static tokenInfo *newTokenFrom (tokenInfo *const token)
+static tokenInfo *newTokenFromFull (tokenInfo *const token, bool copyStr)
 {
 	tokenInfo *result = xMalloc (1, tokenInfo);
 	*result = *token;
-	result->string = vStringNewCopy (token->string);
+	result->string = copyStr? vStringNewCopy (token->string): vStringNew();
 	token->secondary = NULL;
 	token->parentType = NULL;
 	token->signature = NULL;
 	return result;
 }
 
+static tokenInfo *newTokenFrom (tokenInfo *const token)
+{
+	return newTokenFromFull (token, true);
+}
+
 static tokenInfo *newAnonTokenFrom (tokenInfo *const token, unsigned int uTagKind)
 {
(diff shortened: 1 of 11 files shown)


parsers/fortran.c
```diff
@@ -479,6 +481,15 @@ static tokenInfo *newTokenFrom (tokenInfo *const token)
 	return result;
 }
 
+static tokenInfo *newAnonTokenFrom (tokenInfo *const token, unsigned int uTagKind)
+{
+	tokenInfo *result = newTokenFrom (token);
```
masatake, line 485: Could you add the following change:
```
-static tokenInfo *newTokenFrom (tokenInfo *const token)
+static tokenInfo *newTokenFromFull (tokenInfo *const token, bool noCopyStr)
 {
 	tokenInfo *result = xMalloc (1, tokenInfo);
 	*result = *token;
-	result->string = vStringNewCopy (token->string);
+	result->string = noCopyStr? vStringNewCopy (token->string): vStringNew();
 	token->secondary = NULL;
 	token->parentType = NULL;
 	token->signature = NULL;
 	return result;
 }
 
+static tokenInfo *newTokenFrom (tokenInfo *const token)
+{
+	return newTokenFromFull (token, true);
+}
+
```

So you can use `newTokenFromFull (token, false` here.
As the result you don't have to call `vStringClear (result->string);`.

masatake: Other than this minor point, all changes look fine.

techee: Done.

masatake:
I will reorder the commits and merge them.

masatake:
I wonder my pull requests are failed always on some CI platforms.
On the other hand, your pull requests are successful always on the same CI platforms.

masatake:
Merged via #3240. Thank you.

techee:
> I wonder my pull requests are failed always on some CI platforms.
> On the other hand, your pull requests are successful always on the same CI platforms.

I had to re-run them to make them succeed - it seems github had some problems yesterday.

PR #3238, opened 2021-12-27
Asm: extract macro parameters

masatake:


masatake:
I could not overcome the trouble with CD platform.

PR #3239, opened 2021-12-27
Tex: parse some more latex macros

techee:
In addition to the macros already parsed, the old Geany readline-based parser also parses the following latex macros which may be useful:
- `\renewcommand`, `\providecommand`
- tex `\def\cmd{param}`
- `\DeclareMathOperator`
- `\newenvironment`, `\renewenvironment`
- `\newtheorem`

This pul request adds these and introduces the corresponding kinds for them.

### DIFF
--- a/Units/parser-tex.r/newcommand.d/expected.tags
+++ b/Units/parser-tex.r/newcommand.d/expected.tags
@@ -1,3 +1,5 @@
+\\bar	input.tex	/^\\providecommand{\\bar}{\\section{#1}}$/;"	C
+\\foo	input.tex	/^\\renewcommand{\\foo}{\\section{#1}}$/;"	C
 \\mysection0	input.tex	/^\\newcommand{\\mysection0}{\\section{#1}}$/;"	C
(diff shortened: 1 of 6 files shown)


masatake:
I took much time to design `TexParseStrategy`. So I’m glad that one other than I utilizes the data type.

Could you split dc57692 smaller ones as I wrote in https://github.com/universal-ctags/ctags/pull/3236#issuecomment-1001800293 .
Could you use `Tex` as the common prefix for the headers of the commits that changes the parser?
If you made a stand-alone commit changing or adding a test case, please use `Units(Tex)` as prefix in the header for the commit.

```
Units(Tex): add unit tests for the newly introduced kinds
Tex: fix TEX_NAME_FLAG_OPTIONAL for the last macro parameter
Tex: parse \newtheorem
Tex: parse \newenvironment and \renewenvironment
Tex: parse \DeclareMathOperator{cmd}{def}
Tex: parse \def\cmd{replacement}
Tex: add \renewcommand and \providecommand
```

techee:
@masatake Done.

masatake:
Thank you very much.

PR #3240, opened 2021-12-27
Fortran: minor changes for merging #3237

masatake:
Close #3237.

* The order of commits were rearranged, and
* s/fortran/Fortran/ in commit headers.


masatake:
Thank you very much.

PR #3243, opened 2021-12-30
ada: remove setjmp() longjmp()

techee:
setjmp() and longjmp() may cause memory leaks and are better to avoid.
The ada parser actually doesn't seem to use them normally, there are proper
checks for EOF in the code, and longjmp() is used only when something
goes wrong after 1000 hits of EOF.

I went through all the loops and all of them seem to check for EOF so
there should be no need for setjmp()/longjmp().

We currently don't use the ada parser in Geany but since ada is supported, it might be interesting for us in the future and we try to avoid setjmp()/longjmp() because Geany is a long-running process and leaks caused by setjmp/longjmp may accumulate over time.

### DIFF
--- a/parsers/ada.c
+++ b/parsers/ada.c
@@ -95,7 +95,6 @@
 
 #include <string.h>     /* to declare strxxx() functions */
 #include <ctype.h>      /* to define isxxx() macros */
-#include <setjmp.h>
 
 #include "parse.h"      /* always include */
 #include "read.h"       /* to define file readLineFromInputFile() */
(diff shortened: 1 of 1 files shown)


parsers/ada.c
```diff
@@ -624,17 +611,8 @@ static void readNewLine (void)
 		if (line == NULL)
 		{
 			lineLen = 0;
-			exception = EXCEPTION_EOF;
```
masatake, line 627: Removing this mysterious loop is one of mankind's long-cherished wish like a world without war.
Thank you very much fo taking your time.

masatake:
Thank you.

parsers/ada.c
techee: Yeah, this was really mysterious when looking at it for the first time and I assumed the longjmp() was executed during normal parsing. But it turned out to be just a fail-safe mechanism if the parser started cycling somewhere and since none of the other parsers uses anything like that, better to avoid it.

PR #3245, opened 2021-12-31
yaml: implement basic Swagger/OpenAPI subparser

segoon:
The parser is able to parse both Swagger 2.0 and Openapi 3.0.x.
The code is partly based on ansible playbook parser.  It was tested
using the following vim's TagBar parser declaration:

    let g:tagbar_type_yaml = {
                        \  'ctagstype': 'openapi',
                        \  'kinds': [
                        \          'p:path',
                        \          'd:schema',
                        \          'P:parameter',
                        \          'R:response',
                        \          ]
                        \ }


parsers/openapi.c
```diff
@@ -0,0 +1,361 @@
+/*
+*
+*   Copyright (c) 2016, Masatake YAMATO
+*   Copyright (c) 2016, Red Hat, K.K.
+*   Copyright (c) 2022, Vasily Kulikov
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
```
masatake, line 9: Could you add references for the target language or input like:
https://github.com/universal-ctags/ctags/blob/82154927e9a31a088f2fcc9b00e22d279d362b72/parsers/gdscript.c#L12

`git push --force` is welcome.

masatake:
Writing a subparser on the YAML parser is an incredible challenge!
Let's jump into the wild area.

masatake:
Is there any difference between the targets, Swagger 2.0 and Openapi 3.0.x?

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+	const enum openapiKeys* keys;
+	size_t countKeys;
+};
+
+static const enum openapiKeys pathKeys[] = {
+	KEY_UNKNOWN,
+	KEY_PATHS,
+};
+
+static const enum openapiKeys responses3Keys[] = {
```
segoon, line 173: > Is there any difference between the targets, Swagger 2.0 and Openapi 3.0.x?

They are 2 different revisions of the protocol (Swagger == OpenAPI 2.0). They are mostly the same except for several details (not interesting for ctags) and several sections have a distinct path in yaml (e.g. "responses" in swagger and "components.responses" for openapi). I tried to express it in `XXXKeys` names with 2 (swagger) or 3 (openapi). `pathKeys` is the same for both.

Units/parser-openapi.r/swagger.d/input.yaml
```diff
@@ -0,0 +1,64 @@
+swagger: '2.0'
+info:
+    description: test
+    title: test
```
masatake, line 4: How about tagging the title?

Units/parser-openapi.r/openapi.d/input.yaml
```diff
@@ -0,0 +1,55 @@
+openapi: 3.0.0
+info:
+    title: test
+    description: test
+    version: '1.0'
+
+servers:
+  - url: http://example.com
```
masatake, line 8: How about tagging the URLs under servers?

Units/parser-openapi.r/openapi.d/input.yaml
```diff
@@ -0,0 +1,55 @@
+openapi: 3.0.0
+info:
+    title: test
```
masatake, line 3: How about tagging the title?

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+				.inputEnd = inputEnd,
+			},
+			.newTokenNotfify = newTokenCallback
+		},
+	};
+	static parserDependency dependencies [] = {
+		{ DEPTYPE_SUBPARSER, "Yaml", &openapiSubparser },
+	};
+
+	parserDefinition* const def = parserNew ("OpenApi");
```
masatake, line 350: It seems that OpenAPI may be better than OpenApi.

masatake:
Though I named the yaml parser "Yaml",  I think "OpenAPI" is better than "OpenApi" as the name of your parser.

(The review is in progress.)

Units/parser-openapi.r/swagger.d/input.yaml
segoon: If it makes sense to tag all user-defined entities, then ok.
But then more fields must be tagged:
- info itself
- servers children
- webhooks children
- security children
- tags children

I don't think the following entities worth adding tags (too minor IMHO):
- openapi
- jsonSchemaDialect
- externalDocs

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#openapi-object

parsers/openapi.c
segoon: ok, no problem

Units/parser-openapi.r/swagger.d/input.yaml
segoon: Also the items of "components": https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-6

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+
+
+	def->dependencies = dependencies;
+	def->dependencyCount = ARRAY_SIZE (dependencies);
+
+	def->kindTable	= OpenApiKinds;
+	def->kindCount = ARRAY_SIZE (OpenApiKinds);
+	def->parser	= findOpenApiTags;
+	def->useCork = CORK_NIL;
```
masatake, line 359: `CORK_NIL` is not suitable value for `useCork` field.
The `useCork` field is initialized with 0 in parserNew. So you don't have to give a value to the field as far as your parser doesn't need cork indexes. Let's remove this line.

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+	return def;
+}
+
+/* vi:set tabstop=4 shiftwidth=4: */
```
masatake, line 363: Is .editorconfig not enough?

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+		definitions2Keys,
+		ARRAY_SIZE (definitions2Keys),
+	}
+};
+
+static void handleKey(struct sOpenApiSubparser *openapi,
+											 yaml_token_t *token)
+{
+	int i;
+	for (i = 0; i < ARRAY_SIZE(tagSources); i++) {
```
masatake, line 248: Will you maintain this parser？
If the answer is "yes", you can choose your favorite C coding style.
If the answer is "no", I would like you to follow the ctags' coding style like:
```
   for (i = 0; i < ARRAY_SIZE(tagSources); i++)
   {
      ...
```
See the position of `{`.

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+		const struct tagSource* ts = &tagSources[i];
+
+		if (stateStackMatch(
+					openapi->type_stack,
+					ts->keys,
+					ts->countKeys
+					)) {
```
masatake, line 255: Impressive technique. I must rewrite the ansibleplaybook.c with this technique.

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+
+static void inputStart(subparser *s)
+{
+	((struct sOpenApiSubparser*)s)->play_detection_state = DSTAT_INITIAL;
+	((struct sOpenApiSubparser*)s)->type_stack = NULL;
+}
+
+static void inputEnd(subparser *s)
+{
+	Assert (((struct sOpenApiSubparser*)s)->type_stack == NULL);
```
masatake, line 325: Even if a broken input is given to the parser, we can expect the type_stack becomes empty at the end of parsing?

If we cannot expect it, we should clear the stack here instead of doing "Assert".

parsers/openapi.c
```diff
@@ -0,0 +1,363 @@
...
+	if ((token->data.scalar.length == len)
+		&& (strncmp (val, (char *)token->data.scalar.value, len) == 0))
+		return true;
+	else
+		return false;
+}
+
+static enum openapiKeys parseKey(yaml_token_t *token)
+{
+  if (scalarNeq(token, 10, "components"))
```
masatake, line 134: For mapping from "string" to a key, you can use functions defined in main/keyword.h.
I write this here is not for performance. With the function, the code of this parser can be more declarative.

masatake:
This parser is really nice. The technique used in this parser excites me.

In past years, I have been thinking about YPATH, something like XPath but for YAML.
However, I cannot find the time and the way how to implement such a thing.
This parser illustrates the way.
Though the code in this pull request is for OpenAPI. However, most part of the code can be moved to yaml.c/yaml.h.
It will make developing a subparser on the YAML parser much easier.


I'm reading openapi spec slowly. What should be tagged or should not be tagged is not so obvious to me.
It will take more time.

parsers/openapi.c
segoon: fixed

parsers/openapi.c
segoon: it was a copy-pasted line from ansible parser, removed

parsers/openapi.c
segoon: oh, fixed

parsers/openapi.c
segoon: renamed

Units/parser-openapi.r/swagger.d/input.yaml
segoon: added

parsers/openapi.c
segoon: thanks :)

parsers/openapi.c
segoon: ok, done

parsers/openapi.c
masatake: Oh, sorry, I wonder where it came from because I don't use vim:-).

masatake:
@segoon, thank you for updating.
I'm not good at English. So I converted what I wrote here as review comments to C language. See  #3254.
It seems that the many parts of #3254 are already included in the updating.

Units/parser-openapi.r/swagger.d/input.yaml
masatake: The way I wrote is wrong.
In 97915e70b75ca702b916278907d10476394b516d, your parser extracts "title".
However, what I expected is your parser extract "test" in "title: test".

parsers/openapi.c
masatake: I see. Thank you.

masatake:
handleKey and handleValue do the same with different tables.
How about merging them like:
```diff
diff --git a/parsers/openapi.c b/parsers/openapi.c
index 9e26aecdf..493f32ed1 100644
--- a/parsers/openapi.c
+++ b/parsers/openapi.c
@@ -447,13 +447,15 @@ const struct tagSource tagValueSources[] = {
    },
 };
 
-static void handleKey(struct sOpenApiSubparser *openapi,
-                                            yaml_token_t *token)
+static void handleToken(struct sOpenApiSubparser *openapi,
+                       yaml_token_t *token,
+                       const struct tagSource  tss[],
+                       size_t ts_count)
 {
    int i;
-   for (i = 0; i < ARRAY_SIZE(tagSources); i++)
+   for (i = 0; i < ts_count; i++)
    {
-       const struct tagSource* ts = &tagSources[i];
+       const struct tagSource* ts = &tss[i];
 
        if (stateStackMatch(
                    openapi->type_stack,
@@ -473,31 +475,17 @@ static void handleKey(struct sOpenApiSubparser *openapi,
    }
 }
 
+static void handleKey(struct sOpenApiSubparser *openapi,
+                                            yaml_token_t *token)
+{
+   handleToken (openapi, token, tagSources, ARRAY_SIZE(tagSources));
+}
+
 
 static void handleValue(struct sOpenApiSubparser *openapi,
                        yaml_token_t *token)
 {
-   int i;
-   for (i = 0; i < ARRAY_SIZE(tagValueSources); i++)
-   {
-       const struct tagSource* ts = &tagValueSources[i];
-
-       if (stateStackMatch(
-                   openapi->type_stack,
-                   ts->keys,
-                   ts->countKeys
-           ))
-       {
-           // printf("match value! i=%d\n", i);
-
-           tagEntryInfo tag;
-           initTagEntry (&tag, (char *)token->data.scalar.value, ts->kind);
-           attachYamlPosition (&tag, token, false);
-
-           makeTagEntry (&tag);
-           break;
-       }
-   }
+   handleToken (openapi, token, tagValueSources, ARRAY_SIZE(tagValueSources));
 }
 
 static void openapiPlayStateMachine (struct sOpenApiSubparser *openapi,
```

masatake:
I am sorry for not providing the details all at once.
How do you think about #3254 ?
If you are o.k., I will merge #3254.
We can polish the parser after merging it.

segoon:
I'm OK with that.

masatake:
#3254 is merged. Thank you.
Could you make a new pull request for the improvements?

segoon:
here: https://github.com/universal-ctags/ctags/pull/3258

PR #3248, opened 2022-01-02
readtags: add an option to read tags file data from stdin

masatake:
With `-t -` option, readtags reads tags file data from stdin.


masatake:
See #3251.

PR #3249, opened 2022-01-02
Basic: various improvements from Geany

techee:
There were some improvements in the Geany parser, mostly related to extraction of variables in `dim` statements.

Also, the uctags parser determined the used basic dialect based on file extension which wouldn't work very well for Geany as it parses files using MIO buffer. What the parser did appears quite a non-standard approach with the file extensions and I haven't seen it in any other parser. What I did is that I basically just merged all the basic dialect keywords into one and allowed both labels like `label:` and `.label` for any kind of file - hope it's OK.

I added the `dim` extraction code as it was in Geany and then made a few commits on top of that as there were some minor problems in the code and finally made a commit to cleanup the code a bit.

### DIFF
--- a/parsers/basic.c
+++ b/parsers/basic.c
@@ -89,7 +89,7 @@ static char const *extract_name (char const *pos, vString * name)
 	while (isspace (*pos))
 		pos++;
 	vStringClear (name);
-	for (; *pos && !isspace (*pos) && *pos != '(' && *pos != ','; pos++)
+	for (; *pos && !isspace (*pos) && *pos != '(' && *pos != ',' && *pos != '='; pos++)
 		vStringPut (name, *pos);
 	return pos;
 }
--- a/parsers/basic.c
+++ b/parsers/basic.c
@@ -81,6 +81,41 @@ static KeyWord basic_keywords[] = {
  *   FUNCTION DEFINITIONS
  */
 
+static const char *skipToMatching (char begin, char end, const char *pos)
+{
+	int counter = 1;
+	pos++;
+	while (*pos && counter > 0)
+	{
+		if (*pos == end)
+			counter--;
+		else if (*pos == begin)
+			counter++;
+		else if (*pos == '"')
+			pos = skipToMatching ('"', '"', pos) - 1;
+		pos++;
+	}
+	return pos;
+}
+
+static const char *nextPos (const char *pos)
+{
+	if (*pos == '\0')
+		return pos;
+
+	pos++;
+	switch (*pos)
+	{
+		case '(':
+			pos = skipToMatching ('(', ')', pos);
+			break;
+		case '"':
+			pos = skipToMatching ('"', '"', pos);
+			break;
+	}
+	return pos;
+}
+
 static bool isIdentChar (char c)
 {
 	return c && !isspace (c) && c != '(' && c != ',' && c != '=';
@@ -134,7 +169,7 @@ static void extract_dim (char const *pos, BasicKind kind)
 	{
 		/* skip all we don't need(e.g. "..., new_array(5), " we skip "(5)") */
 		while (*pos != ',' && *pos != '\'' && *pos)
-			pos++;
+			pos = nextPos (pos);
 
 		if (*pos == '\'')
 			break; /* break if we are in a comment */
--- a/parsers/basic.c
+++ b/parsers/basic.c
@@ -37,7 +37,6 @@ typedef enum {
 typedef struct {
 	char const *token;
 	BasicKind kind;
-	int skip;
 } KeyWord;
 
 static kindDefinition BasicKinds[] = {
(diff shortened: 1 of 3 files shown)


parsers/basic.c
```diff
@@ -83,6 +82,83 @@ static KeyWord basic_keywords[] = {
...
+{
+	const char *old_pos = pos;
+	while (isspace (*pos))
+		pos++;
+
+	/* create tags only if there is some space between the keyword and the identifier */
+	if (old_pos == pos)
+		return 0;
+
+	vStringClear (name);
```
masatake: How about moving this vStringClear to ...

parsers/basic.c
```diff
@@ -83,6 +82,83 @@ static KeyWord basic_keywords[] = {
...
+			pos++;
+	}
+	/*	capture "dim as string * 4096 chunk" */
+	if (strncmp (pos, "*", 1) == 0)
+	{
+		pos += 1; /* skip "*" */
+		while (isspace (*pos) || isdigit(*pos) || ispunct(*pos))
+			pos++;
+	}
+
```
masatake, line 122: ...here?

parsers/basic.c
```diff
@@ -83,6 +82,83 @@ static KeyWord basic_keywords[] = {
...
+		if (*pos == '\'')
+			return 0; /* break if we are in a comment */
+
+		vStringClear (name);
+		for (; *pos && !isspace (*pos) && *pos != '(' && *pos != ',' && *pos != '='; pos++)
+			vStringPut (name, *pos);
+		makeSimpleTag (name, kind);
+	}
+
+	vStringDelete (name);
```
masatake, line 149: Is it o.k. to delete `name` here?
I think this deletion should be at ...(B)

parsers/basic.c
```diff
@@ -109,6 +185,9 @@ static int match_keyword (const char *p, KeyWord const *kw)
 	name = vStringNew ();
 	p += i;
 
+	if (kw->kind == K_VARIABLE)
+		return extract_dim (p, name, kw->kind); /* extract_dim adds the found tag(s) */
```
masatake: (B)... here like:
```
if (kw->kind == K_VARIABLE)
{
    int r = extract_dim (p, name, kw->kind);
    vStringDelete (name);
    return r;
}

File: parsers/basic.c [hunk 5]
```diff
@@ -140,13 +140,13 @@ static int extract_dim (char const *pos, vString * name, BasicKind kind)
 			pos++;
 
 		if (*pos == '\'')
-			return 0; /* break if we are in a comment */
+			break; /* break if we are in a comment */
```
masatake, line 135: You made a separate commit for these changes.
However, I think this change should be part of the commit " basic: add more advanced extraction of dim-like variables".

If you extend the commit introducing `extract_dim`,  or add a new feature to the commit, making a new commit is greatly appreciated. However, it seems that these changes fix bugs in the base commit, " basic: add more advanced extraction of dim-like variables". In such a case, the commits should be unified.

parsers/basic.c
```diff
@@ -114,7 +114,7 @@ static int extract_dim (char const *pos, vString * name, BasicKind kind)
 		/* now we are at the name */
 	}
 	/* capture "dim as foo ptr bar" */
-	if (strncasecmp (pos, "ptr", 3) == 0 && isspace(*(pos+4)))
+	if (strncasecmp (pos, "ptr", 3) == 0 && isspace(*(pos+3)))
```
masatake, line 109: I think this change should be merged to the commit " basic: add more advanced extraction of dim-like variables".

parsers/basic.c
masatake: After reading the last commit, putting `vStringDelete (name);` here makes sense.
However, I don't like these commit sequences. 

In this commit, I think you did two things: (1) fixing the memory management bug in extract_dim introduced in "basic: add more advanced extraction of dim-like variables", and (2) reorganizing the way to manage a vString storing "name".

(1) should be merged to  "basic: add more advanced extraction of dim-like variables".
So you can focus on (2) in this commit.

I think what I wrote here was a  kind of "Preaching to Buddha".

masatake:
The result of the changes looks good.
However, I think the sequence of commits can be improved.

masatake:
I must say "thank you" first of all.

Do you have any real code written in Basic?
As you may know, I'm gathering "real code" as corpus at https://github.com/universal-ctags/codebase.
I evaluated the improvement for Markdown with the coebase.

I'm using `size:>100000  language:BASIC` as a search expression at Github's code search.

What I found was:
https://github.com/jonatasdemello/gwbasic

If you have any recommendations, I would like to know them.

techee:
> However, I think the sequence of commits can be improved.

It's because I first tried to add the code we had in Geany - which was up to the commit `basic: remove now unused 'skip' KeyWord member` and afterwards started fixing the problems I noticed (which were also present in the Geany code). I'll try to improve it a bit.

> Do you have any real code written in Basic?

Nope, this is my first experience with Basic and hope the last ;-)

Apart from `c.c`, this is probably all we had in Geany that could be interesting for ctags. There's also this Matlab parser

https://github.com/geany/geany/blob/master/ctags/parsers/geany_matlab.c

which doesn't extract all the kinds like the regex parser that ctags uses. Right now I'm not very motivated to improve this parser myself (and we'll probably keep this parser for now because of performance reasons).

techee:
@masatake I hope it's better now.

techee:
I added a fix of this issue we had in Geany: https://github.com/geany/geany/issues/2461

masatake (approved):

masatake:
Thank you very much. Looks fine. I will merge soon.

PR #3250, opened 2022-01-03
main: experiment to detect memory access behind EOL

techee:
Currently when reading the input file, ctags reads it line by line and
maintains a (always growing) vString value for the current line where
characters of the line are stored. Since the string never shrinks,
it may be very long and may contain valid code from previous lines
behind the '\0' character. If parsers contain a bug and read behind
the terminating '\0', this may be undetected by valgrind because
the string is legally allocated.

This experiment re-allocates the string after reading a line to the
exact size of the line which should make valgrind complain if the string
is accessed behind the terminating '\0' by some parser.

Not sure why but the allocated buffer has to be at least 2 bytes long,
otherwise tests start to fail.

### DIFF
--- a/main/read.c
+++ b/main/read.c
@@ -925,6 +925,9 @@ static vString *iFileGetLine (bool chop_newline)
 		if (chopped && !chop_newline)
 			vStringPutNewlinAgainUnsafe (File.line);
 
+		File.line->size = File.line->length == 0 ? 2 : File.line->length + 1;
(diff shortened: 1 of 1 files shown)


techee:
This was inspired by

https://github.com/universal-ctags/ctags/pull/3249/commits/f9cbbba2f22121ada102738de2967155ad65173e

techee:
Hmm, "pity", the valgrind test passed so no bugs were discovered by this. I was "hoping" for some failures :-). Closing.

PR #3252, opened 2022-01-04
Remove GLSL and Ferite from the list of parsers in Geany

techee:
The GLSL parser is just a synonym for the C parser, see

https://github.com/geany/geany/pull/3073

The Ferite language is dead and isn't worth supporting, see

https://github.com/geany/geany/pull/3075


masatake:
Thank you.

PR #3254, opened 2022-01-05
Review result for #3245

masatake:


masatake:
I wonder we can use "nesting level" API as the replacement for the stack.

PR #3256, opened 2022-01-06
buildsys,mingw: Fix link error with PCRE2

k-takata:
Fixes: https://github.com/universal-ctags/ctags-win32/issues/14

When static linking was used and PCRE2 was enabled, link errors occurred.
Fixed it by adding `-DPCRE2_STATIC` to CFLAGS.


masatake (approved):
Thank you.

PR #3257, opened 2022-01-06
packaging,win32: Add PCRE2's LICENCE file

k-takata:
PCRE2 will be linked after #3256.
We should include PCRE2's LICENCE file in the Windows zip package.


masatake (approved):
Thank you for taking care of this.

PR #3259, opened 2022-01-07
packaging,win32: Install autotools

k-takata:
Fixes: universal-ctags/ctags-win32#14

Fix that autotools were not installed.


masatake (approved):

masatake:
Thank you.

PR #3260, opened 2022-01-07
build-sys: Don't use weak symbols on MSYS2

k-takata:
This replaces #3228.

Weak symbols don't work also on MSYS2. Stop using the feature.

`$host` on Cygwin and MSYS2 looks like `x86_64-pc-cygwin` and
`x86_64-pc-msys`. No need to add an `*` at the end of `*-cygwin`


k-takata:
MSYS2 tests passed: https://github.com/universal-ctags/ctags/runs/4738618405?check_suite_focus=true

masatake (approved):

PR #3274, opened 2022-01-16
latex: Add reference tags for environment usage introduced by \begin{MyEnv}

techee:
In Geany we also have environment usages reported by the current parser. I'm not sure if it's interesting for uctags too - I created this pull request where I added these as reference tags. I'm not sure about whether I used the right naming for the role. Also, I made these disabled by default - not sure what the current convention enabling/disabling these kinds of tags is - please check.

### DIFF
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -92,6 +92,7 @@ SystemdUnit    u/unit            RequiredBy         on      referred in Required
 SystemdUnit    u/unit            Requires           on      referred in Requires key
 SystemdUnit    u/unit            WantedBy           on      referred in WantedBy key
 SystemdUnit    u/unit            Wants              on      referred in Wants key
+Tex            e/environment     used               off     environment usage introduced by \begin{MyEnv}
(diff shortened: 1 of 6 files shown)


masatake:
Sorry to be late. I have been thinking about how the name of the role should be. I think "used" is good.

Very interesting.
There is no well-established convention yet. So I would like to accept it as is.
These kinds of trials are welcome. Especially if the trial comes from real demand in a ctags client!

masatake:
It will be nice to add a code testing extracting reference tags in mini-geany.c.

techee:
> There is no well-established convention yet. So I would like to accept it as is.
These kinds of trials are welcome. Especially if the trial comes from real demand in a ctags client!

Yeah, I wasn't sure whether it's something I should submit to uctags as it's a bit strange (the tag names are the same e.g. 'theorem') but then I thought it might be useful for e.g. finding all theorems in document or similar environments.

> It will be nice to add a code testing extracting reference tags in mini-geany.c.

I definitely plan that but there are still some unmerged pull requests in Geany that call ctags functions and I'll do that once we have all the related pull requests merged.

Issue #3271, opened 2022-01-16
freebasic support additions


Glog78:
i had some troubles with the basic ctags so did a small patch which fixed them for me, Here is the patch -> https://termbin.com/m6eb

a.) added declare sub and declare function keywords << not a prefect solution but it makes it so library imported function can be found in the tags.
b.) added namespace support so functions and variables show up under there correct namepsace

it's hacky but for me it works okish, Thanks for the work on ctags cause not every language has already lsif or lsp support.

masatake:
Hi, could you make a pull request for the change?
In addition, we need a test case. https://docs.ctags.io/en/latest/testing-parser.html#units-test-facility explains how to write test cases.

PR #3276, opened 2022-01-17
Basic additions

Glog78:
https://github.com/universal-ctags/ctags/issues/3271#issue-1104975213

this is the requested pull request. i hope it helps

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,9 @@ gnulib/unistd.h
 gnulib/wchar.h
 gnulib/wctype.h
 gnulib/sys/types.h
+compile_commands.json
+libutil.a
+m6eb
+peg/thrift.c
+peg/thrift.h
+.cache/
--- a/parsers/basic.c
+++ b/parsers/basic.c
@@ -326,6 +326,11 @@ static void findBasicTags (void)
 		else
 			match_colon_label (p);
 	}
+
+	/* if there are still lost pointers after the file has been parsed */
+	for(int i=0;i<namespace_depth;i++)
+		vStringDelete(namespaces[i]);
+	namespace_depth=0;
 }
 
 parserDefinition *BasicParser (void)
--- a/parsers/basic.c
+++ b/parsers/basic.c
@@ -31,7 +31,9 @@ typedef enum {
 	K_LABEL,
 	K_TYPE,
 	K_VARIABLE,
-	K_ENUM
+	K_ENUM,
+	K_NAMESPACE,
+	K_NAMESPACE_END
 } BasicKind;
 
 typedef struct {
@@ -45,7 +47,8 @@ static kindDefinition BasicKinds[] = {
 	{true, 'l', "label", "labels"},
 	{true, 't', "type", "types"},
 	{true, 'v', "variable", "variables"},
-	{true, 'g', "enum", "enumerations"}
+	{true, 'g', "enum", "enumerations"},
+  {true, 'n', "namespace", "namespace"}
 };
 
 static KeyWord basic_keywords[] = {
@@ -65,6 +68,12 @@ static KeyWord basic_keywords[] = {
 	{"type", K_TYPE},
 	{"enum", K_ENUM},
 
+	/* freebasic */
+	{"declare sub", K_FUNCTION},
+	{"declare function", K_FUNCTION},
+	{"namespace", K_NAMESPACE},
+	{"end namespace", K_NAMESPACE_END},
+
 	/* blitzbasic, purebasic */
 	{"global", K_VARIABLE},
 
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -117,3 +117,9 @@ gnulib/unistd.h
 gnulib/wchar.h
 gnulib/wctype.h
 gnulib/sys/types.h
+compile_commands.json
+libutil.a
+m6eb
+peg/thrift.c
+peg/thrift.h
+.cache/
```
masatake, line 125: These changes may be nothing to do with improving the support for Basic.
The change itself is acceptable. So could you make a commit for this change or make a pull request for this change?

parsers/basic.c
```diff
@@ -31,7 +31,9 @@ typedef enum {
 	K_LABEL,
 	K_TYPE,
 	K_VARIABLE,
-	K_ENUM
+	K_ENUM,
+	K_NAMESPACE,
+	K_NAMESPACE_END
```
masatake, line 36: Introducing `K_NAMESPACE_END` for representing a keyword is good.
However, introducing `K_NAMESPACE_END` is added to `BasicKind`. Adding `K_NAMESPACE_END` doesn't make sense. Coupling kinds and keywords is not a good idea (though many parsers do so.)
We must split the enum for keywords and that for kinds.

parsers/basic.c
```diff
@@ -162,7 +190,7 @@ static void extract_dim (char const *pos, BasicKind kind)
 
 	for (; isIdentChar (*pos); pos++)
 		vStringPut (name, *pos);
-	makeSimpleTag (name, kind);
+	makeSimpleTagWithNamespace (name, kind);
```
masatake, line 193: In ctags, we call a name combining with its namespace a qualified name.
Qualified names should be emitted only when `--extras=+q` option is given.
If the option is not provided, ctags should print only non-qualified name.
I'm not good at English. So let's me show an example.

input.c:
```c
struct point {
  float x, y;
};
```
```
$ ./ctags -o - /tmp/input.c
point	/tmp/input.c	/^struct point {$/;"	s	file:
x	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
y	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
$ ./ctags -o - --extras=+q /tmp/input.c
point	/tmp/input.c	/^struct point {$/;"	s	file:
point::x	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
point::y	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
x	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
y	/tmp/input.c	/^  float x,y;$/;"	m	struct:point	typeref:typename:float	file:
```
Here `point::x` and `point::y` are the example of the qualified names.
Your `makeSimpleTagWithNamespace` emits qualified names even when `--extras=+` is not given.
It breaks the consistency of the sematics of the ctags' cmdline.

Instead I would like you to look at the `struct:point` in the example output.
We call it "scope field". The namespace for a name should be stored to the scope field of the name.

Storeing the namespace to the scope field of the name is much important than emitting qualified names because, if the scope field for a name is filled, a client tool reading a tags file can generate qualified names by itself.

masatake:
@Glog78, thank you for making a pull request.
I found the original code of basic.c is a bit older.
Your change is stacked on the older code. So I found some inconsistency with the other parts of ctags.

I will update the basic.c and I will adapt this pull request to the new code base.
I will make a pull request when I finish the updating and adapting.
Can I ask you to review the pull request?

.gitignore
Glog78: if you want me too ... the problem was that those file got created while using clangd (lsp) + compile for testing on my maching. So i put them in the gitignore so they won't be inside of the commit ;)

Glog78:
Yeah you can and we also can drop this pull request. I was kindly asked in issue to provide something. I am also not sure if it is clever declare sub and declare function as keyword ... cause basically that is in an extern scope and therefor a definition of a sub or function. If it would be without the extern scope -> it would basically mean that somewhere else in the code the definition of the sub / function will follow. 

Also i am not sure if there would be a need for different basic dialects. It seems that most basic dialects implement things a little different.

external linking in freebasic -> https://www.freebasic.net/wiki/KeyPgExtern
external linking in qb64 -> https://wiki.qb64.org/wiki/DECLARE_DYNAMIC_LIBRARY

i am nearly sure that there are even more versions out there ...

Glog78:
see Basic: extract namespaces (partially derived from #3276) | cleaner implementation for namespaces.

PR #3281, opened 2022-01-19
add installation instuction for clear linux

nobodyatandnothing:


masatake:
Thank you. But we will not introduce the way to install binary packages in README.md because this organization (universal-ctags) doesn't maintain them.
Binary packages for Windows, Mac, and Snap are introduced in README.md because the organization maintains these binary packages.

The instructions for building ctags on clear Linux are welcome. Such instructions should go to https://github.com/universal-ctags/ctags/blob/master/docs/autotools.rst.

PR #3284, opened 2022-01-22
Basic: extract namespaces (partially derived from  #3276)

masatake:
This pull request is inspired by and derived from #3276 submitted by @Glog78 partially.

This pull request doesn't deal with declarations.
The pull request for dealing declarations comes after merging this pull request.

PR #3289, opened 2022-01-29
Basic: extract declarations as reference tags (partially derived from #3276)

masatake:
See #3284.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ depcomp
 gdb-backtrace.txt
 install-sh
 libctags.a
+libutil.a
 Makefile.in
 mini-geany
 mini-geany.exe
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -16,6 +16,7 @@ Automake       d/directory       ltlibrary          on      directory for LTLIBR
 Automake       d/directory       man                on      directory for MANS primary
 Automake       d/directory       program            on      directory for PROGRAMS primary
 Automake       d/directory       script             on      directory for SCRIPTS primary
+Basic          f/function        decl               on      declared
 C              d/macro           condition          off     used in part of #if/#ifdef/#elif conditions
 C              d/macro           undef              on      undefined
 C              h/header          local              on      local header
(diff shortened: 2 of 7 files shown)


masatake:
@Glog78, I could not make a decision about the way for tagging aliases.
This change just extracts declarations as reference tags.

masatake:
@Glog78, this pull request is acceptable?

For trying this pull request, see the instructions in https://docs.ctags.io/en/latest/contributions.html#testing-a-pr-locally-before-being-merged .

Glog78:
@masatake  will check during the weekend ... sorry i am busy with a work related project

masatake:
> @masatake will check during the weekend ... sorry i am busy with a work related project

No problem. Thank you for taking your time for improving ctags.

Glog78:
ok i found quiet a few interesting things....

a.) ptr does count as keyword ? -> https://documentation.help/FreeBASIC/KeyPgPtr.html
b.) i have a variable called subwin :) i guess you should see  the result

i guess we aren't there yet ...

https://github.com/Glog78/freebasic_example 
^^ example project
https://github.com/quangnguyen30192/cmp-nvim-tags
^^ the completion plugin i use to use tags

masatake:
> ok i found quiet a few interesting things....

> a.) ptr does count as keyword ? -> https://documentation.help/FreeBASIC/KeyPgPtr.html

I don't understand this one well.
Do you say whether our parser deals `ptr` as a keyword or not changes the output of ctags?

b.) i have a variable called subwin :) i guess you should see the result

I don't understand this one well.
```
 		declare	function	CreateSubWindow alias "subwin" (byval as win ptr, byval as integer, byval as integer, byval as integer, byval as integer) as win ptr
```

Are you taking about `"subwin"` in this line?

After thinking again and again, I cannot find a good way to deal with aliases because we cannot say whether an alias is a definition or reference. So in this pull request, I don't touch these aliases.

Glog78:
forget my comment ... i was using an old tags file for some reason
the current created tags file looks clean to me :+1: 
again sorry for making you nervous

Glog78:
One question shouldn't be the variables from type _win_st somehow connected to this type ?
currently they are only connected to the namespace if i see it correctly 

like you would want to write 
dim test as curses.curses_priv._win_st
test._notimeout ....

masatake:
Could you use GitHUb's markdown notation like:
```
_win_st
```
or `_win_st`

It helps me to understand your sentences. Sorry for the inconvenience.

Glog78:
We improved the current situation, so and i am not aware that we broke anything which worked before. So yeah i would guess we could merge, even as you noted there is parts missing.

masatake:
>  i am not aware that we broke anything which worked before.

I would like to hear more about this. I guess what you wrote is something to do with your comment:

> One question shouldn't be the variables from type _win_st (any other type for this matter too) somehow connected to this type ?
> currently they are only connected to the namespace if i see it correctly
> 
> like you would want to write
> dim test as curses.curses_priv._win_st
> test._notimeout ....

I didn't understand this one well.
Are you talking about filling `typeref` field?

masatake:
@Glog78, feel free to open an issue for the topic you mentioned in the comments.

PR #3290, opened 2022-01-31
main: fix code resetting fields

### LINKED ISSUE #3283: main: bug in resetting fields
Hello to all,

C++ parser:

```
$ ./ctags.exe --options=NONE -n --kinds-C++=cnsuf --kinds-C++=+p --fields=zKZStse --fields=+a --fields-C++={properties} -u ../test.cpp
```

The content of input file:

```C++
MyClass::MyClass()
{}

void	int foo( int x ){
	return 1;
}

static int foo1( int x ) { return 2; }

static MyClass::foo( int x )		{ return x; }

```

```
ctags.exe: Notice: No options will be read from files or environment
ctags.exe: Warning: Cannot disable fixed field: 'N'{name} in ctags output mode
ctags.exe: Warning: Cannot disable fixed field: 'F'{input} in ctags output mode
ctags.exe: Warning: Cannot disable fixed field: 'P'{pattern} in ctags output mode

```
**Apologies, the previous comment was wrong**
I'm not sure this is a bug, but it's quite strange.
If I add **--fields=+a** to the previous setting, warnings are shown.
If I add 'a' in the first --fields option, i.e, **--fields=zKZStsea**,  it works without warnings.

I found this, because I use an options file for parsing and in some cases want to add access fields option and prototype kind,
but it fails the same way in command line as shown.

Kind regards,
Patrice.
```

```

The version of ctags:

```
$ ctags --version
  Compiled: Dec 23 2021, 14:34:59
  URL: https://ctags.io/
  Optional compiled features: +win32, +wildcards, +regex, +internal-sort, +unix-path-separator, +option-directory, +case-insensitive-filenames, +optscript
```

How do you get ctags binary:

(
Building it locally, via github on Windows,
)

masatake:
Close #3283.

Quoted from #3283:

  $ ./ctags.exe --options=NONE -n --kinds-C++=cnsuf --kinds-C++=+p --fields=zKZStse --fields=+a --fields-C++={properties} -u ../test.cpp
  ...
  ctags.exe: Notice: No options will be read from files or environment
  ctags.exe: Warning: Cannot disable fixed field: 'N'{name} in ctags output mode
  ctags.exe: Warning: Cannot disable fixed field: 'F'{input} in ctags output mode
  ctags.exe: Warning: Cannot disable fixed field: 'P'{pattern} in ctags output mode

This change suppresses these unexpected warnings.


masatake:
It seems that repositories for centos8 are gone.
We have to switch to ubi8.

PR #3292, opened 2022-02-02
lregex: consider aliases of parsers when finding a guest parser

masatake:


masatake:
@Pierre-Andre, I guess this is what you want.

PR #3258, opened 2022-01-06
yaml: add several OpenAPI features

segoon:

PR #3295, opened 2022-02-05
Yaml,OpenAPI: make OpenAPI parser's stack related code reusable

masatake:
The technique used in OpenAPI may be helpful to implement another YAML subparser.
So, I moved the code implementing the technique to the YAML parser experimentally in this change.

The original code uses an array of keywords for pointing out interesting tokens.
This change uses a string for the same purpose. The routines in this change convert the string to an array of keywords.
Till a real specification and its implementation becomes available, we call the format of the string Ypath.


masatake:
I will rebase the changes of #3258 on this pull request.

segoon (approved):

PR #3297, opened 2022-02-14
Elm,Units: Type and constructor tags exclude parameters etc.

niksilver:
This has been tested with

```
make units UNITS=simple-elm
```

which passes.

This is based on a problem encountered in Vim using ctags. It's possible this problem is entirely different (or maybe isn't even a problem!) with a different Vim configuration, a different editor, or a different use case.

To understand the problem being addressed, consider the following code:

```elm
type Param a
    = Cons a
    | Other a

baz : Param Int
baz =
    Cons 42
```

If the cursor is positioned under the `Cons` of the last line, then hitting `Ctrl-]` should take us to the definition of `Cons` on line 2. However, this fails because Vim identifies the tag under the cursor as the pure alphanumeric string `Cons`, but the tag in the tag file is `Cons a`.

Similarly for when the cursor is placed under `Param` in the third line from the bottom.

Also, if the definition of `Param` was

```elm
type Param a =
    Cons a
    | Other a
```

then its tag would be recorded as `Param a =`. Even though that's not the recommended way to format Elm it would still be helpful to generate a useful tag.

So this change is to have tags for types and constructors limited to the initial alphanumeric part, which solves all these problems.


masatake:
@niksilver, thank you.
type variables (type parameters?) like `a` can be stored to `signature:` field.
Do you have an interest in filling the field?

niksilver:
Hello @masatake. Thank you for accepting my code change.

I'm currently trying to build a PackCC parser for Elm which will be more robust than the current regexp parser. As part of that I will see if I can add your suggestions. However, it's very slow going as I'm only a hobbyist developer. So I dare not give a timeframe, but it is something I'm working on.

masatake:
> I'm currently trying to build a PackCC parser for Elm which will be more robust than the current regexp parser. As part of that I will see if I can add your suggestions. However, it's very slow going as I'm only a hobbyist developer. So I dare not give a timeframe, but it is something I'm working on.

Interesting. 

Have you ever considered using the `--_mtable-regex-<LANG>` option?
https://docs.ctags.io/en/latest/optlib.html?highlight=mtable-regex#adding-a-regex-to-a-regex-table

If you are interested in the option, I will write a small initial version of mtable-regex based Elm parser.

niksilver:
Thanks, @masatake, that looks very promising. I'm going to pursue the PackCC parser for a while longer, but will revert to the mtable regexp approach if I feel like I'm losing the battle.

masatake:
@niksilver does your ctags link to pcre2?
I would like to write an initial version of mtable based Elm parser using pcre2.

niksilver:
Apparently no pcre2...

```
$ ./ctags --list-features
#NAME             DESCRIPTION
gnulib_regex      linked with the Gnulib regular expression library
iconv             can convert input/output encodings
option-directory  TO BE WRITTEN
optscript         can use the interpreter
packcc            has peg based parser(s)
regex             can use regular expression based pattern matching
wildcards         can use glob matching
```

PR #3309, opened 2022-03-14
RMarkdown: new parser

masatake:
The parser supports `{r label}` syntax as described in https://bookdown.org/yihui/rmarkdown/.
`{r` part is parsed to choose a guest.
`label` in `label}` part is tagged as a chunkLabel kind object.


masatake:
#3032 must be solved .

PR #3314, opened 2022-03-25
Ruby: extract Class.new as anonymouse tags

masatake:


masatake:
We should have an implementation of trie to make line-oriented parsers faster.

PR #3316, opened 2022-03-26
RDoc: new parser

masatake:
Rdoc:
- [x] fill scope fields
- [x] fill end fields
- [x] set separtor

POD:
- [x] fill scope fields
- [x] fill end fields
- [x] set separtor

...


masatake:
"end" fields must be filled.

PR #3320, opened 2022-03-30
Elm: minor changes

masatake:


masatake:
I'm not sure whether using access: field for storing the information about ports is good or not.
I'll keep this as a draft pull request.

PR #3322, opened 2022-03-31
Elm: fix wrongly shared role definitions between kinds

masatake:
@niksilver, I forgot one important thing in the last review; roles must be defined per kind.
Kinds cannot share a role definition.

```
$ ./ctags --list-roles=Elm
#KIND(L/N)     NAME     ENABLED DESCRIPTION
c/constructor  exposed  on      item exposed from a module
f/function     exposed  on      item exposed from a module
m/module       imported on      module imported
t/type         exposed  on      item exposed from a module
```
is better than
```
$ ./ctags --list-roles=Elm
#KIND(L/N)     NAME     ENABLED DESCRIPTION
c/constructor  exposed  on      item exposed from a module
c/constructor  imported on      module imported
f/function     exposed  on      item exposed from a module
f/function     imported on      module imported
m/module       exposed  on      item exposed from a module
m/module       imported on      module imported
t/type         exposed  on      item exposed from a module
t/type         imported on      module imported
```
, isn't it?

### DIFF
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -41,13 +41,9 @@ DTD            p/parameterEntity condition          on      conditions
 DTD            p/parameterEntity elementName        on      element names
 DTD            p/parameterEntity partOfAttDef       on      part of attribute definition
 Elm            c/constructor     exposed            on      item exposed from a module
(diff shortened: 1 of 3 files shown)


niksilver:
> roles must be defined per kind

I didn't know that. Your change is clearly an improvement - thanks.

PR #3329, opened 2022-04-07
JavaScript: allow class methods get and set

jafl:
`get` and `set` are used to define getters and setters for a class.  However, `get` and `set` can also be method names.  This patch adds support for the latter.


gnulib/.gitignore
```diff
@@ -0,0 +1,2 @@
+fnmatch.h
+stdint.h
```
masatake, line 2: Very minor item.
Could you insert a newline at the end of the last item?

masatake:
Could you remove the last commit [0b5c16a](https://github.com/universal-ctags/ctags/pull/3329/commits/0b5c16a5769e65cd3abbd6bdf66827ffbe52690d)?

Saying frankly, I don't understand what it is. However, I guess, it is not a "must" for the purpose of this pull request.
If I'm wrong, please let me know.

masatake:
@b4n, IF you have time, I would like to review this pull request.

@jafl, thank you.

parsers/jscript.c
```diff
@@ -1653,15 +1653,45 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
+				{
+					TRACE_PRINT("setter");
+					is_setter = true;
+				}
+				deleteToken (savedToken);
+			}
+			else if (isKeyword (token, KEYWORD_get))
+			{
+					TRACE_PRINT("getter");
+				is_getter = true;
```
masatake, line 1686: The style of indentation looks strange. Maybe this depends on a tool you use or the web interface of github.
Anyway, could you look into this?

parsers/jscript.c
```diff
@@ -1653,15 +1653,45 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
+				readToken (token);
+			}
+			else
+			{
+					TRACE_PRINT("setter");
+				is_setter = true;
```
masatake, line 1692: The style of indentation looks strange. Maybe this depends on a tool you use or the web interface of github.
Anyway, could you look into this?

parsers/jscript.c
```diff
@@ -1653,15 +1653,45 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
-		else if (isType(token, TOKEN_KEYWORD) && isKeyword (token, KEYWORD_set))
-		{
-			is_setter = true;
-			readToken (token);
+			if (NextToken == NULL)
+			{
+				tokenInfo *savedToken = newToken ();
+				copyToken (savedToken, token, true);
+				readToken (token);
+				if (isType(token, TOKEN_OPEN_PAREN) && NextToken == NULL)
```
masatake, line 1664: I must verify when whether NextToken can be != NULL in readToken.
If there is such a control flow, we need a test case covering the case.
I must verify this.

parsers/jscript.c
jafl: Fixed by removing the unintentional debugging code

parsers/jscript.c
jafl: Fixed by removing the unintentional debugging code

parsers/jscript.c
jafl: You are correct.  `NextToken` will always be `NULL`.  I updated the commit.

jafl:
Once everything else looks good, I will include the `.gitignore` for `gnulib`, with a newline on the end.

parsers/jscript.c
masatake: Could you put Assert (NextToken == NULL);  ?

masatake:
How about the following change?

A. if the token just read is "TOKEN_OPEN_PAREN", NextToken must be NULL.
    Based on this assumption, we can remove 'NextToken == NULL' from the condition.

B. this parser assumes `if(isKeyword(token, ...) ) {...` has the same meaning `if (isType (token, TOKEN_KEYWORD) && 
isKeyword(token, ...) {...`. Therefore, if you assign non-TOKEN_KEYWORD to `token->type`, you may have to set  `KEYWORD_NONE` to `token->keyword`.

```
index a1be93576..1b8bc3108 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -1659,13 +1659,14 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 			tokenInfo *savedToken = newToken ();
 			copyToken (savedToken, token, true);
 			readToken (token);
-			if (isType(token, TOKEN_OPEN_PAREN) && NextToken == NULL)
+			if (isType(token, TOKEN_OPEN_PAREN))
 			{
 				Assert (NextToken == NULL);
 				NextToken = newToken ();
 				copyToken (NextToken, token, false);	/* save token for next read */
 				copyToken (token, savedToken, true);	/* restore token to process */
 				token->type = TOKEN_IDENTIFIER;			/* process as identifier */
+				token->keyword	= KEYWORD_NONE;
 			}
 			else if (isKeyword (savedToken, KEYWORD_get))
 			{
```

jafl:
Thanks for catching that!

masatake:
Could you wait for a week till merging?
I wonder whether we can get a review from the main developer of the parser.

jafl:
@masatake Note that I cannot merge any pull requests. You will have to do it when you feel comfortable.

masatake:
@jafl, yes, I know.

I approved your pull request explicitly before merging just for recording what I did as the maintainer of this repository. I will merge your pull request this weekend.

masatake:
Thank you.

PR #3331, opened 2022-04-07
JavaScript: always parse the contents of a function block

jafl:
The new test fails without this patch.


masatake:
Thank you for updating the subject. Yes, as the prefix of subject, I expected the name of the parser appeared in the output of `ctags --list-languages`.

masatake (approved):

masatake:
Could you wait for a week till merging?
I wonder whether we can get a review from the main developer of the parser.

Issue #3327, opened 2022-04-07
Splitting c.c into two parsers (c/c++ parser & the parser of the rest)


techee:
I have been thinking about how to proceed with the unification of the c.c parser in uctags and Geany and I think it would help, if we could separate the c/c++ part from the rest of the parsers and concentrate only on the non-c/c++ part. I'm not sure if the c/c++ parsers from c.c are still interesting for uctags e.g. for backwards compatibility but they are something we aren't interested in in Geany. I would suggest the following procedure:

1. Create a copy of `c.c`, e.g. called `c-based.c` that will serve as the base of the non-c/c++ parser.
2. Switch ctags to use `c-based.c` for parsing non-c/c++ languages and make sure ctags compiles and unit tests pass.
3. Remove all the definitions at the top of the parser related to other languages than c/c++ from `c.c`, follow compilation errors and remove the related code from the parser until it compiles.
4. Do the inverse of (2) for `c-based.c`, i.e. remove c/c++ parts and continue like in (2).
5. Do some cleanups.

At that point `c.c` would be left with the old parser and would probably not receive any further maintenance since it's superseded with the cxx parser while `c-based.c` would be the maintained parser for the rest of the languages.

I would do the same for Geany (except for keeping `c.c` as we won't need that one) so we would end up with a simplified `c-based.c` that would hopefully be easier to sync.

I haven't checked but is it still possible to use the old c/c++ parser instead of the cxx parser and is this one covered by some unit tests?

@masatake What do you think about this proposal?

masatake:
Nice idea.
The OldC and OldC++ parsers can be still runnable if you pass an option like --languages=+OldC.
However, we don't need the OldC and OldC++ parsers anymore because C and C++ parsers in the cxx directory work very well.

techee:
Closing in favor of #3332

PR #3332, opened 2022-04-08
split c.c into two parts

techee:
This is what I proposed in

https://github.com/universal-ctags/ctags/issues/3327

with some modifications:

1. I got a bit lazy and didn't remove any code from `c.c` which I left more or less as it was before - I could remove the extra languages if desired, I'm just not sure if it's worth the work since this file probably won't be actively maintained.
2. I also removed Vera from `c-based.c` (and use the parser from `c.c`). The language seems to be mostly dead without any proper documentation and this makes `c-based.c` hard to maintain because nobody knows properly what code is there because of Vera and what it should do. `c.c` seems to be a good graveyard for Vera to me.

I really just fixed the compilation errors and didn't go through the parser in detail - I'm pretty sure that there's some code left which is only specific to C/C++ (but which isn't explicitly guarded for these languages) so there probably are some things which are unnecessary.

### DIFF
--- a/parsers/c-based.c
+++ b/parsers/c-based.c
@@ -1,0 +1,3686 @@
+/*
+*   Copyright (c) 1996-2003, Darren Hiebert
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   This module contains functions for parsing and scanning C, C++, C#, D and Java
+*   source files.
+*/
+
+/*
+*   INCLUDE FILES
+*/
+#include "general.h"        /* must always come first */
+
+#include <string.h>
+#include <setjmp.h>
+
+#include "debug.h"
+#include "entry.h"
+#include "cpreprocessor.h"
+#include "keyword.h"
+#include "options.h"
+#include "parse.h"
+#include "read.h"
+#include "routines.h"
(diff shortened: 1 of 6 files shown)


masatake:
I'm negative about removing the Vera parser.

How about doing as the following?

1. copy c-base.c from c.c as you did.
1.1 remove C/C++ specific code from c-base.c as you did.
1.2 remove parser definitions other than C/C++ from c.c as you did.
2. copy vera.c from c-base.c.
2.1 remove parser definitions other than Vera from vera.c.
3. remove Vera from c-base.c as you did.
4. do as you planed in #3327.

`2`. and `2.1.` are the extra steps. You don't have to touch vera.c more, of course.

techee:
@masatake Done.

masatake:
Could you rebase this pull request on the latest master?
@leleliu008 and I fixed the trouble on CI/CD environment targeting FreeBSD / testing (pull_request) .

techee:
> Could you rebase this pull request on the latest master?

Done.

masatake:
Thank you. Please, wait for awhile.

masatake (approved):
Thank you very much for updating vera.c.

masatake:
Thank you very much.

PR #3337, opened 2022-04-10
buildsys: improve docs/Makefile

leleliu008:
when user just run `make`, no action name is given, we should show help messages.

checking if sphinx-build is installed is involved in certain actions, not all,
for example, if user run `make clean` or `make help`, we do not need to check sphinx-build if is installed.

old code use `which` command, `which` command is not a builtin command in some shell. it is not pre-installed in some os.
we usually use `command` command to check if a command is exist.
`command` command is a builtin command in all different shells.

I removed many lines `@echo`, becuase per line shell code runs in a seperate shell.
I removed them let them run in one shell, it can improve performance.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -8,11 +8,6 @@ PAPER         =
 BUILDDIR      = _build
 EXTDIR        = _ext
 
-# User-friendly check for sphinx-build
-ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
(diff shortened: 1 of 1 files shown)


masatake (approved):
The file was generated from sphinx-autogen, as far as I can remember correctly.
So ideally, the patch should be submitted to the sphinx project.

If what I wrote here is correct, could you add the following sentence to the commit log?
```
TODO: report this to sphinx project.
```

When I cannot find any tasks for improving ctags, I will report:-P.

masatake:
Thank you.

PR #3345, opened 2022-04-12
latex: Clear "lastX" static variables before parsing

techee:
Right now the values in these static variables are preserved even when
parsing another file which is a problem.

For instance consider `a.tex` containing
```
\part{PartA}
\section{SectionA}
\subsection{SubsectionA}
```
and `b.tex` containing
```
\section{SectionB}
\subsection{SubsectionB}
```
The output of `ctags a.tex b.tex` in this case is
```
PartA	a.tex	/^\\part{PartA}$/;"	p
SectionA	a.tex	/^\\section{SectionA}$/;"	s	part:PartA
SectionB	b.tex	/^\\section{SectionB}$/;"	s	part:PartA
SubsectionA	a.tex	/^\\subsection{SubsectionA}$/;"	u	section:PartA""SectionA
SubsectionB	b.tex	/^\\subsection{SubsectionB}$/;"	u	section:PartA""SectionB
```
For instance for `SectionB` the scope is reported as `PartA` even though
there's no `PartA` in the file.

PR #3347, opened 2022-04-12
test case for #3345

masatake:
@techee, could you include this to your #3345?


techee:
Thanks, just added (I wasn't sure if it was possible to create a test for multiple files).

Issue #3341, opened 2022-04-11
man/Makefile.am use GNU extension syntax which is not compatible with BSD make


leleliu008:
`man/Makefile.am` use `Pattern Rules` syntax which is a `GNU make` extension and not supported by `BSD make`.

https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html
https://wiki.netbsd.org/projects/project/bsdmake-enhancements/

leleliu008:
I find that there is no testing about man-pages in our GitHub Actions workflows.

PR #3349, opened 2022-04-14
buildsys: rewite man/Makefile.am

leleliu008:
`man/Makefile.am` use `GNU make` extension syntax which is not compatible with `BSD make`.

`GNU make` support builtin functions and user defined functions, `BSD make` do not support functions.

`GNU make` Pattern-Rules syntax is not compatible with `BSD make`'s.

most `GNU make` supported Automatic Variables that `BSD make` do not support. both support `$@`, but `BSD make` do not support `$<`, `$^` which is often used.

both support `shell assignment operator`(!=), but this feature included in `GNU make` 4.0, which was released in 2013-10-09.
macOS pre-install `GNU make` version is 3.8.1

to summarize, if we want to support `BSD make`, we should do most jobs in shell script.
Makefile just act as a proxy which invoke shell script to do real jobs.

https://www.gnu.org/software/make/manual/html_node/Functions.html
https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html
https://wiki.netbsd.org/projects/project/bsdmake-enhancements/
https://lwn.net/Articles/569832/

### DIFF
--- a/.github/workflows/testing-alpine.yml
+++ b/.github/workflows/testing-alpine.yml
@@ -20,6 +20,8 @@ jobs:
     steps:
       - run: apk add libc-dev gcc make automake autoconf pkgconf python3 file diffutils git
 
+      - run: git config --global --add safe.directory /__w/ctags/ctags
+
       - uses: actions/checkout@v2
 
       - run: cc --version
--- a/makefiles/testing.mak
+++ b/makefiles/testing.mak
@@ -318,12 +318,12 @@ check-genfile:
 	fi
 if HAVE_RST2MAN
 # man/*.in : committed for man pages to be genrated without rst2man
-#   make clean-docs remove both man/*.in and docs/man/*.rst
+#   make clean-docs remove both man/*.rst and docs/man/*.rst
 	$(chkgen_verbose)$(MAKE) -C man clean-docs
-	$(chkgen_verbose)$(MAKE) -C man man-in
+	$(chkgen_verbose)$(MAKE) -C man
 	$(chkgen_verbose)if ! git diff --exit-code -- man; then \
 		echo "Files under man/ are not up to date." ; \
-		echo "Please execute 'make -C man man-in' and commit them." ; \
+		echo "Please execute 'make -C man' and commit them." ; \
 		exit 1 ; \
 	else \
 		echo "Files under man are up to date." ; \
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -1,6 +1,5 @@
 #
 # This is a Makefile.am read by the Autotools.
-# GNU Make is needed.
 #
 #	Copyright (c) 2017, Masatake YAMMATO
 #	Copyright (c) 2017, Red Hat, Inc.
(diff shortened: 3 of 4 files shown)


leleliu008:
Untitled 2

leleliu008:
close #3341

leleliu008:
> It took me more than an hour to remember these code.

Yes, that why I add some links about automake, automake use many implicit rules, if we don't konw about them or forget them, it's difficult to understand the whole logical.

> This is the point we have to discuss.
>
> I've been using Unix-like system more than 30 years. I never use BSD make. I have no idea how it is really import to support it.
I've been encouraging young colleagues to use the features of GNU make (and other GNU tools) to be more productive.

I clearly remeber that this project support `BSD make`. when I first time to contribute to this project, our CI report errors because I used a `GNU make` supported Automatic Variables `$<`. Our CI has `BSD make` checking: https://github.com/universal-ctags/ctags/blob/master/circle.yml#L69

more important is that FreeBSD 13 will trigger following errors when building ctags.
```
make[2]: "/usr/home/vagrant/ctags/man/Makefile" line 1056: Unknown modifier '<'
make[2]: "/usr/home/vagrant/ctags/man/Makefile" line 1056: Unknown modifier ']'
make[2]: Fatal errors encountered -- cannot continue
make[2]: stopped in /usr/home/vagrant/ctags/man
*** Error code 1
```

If we do not support `BSD make`, we must notice user if `BSD make` is used, also must notice deveopers and end-users in our documents.

> > both support shell assignment operator(!=), but this feature included in GNU make 4.0, which was released in 2013-10-09.
> > macOS pre-install GNU make version is 3.8.1
> 
> I don't know about this feature. Are you talking about the `!=` in the following file?
> 
> ```
>     if test -d $(srcdir)/$$file && test $$d != $(srcdir); then \
> ```
> 
> `!=` here is an argument of `test(1)` command.

"!=" shell assignment operator as an alternative to the $(shell ...) function.  Implemented for compatibility with BSD makefiles., $(shell ...)  is frequently used in `GNU make`, but is not supported by `BSD make`.

with "!=" shell assignment operator, we can do very thing in shell. Unfortunately,  this feature included in GNU make 4.0, and on macOS, pre-install GNU make version is 3.8.1

leleliu008:
Screen Shot 2022-04-15 at 3 28 01 PM

.github/workflows/testing-alpine.yml
```diff
@@ -20,6 +20,8 @@ jobs:
     steps:
       - run: apk add libc-dev gcc make automake autoconf pkgconf python3 file diffutils git
 
+      - run: git config --global --add safe.directory /__w/ctags/ctags
+
```
masatake, line 24: What's this?
Is this something to do with BSD make?
I guess this is not for BSD make. Just for fixing an issue building on alpine.
If my guessing is correct, could you consider a pull  request only for alpine?
So, we can merge it quickly. This fix is needed to make #3350 forward.

leleliu008: yes, you are right. I will file a new PR.

masatake:
@hirooih  

> We are not distributing the tar.gz file, because It's not worth the effort to distribute　it.

You are correct.  

@leleliu008 

> If we do not support BSD make, we must notice user if BSD make is used, also must notice deveopers and end-users in our documents.

You are correct.

I'm very comfortable with @hirooih's works, the build file generators for windows, and script synchronizing man -> docs/man. I cannot develop ctags without them. If supporting BSD make breaks the two, I don't want to support BSD make.

leleliu008:
@masatake which file is generated for windows? man/Makefile? Do not support `BSD make` means we just use features that `GNU make` and `BSD make`  both are supported.

hirooih:
@leleliu008 

Before replying to your comments we have to share one more bad news.
`man/Makefile.am` is not only Makefile which is depending on GNU make.
We also have `win32/Makefile`.  This is much complicated one than `man/Makefile.am` and it is much harder to implement by other tools.

***

> If we do not support BSD make, we must notice user if BSD make is used, also must notice developers and end-users in our documents.

I agree with you.  We may need to consider this.

***

>>> if we want to support BSD make,

>> This is the point we have to discuss.

> Our CI has BSD make checking: https://github.com/universal-ctags/ctags/blob/master/circle.yml#L69

I am confused.  According to [this](https://www.freebsd.org/cgi/man.cgi?query=bmake&sektion=1&manpath=freebsd-release-ports) 

> bmake is derived from NetBSD make(1).

Why are we checking "NetBSD make" on Fedora Linux (which should have GNU make)?

> more important is that FreeBSD 13 will trigger following errors when building ctags.

Did FreeBSD 13 stop supporting  GNU make?  I believe the package system (like apt on Ubuntu-Linux) should support it?

***

> both support shell assignment operator(!=), but this feature included in GNU make 4.0, which was released in 2013-10-09.
macOS pre-install GNU make version is 3.8.1

Now I understand what you meant by this sentence.
You did not have to add `man/build.sh` if we could use shell assignment operator(!=).

According to [GNU Make manual, Setting variable](https://www.gnu.org/software/make/manual/html_node/Setting.html), it is close to `FOO := $(shell ...)`.

> If the result of the execution could produce a $, and you don’t intend what follows that to be interpreted as a make variable or function reference, then you must replace every $ with $$ as part of the execution.

In many use cases, this will be a severe constraint.  `man/Makefile.am` maybe OK, but it must be very hard to implement `win32/Makefile` only by this feature.

leleliu008:
on FreeBSD, default make is BSD make, it is pre-isntalled. GNU make can be installed via package manager, but our document do not notice users that tyey must install GNU make.

masatake:
Let's support only GNU make.

> more important, out CI has BSD make checking. so, supporting BSD make policy also make me confused. IMO, I wish only GNU make is supported.

I'm sorry for making you confused. I should remove `fedora30_bmake` from circle.yml.

leleliu008:
I read the `win32/Makefile` souce code. It use GNU makefile syntax, It will be run on `Cyginw` or `MSYS2`,  not be directly run in Windows Terminal or CMD, so there is no any probelm to use shell script. I move man/Makefile.am in shell script will not break win32 platform.

Although people can install BSD make via pacman on `MSYS2`, I think most people do not do that. we can notice users that  building ctags on Windows only supporting GNU make.

only supporting GNU make will benifit developers. Because most developers don't known BSD make that well.

hirooih:
@leleliu008,

> It will be run on Cyginw or MSYS2,

This is not correct. It will be run on any Unix-like environment, Linux, FreeBSD, MacOS, etc.
See my comments on #3353 for some more details.

> not be directly run in Windows Terminal or CMD

This is correct.

leleliu008:
Yes, It can be run on any Unix-like system for developers. For end-users, it will be run on Windows(`Cyginw` or `MSYS2`)

leleliu008:
I think we should make a distinction between developers and end-users. we are developers, we are not end-users, end-users is people who use our project, we should make our project easy to build, because end-users shouldn't known about much details, but we (developers) should know.

leleliu008:
I think documents should divide into two part, one for developers, another for end-users.

hirooih:
> I think documents should divide into two part, one for developers, another for end-users.

I divided it as you wrote already.

http://docs.ctags.io/en/latest/

The first half is for end users.  The second half, after "Extending ctags with a parser written in C", is for developers.

leleliu008:
@hirooih Thank you for your great work.

leleliu008:
Screenshot from 2022-04-16 12-52-16

here you do not motion win32/Makefile, I guess win32/Makefile is for developers, not for end-users, actually, you do not want end-users build from source, beacuse you wish end-users use daily builds. am I correct?

leleliu008:
why we provide prebuild-binaries for Windows? but do not provide prebuild-binaries for other platforms? Because of the License?

hirooih:
> I correct?

No.
The section "Building ctags" is the first part of the user part.

`win32/Makefile` generates files used by `namke -f mk_mvc.mak`.  The files are one of source files for end users.

> why we provide prebuild-binaries for Windows?

Because we have a volunteer doing it.

> but do not provide prebuild-binaries for other platforms?

Because we don't have a volunteer doing it.

leleliu008:
But in document do no say user should run `make` to generate ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters, not say `make` must be `GNU make`, not say `make` must be run in Unix-like Environment(`Cyginw` or `MSYS2`)

leleliu008:
`nmake -f mk_mvc.mak` must be run in Windows Terminal or CMD. why mk_mvc.mak use windows commands but win32/Makefile use Unix commands?

leleliu008:
maybe we should list which feature need volunteer in our documents to find volunteer.

hirooih:
> win32/Makefile generates files used by namke -f mk_mvc.mak. The files are one of source files for end users.

Let me repeat this specifically.

A developer updates ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters by using win32/Makefile on Unix-environment, and commit the files in the Git repository.
A end user run `nmake -f mk_mvc.mak` in Windows Terminal or CMD.  During this process ctags_vs2013.vcxproj and ctags_vs2013.vcxproj.filters are used.

I hope this helps.

leleliu008:
I appreciate your patience in answering all my questions, and guiding me through the process. Frankly speaking, I actually wanna know why? As a developer, I can accapt it, as an end-user, I'm confused.

PR #3340, opened 2022-04-11
GitHub Actions: update freebsd vm

masatake:
Signed-off-by: Masatake YAMATO <yamato@redhat.com>

PR #3336, opened 2022-04-10
GitHub Actions: change FreeBSD setup action

leleliu008:
old FreeBSD setup action is too complex, and can not set FreeBSD version.
new FreeBSD setup action is already used for setup OpenBSD and NetBSD.

old FreeBSD setup action: https://github.com/vmactions/freebsd-vm
new FreeBSD setup action: https://github.com/leleliu008/github-actions-vagrant

Signed-off-by: leleliu008 <leleliu008@gmail.com>

PR #3353, opened 2022-04-15
Limit the support for building with bsd make

masatake:
See #3349.

TODO for merging this:

- [X] make everything GREEN
- [x] revise https://github.com/universal-ctags/ctags/commit/e925b5597d8d9923c122ee29e12b65cdc1e182c3 
   https://github.com/universal-ctags/ctags/commit/e925b5597d8d9923c122ee29e12b65cdc1e182c3 ~should be converted to two commits:~

    ~skip the test case requiring GNU grep if the grep implementation is not GNU grep, and~
    ~install gnugrep in FreeBSD.~
    Just skipping the test case if grep doesn't work expectedly.
    
- [x] Use [eli-schwartz](https://github.com/eli-schwartz)'s Makefile trick for running fedora30_bmake. `fedora30_bmake` is fo verifying ctags command can be built even with bmake. We will not advertise we support building ctags common with bmake. It i just for keeping us NOT using gmake features too aggressively.  

TODO solved in another pull request:

- [ ] update README.md (#3354)
- [ ] require GNU grep and GNU diff more explicit way
- [ ] run "make check" on NetBSD and OpenBSD (#3341)


masatake:
@techee, we have a plan to remove supporting BSD make.
For generating man pages and build-scripts used in win32, we assume GNU make.
As a Geany developer, do you have any concert about this direction?

hirooih:
@masatake san,

I've checked why `fedora30_bmake` which run `make check` does not fail.
It is because `man/Makefile` is used only when `rst2man` is installed and `check-genfile` target in `testing.mak' checks win32/ctags_vs2013.vcxproj* only when GNU make is used.  It makes sense to me.

I think these tests are still valuable.  How about just updating comments something like this?;

```diff
diff --git a/circle.yml b/circle.yml
index be85e0b9..3ce0536d 100644
--- a/circle.yml
+++ b/circle.yml
@@ -66,16 +66,19 @@ jobs:
              make roundtrip
 
   #
-  # On Fedora 31 bmake package cannot be installed because a package
-  # required by bmake is not available. More inspection is needed.
-  # Fedora 30 is the last version where bmake can be installed.
+  # We assume GNU make as the `make` command as we use GNU make extension
+  # aggressively in man/Makefile and win32/Makefile.
   #
-  # Though we use GNU make extension aggressively in man/Makefile,
-  # we can still build ctags with bmake. The man/Makefile is for
-  # generating man pages and associated html files without using
-  # autotools. Makefile generated by autotools converts .rst.in files
-  # to .rst files. It injects proper name of ctags executable.
-  # man/Makefile doesn't do so.
+  # man/Makefile is used to generate man files only when rst2man is installed. 
+  # win32/Makefile is used only when either source.mk or win32/*.in is updated.
+  # In other cases we can still build ctags without GNU make.
+  #
+  # The following tests use bmake (NetBSD make) to check if GNU make extensions
+  # are not used unintentionally.
+  #
+  # On Fedora 31 bmake package cannot be installed because a package required by
+  # bmake is not available. More inspection is needed. Fedora 30 is the last
+  # version where bmake can be installed.
   #
    fedora30_bmake:
      working_directory: ~/universal-ctags
```

techee:
> @techee, we have a plan to remove supporting BSD make.
For generating man pages and build-scripts used in win32, we assume GNU make.
As a Geany developer, do you have any concert about this direction?

We  don't use uctags build for Geany - ctags is built using the Geany build system together with the rest of Geany (autoconf and now also meson). To be honest, I don't know what make we support in Geany but it shouldn't matter for this PR. So no problem here.

leleliu008:
If we only support GNU make, maybe use `GNUmakefile.am` and `GNUmakefile` is better.

https://www.gnu.org/software/make/manual/make.html#What-Name-to-Give-Your-Makefile

if we do not want to change these names, I think we should notcie users something like following after Makefile is generated:

===========================================================================
now you can run make or gmake, when you run as make, besure it is GNU make.
===========================================================================

masatake:
Thank you for input.
I will update my changes.

leleliu008:
I think we can add following in our README.md

## dependencies
|tool|required?|purpose|note|
|----|---------|-------|----|
|[automake](https://www.gnu.org/software/automake/)|required |for generating Makefile.in from Makefile.am||
|[autoconf](https://www.gnu.org/software/autoconf/)|required |for generating configure script||
|[pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/)|required|for finding libraries|[pkgconf](http://pkgconf.org/) can be installed as a alternative.||
|[gmake](https://www.gnu.org/software/gmake/)|required |bmake is not supported.||
|[python>=3.5](https://www.python.org/)|optional|for providing a runtime for `docutils`||
|[docutils](https://docutils.sourceforge.io/)|optional|for generating manpages|if can't found, manpages will not be installed.|

|library|required?|purpose|note|
|-|-|-|-|
|[libiconv](http://www.gnu.org/software/libiconv/)|optional|text coding convert support.|this library have been already included in some platform.|
|[pcre2](https://github.com/PhilipHazel/pcre2)|optional|regex support|default regex library is [gnu-regex](https://www.gnu.org/software/regex/) which is included in our source code. if you use pcre2, you should know the differences between [POSIX Regular Expresion syntax](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html) and [Perl5 Regular Expresion syntax](https://perldoc.perl.org/perlre).|
|[libxml2](http://xmlsoft.org/)|optional|XML support||
|[libyaml](https://github.com/yaml/libyaml/)|optional|YAML support||
|[jansson](https://github.com/akheron/jansson)|optional|JSON support||
|[libseccomp](https://github.com/seccomp/libseccomp/)|optional|sandbox support|only available on GNU/Linux|


Screenshot from 2022-04-17 12-59-06

**here motioned libtool, actually, libtool is not used in our project.**

leleliu008:
Screenshot from 2022-04-17 12-59-06

**here motioned libtool, actually, libtool is not used in our project.**

masatake:
@hirooih:

> I've checked why fedora30_bmake which run make check does not fail.
> It is because man/Makefile is used only when rst2man is installed and check-genfile target in `testing.mak' checks win32/ctags_vs2013.vcxproj* only when GNU make is used. It makes sense to me.
> 
> I think these tests are still valuable. How about just updating comments something like this?;
...

At the first glance, this is a good idea. However, I don't want to pay the cost of the maintenance for keeping our files working well with make.

masatake:
I will consider putting the dependency table in another pull request.

README.md
```diff
@@ -57,7 +57,7 @@ code from GitHub.
 
 ## How to build and install ##
 
-To build with Autotools (Autoconf, Automake, and Libtool) on GNU/Linux, OSX, or Windows 10 WSL,
+To build with Autotools (Autoconf, and Automake) on GNU/Linux, OSX, or Windows 10 WSL,
```
hirooih, line 60: Now we don't need a semicolon.
> (Autoconf and Automake)

hirooih (commented):
@masatake 
Do you mandate to rename Makefile to GNUmakefile if it uses GNU make extensions?
I've never seen a Makefile named as GNUmakefile before. I don't see any actual merit of changing name.

I will follow your decision.

masatake:
>  Do you mandate to rename Makefile to GNUmakefile if it uses GNU make extensions?
Yes.
I tried renaming all Makefile to GNUmakefile. However, I got another failure in CI/CD environments.
So I limited the renaming.

> I've never seen a Makefile named as GNUmakefile before. I don't see any actual merit of changing name.

I'm not sure how it is popular. Anyway, it is explained in the official GNU make manual:
https://www.gnu.org/software/make/manual/make.html#Makefile-Names

We can let bmake fail explicitly with the renaming because bmake cannot find Makefile.
If we don't rename the file, bmake can process the Makefile, and it will fail in a syntax error.
In the former case, a user trying building ctags may not report the failure as an issue; I expect the person understand we don't support bmake when finding GNUmakefile in the output of "$ ls man".

In the latter case, the person may report the failure as an issue with a title like "Wrong syntax in Makefile".

README.md
masatake: Thank you. I will update my change.

leleliu008:
@masatake It seems that you forgot `testing-freebsd.yml`. or you intentionally left it out for some reasons?

masatake:
> @masatake It seems that you forgot testing-freebsd.yml. or you intentionally left it out for some reasons?

I forgot it. I wonder why "run in FreeBSD VM" doesn't report failure. I will inspect the reason.

eli-schwartz:
I've seen GNUmakefile used before. I've also seen it with the following neat trick for BSD support (not BSD make support! Just BSD support):

Create the following Makefile:
```make
all:
	gmake $@

.DEFAULT:
	gmake $@

.PHONY: all
```

On systems where `make` is GNU make, it ignores this Makefile entirely, because GNUmakefile is preferred, and everything proceeds as normal.

On systems where `make` is not GNU make, and users accidentally run some *other* version of Make, the GNUmakefile is (obviously) not respected. The Makefile is read instead. And the Makefile says "no matter what rule is passed, re-run it using the gmake program".

leleliu008:
@masatake the reason I have written in https://github.com/universal-ctags/ctags/pull/3340#issuecomment-1095115445

leleliu008:
It's time to merge #3336

masatake:
@leleliu008, oh I see. I will look into the pull requests this after noon JST.

masatake:
@leleliu008, I cherry-pick'ed your change in #3336 into here, #3353.

masatake:
Hidden bugs are exposed. I will inspect them.

hirooih:
@masatake

I'm writing this to make my intension in my previous post clear. I'm not against your decision.

> We can let bmake fail explicitly with the renaming because bmake cannot find Makefile.
> If we don't rename the file, bmake can process the Makefile, and it will fail in a syntax error.
> In the former case, a user trying building ctags may not report the failure as an issue; I expect the person understand we don't support bmake when finding GNUmakefile in the output of "$ ls man".
> 
> In the latter case, the person may report the failure as an issue with a title like "Wrong syntax in Makefile".

Both cause error.

I guess you meant running `ls` command by "$ ls man".  This works only if the user knows that GNUmakefile is only for GNU make.
It makes a difference if he/she knows about this feature AND reports without reading the README.md.  I think this is very rare case.

And this change makes testing with `bmake` more important to check we must rename Makefile to GNUmakefile or not.

@eli-schwartz 

Interesting trick, thank you.  Shall we add man/Makefile and win32/Makefile?

masatake:
GNU:
```
$ echo ab | grep 'a\|b'
ab
```

OpenBSD:
```
$ echo ab | grep 'a\|b'
$ echo ab | grep 'a|b'
```

How can I use `|` operator of regex in OpenBSD?

hirooih:
You have to use egrep.

[FreeBSED Manual Pages, re_format](https://www.freebsd.org/cgi/man.cgi?query=re_format&sektion=7&apropos=0&manpath=FreeBSD+13.0-RELEASE+and+Ports)

> Regular expressions ("REs"), as defined in	IEEE Std 1003.2	("POSIX.2"), come in
> two forms:	modern REs (roughly those of egrep(1); 1003.2 calls these "extended"
> REs) and obsolete	REs (roughly those of ed(1); 

...

> A (modern)	RE is one<**> or more non-empty<**> branches, separated	by
> `|'.  It matches anything that matches one	of the branches.

from man of GNU grep

> In GNU  grep  there is  no difference in available functionality between basic and extended
> syntaxes.  In other implementations, basic regular expressions are less powerful.

Using only GNU tools on any OSs makes me happy :-)

masatake:
@hirooih, thank you.

In the document you introduced, I found a critical sentence:
```
Obsolete ("basic")	regular	expressions differ in several respects. `|' is an ordinary character and there	is no equivalent for its functionality.
```

I'm very surprised at this.

> Using only GNU tools on any OSs makes me happy :-)

.. it seems you are correct.

To make merge this pull request quickly, I will introduce a quick hack.

eli-schwartz:
`egrep` is a lie, and `grep -E` is how you're supposed to do it.

The former is obsolescent since at least 1997 and is not available everywhere. Some systems do have it as a shell script that runs `grep -E "$@"`.

leleliu008:
> @leleliu008, I cherry-pick'ed your change in #3336 into here, #3353.

Feel free to do whatever you like, as long as it can help you progress at this PR as quickly as possible.

masatake:
By specifying GGREP=/usr/local/bin/grep on FreeBSD, I successfully solved one of the failures.
But `Tmain/e-ctags-output.d` still fails. It is not reproduced on my freebsd13 VM running on local KVM.
I have to inspect the failure on the CI environment with printf.

WARNING: I will do "push --force" many  "TEMP" commits.

masatake:
@eli-schwartz, thank you for input.

leleliu008 (approved):

masatake:
In [e925b55](https://github.com/universal-ctags/ctags/pull/3353/commits/e925b5597d8d9923c122ee29e12b65cdc1e182c3), I introduced a new environment variable, GGREP to use GNU grep in FreeBSD.
Introducing GGREP is not a bad idea. However, on FreeBSD, GNU grep is installed to `/usr/local/bin/grep`, so I guess we don't have to specify its absolute path to GGREP; we may not need GGREP.

e925b55 should be converted to two commits:
1. skip the test case requiring GNU grep if the grep implementation is not GNU grep, and
2. install gnugrep in FreeBSD.

masatake:
@hirooih,

> And this change makes testing with bmake more important to check we must rename Makefile to GNUmakefile or not.

I cannot parse the sentence well.
Do you mean:

>  this change makes testing with bmake more important whether we must rename Makefile to GNUmakefile or not.

?

If the answer is yes,  I'm not sure what "this change" points to. The change removing "fedora34_bmake"?

hirooih:
@masatake 

>  I'm not sure what "this change" points to. The change removing "fedora34_bmake"?

My sentence was not clear, sorry.

"This change" pointed to "mandating to rename Makefile to GNUmakefile if it uses GNU make extensions".

***

Why don't define "GREP = grep -E" in Makefile?  I think this is what we need.

@eli-schwartz

Thank you for your info.

> egrep is a lie, and grep -E is how you're supposed to do it.
> 
> The former is obsolescent since at least 1997 and is not available everywhere.

This is not correct.  On my Ubuntu 20.04.

```
$ type egrep
egrep is /usr/bin/egrep
```

According to [Why these variants of "egrep, fgrep, and rgrep" are deprecated? [closed]](https://stackoverflow.com/questions/45461653/why-these-variants-of-egrep-fgrep-and-rgrep-are-deprecated)

egrep and fgrep were removed from POSIX. 

> These variants are deprecated, but are provided for backward compatibility.

Anyway I agree with you.  We should use `grep -E` instead of `egrep`.

eli-schwartz:
> This is not correct. On my Ubuntu 20.04.

My statement was about whether it is "everywhere". If "everywhere" has become synonymous with "what Ubuntu does", then that is news to me...

Ubuntu is certainly somewhere, and it does provide compatibility shims (for egret and fgrep), like I said. More specifically, GNU grep installs those shims, and Ubuntu doesn't remove them. In GNU grep 3.8, those commands will issue an obsolescence warning every time you use them, and their documentation (which currently states that they exist solely for backwards compat with pre-POSIX scripts) has been removed.

There are most definitely systems out there that don't possess the commands at all, and those systems, too, are somewhere.

Together, the systems that do have it and the systems that don't have it combine into "everywhere", which as a representative group cannot be said to guarantee that it exists and is had.

> egrep and fgrep were removed from POSIX.

grep was added in SUSv2 (which is not POSIX! v3 is POSIX) and, at the same time, egrep and fgrep were added to a "legacy" section, which is optional and the use of is recommended against.

SUSv3 (also known as POSIX 2001) did not have egrep or fgrep, hence they have never been part of POSIX, even if POSIX does mention in the footnotes for grep itself (which is part of POSIX) that they are "likely" to exist as vendor extensions "for historical reasons".

The stackoverflow question does not claim that egrep/fgrep was removed from POSIX. ;) 

Anyway yes, we are all agreed to not use it (and if you do use it, once GNU grep 3.8 is released you'll start getting ugly warnings on stderr telling you to stop).

hirooih:
@eli-schwartz 

I'm sorry for my poor English skill.

I read the following sentence;

> is not available everywhere.

as

> is not available anywhere.

> Anyway yes, we are all agreed to not use it 

Yes.

> (and if you do use it, once GNU grep 3.8 is released you'll start getting ugly warnings on stderr telling you to stop).

No problem.  I always use GNU tools. I need only `grep`:-)

Tmain/utils.sh
```diff
@@ -135,3 +135,10 @@ direq ()
     [ "$(cd ${1} && pwd)" = "$(cd ${2} && pwd)" ]
     return $?
 }
+
+skip_if_grep_doesnt_work_expectedly()
+{
+	if ! ( echo 'MatLab   *.m' | grep '\*\.m\>.*$' > /dev/null ); then
```
masatake, line 141: Ugly.

masatake:
All green again.

Tmain/utils.sh
eli-schwartz: I don't understand why you are jumping through such excessive and unwieldy hoops.

The issue is that POSIX `grep -E` is not the default behavior of POSIX grep, but it is the default for GNU grep. Right? So... why not just explicitly specify -E everywhere?

hirooih:
> So... why not just explicitly specify -E everywhere?

Yes. This is what I wrote as;

> Why don't define "GREP = grep -E" in Makefile? I think this is what we need.

(I should write "Why don't we ...".  Sorry.)

Tmain/utils.sh
masatake: Thank you. It seems that grep -E doesn't work well on FreeBSD 13 though it works well on FreeBSD 12.

masatake: grep command on FreeBSD13 has -E option.
`grep --version` reports:
```
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
```
`grep --help` says:
```
usage: grep [-abcDEFGHhIiLlmnOoPqRSsUVvwxz] [-A num] [-B num] [-C[num]]
+	[-e pattern] [-f file] [--binary-files=value] [--color=when]
+	[--context[=num]] [--directories=action] [--label] [--line-buffered]
+	[--null] [pattern] [file ...]
```

```
echo 'MatLab   *.m' | grep -E '\*\.m\>.*$'
```
This pipeline prints nothing.
This is the reason why the failure occurs only on FreeBSD.

The pattern may include a subpattern supported in a modern GNU extension.

masatake: I collected the version information of grep on FreeBSD12:
```
grep (GNU grep) 2.5.1-FreeBSD
```

It seems that the pattern used in the test sports an incompatible aspect of the compatible implementation.

If you know an alternative pattern for the pattern, a pull request is welcome.
However, this pull request becomes a showstopper in parser developments.
So I will merge the original change.

masatake: I use -E after testing: 2457f86.

masatake:
Thank you for people advising and reviewing.
I will merge this.

You may have more things to say:-P.
In that case, open an issue or send a pull request.

hirooih:
@masatake 

Otsukare-sama-desita.  (Thank you for your hard work.)

> echo 'MatLab   *.m' | grep -E '\*\.m\>.*$'
> This pipeline prints nothing.
> This is the reason why the failure occurs only on FreeBSD.

Now I understand the issue. And this is interesting.

[FreeBSD 13.0-RELEASE Release Notes](https://www.freebsd.org/releases/13.0R/relnotes/)

> The BSD version of [grep(1)](https://www.freebsd.org/cgi/man.cgi?query=grep&sektion=1&format=html) is now installed by default. The obsolete GNU version that was the previous default has been removed.

Oh my goodness...

[grep(1)](https://www.freebsd.org/cgi/man.cgi?query=grep&sektion=1&format=html)

> See [re_format(7)](https://www.freebsd.org/cgi/man.cgi?query=re_format&sektion=7&apropos=0&manpath=FreeBSD+13.0-RELEASE+and+Ports) for more information on regular expressions.

[re_format(7)](https://www.freebsd.org/cgi/man.cgi?query=re_format&sektion=7&apropos=0&manpath=FreeBSD+13.0-RELEASE+and+Ports)

> There are two special cases<**> of bracket expressions: ...
> The additional word delimiters `\<` and `\>` are provided to ease compatibility with traditional SVR4  systems but are not portable and should be avoided.

"should be avoided" but it should work. It seems this is a bug of "the BSD version of grep".

```
. ../utils.sh
if ! ( echo 'MatLab   *.m' | grep -E '\*\.m\>.*$' > /dev/null); then
	skip "'grep -E' needed in this case doesn't work expectedly"
fi
```

This is a good idea.

PS
Another surprise in FreeBSD 13.0-RELEASE Release Notes.

> Removed the obsolete binutils 2.17 and [gcc(1)](https://www.freebsd.org/cgi/man.cgi?query=gcc&sektion=1&format=html) 4.2.1 from the tree. All supported architectures now use the LLVM/clang toolchain.

eli-schwartz:
POSIX doesn't define word boundaries, but given FreeBSD does seem to be violating its own documentation there that's hopefully something slightly fixable... :)

Issue #1933, opened 2018-11-13
JavaScript: Object functions containing function calls are not recognized 


gpavlov2016:
Example:
```
testobj = {
  func1: () => {
    console.log('test1');
  },
  func2: () => {
    console.log('test2');
  }
};
```
In this case func1 is recognized but func2 is not.
Interesting that if the call inside func1 is removed then both func1 and func2 are recognized:
```
testobj = {
  func1: () => {
  },
  func2: () => {
    console.log('test2');
  }
};

```

masatake:
What language are you talking about?

gpavlov2016:
JavaScript ES6

gpavlov2016:
> What language are you talking about?

JavaScript ES6

masatake:
@b4n, can I ask you to look at this?

masatake:
```
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 458b5cb1..bb75637d 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
Following patch makes ctags capture func2. Howerver, I cannot convince myself this is the correct fix.
Let's wait how @b4n says.

@@ -1173,13 +1207,13 @@ static void addToScope (tokenInfo* const token, const vString* const extra)
  */
 
 static bool findCmdTerm (tokenInfo *const token, bool include_newlines,
-                            bool include_commas)
+                            bool include_commas, bool ignore_semicolon)
 {
 	/*
 	 * Read until we find either a semicolon or closing brace.
 	 * Any nested braces will be handled within.
 	 */
-	while (! isType (token, TOKEN_SEMICOLON) &&
+	while (! ((! ignore_semicolon) && isType (token, TOKEN_SEMICOLON)) &&
 		   ! isType (token, TOKEN_CLOSE_CURLY) &&
 		   ! (include_commas && isType (token, TOKEN_COMMA)) &&
 		   ! isType (token, TOKEN_EOF))
@@ -1381,7 +1415,7 @@ static bool parseIf (tokenInfo *const token)
 	{
 		/* The next token should only be read if this statement had its own
 		 * terminator */
-		read_next_token = findCmdTerm (token, true, false);
+		read_next_token = findCmdTerm (token, true, false, false);
 	}
 	return read_next_token;
 }
@@ -1440,7 +1474,7 @@ static void parseFunction (tokenInfo *const token)
 			makeFunctionTagCommon (name, signature, is_generator, is_anonymous);
 	}
 
[...]

Issue #1389, opened 2017-05-18
Javascript: capturing methods


masatake:
`init` in following lines is not captured. 
What kind of heuristic can I use to capture it?
Could you give me hints?

```
var Namespace = Ember.Namespace = Ember.Object.extend({
  isNamespace: true,

  init: function() {
    Ember.Namespace.NAMESPACES.push(this);
    Ember.Namespace.PROCESSED = false;
  }
});
```

ptzz:
This does the trick for exuberant ctags: https://github.com/romainl/ctags-patterns-for-javascript.

Might need some adaptation to work with universal ctags.

I was planning to switch from exuberant ctags to universal ctags due to the poor javascript support in exuberant ctags. Unfortunately, for my use cases universal ctags doesn't improve upon it much. This and the lack of releases (i.e. slow and unpredictable install process) made me stay with exuberant ctags. Discussions about versioning and releases have been going on for almost three years (#228) but seem to be going nowhere. Hampers adoption for sure.

b4n:
Technically `init` is a method on the unnamed object passed as argument to `Ember.Object.extend`, so the right solution would be to tag the unnamed object and simply parse the `{}` as usual.  But people might not expect the result, because JS is a so dynamic language you actually can't know much before actually *running* the code…

masatake:
> This does the trick for exuberant ctags: https://github.com/romainl/ctags-patterns-for-javascript.

Interesting project.
```
--regex-javascript=/^[ \t]*([A-Za-z0-9_$]+)[ \t]*[:=][ \t]*[\(]*function[ \t]*\(/\1/M,Method,Methods/
```

This pattern captures `init` well. However, it doesn't fill its scope field.
Capturing a method without scope information is the same as capturing it as a function.
I wonder how useful it is.
The javascript parser of u-ctags can make more useful result, can't it?

@b4n, I get following idea based on your comment:

I will introduce Ember subparser based on Javascript parser.

If Ember parser detects an unnamed object is passed to `Ember.Object.extend`, Ember parser tags it as a class kind item with an anonymous name. As the result Javascript parser captures init as a method.

I'm not sure whether Ember subparser should be introduced or not. If this code pattern is so popular, more generic mechanism is needed for tagging. 

Per parser is the last resort in u-ctags. `--param-Javascript:func-taking-anon-object-as-vtable` can be introduced.

`--param-Javascript:func-taking-anon-object-as-vtable=Ember.Object.extend:1` means 
[...]

PR #3357, opened 2022-04-17
JavaScript: capture methods inside objects

jafl:
This fixes issues #780, #1389, #1933


masatake:
Thank you.
I will point out more minor things for the person who has contributed to the project more than twice:-)

We use some minor conventions in the way to write commit headers.
About 2f9819c, could you change the header to
```
Javascript,cosmetic: fix indentation
```
See http://docs.ctags.io/en/latest/contributions.html?highlight=cosmetic#title-of-commit-log-and-pull-request .

parsers/jscript.c
```diff
@@ -1658,27 +1652,27 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 		else if (isType (token, TOKEN_KEYWORD) &&
 				 (isKeyword (token, KEYWORD_get) || isKeyword (token, KEYWORD_set)))
 		{
-			tokenInfo *savedToken = newToken ();
-			copyToken (savedToken, token, true);
+			tokenInfo *saved_token = newToken ();
```
masatake, line 1679: Could you make another commit for changing the variable name?

BTW, we don't define a convention about the names having an in-file scope.
This change, /saveToken/saved_token/ is acceptable.

masatake:
Could you see the result of testing at "run units target under VALGRIND / testing"?
I'm not sure whether a contributor can see the result of the test because I'm a member of this organization.

It reported a memory leak.

Are you working on GNU/Linux?
If yes, you can run the test cases locally.

```
1. install valgrind.
2. run "make units LANGUAGES=JavaScript VG=1"
```

jafl:
Testing the fix for the memory leak.  I split out the code cleanup into a separate PR.  Once that is approved and merged, then I will update this PR with only the functional changes.

masatake:
Thank you for updating the pull reuqest. I'll find the time for reviewing.

masatake:
I'll look into this.

masatake:
I have tried. Your change works very well.
As usual, I would like to start reviewing the change from very minor topics.

Units/parser-javascript.r/github-issue-1389.d/input.js
```diff
@@ -0,0 +1,10 @@
+var Namespace = Ember.Namespace = Ember.Object.extend=({
+  isNamespace: true,
+
+  init: function() {
+    Ember.Namespace.NAMESPACES.push(this);
+    Ember.Namespace.PROCESSED = false;
+  }
+});
+
+// https://github.com/universal-ctags/ctags/issues/1389
```
masatake, line 10: Thank you!

Units/parser-javascript.r/github-issue-1933.d/input.js
```diff
@@ -0,0 +1,11 @@
...
+  func1: () => {
+    console.log('test1');
+  },
+  func2: () => {
+    console.log('test2');
+  },
+  func3: x => {
+    console.log(x);
+  }
+};
```
masatake, line 11: As you did in [github-issue-1389.d/input.js](https://github.com/universal-ctags/ctags/pull/3357/commits/f624b4ecd2c48873637dda9ee93e15ffcef7fc5c#diff-bff42b656ff5c564899cc685664805ba3b9d1f35d57c1c5b431f881b53a63cd4), let's add the reference something like:

```js
// Based on #1933 submitted by @gpavlov2016
```

Units/parser-javascript.r/github-issue-780.d/input.js
```diff
@@ -0,0 +1,9 @@
+import Foo from './foo';
+
+const Bar = Foo.create({
+  jump() {
+    return 123;
+  }
+})
+
+export default Bar;
```
masatake, line 9:
```js
// Taken from #780 submitted by @marlun.
```

parsers/jscript.c
```diff
@@ -1741,10 +1775,12 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 					if (isKeyword (token, KEYWORD_async))
 						readToken (token);
 				}
+
+				vString * signature = NULL;
```
masatake, line 1779: You didn't assign a vString here, you assigned at ... *1*

parsers/jscript.c
```diff
@@ -1741,10 +1775,12 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
 				if ( is_shorthand || isKeyword (token, KEYWORD_function) )
 				{
 					TRACE_PRINT("Seems to be a function or shorthand");
-					vString *const signature = vStringNew ();
+					signature = vStringNew ();
```
masatake, line 1783: *1*, here. However, I found that this delayed memory allocation makes reading the code harder.
Even if we may not use the allocated object, how about allocating the object at ...*2*

parsers/jscript.c
masatake: *2*, here. So we can make the code simplified like:

```
vString * signature = vStringNew();
{
}
vStringDelete(signature);
```
This function is already enough long. So I would like to avoid introducing extra complexity.

I would like to call the approach I suggested here *1New1Delete* in this comment.

parsers/jscript.c
```diff
@@ -1805,17 +1841,38 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 						}
 						else if (isType (token, TOKEN_OPEN_PAREN))
 						{
-							skipArgumentList (token, false, NULL);
+							vStringDelete (signature);
+							signature = vStringNew ();
```
masatake, line 1845: If you follow the 1new1delete approach, you can convert these vStringDelete and vStringNew into vStringClear.

parsers/jscript.c
```diff
@@ -1805,17 +1841,38 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
...
 						else if (isType (token, TOKEN_OPEN_SQUARE))
 						{
 							skipArrayList (token, false);
 						}
+						else if (isType (token, TOKEN_ARROW))
+						{
+							TRACE_PRINT("Seems to be an anonymous function");
+							if (signature == NULL)
+							{
+								signature = vStringNew ();
```
masatake, line 1857: If you follow the 1new1delete approach, you can delete some lines here.
```C
if (vStringIsEmpty(signature) && isType (saved_token, TOKEN_IDENTIFIER)) 
{
   	vStringPut (signature, '(');
...
```

parsers/jscript.c
masatake: like:
```C
   vStringClear (signature);
```

masatake:
Nice work: closing three issues with one request!
I made some comments about vstring lifecycle management.

```
[yamato@dev64]~/var/ctags-github% cat /tmp/foo.js 
var f = function(x) { return x }
var o = f({
    init: function () {
    },
    fini: function () {
    },
})
[yamato@dev64]~/var/ctags-github% ./ctags --options=NONE --sort=no -o - /tmp/foo.js
f	/tmp/foo.js	/^var f = function(x) { return x }$/;"	f
init	/tmp/foo.js	/^    init: function () {$/;"	m
fini	/tmp/foo.js	/^    fini: function () {$/;"	m
```

This is the output of ctags with your change.
We cannot know the siblingness  between `init` and `fini`.

```diff
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 8999b65ed..906f1aeb3 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -1144,6 +1144,20 @@ static void readToken (tokenInfo *const token)
  *  Token parsing functions
  */
 
+static void parseMethodsInAnonymousClass (tokenInfo *const token)
+{
+   tokenInfo *const anon_class = newToken ();
+   copyToken (anon_class, token, true);
+   anonGenerate (anon_class->string, "AnonymousClass", JSTAG_CLASS);
+   anon_class->type = TOKEN_IDENTIFIER;
+
+   bool has_methods = parseMethods (token, anon_class, false);
+
+   if (has_methods)
+       makeJsTagCommon (anon_class, JSTAG_CLASS, NULL, NULL, true);
+   deleteToken (anon_class);
+}
+
 static void skipArgumentList (tokenInfo *const token, bool include_newlines, vString *const repr)
 {
    if (isType (token, TOKEN_OPEN_PAREN))   /* arguments? */
@@ -1165,7 +1179,7 @@ static void skipArgumentList (tokenInfo *const token, bool include_newlines, vSt
                if (prev_token_type == TOKEN_ARROW)
                    parseBlock (token, NULL);
                else
-                   parseMethods (token, NULL, false);
+                   parseMethodsInAnonymousClass (token);
            }
            else if (isKeyword (token, KEYWORD_function))
                parseFunction (token);
@@ -1200,7 +1214,7 @@ static void skipArrayList (tokenInfo *const token, bool include_newlines)
                if (prev_token_type == TOKEN_ARROW)
                    parseBlock (token, NULL);
                else
-                   parseMethods (token, NULL, false);
+                   parseMethodsInAnonymousClass (token);
            }
 
            prev_token_type = token->type;
```

With this extra change, ctags emits:
```
[yamato@dev64]~/var/ctags-github% ./ctags --options=NONE --sort=no -o - /tmp/foo.js
f	/tmp/foo.js	/^var f = function(x) { return x }$/;"	f
init	/tmp/foo.js	/^    init: function () {$/;"	m	class:AnonymousClass49f7b9030101
fini	/tmp/foo.js	/^    fini: function () {$/;"	m	class:AnonymousClass49f7b9030101
AnonymousClass49f7b9030101	/tmp/foo.js	/^var o = f({$/;"	c
```
We can know init and fini in the same scope.
I will make a pull request for tagging anonymous classes after merging your change.

jafl:
The reason for the signature complexity is the "goto function" statement.  However, since you don't mind always allocating the string, even if it's never used, then I will use your suggestions.

Regarding your patch, that is not an anonymous class.  That is an anonymous object.  There is no reason to assume that the functions are connected.

masatake (approved):

masatake:
Much easier to read with pre-allocating vString.

> Regarding your patch, that is not an anonymous class. That is an anonymous object. 
>  There is no reason to assume that the functions are connected.

The functions? They are tagged as methods.
The methods are connected with the anonymous object, aren't they?

jafl:
> > Regarding your patch, that is not an anonymous class. That is an anonymous object.
> > There is no reason to assume that the functions are connected.
> 
> The functions? They are tagged as methods. The methods are connected with the anonymous object, aren't they?

I checked, and the code does create `anonymousObject` in `parseStatement`.  So the patch makes sense.  Just replace "class" with "object".  Should I include it in #3363?

jafl:
I added your patch to this PR, since it's directly related.

masatake:
> I added your patch to this PR, since it's directly related.

Thank you. But, could you use `class` kind instead of `variable` kind IN THIS PULL REQUEST?

Instead, let's change the kind to `class` from `variable` in #3363. So each pull request becomes consistent.

masatake:
Thank you!

PR #2115, opened 2019-06-15
[WIP] JavaScript: introduce cork API

masatake:
Doing more interesting things with JavaScript parser like supporting VueJS, cork indexes are needed.

Some wrong (I think) expected.tags file are fixed. Introducing cork API exposes the
bugs about values for scope fields.

@b4n, I cannot make the change ( 1e6e0f3 ) smaller without failing the test cases for JavaScript.
I would like you to see expected.tags files I modified in this pull request.

PR #3363, opened 2022-04-19
JavaScript: fix the definition of a class

jafl:
Before ES6, there was no way to tell if an individual function was a class constructor.  This is why the convention was set that class names start with uppercase, and function names start with lowercase.

This patch also stops tagging objects as classes.  Objects in JavaScript are just like any other value.  One cannot "new" an object, only a constructor function.

### DIFF
--- a/Units/parser-javascript.r/1795612.js.d/expected.tags
+++ b/Units/parser-javascript.r/1795612.js.d/expected.tags
@@ -1,4 +1,4 @@
-RPC	input.js	/^Test.RPC =$/;"	c	class:Test
-asyncMethod	input.js	/^	asyncMethod: function($/;"	m	class:Test.RPC
-asyncRequest	input.js	/^	asyncRequest: function($/;"	m	class:Test.RPC
-request_id	input.js	/^	request_id: 0,$/;"	p	class:Test.RPC
+Test.RPC	input.js	/^Test.RPC =$/;"	p
+asyncMethod	input.js	/^	asyncMethod: function($/;"	m	property:Test.RPC
+asyncRequest	input.js	/^	asyncRequest: function($/;"	m	property:Test.RPC
+request_id	input.js	/^	request_id: 0,$/;"	p	property:Test.RPC
--- a/Units/parser-javascript.r/bug1950327.js.d/expected.tags
+++ b/Units/parser-javascript.r/bug1950327.js.d/expected.tags
@@ -1,13 +1,13 @@
-*	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	class:container.dirtyTab
+*	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	property:container.dirtyTab
 Different	input.js	/^Different.prototype = {$/;"	c
 TabChrome	input.js	/^TabChrome.prototype = {$/;"	c
+container.dirtyTab	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p
 createTabTile	input.js	/^	createTabTile: function(browser) $/;"	m	class:Different
 createTabTile	input.js	/^	createTabTile: function(browser) $/;"	m	class:TabChrome
 destroyTabTile	input.js	/^	destroyTabTile: function(tile)$/;"	m	class:Different
 destroyTabTile	input.js	/^	destroyTabTile: function(tile)$/;"	m	class:TabChrome
-dirtyTab	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	c	class:container
 init	input.js	/^	init: function() $/;"	m	class:Different
 init	input.js	/^	init: function() $/;"	m	class:TabChrome
-snapshot	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	class:container.dirtyTab
-title	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	class:container.dirtyTab
-url	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	class:container.dirtyTab
+snapshot	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	property:container.dirtyTab
+title	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	property:container.dirtyTab
+url	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	property:container.dirtyTab
--- a/Units/parser-javascript.r/github-issue-780.d/expected.tags
+++ b/Units/parser-javascript.r/github-issue-780.d/expected.tags
@@ -1,3 +1,3 @@
-AnonymousClass6b70cce50101	input.js	/^const Bar = Foo.create({$/;"	c
 Bar	input.js	/^const Bar = Foo.create({$/;"	C
-jump	input.js	/^  jump() {$/;"	m	class:AnonymousClass6b70cce50101
+anonymousObject6b70cce50105	input.js	/^const Bar = Foo.create({$/;"	v
+jump	input.js	/^  jump() {$/;"	m	variable:anonymousObject6b70cce50105
--- a/Units/parser-javascript.r/js-empty-class-name.d/expected.tags
+++ b/Units/parser-javascript.r/js-empty-class-name.d/expected.tags
@@ -1,2 +1,2 @@
-prop	input.js	/^var {prop} = { prop: "value" };$/;"	p	class:AnonymousClass4ca5b60a0101
-AnonymousClass4ca5b60a0101	input.js	/^var {prop} = { prop: "value" };$/;"	c
+prop	input.js	/^var {prop} = { prop: "value" };$/;"	p	variable:anonymousObject4ca5b60a0105
+anonymousObject4ca5b60a0105	input.js	/^var {prop} = { prop: "value" };$/;"	v
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -330,6 +330,7 @@ static void parseUI5 (tokenInfo *const token);
 static const char *tokenTypeName(enum eTokenType e);
 static const char* getNameStringForCorkIndex(int index);
 static const char* getKindStringForCorkIndex(int index);
+static const char *kindName(jsKind kind);
 // #define DO_TRACING_USE_DUMP_TOKEN
 #ifdef DO_TRACING_USE_DUMP_TOKEN
 static void dumpToken (const tokenInfo *const token);
@@ -527,7 +528,7 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 	{
 		const char *scopeStr = getNameStringForCorkIndex (scope);
 		const char *scopeKindStr = getKindStringForCorkIndex (scope);
-		TRACE_PRINT("Emitting tag for symbol '%s' of kind %02x with scope '%s:%s'",name,kind, scopeKindStr, scopeStr);
+		TRACE_PRINT("Emitting tag for symbol '%s' of kind %s with scope '%s:%s'", name, kindName(kind), scopeKindStr, scopeStr);
 	}
 #endif
 
@@ -590,6 +591,17 @@ static int makeFunctionTagCommon (tokenInfo *const token, vString *const signatu
 static int makeFunctionTag (tokenInfo *const token, vString *const signature, bool generator)
 {
 	return makeFunctionTagCommon (token, signature, generator, false);
+}
+
+static bool isClassName (tokenInfo *const name)
+{
+	char * p = strrchr(vStringValue (name->string), '.');
+	if (p == NULL)
+		p = vStringValue (name->string);
+	else
+		p++;
+
+	return isupper((unsigned char) *p);
 }
 
 /*
@@ -1293,7 +1305,10 @@ static void readTokenFull (tokenInfo *const token, bool include_newlines, vStrin
 static void readTokenFullDebug (tokenInfo *const token, bool include_newlines, vString *const repr)
 {
 	readTokenFull (token, include_newlines, repr);
-	TRACE_PRINT("token '%s' of type %02x with scope '%s'",vStringValue(token->string),token->type, vStringValue(token->scope));
+
+	const char *scopeStr = getNameStringForCorkIndex (token->scope);
+	const char *scopeKindStr = getKindStringForCorkIndex (token->scope);
+	TRACE_PRINT("token '%s' of type %s with scope '%s' of kind %s", vStringValue(token->string), tokenTypeName(token->type), scopeStr, scopeKindStr);
 }
 # define readTokenFull readTokenFullDebug
 #endif
@@ -1307,19 +1322,19 @@ static void readToken (tokenInfo *const token)
  *	 Token parsing functions
  */
 
-static int parseMethodsInAnonymousClass (tokenInfo *const token)
+static int parseMethodsInAnonymousObject (tokenInfo *const token)
 {
 	int index = CORK_NIL;
 
-	tokenInfo *const anon_class = newToken ();
-	copyToken (anon_class, token, true);
-	anonGenerate (anon_class->string, "AnonymousClass", JSTAG_CLASS);
-	anon_class->type = TOKEN_IDENTIFIER;
+	tokenInfo *const anon_object = newToken ();
+	copyToken (anon_object, token, true);
+	anonGenerate (anon_object->string, "anonymousObject", JSTAG_VARIABLE);
+	anon_object->type = TOKEN_IDENTIFIER;
 
-	index = makeJsTagCommon (anon_class, JSTAG_CLASS, NULL, NULL, true, true);
+	index = makeJsTagCommon (anon_object, JSTAG_VARIABLE, NULL, NULL, true, true);
 	if (! parseMethods (token, index, false))
 	{
-		/* If no method is found, the anonymous class
+		/* If no method is found, the anonymous object
 		 * should not be tagged.
 		 */
 		tagEntryInfo *e = getEntryInCorkQueue (index);
@@ -1328,7 +1343,7 @@ static int parseMethodsInAnonymousClass (tokenInfo *const token)
 		index = CORK_NIL;
 	}
 
-	deleteToken (anon_class);
+	deleteToken (anon_object);
 
 	return index;
 }
@@ -1354,7 +1369,7 @@ static void skipArgumentList (tokenInfo *const token, bool include_newlines, vSt
 				if (prev_token_type == TOKEN_ARROW)
 					parseBlock (token, CORK_NIL);
 				else
-					parseMethodsInAnonymousClass (token);
+					parseMethodsInAnonymousObject (token);
 			}
 			else if (isKeyword (token, KEYWORD_function))
 				parseFunction (token);
@@ -1389,7 +1404,7 @@ static void skipArrayList (tokenInfo *const token, bool include_newlines)
 				if (prev_token_type == TOKEN_ARROW)
 					parseBlock (token, CORK_NIL);
 				else
-					parseMethodsInAnonymousClass (token);
+					parseMethodsInAnonymousObject (token);
 			}
 
 			prev_token_type = token->type;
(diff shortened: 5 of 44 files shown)


jafl:
I'm working on fixing the unit tests.

jafl:
I can't figure out why codecov/project is marked fail.

masatake:
In this project, the result reported by codecov is an optional requirement for merging. The coverage is enough high about your change.

masatake:
This change significantly impacts people developing/using IDE that uses ctags internally.
I would like to say this is not a bug fix or enhancement. The change is a kind of "changing the design of parser".
I don't mean I object to the change. However,...
Changing the definition of classes may change how the "tree" at IDE's side bar is rendered.

My JavaScript knowledge is limited, and I don't have much chance to read source code written in JavaScript with ctags.
So I don't have much thought about this change. 
Instead, I would like @jafl to do what he thinks the parser is better; he has already submitted good changes.
I'm always looking for a good parser maintainer knowing the target language.

@b4n, the de-factor parser maintainer of JavaScript, is absent these days. So a person having knowledge, activeness, and motivation should take over the maintainership.  

My only concern is the impact on the other software.
Especially on Geany. Universal Ctags is owing to development efforts at the Geany project.
So I would like to hear the comments from the people on the project.

@jafl, do you use an editor or IDE with ctags?
If yes, how about the impact of this change on the client tool you use?

jafl:
I started thinking about this because my IDE was showing incorrect results, classifying functions and objects as classes.  Since it could have a big impact, I spent days debating whether or not to submit it.  Ultimately, I decided it was the right thing to do because the parser should match the language specification.

It would be great to hear from others if/how this affects their IDE's.

I am also willing to be the maintainer for the JavaScript parser, if that is needed.

parsers/jscript.c
```diff
@@ -1940,8 +1950,7 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
 	bool is_const = false;
 	bool is_terminated = true;
 	bool is_global = false;
-	bool has_methods = false;
-	vString *	fulltag;
+	vString * fulltag;
```
masatake, line 1953: Very very minor item. 
Could you this change from this commit?
Doing this in another commit is acceptable, of course.

parsers/jscript.c
```diff
@@ -427,6 +427,17 @@ static void makeFunctionTag (tokenInfo *const token, vString *const signature, b
 	makeFunctionTagCommon (token, signature, generator, false);
 }
 
+static bool isClassName (tokenInfo *const name)
+{
+	char * p = strrchr(name->string->buffer, '.');
```
masatake, line 432: Please, access the internal of vString directly only when you have a strong reason.
Instead, let's use vStringValue like:
```
vStringValue (name->string);
```

parsers/jscript.c
```diff
@@ -427,6 +427,17 @@ static void makeFunctionTag (tokenInfo *const token, vString *const signature, b
...
+	if (p == NULL)
+		p = name->string->buffer;
```
masatake, line 434: Let's use vStringValue.
If you find lines that access `->buffer` directly in jscript.c (and the other files), they can be candidates for fixing in another pull request:-)

parsers/jscript.c
jafl: Thanks!  That's much better.

parsers/jscript.c
```diff
@@ -2313,21 +2323,17 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
...
-				anonGenerate (name->string, "AnonymousClass", JSTAG_CLASS);
-			has_methods = parseMethods(token, name, false);
-			if (has_methods)
-				makeJsTagCommon (name, JSTAG_CLASS, NULL, NULL, anonClass);
-			else
-			{
+			bool anonObject = vStringIsEmpty (name->string);
+			if (anonObject)
+				anonGenerate (name->string, "anonymousObject", JSTAG_VARIABLE);
+			parseMethods(token, name, false);
```
masatake, line 2336: With the change, the value returned from parserMethods are ignored, and the tag for the anonymous object is not made. I think this is not a good idea. Could you keep the original code with just changing the kind like:
```c
			has_methods = parseMethods(token, name, false);
			if (has_methods)
				makeJsTagCommon (name, JSTAG_VARIABLE, NULL, NULL, anonObject);
			else
			{
```
?

parsers/jscript.c
```diff
@@ -1940,7 +1950,6 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
 	bool is_const = false;
 	bool is_terminated = true;
 	bool is_global = false;
-	bool has_methods = false;
```
masatake, line 1943: To extract an object having a method or a property as an anonymous tag, we must keep this local variable. See below.

masatake:
@techee, @eht16, and the other Geany developers.
I would like to merge this pull request.
However, this change may have a big impact on Geany.
Could you give us comments?

masatake:
I wonder we need a mechanism for declaring the version of a parser like:

```C
extern parserDefinition* JavaScriptParser (void)
{
	// .jsx files are JSX: https://facebook.github.io/jsx/
	// which have JS function definitions, so we just use the JS parser
	static const char *const extensions [] = { "js", "jsx", "mjs", NULL };
	static const char *const aliases [] = { "js", "node", "nodejs",
	                                        "seed", "gjs",
											/* Used in PostgreSQL
											 * https://github.com/plv8/plv8 */
											"v8",
											NULL };
	parserDefinition *const def = parserNew ("JavaScript");
        def->version = {[CURRENT] = 1, [REVISION] = 0, [AGE] = 0}; // as describe in info page of libtool
...
```

The parser version information may be emitted like:
```
$ cat tags
...
!_TAG_PARSER_VERSION!JavaScript	current	/1/
!_TAG_PARSER_VERSION!JavaScript	revision	/0/
!_TAG_PARSER_VERSION!JavaScript	age	/0/
...
```

Geany includes ctags code. So Geany doesn't need these pseudo tags about version information .
Other client tools reading tags file can utilize the pseudo tags.

jafl:
Should the parser version be included in this PR?

masatake:
> Should the parser version be included in this PR?

No. Sorry for making you confused. It is just an idea.

jafl:
I will merge the updates from #3357 into this PR after that PR is merged.

Units/parser-javascript.r/js-empty-class-name.d/expected.tags
```diff
@@ -1,2 +1,2 @@
-prop	input.js	/^var {prop} = { prop: "value" };$/;"	p	class:AnonymousClass4ca5b60a0101
-AnonymousClass4ca5b60a0101	input.js	/^var {prop} = { prop: "value" };$/;"	c
+prop	input.js	/^var {prop} = { prop: "value" };$/;"	p	class:anonymousObject4ca5b60a0105
```
masatake, line 1: The value for scope field is `class:anonymousObject4ca5b60a0105`.
However, `anonymousObject4ca5b60a0105` is tagged as a variable.
So I think the value of the field should be:
```
variable:anonymousObject4ca5b60a0105
```

Units/parser-javascript.r/github-issue-780.d/expected.tags
```diff
@@ -1,3 +1,3 @@
-AnonymousClass6b70cce50101	input.js	/^const Bar = Foo.create({$/;"	c
 Bar	input.js	/^const Bar = Foo.create({$/;"	C
-jump	input.js	/^  jump() {$/;"	m	class:AnonymousClass6b70cce50101
+anonymousObject6b70cce50105	input.js	/^const Bar = Foo.create({$/;"	v
+jump	input.js	/^  jump() {$/;"	m	class:anonymousObject6b70cce50105
```
masatake, line 3: Though `anonymousObject6b70cce50105` is tagged as a variable, it is referred as a class in the tag of `jump`.

techee:
@masatake Is this just a matter of renaming anonymous tags and reporting some with a different kind? A don't see any problem with that, we'll just have to update our kind mappings but we are prepared for such situations.

Units/parser-javascript.r/js-empty-class-name.d/expected.tags
jafl: Good point.  This requires that I tackle a FIXME in the code.  The patch is going to get bigger :)

masatake:
> @masatake Is this just a matter of renaming anonymous tags and reporting some with a different kind? A don't see any problem with that, we'll just have to update our kind mappings but we are prepared for such situations.

As far as the source code area I reviewed,  we don't have to rename the anonymous tags in this case.
Even if we have to rename them, it is easy.

We have to update ` extensionFields.scopeKindIndex` as you already know.

After reading your code, updating the `socpeKindIndex` is more complicated than I assumed.

I want to carefully review your new change (difference between  1bab2a2 and 5acd011). 
Could you do "push --force" to replace the current single commit with the original and new changes?
`git reflog` may help you.

masatake:
@techee 

> @masatake Is this just a matter of renaming anonymous tags and reporting some with a different kind? A don't see any problem with that, we'll just have to update our kind mappings but we are prepared for such situations.

The name for anonymous tags may not be an issue here.


In the original code, methods and properties have a class in their scope fields:

input.js:
```
var o = {
    init: function () {
    },
    fini: function () {
    },
};
var p = 1
```

tags output with the original code:
```
init	foo.js	/^    init: function () {$/;"	kind:method	scope:class:o	signature:()	roles:def
fini	foo.js	/^    fini: function () {$/;"	kind:method	scope:class:o	signature:()	roles:def
o	foo.js	/^var o = {$/;"	kind:class	roles:def
p	foo.js	/^var p = 1$/;"	kind:variable	roles:def
```

tags output with this change:
```
init	foo.js	/^    init: function () {$/;"	kind:method	scope:variable:o	signature:()	roles:def
fini	foo.js	/^    fini: function () {$/;"	kind:method	scope:variable:o	signature:()	roles:def
o	foo.js	/^var o = {$/;"	kind:variable	lroles:def
p	foo.js	/^var p = 1$/;"	kind:variable	roles:def
```

If Geany expects methods (and properties) may have "class" as their scopes, this pull request breaks the expectation.
With this request, the tags output lost some information. With the original code, `p` and `q` have different kinds. With this pull request, `p` and `q` have the same kind, `variable`. I wonder if this is a matter or not in Geany.

masatake:
@jafl 

After reading your code, updating the `socpeKindIndex` is more complicated than I assumed.

I want to carefully review your new change (difference between  1bab2a2 and 5acd011). 
Could you do "push --force" to replace the current single commit with the original and new changes?
`git reflog` may help you.

masatake:
```
var o = {
    init: function () {
    },
    fini: function () {
    },
};
var p = 1
```

In my understanding, What @jafl wants to say is `o` is not a `class`.
This is understandable even with my limited knowledge of JavaScript.

I will talk about the next steps AFTER merging this pull request.
In lexically, `o` is a variable. However, I think people using Javascript may want to say `o` is an object.
In the other words, they expect the following tags output:
```
init	foo.js	/^    init: function () {$/;"	kind:method	scope:object:o	signature:()	roles:def
fini	foo.js	/^    fini: function () {$/;"	kind:method	scope:object:o	signature:()	roles:def
o	foo.js	/^var o = {$/;"	kind:object	roles:def
p	foo.js	/^var p = 1$/;"	kind:variable	roles:def
```

I'm comfortable with `o` and `p` being tagged differently.
 
The algorithm emitting `o` is:
0. introduce a new kind "o/object".
1. make a tag for `o` with `variable` kind as this pull request does.
2. count the methods and properties defined in `o` => 2.
3. change the kind of `o` to `object` if the counter > 0.
4. update the scope fields of tags having `o` as their scope.

The following case doesn't work:
```
var gen = function () {
 return {
    init: function () {
    },
    fini: function () {
    }
  };
}
var o = get()
var p = 1
```

In this case `o` is tagged as a variable. 

How about an array? Introducing the "array" kind?
I'm not sure:-)

techee:
> If Geany expects methods (and properties) may have "class" as their scopes, this pull request breaks the expectation.
With this request, the tags output lost some information. With the original code, p and q have different kinds. With this pull request, p and q have the same kind, variable. I wonder if this is a matter or not in Geany.

@masatake I don't think this really matters to Geany, it should be no problem for us.

masatake:
@jafl, thank you.

masatake:
Now I understood your change. 
Introducing corkAPI may be better. But the way you took is acceptable.

I will review  5acd011 directly. We don't need 1bab2a2 anymore.

parsers/jscript.c
```diff
@@ -2097,7 +2112,9 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
 			/* Assume it's an assignment to a global name (e.g. a class) using
 			 * its fully qualified name, so strip the scope.
 			 * FIXME: resolve the scope so we can make more than an assumption. */
+			token->scopeKind = JSTAG_NONE;
 			vStringClear (token->scope);
```
masatake, line 2116: These two lines imply it is better to introduce a helper function named `clearTokenScope`.
```c
static void clearTokenScope(tokenInfo *token)
{
   vStringClear (token->scope);
   token->scopeKinde = KIND_GHOST_INDEX;
}
```

parsers/jscript.c
```diff
@@ -1913,6 +1926,7 @@ static bool parseMethods (tokenInfo *const token, const tokenInfo *const class,
 	findCmdTerm (token, false, false);
 
 cleanUp:
+	token->scopeKind = saveScopeKind;
 	vStringCopy (token->scope, saveScope);
```
masatake, line 1930: How about introducing something like `fillTokenScope(tokenInfo *token, vString *saveScope, int saveScopeKind);`?

masatake: `setScope` may be better as the name of the function.

parsers/jscript.c
masatake: `clearScope` or `resetScope` may be better.

Units/review-needed.r/simple.js.t/expected.tags
```diff
@@ -1,5 +1,5 @@
 Database	input.js	/^Database.prototype.getTodaysDate = Database_getTodaysDate;$/;"	c
-ValidClassOne	input.js	/^testlib.extras.ValidClassOne = function(a,b) { $/;"	c	class:testlib.extras
+ValidClassOne	input.js	/^testlib.extras.ValidClassOne = function(a,b) { $/;"	c	scope:testlib.extras
```
masatake, line 2: Very interesting.
I have many comments about this line.
...

masatake:
When, I understand the reason why you introduce the "scope" kind, I wonder what we should do.

This pull request will become much larger. The goal of change becomes ambiguous.
... I think introducing #2115 first is better. This pull request, switching class to variable kinds, may come later.

I would like to try to rebase #2115 to the latest master. So I would like you to review my approach.
Could you wait for a week or so?

In addition to corkAPI, I would like you to know the concepts, roles and reference tags

See http://docs.ctags.io/en/latest/man/ctags-lang-python.7.html and http://docs.ctags.io/en/latest/man/ctags.1.html?highlight=roles#roles .

With combining these concepts, we can fill the scope fields of tags cleanly.

Units/review-needed.r/simple.js.t/expected.tags
jafl: Yes, this is a tricky issue.  The parser doesn't always know what the thing is.  Perhaps the cork queue will help solve this?

jafl:
NetBSD unit tests failed due to "Transient resolver failure"

masatake:
@jafl, unexpected failures on *BSD happen frequently. In such cases, could you press "Re-run jobs" by yourself? The question is whether you can do it with your account.

masatake:
Oh, I already did "Re-run jobs" though I wanted you to try.

jafl:
I do not see a "Re-run jobs" button.

masatake:
> I do not see a "Re-run jobs" button.

I see. I'm looking for a way to make it visible to you.

masatake:
Like #3424, I guess the change for [Units/parser-javascript.r/1880687.js.d/input.js](https://github.com/universal-ctags/ctags/pull/3363/files#diff-f31b95120b6ea91dcf46b0b2d091314d2ef1cb8bbb74a85c3ece5a63e166f675) in this pull request is not intentional. Could you rebase this pull request on the latest master branch?

parsers/jscript.c
```diff
@@ -592,6 +593,17 @@ static int makeFunctionTag (tokenInfo *const token, vString *const signature, bo
...
 
+static bool isClassName (tokenInfo *const name)
+{
+	char * p = strrchr(vStringValue (name->string), '.');
+	if (p == NULL)
+		p = vStringValue (name->string);
+	else
+		p++;
+
+	return isupper(*p);
```
masatake, line 604: In my understanding, a cast operator `(unsigned int)` is needed in the front of `*p`.
isupper() expects int but only -1 is the acceptable negative value; the behavior of isupper() is not defined in the standard C spec.
On the other hand, we cannot have an assume whether char is `unsigned char` or `singed char`. So `*p` can be a negative value other than -1.

See https://github.com/universal-ctags/ctags/pull/1697 .

jafl: Good point.  Fixed.

jafl:
@masatake You will have to re-run the BSD tests.

masatake (approved):

masatake:
If one in the project becomes absent, the God (, something you believe, or something supernatural) dispatches another hacker to this project. It is known as the "law of conservation of the number of ctags developers".

@jafl, please, merge this!

jafl:
I don't see a button for merging.

masatake:
Thank you.

Units/parser-javascript.r/1795612.js.d/expected.tags
```diff
@@ -1,4 +1,4 @@
-RPC	input.js	/^Test.RPC =$/;"	c	class:Test
-asyncMethod	input.js	/^	asyncMethod: function($/;"	m	class:Test.RPC
-asyncRequest	input.js	/^	asyncRequest: function($/;"	m	class:Test.RPC
-request_id	input.js	/^	request_id: 0,$/;"	p	class:Test.RPC
+Test.RPC	input.js	/^Test.RPC =$/;"	p
```
b4n, line 1: Is this change expected?  Admittedly `Test` is not actually declared in this file so creating a tag for it is a bit artificial, but this change looks a bit unexpected (it's at least not obvious to me which change did this, so I can't tell if it was intentional).

I don't really mind either way as there's no definite answer (it's not even proper JS given `Test` is not even imported or anything), but I'm wondering if this was on purpose or an accident.

masatake: This looks like a bug about the granularity of tagging.

Looking at this again, a tag (`Test.RPC`) with more than one token (`Test` and `RPC`) looks strange. (as far as comparing with 
Units/parser-javascript.r/js-class-related-unterminated.d/expected.tags)

My expectation (--extras=+{reference} --fields=+{roles}{scope}):

```
RPC	input.js	/^Test.RPC =$/;"	p	scope:variable:Test
Test	input.js	/^Test.RPC =$/;"	v	roles:chainElt
```

I introduced the `chainElt` roles for avoiding more-than-one-token tag entries.


 
There is room for discussion about the kinds (p for `RPC` and v for `Test`), but there will be no perfect answer. It is up to the parser maintainer.

masatake: I'll convert this to an issue.

Units/parser-javascript.r/bug1950327.js.d/expected.tags
```diff
@@ -1,13 +1,13 @@
-*	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	class:container.dirtyTab
+*	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	property:container.dirtyTab
 Different	input.js	/^Different.prototype = {$/;"	c
 TabChrome	input.js	/^TabChrome.prototype = {$/;"	c
+container.dirtyTab	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p
```
masatake, line 4: This is just another two-tokens tag.
I think this should be:
```
dirTab	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	p	variable:container
container	input.js	/^		container.dirtyTab = {'url': false, 'title':false, 'snapshot':false, '*': false}		$/;"	v	roles:chainElt
```

What the scope for `cotainer` is is another issue.

Units/parser-javascript.r/1795612.js.d/expected.tags
masatake:
```diff
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 3788cc090..f54c432da 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -485,7 +485,7 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 
 	const char *p;
 	char *name_chain = NULL;
-	if (!token->dynamicProp && kind != JSTAG_PROPERTY &&  (p = strrchr (name, '.')) != NULL )
+	if (!token->dynamicProp && (p = strrchr (name, '.')) != NULL )
 	{
 		if ((p - name) != 0)
 			name_chain = eStrndup (name, (size_t) (p - name));
```

With this, we can avoid the two-tokens tags.
It is easy to fix the parser as we want it, but it is not easy to decide how the parser should be.

masatake: We should open two issues. I will open issues for them.

PR #3367, opened 2022-04-25
perl6: Add more file name extensions

techee:
See https://en.wikipedia.org/wiki/Raku_(programming_language)

### DIFF
--- a/parsers/perl6.c
+++ b/parsers/perl6.c
@@ -325,7 +325,8 @@ findPerl6Tags (void)
 parserDefinition *
 Perl6Parser (void)
 {
-    static const char *const extensions[] = { "p6", "pm6", "pm", "pl6", NULL };
+    static const char *const extensions[] = { "p6", "pm6", "pm", "pl6",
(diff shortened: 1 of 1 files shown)


masatake:
Private NOTE: https://docs.raku.org/language/pod

parsers/perl6.c
```diff
@@ -325,7 +325,8 @@ findPerl6Tags (void)
 parserDefinition *
 Perl6Parser (void)
 {
-    static const char *const extensions[] = { "p6", "pm6", "pm", "pl6", NULL };
+    static const char *const extensions[] = { "p6", "pm6", "pm", "pl6",
+		"pod6", "t6", "raku", "rakumod", "rakudoc", "rakutest", NULL };
```
masatake, line 329: In  my understanding pod6 and rakumod are for extensions for a documentation language, not for Perl6.
I guess our perl6 parser doesn't handle input fils for the documentation language.

techee:
Yes, you are right, I just removed it (together with `rakudoc` which is the same).

masatake:
Thank you.

Issue #3298, opened 2022-02-23
Adoption of an icon


tartley:
[The snap package of universal-ctags](https://snapcraft.io/universal-ctags) currently displays with a grey circle default (ie unset) icon. I'm considering adopting this allegedly public-domain icon for it:

![](https://icons.iconarchive.com/icons/paomedia/small-n-flat/128/tag-icon.png)

from https://iconarchive.com/show/small-n-flat-icons-by-paomedia/tag-icon.html
(Maybe customizing it first, by replacing the three grey lines with a U ?)

I thought I'd announce it here, so that if anyone else cares, we could converge on a single icon used in each of the repos where it makes sense. I don't know that it does make sense anywhere else, except perhaps the blog? But I thought over-communicating is better than under-communicating.

masatake:
Thank you for taking care of this kind of topic.
>  (Maybe customizing it first, by replacing the three grey lines with a U ?)

My favor is `;"`.

`;"` is not so catchy. However, it represents the heart of ctags.
See https://docs.ctags.io/en/latest/man/tags.5.html#proposal .

```
Optionally:

;”
    semicolon + doublequote: Ends the tagaddress in way that looks like the start of a comment to Vi.
{tagfield}
    See below.

A tagfield has a name, a colon, and a value: “name:value”.
```

tartley:
Finally gave this edit a shot. Thoughts welcome:

![](https://www.tartley.com/files/universal-ctags.04-font.svg)

masatake:
Thank you!
Could you make a pull request putting the SVG file to misc dir as universal-ctags.svg?
Could you upload a properly scaled PNG file as the icon for this organization? (if you don't have proper access right, I would like to upload to this issue as an attachment).

masatake:
I cannot wait for having our own icon on our GitHub page!
I uploaded a png image converted from the SVG file with Inkscape.
Please, look at our organization page.
@fishman, the repository you founded grows to have its own icon!

tartley:
OMG that's amazing. Thanks for beating me to it. PR incoming...

tartley:
Also applied the svg here https://snapcraft.io/universal-ctags

PR #3371, opened 2022-04-29
Add icon to misc/universal-ctags.svg

tartley:
This completes https://github.com/universal-ctags/ctags/issues/3298


masatake:
Thank you!

PR #3375, opened 2022-05-02
Cxx: use a trashbox for deleteing tokens allocated for anonymous parameters

### LINKED ISSUE #3372: cxx: Memory leak for anonymous function parameters
When running a slightly older version of ctags (in Geany) with valgrind, I'm getting this:
```
==27073== 288 (176 direct, 112 indirect) bytes in 2 blocks are definitely lost in loss record 15,678 of 17,610
==27073==    at 0x48490C0: malloc (vg_replace_malloc.c:381)
==27073==    by 0x4B13233: eMalloc (routines.c:220)
==27073==    by 0x4AB7433: createToken (cxx_token.c:30)
==27073==    by 0x4AEE40F: objPoolGet (objpool.c:66)
==27073==    by 0x4AB7647: cxxTokenCreate (cxx_token.c:78)
==27073==    by 0x4AB7647: cxxTokenCreateAnonymousIdentifier (cxx_token.c:142)
==27073==    by 0x4AB1493: cxxParserTokenChainLooksLikeFunctionParameterList.part.0 (cxx_parser_function.c:2179)
==27073==    by 0x4AB18AB: cxxParserTokenChainLooksLikeFunctionParameterList (cxx_parser_function.c:1934)
==27073==    by 0x4AB18AB: cxxParserLookForFunctionSignatureCheckParenthesisAndIdentifier (cxx_parser_function.c:621)
==27073==    by 0x4AB1C7F: cxxParserLookForFunctionSignature (cxx_parser_function.c:966)
==27073==    by 0x4AAF663: cxxParserAnalyzeOtherStatement (cxx_parser.c:1505)
==27073==    by 0x4AAFEB3: cxxParserParseBlockInternal (cxx_parser_block.c:686)
==27073==    by 0x4AAFEB3: cxxParserParseBlock (cxx_parser_block.c:798)
==27073==    by 0x4AAE1C7: cxxParserMain (cxx_parser.c:1886)
==27073==    by 0x4B04D77: createTagsForFile (parse.c:3737)
==27073==    by 0x4B04D77: createTagsWithFallback1 (parse.c:3857)
```
This is for function parameter kinds enabled and it seems to be related to the situation when anonymous function parameter is discovered like in `void foo(int *)`. I suspect that normally `pIdentifier` points to something like token chain that is released by some other code while in this case the token is created and not freed in:
https://github.com/universal-ctags/ctags/blob/a7ae1df23ad048ab2f73a23049454ce96593ffea/parsers/cxx/cxx_parser_function.c#L2179

masatake:
~Close #3372.~
Partially close #3372.

The original code assumed tokens allocated in
cxxParserTokenChainLooksLikeFunctionParameterList() for tagging
anonymous parameters were deleted on its caller side
(cxxParserEmitFunctionParameterTags()).

However, in some conditions, cxxParserEmitFunctionParameterTags() are
not called. As a result, the allocated tokens are not deleted.

To avoid the memory leaking, the deletion cannot depend on
cxxParserEmitFunctionParameterTags(). This change uses per-parsing
trashbox for deleting the anonymous tokens. The objects linked to
the trashbox are destroyed at the end of parsing the current input
file.


techee:
I wanted to try this PR but the valgrind unit test still seems to fail here.

masatake:
> I wanted to try this PR but the valgrind unit test still seems to fail here.

Thank you for pointing out.

It seems that there are more than one bug.
One is fixed this pull request. 

It seems that the input-1.c derived from the issue you reported is not fixed with this PR.

I tried to find simplified input reproducing the same leak again. What I found is:

```
g (a () b c)) {
```

PR #3383, opened 2022-05-12
Cxx: skip __declspec()

### LINKED ISSUE #3382: C/C++: problem with __declspec in the parser
I have the problem that `__declspec(xyz)` attributes for classes are not correctly parsed, even when ignoring them using `-I __declspec+`.
This does seem to [work on exuberant ctags though](https://stackoverflow.com/questions/20382193/ctags-vim-trouble-handling-expressions-with-brackets-in-identifier-list).
Do you have any idea on how to still parse the file correctly (e.g., using an extra preprocessor macro`-D __declspec(x)=`)? It seems as if the two underscores somehow are incompatible with the preprocessor options. 
Thank you!

***

The name of the parser: C/C++

```
$ ctags --options=NONE -I __declspec+ --fields-all="*" --fields-c++=-{macrodef} --fields-c=-{macrodef} --fields=-Prtl --kinds-c=+p --kinds-c++=+p --output-format=json simple.cpp
```

The content of input file:

```cpp
namespace Base
{
    template<typename X, typename Y>
    class __declspec(abc) Foo
        : public B
        , public C
    {
    	void foo() {}
    };

    template<typename Z, typename W>
    class Bar : private D<Z,W>
    {
    	void bar() {}
    };

    template<>
    void Foo<int,int>::foo() {printf("hello world\n");}
}
```

```
{"_type": "tag", "name": "Base", "path": "simple.cpp", "file": true, "line": 1, "kind": "namespace", "extras": "fileScope"}
```

The tags output you expect:

```
{"_type": "tag", "name": "Bar", "path": "simple.cpp", "file": true, "line": 15, "kind": "class", "inherits": "D<Z,W>", "scope": "Base", "scopeKind": "namespace", "extras": "fileScope", "end": 18, "template": "<typename Z,typename W>"}
{"_type": "tag", "name": "Base", "path": "simple.cpp", "file": true, "line": 1, "kind": "namespace", "extras": "fileScope", "end": 22}
{"_type": "tag", "name": "Foo", "path": "simple.cpp", "file": true, "line": 4, "kind": "class", "inherits": "B,C", "scope": "Base", "scopeKind": "namespace", "extras": "fileScope", "end": 9, "template": "<typename X,typename Y>"}
{"_type": "tag", "name": "bar", "path": "simple.cpp", "file": true, "line": 17, "kind": "function", "access": "private", "signature": "()", "scope": "Base::Bar", "scopeKind": "class", "extras": "fileScope", "end": 17}
{"_type": "tag", "name": "foo", "path": "simple.cpp", "file": true, "line": 8, "kind": "function", "access": "private", "signature": "()", "scope": "Base::Foo", "scopeKind": "class", "extras": "fileScope", "end": 8}
{"_type": "tag", "name": "foo", "path": "simple.cpp", "line": 21, "kind": "function", "signature": "()", "scope": "Base::Foo", "scopeKind": "class", "end": 21, "template": "<>", "properties": "scopespecialization,specialization"}
```

The version of ctags:

```
$ ctags --version
Universal Ctags 5.9.0(12c20f1e), Copyright (C) 2015-2022 Universal Ctags Team
  Compiled: Apr 21 2022, 10:33:00
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +gnulib_fnmatch, +gnulib_regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc, +optscript
```

How do you get ctags binary:

I've tried both, macOS and Windows pre-built binaries from the win32 and homebrew projects.

masatake:
Close #3382

### DIFF
--- a/Tmain/input-encoding-option.d/run.sh
+++ b/Tmain/input-encoding-option.d/run.sh
@@ -9,6 +9,9 @@ BUILDDIR=$2
 . ../utils.sh
 
 if ${CTAGS} --quiet --options=NONE --list-features | grep -q iconv; then
+  check_encoding shift_jis
+  check_encoding euc-jp
+  check_encoding utf-8
   if ${CTAGS} --quiet --options=NONE \
 	      --pseudo-tags=-TAG_PROC_CWD \
 	      --input-encoding=utf-8 --input-encoding-java=shift_jis --input-encoding-javascript=euc-jp \
--- a/Tmain/invalid-encoding-option.d/run.sh
+++ b/Tmain/invalid-encoding-option.d/run.sh
@@ -8,6 +8,7 @@ CTAGS=$1
 . ../utils.sh
 
 if ${CTAGS} --quiet --options=NONE --list-features | grep -q iconv; then
+  check_encoding euc-jp
   ${CTAGS} --quiet --options=NONE --input-encoding-java=invalid	--input-encoding-javascript=euc-jp -o - input.js input.java
   exit $?
 else
--- a/Tmain/no-input-encoding-option.d/run.sh
+++ b/Tmain/no-input-encoding-option.d/run.sh
@@ -8,6 +8,7 @@ BUILDDIR=$2
 . ../utils.sh
 
 if ${CTAGS} --quiet --options=NONE --list-features | grep -q iconv; then
+  check_encoding utf-8
   ${CTAGS} --quiet --options=NONE --output-encoding=utf-8 -o $BUILDDIR/tags input.js input.java
   exit $?
 else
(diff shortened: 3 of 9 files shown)


masatake:
Failures unrelated to this change are reported on the ubi8 container, a platform I cannot ignore.

masatake:
The change for fixing is enough compact and self-closed. Even when @pragmaware returns and says, "I don't like this fix," we can revert this commit quickly.

I don't like this change because it throws away the information found in the source code in the early stage of parsing. The change makes it impossible to use the information in the upper stages of parsing and/or to store the information in fields of tag entries. However, this is the same as `__attribute__()`.

Issue #3380, opened 2022-05-09
error linking


zuloloxi:
hello
Used  nmake to generate  exe file, the line command:
       nmake -f mk_mvc.mak 
At the end, have these linking error:
....
Microsoft (R) Incremental Linker Version 14.29.30140.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:ctags.exe
setargv.obj
main/routines.obj
main/trashbox.obj
main/vstring.obj
main/args.obj
main/colprint.obj
main/dependency.obj
main/entry.obj
main/entry_private.obj
main/error.obj
main/field.obj
main/flags.obj
main/fmt.obj
main/htable.obj
main/keyword.obj
main/kind.obj
main/lregex.obj
main/lregex-default.obj
main/lxpath.obj
main/main.obj
main/mbcs.obj
main/nestlevel.obj
main/numarray.obj
main/objpool.obj
main/options.obj
main/param.obj
main/parse.obj
main/portable-scandir.obj
main/promise.obj
main/ptag.obj
main/ptrarray.obj
main/rbtree.obj
main/read.obj
main/script.obj
main/seccomp.obj
main/selectors.obj
main/sort.obj
main/stats.obj
main/strlist.obj
main/trace.obj
main/tokeninfo.obj
main/unwindi.obj
main/writer.obj
main/writer-etags.obj
main/writer-ctags.obj
main/writer-json.obj
main/writer-xref.obj
main/xtag.obj
main/CommonPrelude.obj
main/repoinfo.obj
main/mio.obj
parsers/abaqus.obj
parsers/abc.obj
parsers/ada.obj
parsers/ant.obj
parsers/asciidoc.obj
parsers/asm.obj
parsers/asp.obj
parsers/autoconf.obj
parsers/autoit.obj
parsers/automake.obj
parsers/awk.obj
parsers/basic.obj
parsers/beta.obj
parsers/bibtex.obj
parsers/c-based.obj
parsers/c.obj
parsers/clojure.obj
parsers/css.obj
parsers/cobol.obj
parsers/cpreprocessor.obj
parsers/cxx/cxx.obj
parsers/cxx/cxx_debug.obj
parsers/cxx/cxx_debug_type.obj
parsers/cxx/cxx_keyword.obj
parsers/cxx/cxx_parser.obj
parsers/cxx/cxx_parser_block.obj
parsers/cxx/cxx_parser_function.obj
parsers/cxx/cxx_parser_lambda.obj
[...]

PR #3391, opened 2022-05-18
win32: Adjust config_mvc.h

k-takata:
* MSVC doesn't have iconv (by default).
* MSVC 2013 or later has inttypes.h.
* MSVC doesn't have the 'truncate' function.

These changes don't have any impacts, though.

### DIFF
--- a/win32/config_mvc.h
+++ b/win32/config_mvc.h
@@ -178,10 +178,10 @@
 /* #undef HAVE_FTRUNCATE */
 
 /* Define this value if support multibyte character encoding. */
-#define HAVE_ICONV 1
+/* !!!#define HAVE_ICONV 1 */
 
 /* Define to 1 if you have the <inttypes.h> header file. */
-/* !!!#define HAVE_INTTYPES_H 1 */
+#define HAVE_INTTYPES_H 1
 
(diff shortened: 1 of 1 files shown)


hirooih (approved):
It took me some time to remember what I did.

https://github.com/universal-ctags/ctags/pull/3054#issuecomment-873386035
This is what I did on this file.

> * MSVC 2013 or later has inttypes.h.

At that time I had to do undefine HAVE_INTTYPES_H to pass CI on appveyor.  I think we used MSVC 2013 at that time.  It did not have inttypes.h.

According to #3380 we decided to use VC++ 2015 or later.

I think "MSVC 2015 or later has inttypes.h." is the correct statement.

I see no issues on others.

k-takata:
I have several versions of VC++, and I confirmed that VC++ 2012 doesn't have inttypes.h and VC++ 2013 has it.
And currently, we still use VC++ 2013 on AppVeyor. So, there might be some other issue at that time.

leleliu008:
Yes, `MSVS 2013` added the `inttypes.h`. https://devblogs.microsoft.com/cppblog/c99-library-support-in-visual-studio-2013/

masatake:
THANK YOU VERY MUCH.

PR #3392, opened 2022-05-18
win32: Adjust line endings of some files in the repository

k-takata:
Currently, some files under the win32 directory are committed in the CRLF line endings.
Add `.gitattributes` and make git commit them in LF and checkout them in CRLF.

Only the batch files and Visual Studio project files need to be in the CRLF line endings.
Convert the other files in the win32 directory to LF.

After this change, no `^M` marks will be shown in `git diff` and `git log -p`.


k-takata:
This PR depends on #3391. So, this should be rebased after it is merged.

masatake (approved):

k-takata:
`centos7_make` starts failing after I merged this PR:
https://app.circleci.com/pipelines/github/universal-ctags/ctags/2394/workflows/5a2c8ab2-aad0-4d29-b121-cc920e8487ba/jobs/25763
I don't see the CircleCI tests on this page. Aren't the tests executed when I just create a PR?

Perhaps, the git version coming with CentOS 7 is too old and does not support `.gitattributes`?

masatake:
...I'm surprised at the CircleCI tests were not executed. The test should be executed automatically when creating a PR.

> Perhaps, the git version coming with CentOS 7 is too old and does not support .gitattributes?

If this is true, CentOS 7 is not a good environment for running 'make check-genfile'.
Anyway, I will merge #3401 first.

Issue #3388, opened 2022-05-16
C++: problem with template function return type


delsner:
I have a problem with return types of template functions, where the C/C++ parser misses functions, if they return any non-primitive type. Do you have any idea how to circumvent or patch the problem when parsing C++ code? Thank you!

***

The name of the parser: C/C++

The command line you used to run ctags:

```
$ ctags --options=NONE --fields-all="*" --fields-c++=-{macrodef} --fields-c=-{macrodef} --fields=-Prtl --kinds-c=+p --kinds-c++=+p --output-format=json simple.cpp
```

The content of input file:

```cpp
template<typename A, typename B>
A& foo(A& bar, B const& baz)
{
    return bar;
}
```

The tags output you are not satisfied with:

```
[empty]
```

The tags output you expect:

```
{"_type": "tag", "name": "foo", "path": "simple.cpp", "line": 2, "kind": "function", "signature": "(A & bar,B const & baz)", "end": 5, "template": "<typename A,typename B>"}
```

The version of ctags:

```
$ ctags --version
Universal Ctags 5.9.0(de7fae6a), Copyright (C) 2015-2022 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: May 16 2022, 06:25:51
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +gnulib_fnmatch, +gnulib_regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc, +optscript
```

How do you get ctags binary:

macosx binary taken from Universal-ctags/homebrew-universal-ctags project
Win32 binary taken from Universal-ctags/ctags-win32 project

----

@masatake added:

```
#include <cstddef>

template <std::size_t S>
struct M {
};

template <T>
struct M<static_cast<std::size_t>(-1)> {
};

```

The second `M` is not tagged. Maybe the input exposes the same bug.
With the fix, "removing the code block", proposed by @delsner, the second M is tagged well.
Derived from #3263.

masatake:
This one is hard to fix.

delsner:
Could you maybe elaborate why? I'm not familiar with the code base. How would you approach it?

masatake:
[...]

PR #3393, opened 2022-05-19
C++: Fix parser for template function return types (issue #3388)

severin-kacianka:
This commit fixes issue #3388, where the C++ parser does not correctly parse function return types.

### DIFF
--- a/parsers/cxx/cxx_parser_template.c
+++ b/parsers/cxx/cxx_parser_template.c
@@ -45,26 +45,6 @@ static bool cxxTemplateTokenCheckIsNonTypeAndCompareWord(const CXXToken *t,void 
 	// To be non type the token must NOT be preceded by class/struct/union
 	if(!t->pPrev)
 		return false;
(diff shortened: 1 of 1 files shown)


masatake:
You can run the test cases on your PC: `make units LANGUAGES=C++`.

masatake:
Thank you for trying to fix the issue. However, I think your approach is too ad-hoc.
Deleting the whole code block approach, https://github.com/universal-ctags/ctags/issues/3388#issuecomment-1131781124, looks better for me now.

masatake:
@delsner, could you make your own pull request?
I cannot understand what happens here :-)

When making a new pull request, make it based on the latest master branch so the test case were updated.

delsner:
Alright, will do it later! Sorry for the inconveniences. :-)

masatake:
#3388 fixes the original issue.
@[severin-kacianka](https://github.com/severin-kacianka), thank you for trying to fix the original issue.

PR #3398, opened 2022-05-25
C++  fix invalid test input

masatake:
See #3388.


misc/validators/validator-cxx11
```diff
@@ -0,0 +1,32 @@
...
+STD=c++11
+action=$1
+input=$2
+case "$action" in
+    is_runnable)
+	type gcc > /dev/null 2>&1
+	exit $?
+	;;
+    validate)
+	gcc -fsyntax-only -std=${STD} "$input" > /dev/null 2>&1
```
masatake, line 29: Supressing stderr is not good idea.

masatake:
`The requested URL returned error: 429` is reported.
Just repeating "re-run failed jobs" doesn't help.

PR #3399, opened 2022-05-26
misc: improve the readability of misc/git-tag-maybe.sh

leleliu008:


misc/git-tag-maybe.sh
```diff
@@ -1,36 +1,46 @@
...
+}
+
+run() {
+    printf '%b\n' "${COLOR_PURPLE}==>${COLOR_OFF} ${COLOR_GREEN}$*${COLOR_OFF}"
+    eval "$*"
+}
+
+run git --version
+
 base=5.9
```
masatake, line 21: Could you move the definition of `${base}` to the beginning of the script?

masatake (approved):

masatake:
Let's see what happens this weekend!

leleliu008:
It will be like this:

Screenshot from 2022-05-26 17-58-51

masatake:
Fine!

PR #3400, opened 2022-05-26
AppVeyor,docs: Drop support for mingw.org

### LINKED ISSUE #3397: AppVeyor,docs: drop the support for the original mingw coming from MSYS 1.0
- [ ] delete the environment from appveyor.yml
- [ ] write about the "not supporting"  to README or somewhere

----

@lambertlulala 
`off_t` is always 32 bits on MSVC. So, I think it's better to define our own off_t.

@masatake 
> Is "the original version of mingw" equal to "MSYS2 / testing (MINGW32)" ?

No. It is not installed in the GHA environment and it is installed in `C:\MinGW` in AppVeyor.
It comes with MSYS 1.0.

(edited)
Currently, we are using the original mingw on the `compiler=mingw` environment on AppVeyor:
https://ci.appveyor.com/project/universalctags/ctags/builds/43598413/job/56nvcd9jmdn28370
It might be a good time to drop support for the original mingw.

_Originally posted by @k-takata in https://github.com/universal-ctags/libreadtags/issues/37#issuecomment-1132332037_

k-takata:
Close #3397

* Change the `mingw` build on AppVeyor to use mingw-w64 (from msys2) instead of the original version of mingw.
  (Should we move the `mingw` build from AppVeyor to GitHub Actions?)
* Updated the document.

### DIFF
--- a/docs/windows.rst
+++ b/docs/windows.rst
@@ -28,13 +28,17 @@ Note that ctags cannot be built with Visual Studio older than 2013 anymore. Ther
 GCC
 .............................................................................
 
-There are three flavors of GCC for Windows:
+You can use `MinGW-w64 <https://www.mingw-w64.org/>`_ to build ctags. There are several ways to install MinGW-w64 in Windows.
 
-- MinGW http://www.mingw.org
-- MinGW-w64 http://mingw-w64.sourceforge.net
-- TDM-GCC http://tdm-gcc.tdragon.net
+- MSYS2 https://www.msys2.org/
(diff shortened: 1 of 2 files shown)


k-takata:
It should be a future task to move the test of `make -f mk_mingw.mak` from AppVeyor to GitHub Actions.

k-takata:
FreeBSD 13 is failing. I think it is unrelated to this PR, but why?

masatake (approved):

PR #3401, opened 2022-05-26
test(check-genfile): show diffs when running on a CI environment

masatake:


masatake:
Though I added `git diff`, it showed nothing interesting.

leleliu008:
why show diff twice?

masatake:
> why show diff twice?

I wrongly assumed `--exit-code` suppresses the output.

PR #3405, opened 2022-06-01
GitHub Actions: fix vagrant the requested URL returned error: 429

leleliu008:

### DIFF
--- a/.github/workflows/testing-freebsd.yml
+++ b/.github/workflows/testing-freebsd.yml
@@ -18,6 +18,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    # to resolve The requested URL returned error: 429
+    # I do not known why, but it works indeed.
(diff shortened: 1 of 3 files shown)


masatake:
Even though we rebased some branches related to pull requests on the master branch merging this pull request, the 429 is reproduced. 

I found a page about the 429 of vagrant:
https://github.com/hashicorp/vagrant/blob/main/website/content/vagrant-cloud/request-limits.mdx

Is there a way to dump X-RateLimit-Limit, X-RateLimit-Remaining, and X-RateLimit-Reset https headers when we got the 429 ? I guess what we can do for overcoming the 429 is limited. But I wonder if we can find some hints in the dump of the headers.

leleliu008:
It is almost certainly limited by the server.

add following to our workflow file will output debug info including HTTP prococal.
```
env:
      VAGRANT_LOG: debug
```

PR #3407, opened 2022-06-02
GitHub Actions: add debug vagrant support

leleliu008:

### DIFF
--- a/.github/workflows/testing-freebsd.yml
+++ b/.github/workflows/testing-freebsd.yml
@@ -18,16 +18,13 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    # to resolve The requested URL returned error: 429
-    # I do not known why, but it works indeed.
-    - run: rm -rf ~/.vagrant.d
-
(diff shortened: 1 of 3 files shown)


masatake:
Testing on s390 is a severe bottleneck; it takes more than three hours.
After some trials, I will revert the change for introducing testing on s390 if the trails are never completed.

masatake:
@leleliu008, on the ctags side, this pull request is acceptable, of course. 
However, if I maintain leleliu008/github-actions-vagrant@v1, I will put the code for setting the environment variable to leleliu008/github-actions-vagrant@v1 and add an interface.
On the ctags side, a .yml uses the interface like:


```
    - uses: leleliu008/github-actions-vagrant@v1
      with:
        mem: 2048
        box: generic/openbsd6
        debug: true
...
```

So we can reduce the knowledge about vagrant on the ctag side.
Just an idea.

leleliu008:
OK, I will update leleliu008/github-actions-vagrant@v1

masatake (approved):

PR #3410, opened 2022-06-15
Remove punctuation from SCSS classes & identifers

### LINKED ISSUE #3409: Tags generated for SCSS include characters that prevent completion
[In my original](https://gist.github.com/Roy-Orbison/71bc81f488f85adaeacfb76a7967eda0#file-scss-ctags-L51), the punctuation used to denote classes is excluded from the tag itself. In @masatake's [revised version](https://github.com/universal-ctags/ctags/blob/bfc40942a45132ab6c3efa12f34cc623f602c41b/optlib/scss.ctags#L62), the `%` and `.` are included, which breaks autocompletion in Vim because those characters are not normally part of keywords. One can add them with `isk+=%,.` but this breaks chaining classes as `%foo.bar` is then seen by Vim as a single keyword where humans and uctags do not.

Also,  the insertion of `:` into the keyword characters of `c`, `P` & `i` cause a similar issue to the above, and results in tags like `.foo:not`. These are a bit unexpected because pseudo-classes have language-level completion support.

I don't want to create a PR to undo those changes without knowing why they were made, so could someone please clarify those decisions?

Applying this patch, my tags work as expected with only `isk+=-` in my `.vimrc`.
```patch
--- a/optlib/scss.ctags
+++ b/optlib/scss.ctags
@@ -59,9 +59,9 @@
 --_mtable-regex-SCSS=toplevel/:[^\n{]+[;{]\n?//
 --_mtable-regex-SCSS=toplevel/\$([A-Za-z0-9_-]+)[ \t]*:[ \t]*\(/\1/v/{tenter=map}
 --_mtable-regex-SCSS=toplevel/\$([A-Za-z0-9_-]+)[ \t]*:[^\n]*\n?/\1/v/
---_mtable-regex-SCSS=toplevel/([.][A-Za-z0-9_:-]+)/\1/c/
---_mtable-regex-SCSS=toplevel/(%[A-Za-z0-9_:-]+)/\1/P/
---_mtable-regex-SCSS=toplevel/#([A-Za-z0-9_:-]+)/\1/i/
+--_mtable-regex-SCSS=toplevel/[.]([A-Za-z0-9_-]+)/\1/c/
+--_mtable-regex-SCSS=toplevel/[%]([A-Za-z0-9_-]+)/\1/P/
+--_mtable-regex-SCSS=toplevel/#([A-Za-z0-9_-]+)/\1/i/
 --_mtable-regex-SCSS=toplevel/.//
 --_mtable-regex-SCSS=comment/\*\///{tleave}
 --_mtable-regex-SCSS=comment/.//
```

Roy-Orbison:
Closes #3409.

### DIFF
--- a/Units/parser-scss.r/placeholder.d/expected.tags
+++ b/Units/parser-scss.r/placeholder.d/expected.tags
@@ -1,6 +1,6 @@
-.alert:hover	input.scss	/^.alert:hover, %strong-alert {$/;"	c
-%strong-alert	input.scss	/^.alert:hover, %strong-alert {$/;"	P
-%strong-alert:hover	input.scss	/^%strong-alert:hover {$/;"	P
-%toolbelt	input.scss	/^%toolbelt {$/;"	P
-.action-buttons	input.scss	/^.action-buttons {$/;"	c
-.reset-buttons	input.scss	/^.reset-buttons {$/;"	c
+alert	input.scss	/^.alert:hover, %strong-alert {$/;"	c
+strong-alert	input.scss	/^.alert:hover, %strong-alert {$/;"	P
+strong-alert	input.scss	/^%strong-alert:hover {$/;"	P
+toolbelt	input.scss	/^%toolbelt {$/;"	P
+action-buttons	input.scss	/^.action-buttons {$/;"	c
+reset-buttons	input.scss	/^.reset-buttons {$/;"	c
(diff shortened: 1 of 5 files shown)


masatake:
You can run the test locally with `make units LANGUAGES=SCSS`.

You have to update the following files:

* Units/parser-scss.r/placeholder.d/expected.tags
* Units/parser-scss.r/variable.d/expected.tags
* Units/parser-scss.r/selectors.d/expected.tags

masatake:
Could you add optlib/scss.c to your commit?

masatake:
Could you write the background and idea of the change to your commit log?

masatake:
Can I ask two more things?

* Put `Closes #3409` to the commit log, and
* Use `SCSS: ` as the prefix of the header of the commit log like:
  `SCSS: remove punctuation from classes & identifers`

masatake:
Thank you.

PR #3416, opened 2022-06-20
#3388 Bugfix for cpp template function return types

delsner:
Fixes issue #3388, where the C++ parser does not correctly parse template function return types and therefore misses functions.

This PR removes C++ template parsing code that was originally written to support C++03-specific template syntax. However, as described in #3388 this syntax is not actually valid C++03. Yet, this code prevents correctly parsing non-primitive template function return types.


masatake:
Could you add a test case?
http://docs.ctags.io/en/latest/testing-parser.html#units-test-facility explains how to write a test case for "Units".

delsner:
The codecov coverage threshold check is failing and there also seems to be a flaky test here: https://github.com/universal-ctags/ctags/runs/7044140203; I also added a small fix for a comment that was introduced in #3398 (see aaf6ca24)

delsner:
@masatake any more comments on this PR?

masatake:
[7560461](https://github.com/universal-ctags/ctags/pull/3416/commits/75604613e00bbcaed9effe9a542de3939d509103)
[7560461](https://github.com/universal-ctags/ctags/pull/3416/commits/75604613e00bbcaed9effe9a542de3939d509103)
[96f45ea](https://github.com/universal-ctags/ctags/pull/3416/commits/96f45ea6132fbe477733288c0c9615dda9172c1c)

These can be squashed into one commit. We should not record your try-and-error commits to the ctags history if possible.

About [e7fbffe](https://github.com/universal-ctags/ctags/pull/3416/commits/e7fbffe84bf2bf45207bb88514475e590048c983), I would like you to start the header of the commit log with "C++: ...".

Instead of putting the string `#3388` to the header, could you put `Close #3388.` to the body of the commit log?

You explained the change in the first comment of this pull request.
This is good.  However, the commit log is too short. Could you write the commit log as much like the initial comment like https://github.com/universal-ctags/ctags/commit/55ba7264c0722c9519c0fdfacdf2363ad1c59e94.

masatake:
I forgot to write the most important message; thank you.

I'm not sure whether I can ask this or not but...
You deleted code from cxxParserParseTemplateAngleBracketsInternal().
You will see many comments in cxxParserParseTemplateAngleBracketsInternal().
Could you try making the comments consistent with the deletion?

delsner:
thanks for the feedback, I'll incorporate it tomorrow. I expected to squash the entire PR into a single commit after all, but I'll rewrite the history as requested!

delsner:
@masatake feel free to check again. Regarding making the comments more consistent: I added an explanation on why parsing non-nested greater than signs is not necessary (syntax error) and left a note pointing to the issue where the deleted code used to be.

masatake:
(The failure on Cygwin is nothing to do with this pull request.)

masatake:
I'm trying to fix the failures on Cygwin.

masatake:
Could you rebase your change on the LATEST main branch?

delsner:
> Could you rebase your change on the LATEST main branch?

done, let's see!

parsers/cxx/cxx_parser_template.c
```diff
@@ -421,35 +424,8 @@ cxxParserParseTemplateAngleBracketsInternal(bool bCaptureTypeParameters,int iNes
...
-				{
-					CXX_DEBUG_PRINT("Treating as greater-than sign");
-
-					if(cxxTokenTypeIsOneOf(g_cxx.pToken,CXXTokenTypeOpeningParenthesis))
-						cxxParserUngetCurrentToken(); // needs to be condensed
-
-					continue;
-				}
-
+                // We do not check for the greater than operator, as it is only valid
```
masatake, line 427: A minor request.
Could you use "tab" characters instead of space characters for indentation?

parsers/cxx/cxx_parser_template.c
```diff
@@ -269,7 +269,10 @@ cxxParserParseTemplateAngleBracketsInternal(bool bCaptureTypeParameters,int iNes
 	// around but without proper state (include headers, macro expansion, full type
 	// database etc) we simply can't do the same. However, we try to recover if we
 	// figure out we (or the programmer?) screwed up.
-	//
+    // For 'greater-than' operators, the first non-nested operator is taken as the
```
masatake, line 272: A minor request.
Could you use "tab" characters instead of space characters for indentation?

masatake:
I read the changes. Thank you. It looks good to me. I will merge this pull request.
Before merging, could you see the minor requests?

parsers/cxx/cxx_parser_template.c
delsner: ✅

parsers/cxx/cxx_parser_template.c
delsner: ✅

masatake (approved):

masatake:
Thank you very much.

PR #3418, opened 2022-06-25
tmain: suppress the error message reported when running "iconv -l" 

masatake:
"ioconv -l" of Cygwin running on Git Hub Actions reports "I/O erorr".
The message may come from https://git.savannah.gnu.org/gitweb/?p=libiconv.git;a=blob;f=src/iconv.c;h=e3932e5f18e0484972aa1f92c797e5f13ae52937;hb=HEAD#l1130 .

As far as reading the condition surrounding the code for printing the messasge,
we can ignore it.

### DIFF
--- a/Tmain/utils.sh
+++ b/Tmain/utils.sh
@@ -138,7 +138,7 @@ direq_maybe ()
 
 check_encoding()
 {
-    if iconv -l | grep -qi "$1"; then
+    if iconv -l 2> /dev/null | grep -qi "$1"; then
 		return 0
 	fi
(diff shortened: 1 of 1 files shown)


masatake:
The test case reports:
```
+iconv.exe: I/O error
```

This comes from:
https://git.savannah.gnu.org/gitweb/?p=libiconv.git;a=blob;f=src/iconv.c;h=e3932e5f18e0484972aa1f92c797e5f13ae52937;hb=HEAD#l1130

Strange but we can ignore the error.

PR #3422, opened 2022-06-28
Removed the `$` sign

rehanguha:
This PR will help to for utilising the copy functionality from GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,12 +59,12 @@ code from GitHub.
 
 To build with Autotools (Autoconf and Automake) on GNU/Linux, OSX, or Windows 10 WSL,
 ```
-    $ git clone https://github.com/universal-ctags/ctags.git
-    $ cd ctags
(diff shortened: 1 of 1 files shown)


masatake:
I prefer to keep this `$`. `$` is a great hint for "this is a command-line".
In the context, understanding is more important than optimization of operation.

rehanguha:
> I prefer to keep this `$`. `$` is a great hint for "this is a command-line". In the context, understanding is more important than optimization of operation.

I feel what you are referring makes sense and optimisation of the operation is also important in terms of ease of usability.
Will it make sense if we write a one liner above the code section about **The below commands can be executed as an normal user i.e. we do not need `root` permissions**. (We can alter the statement as per our liking as well.)

rehanguha:
It makes sense. Nothing much to add from my side. :+1:

Issue #3427, opened 2022-07-05
JavaScript: don't make null tags


masatake:
With ctags built with `--enable-debugging`, the following input hits an Assert:
```
const { extendDefaultPlugins } = require( 'svgo' );
```

```
ctags: Warning: ignoring null tag in /tmp/foo.js(line: 1)
ctags: main/entry.c:1454: registerEntry: Assertion `corkIndex != CORK_NIL' failed.
ctags: main/entry.c:1454: parsing /tmp/foo.js:1 as JavaScript
zsh: IOT instruction (core dumped)  ./ctags /tmp/foo.js
```

With the following change, we can avoid the crash:
```
diff --git a/parsers/jscript.c b/parsers/jscript.c
index c226d651b..06a2204b4 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -451,6 +451,8 @@ static int makeJsRefTagsForNameChain (char *name_chain, const tokenInfo *token,
        char *next = strchr(name_chain, '.');
        if (next)
                *next = '\0';
+       if (!*name)
+               return CORK_NIL;
        int index = bestJSEntryInScope (scope, name);
 
        if (index == CORK_NIL)
@@ -488,6 +490,8 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 {
        int index = CORK_NIL;
        const char *name = vStringValue (token->string);
+       if (!*name)
+               return CORK_NIL;
 
        const char *p;
        char *name_chain = NULL;
```

However, I wonder whether we should fix these callees receiving empty names or callers passing empty names.  See is strongly related to #900

Of course, ctags should be able to extract `extendDefaultPlugins`. However, it is not the point of this issue. See #1112.

masatake:
I would like to withdraw this because fixing callers is better.

Issue #1112, opened 2016-09-06
JavaScript: destructural binding


techee:
When playing with various languages I ran into this warning for this file:

https://github.com/facebook/react/blob/master/src/renderers/shared/fiber/ReactFiberBeginWork.js

It looks the code doesn't like lines like

```
var { findNextUnitOfWorkAtPriority } = require('ReactFiberPendingWork');
```

and the warning comes from

https://github.com/universal-ctags/ctags/blob/master/parsers/jscript.c#L1778

because the `name` tag isn't checked whether it's an identifier or something else - it's probably { in this case.

Not a js developer myself so leaving this to someone who knows the correct syntax.

Strangehill:
I also get this warning with array destructuring :
const [lang, setLang] = useState("en")

masatake:
https://github.com/universal-ctags/ctags/pull/2115
I found I tried to fix this issue once.
https://github.com/universal-ctags/ctags/pull/2115/commits/e977d56e5cc0531651c80c1481928b791a3c3ade

Though I opened the pull request, the way to fix the issue in the pull request looks a bit tricky for me.
In addition, as @Strangehill wrote, the fix doesn't handle the array destructuring.

I will fix this but I cannot say when. I have to finish rewriting R parser.

See also #2494.

masatake:
I'm working on this item.
The code in my PC partially works.
```
$ cat /tmp/foo.js
cat /tmp/foo.js
var [a, b, ...rest] = [1, 2, 3];
$ ./ctags -o - /tmp/foo.js 
./ctags -o - /tmp/foo.js 
a	/tmp/foo.js	/^var [a, b, ...rest] = [1, 2, 3];$/;"	v
b	/tmp/foo.js	/^var [a, b, ...rest] = [1, 2, 3];$/;"	v
rest	/tmp/foo.js	/^var [a, b, ...rest] = [1, 2, 3];$/;"	v
$ 
```

Issue #900, opened 2016-04-24
javascript: Got "ctags: Warning: ignoring null tag" 


masatake:
```
[yamato@x201]~/var/ctags-github% cat /tmp/foo.js 
if (true) {
    this[0] = function() {};
}

[yamato@x201]~/var/ctags-github% ./ctags -o - /tmp/foo.js 
ctags: Warning: ignoring null tag in /tmp/foo.js(line: 2)
[yamato@x201]~/var/ctags-github% js /tmp/foo.js 
[yamato@x201]~/var/ctags-github%
```

cristianoliveira:
Is there any update on this?

cristianoliveira:
Apparently it is related to Object destructuring like:

```js
    const { name, email } = user;

```

masatake:
I have not inspected the warning well.
However, the input you gave looks valid as I tested the input on node command
I would like you to open a new issue for the input.
If you are good at JavaScript, I would like to know the specification as reference.
See  https://github.com/universal-ctags/ctags/issues/1949#issuecomment-457829914 how I struggled.

```
const { name, email } = user;
```
`name` and `email` should be tagged well.

```

[yamato@slave]~% node
> var user = {name: 'a', email: 'b'}
var user = {name: 'a', email: 'b'}
undefined
> const { name, email } = user;
const { name, email } = user;
undefined
> name
name
'a'
```

cristianoliveira:
> If you are good at JavaScript, I would like to know the specification as reference.

Not sure if I got your question but the specification for the Object Destructuring is on the ECMAScript specification:
https://www.ecma-international.org/ecma-262/6.0/#sec-destructuring-assignment

:)

masatake:
Thanks. That is what I need.

winkee01:
Why does this issue not be addressed? I've got the same problem.

winkee01:
> Thanks. That is what I need.

how did you remove the warning?

masatake:
> how did you remove the warning?

I didn't remove the warning.

winkee01:
> > how did you remove the warning?
> 
> I didn't remove the warning.

well, I've got the same issue, `{A, B} = some_obj` can be parsed correctly by ctags.

winkee01:
Looks like no one is going to fix this issue

masatake:
> Looks like no one is going to fix this issue

For the destructural binding, I tried once.

```
[jet@living]/tmp% cd /tmp
cd /tmp
[...]

PR #3428, opened 2022-07-05
JavaScript: avoid assertion failures when making an empty tag

masatake:
Related to #900, #1112, and #3427.

When the parser tries making a tag with empty name, an assertion
in registerEntry fails. The parser side should be fixed but it
will take a time. This change puts guard conditions around
registerEntry to avoid the failures.

THIS IS A TEMPORARILY SOLUTION.


jafl (approved):
I was just going to submit a PR for this :)

masatake:
Testing on BSDs becomes a bottleneck of development.

masatake:
> I was just going to submit a PR for this :)

I recognized the assertion failures with `codebase`.
https://docs.ctags.io/en/latest/contributions.html#add-realistic-examples-for-you-parser-to-codebase

You know JavaScript well. So you can run semi-test JavaScript in your brain.
However, I am not. This is not only JavaScript but all other languages than C.
So I have gathered REAL source code for various languages for testing parsers of ctags.

See https://github.com/universal-ctags/codebase. I think this is useful for a developer even if the one knows well about the target language.

PR #3430, opened 2022-07-06
Javascript fixes

jafl:
Make variable the default instead of function; fix args passed to anyKindsEntryInScope()
Categorize properties correctly and tag array properties
Patch to cover another case that looks like a getter/setter, but isn't
Bug fix to avoid crash on kind=ghost
Improve debugging output
Turn on generation of fully-qualified tags


masatake:
It seems that you must update `Tmain/list-roles.d/stdout-expected.txt`.
You can run `make tmain` on your PC.

`misc/review` script may help you to visit the difference between the output of the test case and the expected value of the test case.

masatake:
Surprisingly someone wrote about misc/review.
See https://docs.ctags.io/en/latest/testing-parser.html?highlight=misc%2Freview#reviewing-the-result-of-units-test

parsers/jscript.c
```diff
@@ -3130,6 +3130,7 @@ extern parserDefinition* JavaScriptParser (void)
 	def->keywordTable = JsKeywordTable;
 	def->keywordCount = ARRAY_SIZE (JsKeywordTable);
 	def->useCork	= CORK_QUEUE|CORK_SYMTAB;
+	def->requestAutomaticFQTag = true;
```
masatake, line 3119: How about adding a test case that runs ctags with --extras=+q ?

parsers/jscript.c
```diff
@@ -267,14 +267,14 @@ static roleDefinition JsClassRoles [] = {
...
-	  .referenceOnly = false, ATTACH_ROLES(JsFunctionRoles) },
+	{ true,  'f', "function",	  "functions"        },
 	{ true,  'c', "class",		  "classes",
 	  .referenceOnly = false, ATTACH_ROLES(JsClassRoles)    },
 	{ true,  'm', "method",		  "methods"          },
 	{ true,  'p', "property",	  "properties"       },
 	{ true,  'C', "constant",	  "constants"        },
-	{ true,  'v', "variable",	  "global variables" },
+	{ true,  'v', "variable",	  "global variables",
+	  .referenceOnly = false, ATTACH_ROLES(JsVariableRoles) },
```
masatake, line 277: This change requires updating the stdout-expected.txt files under Tmain.
Run `make tmain` locally.

masatake:
I added some minor comments. But the changes are good; significant improvements.

parsers/jscript.c
jafl: Done

parsers/jscript.c
jafl: Done

masatake (approved):

PR #3434, opened 2022-07-09
JavaScript: minor changes for future development

masatake:
The next one is highlight of this pull request:
```
The token type will be needed for supporting destructing
like:

	let [a, ...rest] = foo();

```


parsers/jscript.c
```diff
@@ -2598,7 +2622,7 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
 					}
 
 					indexForName = isClassName (name) ?
-						makeClassTag (name, signature, NULL):
```
jafl, line 2601: Remove trailing space

parsers/jscript.c
```diff
@@ -1029,7 +1030,30 @@ static void readTokenFullRaw (tokenInfo *const token, bool include_newlines, vSt
...
+				if (d != '.')
+				{
+					ungetcToInputFile (d);
+					ungetcToInputFile ((int)'.');
+					break;
+				}
+
+				token->type = TOKEN_DOTS;
+				if (repr)
+					vStringCatS (repr, "..");
```
jafl, line 1054: Should this be 3 dots?

parsers/jscript.c
```diff
@@ -124,7 +124,8 @@ typedef enum eTokenType {
 	 * Used only in readTokenFull or lower functions. */
 	TOKEN_ATMARK,
 	TOKEN_BINARY_OPERATOR,
-	TOKEN_ARROW
+	TOKEN_ARROW,
+	TOKEN_DOTS,					/* ... */
```
jafl, line 128: Remove trailing comma

jafl (changes requested):

parsers/jscript.c
masatake: This diff line starts from `-`.
This line removes the trailing space you introduced in 6d85089456 ( https://github.com/universal-ctags/ctags/commit/6d85089456ed215ce6b6a673744ae42ccc5e0e99#r78107639 ) .

parsers/jscript.c
masatake: The first dot is already appended to `repr` at line 1023:
```
  1018          if (repr && c != EOF)
  1019          {
  1020                  if (i > 1)
  1021                          vStringPut (repr, ' ');
  1022                  vStringPut (repr, c);
  1023          }
```

parsers/jscript.c
```diff
@@ -1029,7 +1030,30 @@ static void readTokenFullRaw (tokenInfo *const token, bool include_newlines, vSt
...
+				{
+					ungetcToInputFile (d);
+					break;
+				}
+
+				d = getcFromInputFile ();
+				if (d != '.')
+				{
+					ungetcToInputFile (d);
+					ungetcToInputFile ((int)'.');
```
masatake, line 1048: Do I need this cast operation, `(int)`?

parsers/jscript.c
masatake: This is intentional.

Adding ',' at the end of the last element in a list is a good practice because a future diff adding a new element at the end of the list can be smaller; when adding TOKEN_foo, just adding a line is what we have to do like:
```diff
    TOKEN_arrow,
+   TOKEN_foo,
```
We will not have to do like:
```diff
-    TOKEN_arrow
+    TOKEN_arrow,
+    TOKEN_foo
```

With this technique, we can make the diff two lines smaller.
Generally, smaller diff helps people understand the diff.

(I'm talking about the size of a commit. I am not talking about the number of commits in a pull request.)

In my experience, this is a popular technique:
* https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/netlink.h#n186
* https://sourceware.org/git/?p=glibc.git;a=blob;f=include/stdio.h;h=a6f7fd43cb664eb55f4987dfa830466bba38aa5b;hb=HEAD#l149
* https://github.com/llvm/llvm-project/blob/41fba3c107a5bc99065f3bf8b9f5b9d52eab2d98/llvm/include/llvm/BinaryFormat/ELF.h#L322

It is popular technique even in ctags:
* https://github.com/universal-ctags/ctags/blob/master/parsers/jscript.c#L97
* https://github.com/universal-ctags/ctags/blob/2052bbfcb2e06a4728417010bfa6fbb00e16a5bf/parsers/jscript.c#L254
* https://github.com/universal-ctags/ctags/blob/2052bbfcb2e06a4728417010bfa6fbb00e16a5bf/parsers/jscript.c#L281

An exception is a list we don't assume that a developer adds an element to the end of:
```
enum foo {
   A,
   B,
   COUNT_foo
};
```
This code implies a developer should add a new element before `COUNT_foo`.

A parser maintainer like you is the most valuable in this project. In most of all cases, the preferences of the parser maintainer have higher priority than the project's common coding convention. If you don't like the last ',' in the last element in a list, let me know. I will withdraw this change.

To make my intention clear, I will make a new pull request just for this change.

parsers/jscript.c
masatake: I will add a comment explaining why "TWO dots" is enough here.

masatake: See #3436.

parsers/jscript.c
masatake: See #3436.

parsers/jscript.c
masatake: We don't need this. I will remove this.

masatake: Removed in a newly added commit.

parsers/jscript.c
masatake: I added a comment in a newly added commit.

parsers/jscript.c
jafl: Doh!  Sorry!

parsers/jscript.c
jafl: OK, I'll remember that for my future patches.

jafl (approved):

PR #3433, opened 2022-07-08
JavaScript: extract arrow functions

masatake:
Extracting `id` in 
```
var id = (x) => {
  return x;
}
```

PR #3437, opened 2022-07-12
javascript: improved parsing of anonymous functions

jafl:
This is much simpler than #2005


jafl:
@masatake If I do a major refactor of `parseStatement` to break it up, will that cause major pain for any of your other jscript.c PR's?

masatake:
Incredible!

masatake (approved):

masatake:
Let's ignore the testing on NetBSD.

masatake:
> @masatake If I do a major refactor of `parseStatement` to break it up, will that cause major pain for any of your other jscript.c PR's?

I submitted #3433 and #3435. However, I don't care if your future pull request breaks them. Refactoring has a much higher priority!

masatake:
BTW if you want to let users of the JavaScript parser know the design of the parser, you can write a man page for it. See https://docs.ctags.io/en/latest/man-pages.html and https://docs.ctags.io/en/latest/man/ctags-lang-python.7.html as examples.

PR #3445, opened 2022-07-16
javascript: capture arrow functions…

jafl:
…that are not assigned to global variables


parsers/jscript.c
```diff
@@ -2849,11 +2858,15 @@ static bool parseStatement (tokenInfo *const token, bool is_inside_class)
 			}
 			if (isType (token, TOKEN_CLOSE_CURLY))
 				is_terminated = false;
-
```
masatake, line 2852: Minor comment. For the original goal of the pull request, you don't have to delete this line. If you want to delete this line, you can make a separate commit just for deleting the line. I have applied this rule to my changes and contributors' changes.
`-p` option of `git add` helps us split a change into two.

However, about the Javascript parser, you are the maintainer. 
So you don't have to follow the rule.
So this comment is just a comment.

masatake (approved):

parsers/jscript.c
jafl: It was a mistake.  I put it back.

masatake:
Let me review this pull request again. I got a question about the life cycle of `sig`.

masatake:
I solved my question.

masatake (approved):

PR #3439, opened 2022-07-13
GitHub actions: revise .yml files for BSD platforms

masatake:

PR #3446, opened 2022-07-17
GitHub Actions: serialize the testing on BSDs

masatake:
We have gotten "429 Too Many Requests" failures when running test cases
on the BSD instances.

While monitoring which one fails, I found an ambiguous rule. We have 4
BSD targets. Only one or two of them got the 429.

My guess of the reason for 429 is that multiple BSD targets start at
the same time or within a short period; the volume of the download
access from vagrant cloud reaches the limit formulated by Vaygrant
Cloud.

This experimental change serializes the execution of tests on BSDs
to limit the volume.

@leleliu008 provides the actual way for serializing at
https://github.com/universal-ctags/ctags/pull/3439#issuecomment-1183709443

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -45,3 +45,6 @@ end_of_line = crlf
 
 [win32/*.vcxproj*]
 insert_final_newline = false
+
+[.github/workflows/*.yml]
+trim_trailing_whitespace = true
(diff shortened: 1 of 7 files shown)


masatake:
???Tests on NetBSD and OpenBSD do not appear.

leleliu008:
this way have some limitations:

Screenshot from 2022-07-17 20-07-35

if this way do not work, I think we can store these vagrant box files on our GitHub, then we fetch these vagrant box files from GitHub.

leleliu008:
very strange, our workflow files obviously meet the requirements.

masatake:
#3447 fixed the issue I tried to solve with this pull request.

PR #3447, opened 2022-07-17
GitHub Actions: To reduce error: 429 Too Many Requests

leleliu008:
Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context


masatake:
Fine! Thank you.

masatake (approved):

PR #3452, opened 2022-07-27
Add:  Ada 2012 and Ada 2022 keywords ("some" and "parallel")

spr93:


masatake:
Thank you.
Could you add `:` to to
```
Add Ada 2012 and Ada 2022 keywords```
---^
```

Without the change proposed in this pull request, what kind of input cannot ctags handle well?
In the other words, could you add a test case?

masatake:
https://docs.ctags.io/en/latest/testing-parser.html expalins how to write a test case for a parser.

spr93:
I want to make sure that you saw that the only proposed changes are two lines that add two keywords to the existing keyword enum. The keywords are not unusually long; they do not use they have any non-A-Z characters; and they do not make the ada.c keyword enum unique in terms of how many values it has. I don't see how any future issue could be due to a couple new keywords in these circumstances, and a new test, etc. seems overkill.

I'm more than happy to explain how the proposed commit improves ctag's Ada input handling:
One added line is:
ADA_KEYWORD_SOME,		/* Ada 2012 */
Documentation:
http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-2-9.html
http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-4-5-8.html
http://www.ada-auth.org/standards/12rat/html/Rat12-1-3-2.html

The other is
ADA_KEYWORD_PARALLEL,      /* Ada 2022 */
Documentation:
http://www.ada-auth.org/standards/2xrm/html/RM-2-9.html#I1388
http://www.ada-auth.org/standards/2xrm/html/RM-5-6-1.html#S0192

Note that the /* Ada 20x2 */ comments in the proposed commit follow the existing style:  A pre-existing commit followed the same format when it added a reserved word adopted by the Ada 2005 standard:
ADA_KEYWORD_OVERRIDING, /* Ada 2005 */ 

Best regards

masatake:
What I would like to know is why we should add the keywords.
Adding the keywords changes the behavior of ctags?
If yes, we may need a test case.
If not, we don't need this change.

masatake:
@spr93 ping.

masatake:
I could not understand how this pull request improves ctags.
I guess the author assumes ctags is a syntax highlighter like pygments.

PR #3453, opened 2022-07-29
GPerf: new parser

masatake:


masatake:
I wonder why the coverage for gperf.c is 0%.

PR #3456, opened 2022-08-10
ttcn: add "template" ["(omit)" | "(value)" | "(present)"]

neeels:
Modern TTCN3 has TemplateRestriction keywords, which improve detection
of invalid template use at compile time. They simply add one of
"(omit)", "(value)", "(present)" after the "template" keyword.

Example:

 template (value) ts_FOO_Request(integer bar := 0) :=
 	{ member := bar };

 template (present) tr_FOO_Request(template integer bar := *) :=
 	{ member := bar };

Before this patch, u-ctags would not tag these two definitons above,
making my work on the Osmocom test suite hard. (See
https://gitea.osmocom.org/ttcn3/osmo-ttcn3-hacks)

At first I tried to exactly match the three keywords within the braces,
but the token parsing does not support unrolling more than one token. So
if an unmatching token follows after the opening brace, I cannot un-get
the opening brace. The easiest solution, and one that also is tolerant
to keyword typos, is to just accept any braces after "template".

Related: ETSI ES 201 873-1 V4.14.1 § A.1.6.1.3


masatake:
Thank you. Could you add a test case?
See https://docs.ctags.io/en/latest/testing-parser.html about how to write a test case.

masatake:
@akrzyz, could you review this?

akrzyz:
Looks fine but please add test case for TemplateRestriction and test there all three possible restrictions.

neeels:
Thanks for your review!
I added Units/parser-ttcn.r/ttcn-template-restriction.d/
Hope that's all correct.

masatake:
See  #3466.
Could you rebase on the latest master?

neeels:
done

masatake (approved):

masatake:
@neeels thank you for the contribution.
@akrzyz thank you for the review.

Issue #3457, opened 2022-08-11
SystemVerilog: The last coverpoint definition of covergroup caused Error Scope conditionally


roccomao:
When **last coverpoint definition** of covergroup cover a variable for a hierarchical index (for example from interface), in addition, it has bins definition with `{...}`, this will cause error scope of subsequent tags parser. For details, see what I've commented on in the follow code (`recur.sv`).

```systemverilog
class A;
  covergroup A0;
    coverpoint intf.x {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    coverpoint intf.y {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    // empty bins definition, end of ';', everything is OK.
    coverpoint intf.z;
  endgroup

  covergroup A1;
    coverpoint intf.a;
    coverpoint intf.b;
    // bins definition with '{}', but not a variable by hierarchical index, OK.
    coverpoint no_sro {
      bins foo = { 0 };
      bins bar = { 1 };
    }
  endgroup

  covergroup A2;
    coverpoint intf.m {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    coverpoint intf.n {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    // exceptional, index by 'this', OK.
    coverpoint this.t {
      bins foo = { 0 };
      bins bar = { 1 };
    }
  endgroup

  covergroup A3;
    coverpoint intf.r {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    coverpoint intf.s {
      bins foo = { 0 };
      bins bar = { 1 };
    }
    // contain multiple hierarchical variable, OK.
    cross intf.u, intf.v {
      bins foo = { 0 };
      bins bar = { 1 };
    }
  endgroup

  covergroup A4;
    coverpoint intf.p;
    // last definition, hierarchical variable, bins definition with '{}'
    // meet at the same time, subsequent parsers will all go WRONG.
    coverpoint intf.q {
      bins foo = { 0 };
      bins bar = { 1 };
    }
  endgroup

  // Error Scope >> covergroup:A.A4
  covergroup A5;
    coverpoint foo;
    coverpoint bar;
  endgroup
endclass

// Error Scope >> covergroup:A.A4
class B;
  int one, two;
  function new();
  endfunction
endclass

// Error Scope >> covergroup:A.A4
typedef enum { FOO, BAR } E;
```

Outputs Error Hierarchy as follow:

[...]

PR #3460, opened 2022-08-13
Verilog: introduce VERBOSE macro for internal errors

hirooih:
I added a debug macro to debug #3457.
It was quite useful and I found another bug for which I will send a PR.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -42,6 +42,18 @@
 #define IDX_SYSTEMVERILOG   0
 #define IDX_VERILOG         1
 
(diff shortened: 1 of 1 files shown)


hirooih:
There is one coverage failure, but It should be fine. Because this is a fix for cases that should never have happened.

masatake:
Why do we need so many commits for introducing the VERBOSE macro?
As far as reading the commit logs, you started the project introducing the VERBOSE macro two years ago. We need more descriptions for such a large project.

masatake (approved):

PR #3461, opened 2022-08-13
Verilog comment in multi line macro

hirooih:
This is the fix for a bug found by the VERBOSE macro (#3460).
I also made a refactoring.


masatake:
Should I review all 17 commits in this pull request?

masatake (approved):

Issue #3462, opened 2022-08-13
SystemVerilog: "assertions" parsing error when encountered "::"


roccomao:
The name of the parser: **Verilog/SystemVerilog**

In many cases, when defining an assertion (assert, assume, cover, restrict), there is no explicit naming label. If there are other **uncovered or exposed** (not parsed by other tags) colons before it, it will be detected as an assertion. For example:

```systemverilog
import uvm_pkg::*;
// something
assert (foo) else do something;
// something
someclass::method();
// something
assert (bar) else do something;
```

Output tags as follow:

```
uvm_pkg	input.sv	/^assert (foo) else do something;$/;"	A
someclass	input.sv	/^assert (bar) else do something;$/;"	A
```

This is not the expected result. Expected should be empty.

hirooih:
I will take this.

hirooih:
The fix #3462 was merged.
I close this issue.

@roccomao,
Thank you for your report.  It helped me a lot.

PR #3464, opened 2022-08-16
Verilog: fix bug caused by double colon ("::")

hirooih:
This is the fix for #3462.

There were two bugs.
Some refactoring fixes are also included.

### DIFF
--- a/Units/parser-verilog.r/systemverilog-github3462.d/args.ctags
+++ b/Units/parser-verilog.r/systemverilog-github3462.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--kinds-systemverilog=+Q
--- a/Units/parser-verilog.r/systemverilog-github3462.d/expected.tags
+++ b/Units/parser-verilog.r/systemverilog-github3462.d/expected.tags
@@ -1,0 +1,3 @@
+foo	input.sv	/^package foo;$/;"	K
+import_func	input.sv	/^  import "DPI-C" context function int import_func (string str);$/;"	Q	package:foo
+str	input.sv	/^  import "DPI-C" context function int import_func (string str);$/;"	p	prototype:foo.import_func
--- a/Units/parser-verilog.r/systemverilog-github3462.d/input.sv
+++ b/Units/parser-verilog.r/systemverilog-github3462.d/input.sv
@@ -1,0 +1,14 @@
+// SystemVerilog: "assertions" parsing error when encountered "::" #3462
+// https://github.com/universal-ctags/ctags/issues/3462
+
+import uvm_pkg::*;
+// something
+assert (foo) else do something;
+// something
+someclass::method();
+// something
+assert (bar) else do something;
+
+package foo;
+  import "DPI-C" context function int import_func (string str);
+endpackage
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1399,6 +1399,25 @@ static int processAssertion (tokenInfo *const token, int c)
 	return c;
 }
 
+// data_declaration ::=
+//   ...
+//   import < package_identifier :: identifier | package_identifier :: * > ; 
+// dpi_import_export ::=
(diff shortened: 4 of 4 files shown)


masatake:
Do we have to merge whole this pull request to pass the test case introduced in 15fddff?

If the answer is yes, we should move the 15fddff to the end of this pull request.
The 15fddff is at the begging of this pull request.

A test case should be merged after merging the code for fixing the bug reproduced by the test case. The exception is .b test case.

hirooih:
> Do we have to merge whole this pull request to pass the test case introduced in 15fddff?

To be precise, my answer is "no".  I'm following the rule, [Title of commit log and pull request](https://docs.ctags.io/en/latest/contributions.html#toc-entry-22).
We don't have to merge the last 3 commits prefixed by "refactor" or "style" **to pass the test case**.

> If the answer is yes, we should move the https://github.com/universal-ctags/ctags/commit/15fddffaa70a660dce78c92fc484b9bf56296795 to the end of this pull request.

I should have commented on this.

I'd rather suggest making this a new recommendation.
It has clear advantage. This will make it easier to reproduce the bug.

```sh
$ git checkout 15fddff
$ make units CATEGORIES=parser-verilog
... fails ...
$ git checkout verilog-double-colon
```
Missing a test case was one of bugs. Committing the fix first is better than committing it later.

How do you think?
If you don't like it, I am happy to rebase my PR.

masatake:
>> Do we have to merge whole this pull request to pass the test case introduced in https://github.com/universal-ctags/ctags/commit/15fddffaa70a660dce78c92fc484b9bf56296795?

> To be precise, my answer is "no". I'm following the rule, [Title of commit log and pull request](https://docs.ctags.io/en/latest/contributions.html#toc-entry-22).
We don't have to merge the last 3 commits prefixed by "refactor" or "style" to pass the test case.

Please, ignore my question. Including "noise" commits to a pull request is o.k. as far as they are semantically separated well.

What I would like to comment on is the order of commits.

A. fix -> test case for the fix
B. a test case -> fix for the test case
C. put a fix and its test case in a commit 

B is understandable. It may be one of the best practices in test-driven development.

However, I chose A (or C). for this project.
This project has no "release-engineering".
We don't have to take our time for it.

We have no explicit "release". 
However, weekly git-tagging is a kind of release.
None of us cares about the timing of the git-tagging.

So we cannot assume which commit is popular among users.
We cannot assume when a user does "git clone https://github.com/universal-ctags/ctags.git".

A curious user may run "make check".
We don't make a release. Instead, I would like to make "make check" successful for ANY commits. Let's call this policy "make-make-check-successful".

B. breaks the "make-make-check-successful".
This is the reason I like A (or C).

For supporting B style, we have .b test case mechanism.

B' a test case with .b -> fix for it with renaming the .b to .d.

Choose one of A, B', or C.

hirooih:
> If you don't like it, I am happy to rebase my PR.

I've pushed by A.

> We cannot assume when a user does "git clone https://github.com/universal-ctags/ctags.git".

B has no issues about this.
All commits in a pull-request is merged atomically on upstream repository.  A user cannot clone a pull-request partially.

***

The fails on MacOS is due to missing Python 3.9.

Thank you.

masatake:
See  #3466.
Could you rebase the latest master?

hirooih:
Oh, sorry.
Done.

masatake (approved):

masatake:
Thank you for updating the commits. All look good for me.

>  All commits in a pull-request is merged atomically on upstream repository. A user cannot clone a pull-request partially.

How about 'git checkout'?

You wrote:
```
$ git checkout 15fddff
$ make units CATEGORIES=parser-verilog
... fails ...
$ git checkout verilog-double-colon
```

hirooih:
Thank you for your review and apporoval.

> How about 'git checkout'?

If you checkout by branch name, obviously it should be no problem.

> $ git checkout 15fddff

To get the SHA-1 ID, 15fddff, you have to check the output of `git log` and choose the ID from the log.
You know what you are doing.  It should be OK.

PR #3465, opened 2022-08-18
javascript: split ParseStatement into multiple functions

jafl:
also improve code re-use

### DIFF
--- a/Units/review-needed.r/simple.js.t/input.js
+++ b/Units/review-needed.r/simple.js.t/input.js
@@ -3,7 +3,6 @@
  * ctags -f - simple.js
  *
  * functions
- *   extra.validFunctionFour
  *   getHalfOf
  *   getHalfOf.calculate
  *   testlib.core.validFunctionSix
@@ -39,8 +38,6 @@ function validFunctionThree(a,b) {
 }
 
 var my_global_var1 = 33;
-
-function extra.validFunctionFour(a,b) {}
 
 //pseudo-module setup
 testlib = {}
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -321,7 +321,7 @@ static const keywordTable JsKeywordTable [] = {
 /* Recursive functions */
 static void readTokenFull (tokenInfo *const token, bool include_newlines, vString *const repr);
 static void skipArgumentList (tokenInfo *const token, bool include_newlines, vString *const repr);
-static void parseFunction (tokenInfo *const token);
+static bool parseFunction (tokenInfo *const token, tokenInfo *const name, const bool is_inside_class);
 static bool parseBlock (tokenInfo *const token, int parentScope);
 static bool parseMethods (tokenInfo *const token, int classIndex, const bool is_es6_class);
 static bool parseLine (tokenInfo *const token, bool is_inside_class);
@@ -1404,7 +1404,7 @@ static void skipArgumentList (tokenInfo *const token, bool include_newlines, vSt
 					parseMethodsInAnonymousObject (token);
 			}
 			else if (isKeyword (token, KEYWORD_function))
-				parseFunction (token);
+				parseFunction (token, NULL, false);
 
 			prev_token_type = token->type;
 		}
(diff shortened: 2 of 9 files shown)


jafl:
@masatake I can't figure out why the MacOS tests are failing.  Homebrew claims to have python3.9 when I use `brew search`

masatake:
I'm inspecting the issue now.

masatake:
See  #3466.
Could you rebase on the latest master?

jafl:
@masatake Only 2 checks?

parsers/jscript.c
```diff
@@ -1360,7 +1360,7 @@ static int parseMethodsInAnonymousObject (tokenInfo *const token)
 
 	tokenInfo *const anon_object = newToken ();
 	copyToken (anon_object, token, true);
-	anonGenerate (anon_object->string, "anonymousObject", JSTAG_VARIABLE);
+	anonGenerate (anon_object->string, "anonymousObject_", JSTAG_VARIABLE);
```
masatake, line 1363: Do we need this in this commit?

This commit is marked as "refactor". This change is not suitable for such a commit.
I expect if a commit is marked "refactor", most of all test cases are passed successfully without changing the test cases. (An exception is adding a new test case.)

parsers/jscript.c
```diff
@@ -1755,47 +1756,58 @@ static void parseFunction (tokenInfo *const token)
 		is_generator = true;
 		readToken (name);
 	}
+
 	if (isType (name, TOKEN_OPEN_PAREN))
 	{
 		/* anonymous function */
 		copyToken (token, name, false);
-		anonGenerate (name->string, "anonymousFunction", JSTAG_FUNCTION);
+		anonGenerate (name->string, "anonymousFunction_", JSTAG_FUNCTION);
```
masatake, line 1764: Do we need this in this commit?

If you think we need this one, let's make a commit for introducing "_".
You can prepend such a commit to this pull request.
In such a case, I would like you to write the background or reason why "_" is needed in its commit log. I don't say introducing "_" breaks compatibilities. However, I don't say introducing "_" is a small change. Unlike "refactor", it is user-visible change. Therefore, I would like you to explain this to users.

Units/review-needed.r/simple.js.t/input.js
```diff
@@ -40,8 +39,6 @@ function validFunctionThree(a,b) {
 
 var my_global_var1 = 33;
 
-function extra.validFunctionFour(a,b) {}
```
masatake, line 43: Why do we have to remove this?

A. the limitation of this refactor. we will support this syntax in the future.
For A, you can move this fragment to ".b" test case.

B. the parser should not support this syntax for some reason.
For B, could you write the reason or background to the commit log?
This is a "refactor" commit. For such a commit, we expect all test cases for the parser should be passed without modification.

parsers/jscript.c
```diff
@@ -1755,47 +1756,58 @@ static void parseFunction (tokenInfo *const token)
 		is_generator = true;
 		readToken (name);
 	}
+
```
masatake, line 1759: Minor comment. This addition is not a must.

parsers/jscript.c
```diff
@@ -1755,47 +1756,58 @@ static void parseFunction (tokenInfo *const token)
...
-	findCmdTerm (token, false, false);
+	if ( lhs_name == NULL )
+		findCmdTerm (token, false, false);
 
  cleanUp:
 	vStringDelete (signature);
 	deleteToken (name);
 
 	TRACE_LEAVE();
+
```
masatake, line 1809: Minor comment. This addition is not a must.

parsers/jscript.c
```diff
@@ -2255,6 +2267,7 @@ static bool parseES6Class (tokenInfo *const token, const tokenInfo *targetName)
 	deleteToken (className);
 
 	TRACE_LEAVE();
+
```
masatake, line 2270: Minor comment. This addition is not a must.

masatake (changes requested):
Great work.

Do you know git add -p?
If you want to insert empty lines to adjust the code to your favorite, you can make a separate commit with 'git add -p'.

Units/review-needed.r/simple.js.t/input.js
jafl: This is invalid JavaScript.  I will create a separate commit for it.

masatake (approved):

PR #3468, opened 2022-08-21
javascript: create man page

jafl:
I'll add more when I think of something.


masatake:
The way to add a man page is a bit complicated.
See 518bd65952ed954a333dfd03ca942f607fabfff5 as an example.

* add an entry to docs/man-pages.rst,
* add an entry to man/GNUmakefile.am, 
* write man/ctags-lang-<LANG>.7.rst.in, and
* run `make -C man` which generates docs/man/ctags-lang-javascript.7.rst


See https://github.com/universal-ctags/ctags/blob/master/docs/README.md for more details.

masatake (approved):

masatake:
Now the page is visible at https://docs.ctags.io/en/latest/man/ctags-lang-javascript.7.html !
Thank you!

Issue #3469, opened 2022-08-25
Verilog: create tag of instance's reference name


my2817:
I use ctags in my work, with emacs+citre.  I write some HACK code to meet my needs.  And put the info at here if some others are searching for same solutions.

1. add a new kind K_REFERENCE to ctreate tag of instance's reference name. So that we can get a list from the tags file that how many times or where a module been referenced.
2. (HACK) To give info of a instance's reference name. Update  instance's scope field. Append instance's reference name to scope field. 

https://github.com/universal-ctags/ctags/compare/master...my2817:ctags:master

*****

The name of the parser: verilog

The content of input file:

```verilog
module test (/*AUTOARG*/);
   input i_a, i_b;
   output o_c;

   ref1 int1 ();
   ref1 int2 ();
   ref3 # (.A (aaa),
           .B (bbb))
   int3 ();

endmodule // test

```

The tags output you are not satisfied with:

```

```

The tags output you expect:

```

i_a     ./v_test/test.v /^   input i_a, i_b;$/;"        kind:port       line:21 language:Verilog        scope:module:test       roles:def
i_b     ./v_test/test.v /^   input i_a, i_b;$/;"        kind:port       line:21 language:Verilog        scope:module:test       roles:def
int1    ./v_test/test.v /^   ref1 int1 ();$/;"  kind:instance   line:24 language:Verilog        scope:module:test:ref1  roles:def
int2    ./v_test/test.v /^   ref1 int2 ();$/;"  kind:instance   line:25 language:Verilog        scope:module:test:ref1  roles:def
int3    ./v_test/test.v /^   int3 ();$/;"       kind:instance   line:28 language:Verilog        scope:module:test:ref3  roles:def
o_c     ./v_test/test.v /^   output o_c;$/;"    kind:port       line:22 language:Verilog        scope:module:test       roles:def
ref1    ./v_test/test.v /^   ref1 int1 ();$/;"  kind:reference  line:24 language:Verilog        scope:module:test       roles:def
ref1    ./v_test/test.v /^   ref1 int2 ();$/;"  kind:reference  line:25 language:Verilog        scope:module:test       roles:def
ref3    ./v_test/test.v /^   ref3 # (.A (aaa),$/;"      kind:reference  line:26 language:Verilog        scope:module:test       roles:def
test    ./v_test/test.v /^module test (\/*AUTOARG*\/);$/;"      kind:module     line:20 language:Verilog        roles:def
```

masatake:
[...]

PR #3470, opened 2022-08-25
Verilog: fill typeref fields of instances

masatake:
This one is conceptually based on
https://github.com/universal-ctags/ctags/compare/master...my2817:ctags:master
reported and written by @my2817 at #3469.

The test case is also taken from #3469 submitted by @my2817.

This commit does two things:

* fill typeref fields of instances with module names, and
* extract the module names used for defining instances as
  defInstance role of module kind.

TODO: Update ctags-lang-verilog(7).

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -108,6 +108,10 @@ typedef enum {
 	K_NETTYPE,
 } verilogKind;
 
+typedef enum {
+	R_MODULE_DEFINSTANCE,
+} verilogModuleRole;
+
 typedef struct {
 	const char *keyword;
 	verilogKind kind;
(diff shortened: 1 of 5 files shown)


parsers/verilog.c
```diff
@@ -1409,7 +1434,7 @@ static int processAssertion (tokenInfo *const token, int c)
 
 // data_declaration ::=
 //   ...
-//   import < package_identifier :: identifier | package_identifier :: * > ; 
+//   import < package_identifier :: identifier | package_identifier :: * > ;
```
masatake, line 1437: This change is not needed.

masatake:
This pull request proposed two things. To simplify the discussion and review process, I made a new pull request that focused only on one of two. See #3471.

I will rework this pull request after #3471 is merged.

PR #3472, opened 2022-08-28
docs(man): remove unexpected newline from the rendered verilog man page

masatake:


hirooih (approved):
I did not know about this rule.

https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists

> A sublist is a new list so requires a blank line above and below.

I'm afraid that I made more same mistakes...

PR #3476, opened 2022-09-03
PowerShell: add tag generation from class

kumarstack55:
This pull request adds functionality to the existing parser.

Previously, the PowerShell parser generated tags from functions and variables.

By merging this pull request, it will be able to generate tags from classes in addition to those.

Please note that I do not have a lot of experience contributing to the open source community. Also, this is my first time attempting to contribute to u-ctags.

Please let me know if there are any details I should address.

### DIFF
--- a/parsers/powershell.c
+++ b/parsers/powershell.c
@@ -39,12 +39,14 @@ static const char *const accessTypes[] = {
 typedef enum {
 	K_FUNCTION,
 	K_VARIABLE,
+	K_CLASS,
 	COUNT_KIND
 } powerShellKind;
 
 static kindDefinition PowerShellKinds[COUNT_KIND] = {
 	{ true, 'f', "function",	"functions" },
-	{ true, 'v', "variable",	"variables" }
+	{ true, 'v', "variable",	"variables" },
+	{ true, 'c', "class",		"classes" },
 };
 
 
@@ -55,7 +57,8 @@ typedef enum eTokenType {
 	TOKEN_SEMICOLON,
 	TOKEN_COLON,
 	TOKEN_COMMA,
-	TOKEN_KEYWORD,
+	TOKEN_KEYWORD_FUNCTION,
+	TOKEN_KEYWORD_CLASS,
 	TOKEN_OPEN_PAREN,
 	TOKEN_OPERATOR,
 	TOKEN_IDENTIFIER,
@@ -143,6 +146,18 @@ static void makeFunctionTag (const tokenInfo *const token, const vString *const 
 	}
 }
 
+static void makeClassTag (const tokenInfo *const token)
+{
+	if (PowerShellKinds[K_CLASS].enabled)
+	{
+		tagEntryInfo e;
+
+		initPowerShellEntry (&e, token, K_CLASS, NULL);
+
+		makeTagEntry (&e);
+	}
+}
+
 static tokenInfo *newToken (void)
 {
 	tokenInfo *const token = xMalloc (1, tokenInfo);
@@ -218,6 +233,11 @@ static bool isTokenFunction (vString *const name)
 {
 	return (strcasecmp (vStringValue (name), "function") == 0 ||
 			strcasecmp (vStringValue (name), "filter") == 0);
+}
+
+static bool isTokenClass (vString *const name)
+{
+	return strcasecmp (vStringValue (name), "class") == 0;
 }
 
 static bool isSpace (int c)
@@ -358,7 +378,9 @@ getNextChar:
 			{
 				parseIdentifier (token->string, c);
 				if (isTokenFunction (token->string))
-					token->type = TOKEN_KEYWORD;
+					token->type = TOKEN_KEYWORD_FUNCTION;
+				else if (isTokenClass (token->string))
+					token->type = TOKEN_KEYWORD_CLASS;
 				else
 					token->type = TOKEN_IDENTIFIER;
 			}
@@ -456,7 +478,6 @@ static bool parseFunction (tokenInfo *const token)
 				case TOKEN_STRING:			vStringCatS (arglist, "'...'");	break;
 
 				case TOKEN_IDENTIFIER:
-				case TOKEN_KEYWORD:
 				case TOKEN_VARIABLE:
 				{
 					switch (vStringLast (arglist))
@@ -506,6 +527,43 @@ static bool parseFunction (tokenInfo *const token)
 	return readNext;
 }
 
+/* parse a class
+ *
+ * 	class MyClass {}
+ */
+static bool parseClass (tokenInfo *const token)
+{
+	bool readNext = true;
+	tokenInfo *nameFree = NULL;
+	const char *access;
+
+	readToken (token);
+
+	if (token->type != TOKEN_IDENTIFIER)
+		return false;
+
+	nameFree = newToken ();
+	copyToken (nameFree, token, true);
+	readToken (token);
+
+	makeClassTag (nameFree);
+
+	while (token->type != TOKEN_OPEN_CURLY && token->type != TOKEN_EOF)
+	{
+		readToken (token);
+	}
+
+	if (token->type == TOKEN_OPEN_CURLY)
+		enterScope (token, nameFree->string, K_CLASS);
+	else
+		readNext = false;
+
+	if (nameFree)
+		deleteToken (nameFree);
+
+	return readNext;
+}
+
 /* parses declarations of the form
  * 	$var = VALUE
  */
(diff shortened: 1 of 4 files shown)


parsers/powershell.c
```diff
@@ -39,12 +39,14 @@ static const char *const accessTypes[] = {
...
 	K_VARIABLE,
+	K_CLASS,
 	COUNT_KIND
 } powerShellKind;
 
 static kindDefinition PowerShellKinds[COUNT_KIND] = {
 	{ true, 'f', "function",	"functions" },
-	{ true, 'v', "variable",	"variables" }
+	{ true, 'v', "variable",	"variables" },
+	{ true, 'c', "class",		"classes" },
```
masatake, line 49: You put `,` at the end of this entry. Good!

parsers/powershell.c
```diff
@@ -55,7 +57,9 @@ typedef enum eTokenType {
 	TOKEN_SEMICOLON,
 	TOKEN_COLON,
 	TOKEN_COMMA,
-	TOKEN_KEYWORD,
+	TOKEN_KEYWORD_FUNCTION,
+	TOKEN_KEYWORD_CLASS,
+	TOKEN_KEYWORD_OTHER,
```
masatake, line 62: Though TOKEN_KEYWORD_OTHER is defined, it is not used.
Do we need this?

parsers/powershell.c
```diff
@@ -456,7 +479,7 @@ static bool parseFunction (tokenInfo *const token)
 				case TOKEN_STRING:			vStringCatS (arglist, "'...'");	break;
 
 				case TOKEN_IDENTIFIER:
-				case TOKEN_KEYWORD:
+				case TOKEN_KEYWORD_OTHER:
```
masatake, line 482: Where is TOKEN_KEYWORD_OTHER assigned to `->type`.

parsers/powershell.c
```diff
@@ -506,6 +529,38 @@ static bool parseFunction (tokenInfo *const token)
...
+	readToken (token);
+
+	makeClassTag (nameFree);
+
+	if (token->type == TOKEN_OPEN_CURLY)
+		enterScope (token, nameFree->string, K_CLASS);
+	else
+		readNext = false;
+
+	if (nameFree)
```
masatake, line 558: Can nameFree be NULL?
If yes, nameFree->string at line 554 is unsafe.

parsers/powershell.c
```diff
@@ -506,6 +529,38 @@ static bool parseFunction (tokenInfo *const token)
...
+	if (token->type != TOKEN_IDENTIFIER)
+		return false;
+
+	nameFree = newToken ();
+	copyToken (nameFree, token, true);
+	readToken (token);
+
+	makeClassTag (nameFree);
+
+	if (token->type == TOKEN_OPEN_CURLY)
```
masatake, line 553: This may not work well for the input like:
```
class MyException : Exception {
```

If a class definition has {...} block ALWAYS, you can do something like:
```
while (token->type != TOKEN_OPEN_CURLY && token->type != TOKEN_EOF)
{
   readToken (token);
}

if (token->type != TOKEN_OPEN_CURLY)
  enterScope (token, nameFree->string, K_CLASS);
...
```

parsers/powershell.c
masatake: I found the same code pattern in parseFunction().
If you are intrested in fixing parseFunction(), you can do it in this pull request but please make a separate commit.

masatake:
> Please note that I do not have a lot of experience contributing to the open source community. Also, this is my first time attempting to contribute to u-ctags.

I'm proud of u-ctags receiving a contribution from a person like you.

I gave some comments. Feel free to do `git commit --amend` and `git push --force`.

If you are interested in improving the PowerShell parser, I have more comments.

* filling inherits: field
  ```
  MyException	input.ps1	/^class MyException : Exception {$/;"	c	inherits:Exception
  ```

* using keyword API
  You extended the types of token: KEYWORD_FUNCTION, and KEYWORD_CLASS.
  However, keeping the type KEYWORD but adding `keyword` field to tokenInfo.

parsers/powershell.c
kumarstack55: There was an incorrect implementation. Thank you for pointing this out.
Please let me make an attempt to see if I can implement it in this PR.

Since the introduction of classes in PowerShell 5.0, class definitions always seem to include `{}`. Note that in PowerShell 5.1, the earliest currently supported version, the class syntax is:

```powershell
class <class-name> [: [<base-class>][,<interface-list>]] {
    [[<attribute>] [hidden] [static] <property-definition> ...]
    [<class-name>([<constructor-argument-list>])
      {<constructor-statement-list>} ...]
    [[<attribute>] [hidden] [static] <method-definition> ...]
}
```

* PowerShell 5.0 (not supported)
    * https://docs.microsoft.com/en-us/previous-versions/powershell/module/microsoft.powershell.core/about/about_classes?view=powershell-5.0
* PowerShell 5.1
    * https://docs.microsoft.com/ja-jp/powershell/module/microsoft.powershell.core/about/about_classes?view=powershell-5.1
* PowerShell 7.2
    * https://docs.microsoft.com/ja-jp/powershell/module/microsoft.powershell.core/about/about_classes?view=powershell-7.2

kumarstack55:
> If you are interested in improving the PowerShell parser, I have more comments.
>
> * filling inherits: field
>
> ```
> MyException	input.ps1	/^class MyException : Exception {$/;"	c	inherits:Exception
>```

This improvement would be better.
However, I do not intend to make this improvement in this PR.

This is because I cannot think of a good way to do it right now, because I do not know how it should work with multiple inheritance and interface implementations, and how it should be implemented.

kumarstack55:
> * using keyword API
> 
> You extended the types of token: KEYWORD_FUNCTION, and KEYWORD_CLASS.
> However, keeping the type KEYWORD but adding keyword field to tokenInfo.

Thank you for your review.
However, I feel I still do not understand your comment.
Could you please comment so that I may understand.

My current PR is as follows:
* (a) split TOKEN_KEYWORD into TOKEN_KEYWORD_FUNCTION, TOKEN_KEYWORD_CLASS, and TOKEN_KEYWORD_OTHER. The `token->type` attribute determines whether the keyword is a function or a class.
* (b) do not change tokenInfo.

Is the proposed implementation of @masatake as follows. :
* (a) do not split TOKEN_KEYWORD.
* (b) add an attribute like `token->keywordType` to tokenInfo. When `token->type` is TOKEN_KEYWORD, `token->keywordType` attribute determines if the keyword is a function or a class.

@masatake : Is this understanding correct?

parsers/powershell.c
masatake: >  Please let me make an attempt to see if I can implement it in this PR.

Yes, please try.

masatake:
Your understanding is correct.

My proposal about keyword may make this pull reuquest bigger.

So about keyword, I propose the following change in this pr:

* Remove TOKEN_KEYWORD_OTHER IF there is no trouble.
(Introducing TOKEN_KEYWORD_FUNCTION and TOKEN_KEYWORD_CLASS is acceptable.)

parsers/powershell.c
kumarstack55: In [another comment](https://github.com/universal-ctags/ctags/pull/3476#issuecomment-1236221400), masatake suggested leaving `TOKEN_KEYWORD` as it is and adding an attribute like `token->keywordType`.

That proposal shall be addressed in a separate PR in accordance with [this comment](https://github.com/universal-ctags/ctags/pull/3476#issuecomment-1236241369).

In this PR, remove the unused TOKEN_KEYWORD_OTHER.

parsers/powershell.c
kumarstack55: In [another comment](https://github.com/universal-ctags/ctags/pull/3476#issuecomment-1236221400), masatake suggested leaving `TOKEN_KEYWORD` as it is and adding an attribute like `token->keywordType`.

That proposal shall be addressed in a separate PR in accordance with [this comment](https://github.com/universal-ctags/ctags/pull/3476#issuecomment-1236241369).

In this PR, remove the unused TOKEN_KEYWORD_OTHER.

parsers/powershell.c
kumarstack55: I assume that nameFree has a value and that nameFree is not null.
I would remove the `if` line.

parsers/powershell.c
kumarstack55: Contrary to expectations, tags were created for variables in properties and methods within classes.

To keep this PR small and simple, no tags will be created for variables within classes in this PR.

Instead, TODO was left as a comment.

https://github.com/universal-ctags/ctags/pull/3476/commits/d45a7a453df569a5bfb5925182bd2a81b11e1926#diff-42a33750a868afb3e3d32b5f60a7bfe62fb32ee5697058cf74d9812bb5021c31R582

kumarstack55:
Thank you for your comments.

I have finished making changes to the code.
@masatake : Could you please confirm?

If everything is ok, I will squash the commits.
The squash should be the following two commits:

* PowerShell: add tag generation from class
* PowerShell: remove unnecessary if statements

masatake:
> @masatake : Could you please confirm?

Fine. Please squash the commits into the two.

kumarstack55:
I squashed it.
I would appreciate it if you could merge.

masatake (approved):

masatake:
Thank you!

PR #3480, opened 2022-09-10
PowerShell: keep the KEYWORD type

### LINKED ISSUE #3479: PowerShell: Keep the KEYWORD type and add a field for the keyword to tokenInfo
## Goal of this Issue

Based on the discussion in https://github.com/universal-ctags/ctags/pull/3476 , make the following changes already made.

* Do not extend the token type.
* Add field to tokenInfo.

## Current Implementation

* The tokenInfo type has been extended with https://github.com/universal-ctags/ctags/pull/3476 .

## Reference

https://github.com/universal-ctags/ctags/pull/3476#issuecomment-1236164458

> If you are interested in improving the PowerShell parser, I have more comments.

...

> using keyword API
> You extended the types of token: KEYWORD_FUNCTION, and KEYWORD_CLASS.
> However, keeping the type KEYWORD but adding keyword field to tokenInfo.

kumarstack55:
Close #3479


masatake:
Though we have only two keywords in the parser, using functions and a constant defined in keyword.h is better.

parsers/php.c illustrates how to use the functions and the constant.
Let me explain.

0. include keyword.h.
1. define enum for keywords like:
   ```
   enum {
	KEYWORD_abstract,
	KEYWORD_and,
	KEYWORD_as,
   ...    
   ```

2. define keywords that mapping keywords to the enumerators:
   ```
   static const keywordTable PhpKeywordTable[] = {
	/* keyword			keyword ID */
	{ "abstract",		KEYWORD_abstract		},
	{ "and",			KEYWORD_and				},
	{ "as",				KEYWORD_as				},
    ```
3. fill keywordTable field of your parser:
   ```
   extern parserDefinition* PhpParser (void)
   {
	...
	def->keywordTable = PhpKeywordTable;
	def->keywordCount = ARRAY_SIZE (PhpKeywordTable);
   ...
 
   ```
4. call lookupKeyword or lookupCaseKeyword at `default` clause in the case statement in your readToken() function:
   ```
   		default:
			if (! isIdentChar (c))
				token->type = TOKEN_UNDEFINED;
			else
			{
				parseIdentifier (token->string, c);
				if (MayBeKeyword)
					token->keyword = lookupCaseKeyword (vStringValue (token->string), getInputLanguage ());
				else
					token->keyword = KEYWORD_NONE;

				if (token->keyword == KEYWORD_NONE)
					token->type = TOKEN_IDENTIFIER;
				else
					token->type = TOKEN_KEYWORD;
			}
			break;
   ```


Hmmm... the last step is a bit too complicated for the current implementation of the PowerShell parser; MayBeKeyword is not needed:
```
		default:
			if (! isIdentChar (c))
				token->type = TOKEN_UNDEFINED;
			else
			{
				parseIdentifier (token->string, c);
				token->keyword = lookupCaseKeyword (vStringValue (token->string),
													getInputLanguage ());
				if (token->type == KEYWORD_NONE)
					token->type = TOKEN_KEYWORD;
				else
					token->type = TOKEN_IDENTIFIER;
			}
			break;
```

kumarstack55:
Thank you for your detailed comments. Thanks to you, I was able to add a commit that fixes the use of the keyword api.

However, when I ran the git rebase, something went wrong and the MR got some unexpected commits mixed in.

I don't know how to remove the unwanted commits from MR in this situation. I am thinking of rebuilding MR at a later date.

masatake:
Consider `origin` is your repository at github.com, and `upstream` is `https://github.com/universal-ctags/ctags` repository. Such naming can be done in your .git/config.

```
cd ctags
git checkout devel
: making backup.
git checkout -b devel-new
git checkout master
git pull upstream master:master
git checkout devel-new
git rebase master
: pushing devel-new as devel
git push origin devel-new:devel
```

These may be suboptimal procedures.
However, these procedures work fine for years in my ctags development.

masatake:
After thinking again, cherry-picking may be more understandable.

```
cd ctags
git checkout master
git pull upstream master:master
: new branch forked from master
git checkout -b devel-new
git cherry-pick 11791e7
: pushing devel-new as devel
git push origin --force devel-new:devel
```

parsers/powershell.c
```diff
@@ -619,12 +624,20 @@ static void enterScope (tokenInfo *const parentToken,
...
 				break;
 
-			case TOKEN_KEYWORD_FUNCTION:
-				readNext = parseFunction (token);
-				break;
+			case TOKEN_KEYWORD:
+				switch (token->keyword)
+				{
+					case KEYWORD_function:
+					case KEYWORD_filter:
```
masatake, line 631: It seems that there is no difference between filter and function FOR CTAGS though they have their own purpose in PowerShell.

In that case, you can map "function" and "filter" to KEYWORD_function; and we can make KEYWORD_filter obsoleted. In the other words, let's ctags extract a filter object with function kind. Let's call this "unifying-function-and-filter approach".

If you think the difference between function and filter is important for ctags' users,
introducing filter kind is another approach.
Let's call "introducing-a-kind-for-filter approach".

"unifying-function-and-filter approach" is easy, and we can choose "introducing-a-kind-for-filter approach" later. So, I'll make more comments about "unifying-function-and-filter approach" in other places.

parsers/powershell.c
```diff
@@ -72,8 +71,24 @@ typedef enum eTokenType {
...
+	KEYWORD_filter,
+	KEYWORD_class,
+};
+
+/* We need an integer that is not an unsigned to allow KEYWORD_NONE. */
+typedef int keywordId;
+
+static const keywordTable PowerShellKeywordTable[] = {
+	{ "function",	KEYWORD_function },
+	{ "filter",		KEYWORD_filter },
```
masatake, line 85: If you choose "unifying-function-and-filter approach", this line should be:
```
+              { "filter",		KEYWORD_function },
```

parsers/powershell.c
```diff
@@ -72,8 +71,24 @@ typedef enum eTokenType {
 	TOKEN_VARIABLE
 } tokenType;
 
+enum {
+	KEYWORD_function,
+	KEYWORD_filter,
```
masatake, line 76: If you choose "unifying-function-and-filter approach", KEYWORD_filter is not needed.
About "unifying-function-and-filter approach", see below.

masatake:
It seems that KEYWORD_filter is not needed now.
KEYWORD_function can be reused for "filter" keyword.

masatake:
I studied PowerShell a bit.
It seems a good idea for me to introduce "filter" kind.

```
diff --git a/parsers/powershell.c b/parsers/powershell.c
index 08f6f8aec..b20c04ae4 100644
--- a/parsers/powershell.c
+++ b/parsers/powershell.c
@@ -40,6 +40,7 @@ typedef enum {
 	K_FUNCTION,
 	K_VARIABLE,
 	K_CLASS,
+	K_FILTER,
 	COUNT_KIND
 } powerShellKind;
 
@@ -47,6 +48,7 @@ static kindDefinition PowerShellKinds[COUNT_KIND] = {
 	{ true, 'f', "function",	"functions" },
 	{ true, 'v', "variable",	"variables" },
 	{ true, 'c', "class",		"classes" },
+	{ true, 'i', "filter",      "filter" },
 };
 
 
@@ -131,13 +133,13 @@ static void makeSimplePowerShellTag (const tokenInfo *const token, const powerSh
 }
 
 static void makeFunctionTag (const tokenInfo *const token, const vString *const arglist,
-							 const char *const access)
+							 const char *const access, int kind)
 {
-	if (PowerShellKinds[K_FUNCTION].enabled)
+	if (PowerShellKinds[kind].enabled)
 	{
 		tagEntryInfo e;
 
-		initPowerShellEntry (&e, token, K_FUNCTION, access);
+		initPowerShellEntry (&e, token, kind, access);
 
 		if (arglist)
 			e.extensionFields.signature = vStringValue (arglist);
@@ -506,14 +508,14 @@ static bool parseFunction (tokenInfo *const token)
 		}
 		while (token->type != TOKEN_EOF && depth > 0);
 
-		makeFunctionTag (nameFree, arglist, access);
+		makeFunctionTag (nameFree, arglist, access, K_FUNCTION);
 		vStringDelete (arglist);
 
 		readToken (token);
 	}
 	else if (token->type == TOKEN_OPEN_CURLY)
 	{	/* filters doesn't need to have an arglist */
-		makeFunctionTag (nameFree, NULL, access);
+		makeFunctionTag (nameFree, NULL, access, K_FUNCTION);
 	}
 
 	if (token->type == TOKEN_OPEN_CURLY)
```

This will be the first commit in this pull request.

The second commit will be based on 143746bb77cd5fd270f0c2f68d6a1ac367cf6978.
However, the following extra change may be needed:
```
@@ -433,6 +435,7 @@ static bool parseFunction (tokenInfo *const token)
 	bool readNext = true;
 	tokenInfo *nameFree = NULL;
 	const char *access;
+	int kind = token->keyword == KEYWORD_function? K_FUNCTION: K_FILTER;
 
 	readToken (token);
 
@@ -506,18 +509,18 @@ static bool parseFunction (tokenInfo *const token)
 		}
 		while (token->type != TOKEN_EOF && depth > 0);
 
-		makeFunctionTag (nameFree, arglist, access);
+		makeFunctionTag (nameFree, arglist, access, kind);
 		vStringDelete (arglist);
 
 		readToken (token);
 	}
 	else if (token->type == TOKEN_OPEN_CURLY)
 	{	/* filters doesn't need to have an arglist */
-		makeFunctionTag (nameFree, NULL, access, K_FUNCTION);
+		makeFunctionTag (nameFree, NULL, access, kind);
 	}
 
 	if (token->type == TOKEN_OPEN_CURLY)
-		enterScope (token, nameFree->string, K_FUNCTION);
+		enterScope (token, nameFree->string, kind);
 	else
 		readNext = false;
 

```

The value for `kind` local var is given in parseFunction. However, passing the value from its caller may be better.
That means:
```
		case TOKEN_KEYWORD:
				switch (token->keyword)
				{
					case KEYWORD_function:
						readNext = parseFunction (token, K_FUNCTION);
						break;
					case KEYWORD_filter:
						readNext = parseFunction (token, K_FILTER);
						break;
```

parsers/powershell.c
kumarstack55: The short answer is that I think the "introducing-a-kind-for-filter approach" should be chosen.

If a user writes function and filter separately, there is a reason. Therefore, function and filter should be treated differently. 

There are reasons why we should not unify function and filter.

In the Before You Start section of the Contribution Guide, it says that ctag users use ctag for `navigating the source code tree`.
https://docs.ctags.io/en/latest/contributions.html#before-you-start

At least in PowerShell v5.1, function and filter use the same namespace. Thus, if `function func1 {` and `filter func1 {` exist, the later definition is valid.

If there are multiple functions with the same name in the source code, some implementations of the `source code tree` may display strange results. However, if they are of the same kind, the user will easily notice it because func1 will be displayed near func1 in the code tree.

Similarly, the `source code tree` may look strange when there is a function and a filter with the same name in the source code. I would guess that many implementations will have the names appear farther apart, especially when the kind is different. Because of the distance, I would think that it would be difficult to notice that the names are identical.

I think the mechanism for noticing duplicate function and filter names is more the role of a separate linter than the role of ctags.

As a result, I think the "introducing-a-kind-for-filter approach" should be chosen.

kumarstack55: I agree with this comment.
https://github.com/universal-ctags/ctags/pull/3480#issuecomment-1246010526

masatake: So will you update the pull request that follows the "introducing-a-kind-for-filter approach" ?

parsers/powershell.c
kumarstack55: We chose the "introducing-a-kind-for-filter approach". For more information, see https://github.com/universal-ctags/ctags/pull/3480#discussion_r977902309.

parsers/powershell.c
kumarstack55: We chose the "introducing-a-kind-for-filter approach". For more information, see https://github.com/universal-ctags/ctags/pull/3480#discussion_r977902309.

parsers/powershell.c
kumarstack55: Thank you for your kind review and comments.
I have implemented it and git pushed it.
Please check it out.

Please let me know if anything is missing.

masatake (commented):
The change itself is fine.

About the header of the commit log, "PowerShell: introduce filer kind" is better than "PowerShell: use keyword api" is better because introducing a new kind has a larger impact on users than using the keyword API. Using the keyword API is a user-invisible change.

kumarstack55:
> The change itself is fine.
> 
> About the header of the commit log, "PowerShell: introduce filer kind" is better than "PowerShell: use keyword api" is better because introducing a new kind has a larger impact on users than using the keyword API. Using the keyword API is a user-invisible change.

Thank you for reviewing this.

I have modified the header of the commit log.
I have added a description of the keyword api and the discussion in the body of the commit log.
Please check it out.

masatake:
I'm sorry: s/filer/filter/g.

kumarstack55:
Sorry, I missed the typo.
I modified the commit message and pushed it again.

masatake:
Thank you!

PR #3481, opened 2022-09-12
main: extend the bucket array of hashTable dynamically

masatake:
There is a plan to store much more items in a hashtable (htable).

The current implementation, allocating a bucket array of a htable with the size specified when initializing the htable, doesn't fit the purpose because proper size is not predictable.

With this change, the htable itself extends the size of the array twice dynamically when the approximate filling rate of the bucket array of the htable is greater than 80%.

### DIFF
--- a/Tmain/optscript.d/arrayx.expected
+++ b/Tmain/optscript.d/arrayx.expected
@@ -1,0 +1,2 @@
+unmatchedmark
+null
--- a/Tmain/optscript.d/arrayx.ps
+++ b/Tmain/optscript.d/arrayx.ps
@@ -1,0 +1,8 @@
+clear
+{
+    { ] } stopped {
+        _errorname =
+        stop
+    } if
+} stopped pop
+_errorname =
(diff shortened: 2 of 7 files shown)


masatake:
For a name in optscript, a hashtable uses the memory address of the name as its hash value when storing the name to itself. The address is not predictable. So we cannot predict the hash bucket for the name in a hashtable. So we cannot predict how 'forall' alike operations traverse items in a hashtable. So we cannot predict the output from a pretty-printer printing a hashtable.

Understandable.

The question is why the test case has not reported an error till today.

masatake:
> The question is why the test case has not reported an error till today.

The code of optscript had a bug. When making a dictionary like << /a 1 /b 2 >>, optscript made a dictionary with size 1. In the test case, we stored all items in a dictionary with only one hash bucket. The original dictionary didn't do dynamic extending; the test was not for dictionaries but for single-linked lists. The bus was fixed in #3484.

PR #3489, opened 2022-09-25
Bats: Sh parser based new sub parser

masatake:
Reference:
*   https://bats-core.readthedocs.io/en/stable/writing-tests.html


masatake:
Some platforms report errors "brew is not found".
@leleliu008 do you have any idea about fixing these?

leleliu008:
> Some platforms report errors "brew is not found".
> @leleliu008 do you have any idea about fixing these?

homebrew has been removed from `$PATH` on Ubuntu runner images two days ago.

Reference: https://github.com/actions/runner-images/issues/6283

I will fix it.

masatake:
> if we still want to use brew command on Ubuntu runner images, we need to add this command to PATH manually.
>
> if we do not need latest version of softwares, we can use apt-get instead of brew.

Thank you for looking at this topic.

I don't have any strong opinion or preference.
So including reverting my change, I will accept any pull request about this area.

Issue #1577, opened 2017-10-12
require new features with {scope}


Fmajor:
for scope, now we have these flags
```
{scope=push}
{scope=ref}
{scope=pop}
{scope=clear}
{scope=set}
{placeholder}
```
In my custom syntax file, i need to parse the same pattern in different way with different context (or different stack status), so i come up with some new features with {scope}, like
```
{scope=replace}
    replace the top of stack (pop and push, not clear and push as '{scope=set}')
{scope=push(n)}
    push current match nth times
{scope=pop(n)}
    pop current match nth times
```
also, i'd like to use some boolean operation for the scope, like
```
{scope-top-kind=blabla}, {scope-top-kind!=blabla}
{scope-level=blabla}, {scope-level>blabla}, {scope-level<blabla}
    the --regex-X will run only when the expression are true
```
i'm wondering if someone else have the same requirements.

masatake:
Interesting idea but it needs consideration.


```
{scope=replace}
    replace the top of stack (pop and push, not clear and push as '{scope=set}')
```

Implementing this is not difficult. I wonder `{scope=pop}{scope=push}` does the
same as `{scope=replace}`

```
{scope=push(n)}
    push current match nth times
```
Ctags can push a language object which is tagged to the scope stack.
Pushing nth substring directly to the stack is not possible.
Consider scope entry has not only name but also kind.
kind is not specified in your notation.

BTW, matchTagPattern() in main/lregex.c is the function where scope stack is used.
Let's think about the following code of imaginary language:
```
var Parent::item;
```
I guess you may want to make the following tag:

```
item  ipnut.X /^var Parent::item;/;" scope:???:Parent
```

We have to fill ???.

I know how useful the feature you propose is.
However, tool fill the ???, I have to implement the way to handle reference tag in regex parser.
See https://github.com/universal-ctags/ctags/pull/1260

With the reference tag, you can capture "Parent" in the code as a reference tag.
A reference tag has a kind. So we can fill ??? with the kind.

I'm working on implementing the way to capture reference tags via regex parser very slowly.
However, it will take time.

Your proosal may be implemented on that.

Do you have any target language?
I would like to see.

#. an example input file (shorter is better)
#. expected tags output
#. IDEAL command line options to make the output from the input.

You can use any imaginary options in the last item.

BTW, matchTagPattern() is the function in which the scope stack is uesd.

Fmajor:
i am writing Vue, like this (i have delete all the function body..)
[...]

PR #3490, opened 2022-09-27
html: parse javascript broken by php

jafl:
The html parser needs to be able to use php as a guest parser.

### DIFF
--- a/Units/parser-php.r/run-guest-nested.d/args.ctags
+++ b/Units/parser-php.r/run-guest-nested.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=yes
+--extras=+rg
+--fields=+rln
+--language-force=html
--- a/Units/parser-php.r/run-guest-nested.d/input.php
+++ b/Units/parser-php.r/run-guest-nested.d/input.php
@@ -1,0 +1,39 @@
+<html><head><title>X</title>
+<?php // This test input is derived from #2256 submitted by @StephenWall.
+  function draw($x) {
+      echo "$x";
+  }
+?>
+    <meta http-equiv="content-type" content="text/html; charset=ISO-8859-1">
+    <title>My Home Page</title>
+    <link rel="stylesheet" href="css/general.css" type="text/css">
+    <style type="text/css"></stylE>
+  </head>
+  <body>
+    <h1>stuff</h1>
+    <a href="/blah">nowhere</a>
+    <?=draw('nowhere')?>
+    <a href="/blech">somewhere</a>
+    <?=draw('somewhere')?>
+  </body>
+</html>
+<script>
+    var f1 = function (n) {
+	return n + <?php /* some value */ ?>;
+    }
+</script>
+<script>
+    var f2 = function (m) {
+	return m + <% /* some value from JSP */ %>;
+    }
+</script>
+<style>
+.blarg {
+    position: relative;
+}
+</style>
+<?php
+  function nothing($x) {
+      ;
+  }
+?>
--- a/parsers/html.c
+++ b/parsers/html.c
@@ -174,8 +174,8 @@ static const char *tokenTypes[] = {
 	E(NAME),
 	E(STRING),
 	E(TEXT),
-	E(TAG_START),
-	E(TAG_START2),
+	E(OPEN_TAG_START),
+	E(CLOSE_TAG_START),
 	E(TAG_END),
 	E(TAG_END2),
 	E(EQUAL),
@@ -195,6 +195,8 @@ static int Lang_html;
 
 
 static void readTag (tokenInfo *token, vString *text, int depth);
+
+static void skipOtherScriptContent (const int delimiter);
 
 static void readTokenText (tokenInfo *const token, bool collectText)
 {
@@ -255,7 +257,6 @@ getNextChar:
 		case '<':
 		{
 			int d = getcFromInputFile ();
-
 			if (d == '!')
 			{
 				d = getcFromInputFile ();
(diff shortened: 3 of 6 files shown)


masatake:
(I'm working on #3491).

masatake:
Could you rebase this pull request on the latest master?

Units/parser-php.r/run-guest-nested.d/args.ctags
```diff
@@ -0,0 +1,4 @@
+--sort=yes
+--extras=+rg
+--fields=+rln
+--language-force=html
```
masatake, line 4: Could you put a new line at the end?

Units/parser-php.r/run-guest-nested.d/input.php
```diff
@@ -0,0 +1,39 @@
...
+<style>
+.blarg {
+    position: relative;
+}
+</style>
+<?php
+  function nothing($x) {
+      ;
+  }
+?>
```
masatake, line 39: Could you put a new line at the end?

masatake: (If possible, I should put a rule to .editorconfig forcing this.)

Units/parser-php.r/run-guest-nested.d/args.ctags
masatake, line 4: This is not a requirement for merging. Just an idea.

I wonder if we can remove --language-force=html.
If the PHP parser recognizes the input is an HTML file when reading "<HTML", the PHP parser can run the HTML parser as a guest. Then the HTML parser runs PHP parser when reading '<?php'

masatake (approved):

Units/parser-php.r/run-guest-nested.d/args.ctags
jafl: Unfortunately, when PHP is the root parser, it does not find the JavaScript functions.  I don't see any way to fix that in the PHP parser, because it only understands "chunks between PHP code", and cannot treat the entire <script> block as a single unit.

masatake (commented):
Can I withdraw my approval?
I would like to understand the relation between this pull request and my old one (https://github.com/masatake/ctags/commit/f476dfe1d31ba7d25d0a1f50f94a756d3fc6cd34) .

Units/parser-php.r/run-guest-nested.d/args.ctags
```diff
@@ -0,0 +1,4 @@
+--sort=yes
```
masatake, line 1: Do you really want the sorted result?

masatake:
Instead of introducing the new test case, how about extending parser-php.r/run-guest.d/ ?

parsers/html.c
```diff
@@ -157,29 +158,31 @@ typedef enum {
...
-	TOKEN_TAG_START2,	/* </ */
+	TOKEN_OPEN_TAG_START,	/* <  */
+	TOKEN_CLOSE_TAG_START,	/* </ */
 	TOKEN_TAG_END,		/* >  */
 	TOKEN_TAG_END2,		/* /> */
 	TOKEN_EQUAL,
 	TOKEN_COMMENT,
-	TOKEN_OTHER
+	TOKEN_OTHER,
+	TOKEN_OTHER_SCRIPT,
```
masatake, line 168: TOKEN_OTHER_SCRIPT is introduced here. However, it seems that this one is not used in this pull request.

jafl: Removed.  It was left over from a previous iteration of the code.

Units/parser-php.r/run-guest-nested.d/args.ctags
jafl: It's easier to read.

jafl:
@masatake The difference between this and your previous commit is that there is `<?php ... ?>` in the middle of the JavaScript block.  Without my change, the functions in the JavaScript block are not tagged.

masatake (approved):

masatake:
Looks good for me.
Thank you.

Units/parser-php.r/run-guest-nested.d/args.ctags
masatake: I inspected the old code that I almost forgot and your change.
In the current implementation, the PHP parser makes a promise for each HTML fragment.
A javascript block is splitted into HTML fragments. As the result, especially the bottom half of the block cannot be parsed well by JavaScript parser.

Instead of making small promises, we can make the PHP parser makes one large promise when "<HTML" is found WITH masking <?php...> and <%...%> areas by whitespace.

A promise can be built with a filter that overwrites the input stream before passing it to a guest parser as far as the filter doesn't change the length of the input stream.
`attachPromiseModifier` defined in promise.c is the engine applying such a filter.
Let me try to apply this algorithm to delete --language-force=html option:-P.

masatake:
Do you like this crazy mechanism for stacking parsers?

More crazy items can be found in https://github.com/universal-ctags/ctags/issues/1577 .
When the JavaScript parser supports destructing binding, I will work on supporting VUE again.

Units/parser-php.r/run-guest-nested.d/input.php
masatake: (You don't have to take any extra action.)

My comment here is wrong.
About input.*, any contents are acceptable.
The new line at the EOF is generally expected.
However, for the test input, there is no common expectation.
We may want to test ctags with various broken inputs.

Feel free to revert the change you had to introduce for responding to my review comment.

Units/parser-php.r/run-guest-nested.d/args.ctags
masatake: Fypp parser is line-oriented but implements this technique; making a promise for whole the file area to run Fortran parser as a guest.
Line filters attached to the promise mask the preprocessor lines in the area with whitespaces.

masatake:
@jafl don't you merge this?

jafl:
@masatake I thought you were working on a different solution to this issue?

masatake:
> @masatake I thought you were working on a different solution to this issue?

Sorry for making you be confused. I will work on the extra solution after you merge this pull request. So I would like you to merge this pull request if you are o.k.

masatake:
Thank you!

PR #3491, opened 2022-09-27
GitHub Actions: use apt-get instead of brew on Ubuntu

masatake:
It seems that brew command is not available on Ubuntu these days.

### DIFF
--- a/.github/workflows/cross-compile-android-ndk-on-mac.yml
+++ b/.github/workflows/cross-compile-android-ndk-on-mac.yml
@@ -17,9 +17,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


masatake:
@leleliu008, though I sent a review request to you, I will merge this pull request as soon as the building and testing on CI platforms are successful because the failure in cross-compiling has become a stopper of development.

masatake:
Big endian, Valgrind, and BSD are snail environments.
CI on the other environments is successfully finished.

masatake:
Can't we cache the image? of vagrant?
https://stackoverflow.com/questions/66261101/using-vagrant-on-github-actions-ideally-incl-virtualbox

PR #3492, opened 2022-09-28
GitHub Actions: vagrant box using cache

leleliu008:
Reference: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Thanks to `Masatake YAMATO` comment in https://github.com/universal-ctags/ctags/pull/3491#issuecomment-1260089341

### DIFF
--- a/.github/workflows/testing-bsds.yml.bak
+++ b/.github/workflows/testing-bsds.yml.bak
@@ -18,6 +18,13 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - uses: actions/cache@v2
+      with:
+        path: ~/.vagrant.d/boxes
+        key: vagrant-generic-freebsd-${{ matrix.freebsd-version }}
+        restore-keys: |
+          vagrant-generic-freebsd-${{ matrix.freebsd-version }}
+
     # https://app.vagrantup.com/boxes/search?utf8=%E2%9C%93&sort=downloads&provider=&q=freebsd
     # https://github.com/leleliu008/github-actions-vagrant
     - uses: leleliu008/github-actions-vagrant@v2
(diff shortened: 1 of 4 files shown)


masatake:
Excellent! I'm surprised at the change for caching is much smaller than I assumed.

leleliu008:
> Excellent! I'm surprised at the change for caching is much smaller than I assumed.

I first time to use this action, what confused me is that BSD workflows is not run.

masatake (approved):

masatake:
We don't need serializing testing on BSDs anymore, do we?

leleliu008:
@masatake we can use [gh-actions-cache](https://github.com/actions/gh-actions-cache) to manage these caches

Screenshot from 2022-09-28 17-41-50

masatake:
> 

> @masatake we can use [gh-actions-cache](https://github.com/actions/gh-actions-cache) to manage these caches
> 
> Screenshot from 2022-09-28 17-41-50

Does it mean we HAVE to manage the life cycle of caches?
e.g. when releasing a new version of xBSD, we have to throw away the cache for the old one?

leleliu008:
> We don't need serializing testing on BSDs anymore, do we?

Yes, we do not need `.github/workflows/testing-bsds.yml.bak` any more  in theory, but I keep it  without deleting it, if this pull request run well, we can delete it.

masatake:
> Yes, we do not need .github/workflows/testing-bsds.yml.bak any more in theory, but I keep it without deleting it, if this pull request run well, we can delete it.

I strongly agree with you.

masatake:
The testing on S390 is enough slow, so serializing the testing on BSDs cannot be a bottle neck:-P.

leleliu008:
> The testing on S390 is enough slow, so serializing the testing on BSDs cannot be a bottle neck:-P.

`.github/workflows/testing-big-endian.yml` is using `KVM+qemu`, codes run in virtual machine is slower, I will look into this.

PR #3496, opened 2022-10-03
Introduce `--with-rst2man` and `--with-rst2html`

catap:
The idea is quite simple: let allow the user who builds it to suggest which `rst2man` and `rst2html` should be used, instead of guessing.

I've also extended the guess list to  python 3.10.


masatake (approved):

masatake:
Thank you.

Issue #3494, opened 2022-09-29
A new language definition for Zshshell script


psprint:
I've defined a new language that much better parses shell script:
- it allows arbitrary [[:alphnum].+:_-] characters in the name of the function (runnables in shell are allowed the same as the filenames),
- it recognizes multiple variables in single line, e.g.: `local a=1 b=3`.

I'm using it succesfully in `zinit` zsh plugin manager: https://asciinema.org/a/aEqd70X5oR6ruvr1OwrdJ4goG

The def file is at:
https://github.com/zdharma-continuum/zinit/blob/main/share/zsh.ctags

I've somewhere saw some info about `optlib`, that users are encouraged to submit custom language definition, so I'm doing so.

masatake:
Welcome!

About the video about the zsh plugin, I could not understand it well.
My knowledge about the area may be too limited or the clock speed of my brain is too slow.

> I've somewhere saw some info about optlib, that users are encouraged to submit custom language definition, so I'm doing so.

Yes.

I would like to integrate your effort into our source tree in directly or indirectly way.

> I've defined a new language that much better parses shell script:

As you may know, ctags has a parser for shell scripts.
The existing one also has some interesting features:
```
% cat foo
#!/bin/sh

. /etc/site-common

cat > foo.c <<EOF
int main(void) {
 printf("%s, hello world\n", "$USER");
 return 0;
}
EOF

cat > foo.py <<EOF
def hello(user):
    print user + ' hello world'
hello('$USER')
EOF
% ~/bin//ctags --options=NONE -G -o - --extras=+r --fields=+Krl /tmp/foo
ctags: Notice: No options will be read from files or environment
/etc/site-common	/tmp/foo	/^. \/etc\/site-common$/;"	script	language:Sh	roles:loaded
EOF	/tmp/foo	/^EOF$/;"	heredoc	language:Sh	roles:endmarker
EOF	/tmp/foo	/^cat > foo.c <<EOF$/;"	heredoc	language:Sh	roles:def
EOF	/tmp/foo	/^cat > foo.py <<EOF$/;"	heredoc	language:Sh	roles:def
hello	/tmp/foo	/^def hello(user):$/;"	function	language:Python	roles:def
main	/tmp/foo	/^int main(void) {$/;"	function	language:C	typeref:typename:int	roles:def
```

So I would like to extend the shell script parser based on your zsh.ctags.

[...]

PR #3499, opened 2022-10-09
Zsh: new standalone parser understanding autoload built-in

masatake:
Sh parser handled .zsh input.

However, as reported and discussed in #3494, the paresr doesn't handle zsh related items well.

So I introduced the standalone Zsh parser. Supporting multiple
dialects of shell syntax in the Sh parser may make the code
complicated. Splitting the code into the Sh parser and the Zsh
parser allows us to improve the code for zsh more aggressively.

The current implementation supports only "autload" zsh built-in.
"autoload" role is added to "script" kind and "function" kind.

e.g.

    $ cat input.zsh
    autoload /tmp/foo
    $ ./ctags --extras=+r --fields=+rK -o - input.zsh
    /tmp/foo	input.zsh	/^autoload \/tmp\/foo$/;"	script	roles:autoloaded
    foo	input.zsh	/^autoload \/tmp\/foo$/;"	function	roles:autoloaded


masatake:
```
autoload foo
```

`foo` is not a function definition.
So the Zsh parser extracts `foo` as a reference tag.
See the man page and documents about the concept reference tags.

You must add ``--extras=+r`` to extract the references for autoloading functions. ``--field=+r`` shows how the tag is referenced.

PR #3509, opened 2022-10-16
TOML: new PEG based parser

masatake:
The peg file is imported from

	Copyright (c) Tom Preston-Werner
	https://github.com/toml-lang/toml/blob/main/toml.abnf

- [ ] a subparser of Cargo.toml (https://doc.rust-lang.org/cargo/index.html)


peg/toml.peg
```diff
@@ -0,0 +1,273 @@
...
+# -------------------------------------------------------------------------
+# -- Tables
+
+table <- stdTable / arrayTable
+
+stdTable <- '[' ws { tableKeyStart(auxil, false, $0s); } tableKey { tableKeyEnd(auxil); } ws ']'
+
+arrayTable <- '[[' ws { tableKeyStart(auxil, true, $0s); } tableKey { tableKeyEnd(auxil); } ws ']]'
+
+keyval <- { keyvalStart(auxil, $0s); } tableKey { keyvalKeyEnd(auxil); } ws '=' ws val { keyvalValEnd(auxil); }
```
masatake, line 93: Instead of `key`, it seems that we need `tableKey` here for accepting toml-1.0 input .

masatake:
It seems that a bug in pegof.

dolik-rce:
First problem I found is that pegof doesn't currently support unicode escapes in strings, e.g. `underscore <- '\u005F'`. It should be easy to implement.

masatake:
@dolik-rce Thank you.

PR #3522, opened 2022-10-31
GitHubAtions: choco install -y --source cygwin is broken since chocolatey v1.2.0, so we change to the offical way to install cygwin packages

leleliu008:


masatake (approved):

masatake:
Thank you.

PR #3527, opened 2022-11-09
GitHub Actions: (macos) don't install libxml2 and gdb

masatake:


masatake:
#3528 may fix the issue I would like to fix in this pull request.

PR #3528, opened 2022-11-09
GitHub Actions: fix-github-actions-workflow-macos

leleliu008:
python@3.11 symlink 2to3-3.11 to /usr/local/bin/2to3-3.11, but it already exists. so we need use --overwrite option when installing python@3.11

python@3.11 is depended by gdb.


masatake (approved):

masatake:
@leleliu008 Thank you!

PR #3541, opened 2022-11-19
CXX: add constexpr to "properties:" field

### LINKED ISSUE #3539: C++: Output information on `constexpr` and `consteval` functions
Hi!
It seems as if `ctags` currently does not support distinctive tagging of `constexpr` and `consteval` functions (they're merely tagged as regular `function`s).
One possible workaround is to check the `pattern` field for these keywords, but having `ctags` output a (extra) field or property for them would be cleaner. (I personally would prefer adding these keywords to the `properties` field.)
Is there any preliminary or ongoing work on this already?
If not, I might take a stab (and would appreciate any advise for approaching this, as I'm not familiar with the C++ parser).
Thanks!

------------------------------------------------------------------------------------------

The name of the parser: C/C++

```
$ ctags --options=NONE --fields-all="*" --output-format=json const.cpp
```

The content of input file (from [here](https://en.cppreference.com/w/cpp/language/consteval)):

```cpp
// This function might be evaluated at compile-time, if the input
// is known at compile-time. Otherwise, it is executed at run-time.
constexpr unsigned factorial(unsigned n) {
    return n < 2 ? 1 : n * factorial(n - 1);
}

// With consteval we enforce that the function will be evaluated at compile-time.
consteval unsigned combination(unsigned m, unsigned n) {
    return factorial(n) / factorial(m) / factorial(n - m);
}
```

```
{"_type": "tag", "name": "combination", "path": "const.cpp", "pattern": "/^consteval unsigned combination(unsigned m, unsigned n) {$/", "language": "C++", "line": 8, "typeref": "typename:consteval unsigned", "kind": "function", "signature": "(unsigned m,unsigned n)", "roles": "def", "end": 10}
{"_type": "tag", "name": "factorial", "path": "const.cpp", "pattern": "/^constexpr unsigned factorial(unsigned n) {$/", "language": "C++", "line": 3, "typeref": "typename:unsigned", "kind": "function", "signature": "(unsigned n)", "roles": "def", "end": 5}
```

The version of ctags:

```
$ ctags --version
Universal Ctags 5.9.0(7b5c7efa), Copyright (C) 2015-2022 Universal Ctags Team
  Compiled: Jul 13 2022, 17:12:58
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +gnulib_fnmatch, +gnulib_regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc, +optscript
```

How do you get ctags binary:

macosx binary taken from Universal-ctags/homebrew-universal-ctags project

masatake:
Partially close #3539.

### DIFF
--- a/parsers/cxx/cxx_tag.h
+++ b/parsers/cxx/cxx_tag.h
@@ -150,7 +150,9 @@ typedef enum _CXXTagProperty
 	// scoped enum (C++11)
 	CXXTagPropertyScopedEnum = (1 << 16),
 	// function-try-block: int f() try { ... } catch { ... }
-	CXXTagPropertyFunctionTryBlock = (1 << 17)
+	CXXTagPropertyFunctionTryBlock = (1 << 17),
+	// constexpr has been seen.
+	CXXTagPropertyConstexpr = (1 << 18),
 } CXXTagProperty;
 
 // Set the modifiers field of the tag.
--- a/Units/parser-cxx.r/properties-constexpr.d/args.ctags
+++ b/Units/parser-cxx.r/properties-constexpr.d/args.ctags
@@ -1,0 +1,4 @@
+--sort=no
+--kinds-c++=*
+--fields=+x
+--fields-c++=+{properties}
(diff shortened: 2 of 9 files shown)


parsers/cxx/cxx_tag.h
```diff
@@ -150,7 +150,9 @@ typedef enum _CXXTagProperty
 	// scoped enum (C++11)
 	CXXTagPropertyScopedEnum = (1 << 16),
 	// function-try-block: int f() try { ... } catch { ... }
-	CXXTagPropertyFunctionTryBlock = (1 << 17)
+	CXXTagPropertyFunctionTryBlock = (1 << 17),
+	// constexpr hasa been seen.
```
delsner, line 154:
```suggestion
	// constexpr has been seen.
```

PR #3571, opened 2022-12-03
PowerShell: introduce enum kind

### LINKED ISSUE #3570: PowerShell: introduce enum kind
The content of input file:

```powershell
# EnumName1
enum EnumName1 {
    Label11
    Label12 = 10
}

# EnumName2
enum EnumName2 {
    Label21
    Label22 = 20
}

# EnumName3
Enum EnumName3 {
    Label31
    Label32 = 30
}

# EnumName4
[Flags()] enum EnumName4 {
    Label41
    Label42 = 40
}
```

```tags
```

The tags output you expect:

```tags
EnumName1       input.ps1       /^enum EnumName1 {$/;"  e
EnumName2       input.ps1       /^enum EnumName2 {$/;"  e
EnumName3       input.ps1       /^enum EnumName3 {$/;"  e
EnumName4       input.ps1       /^[Flags()] enum EnumName4 {$/;"      e
```

The version of ctags:

```console
$ ctags --version
Universal Ctags 5.9.0(60cc214b), Copyright (C) 2015-2022 Universal Ctags Team
  Compiled: Dec  3 2022, 23:02:53
  URL: https://ctags.io/
  Optional compiled features: +wildcards, +regex, +gnulib_regex, +iconv, +option-directory, +packcc, +optscript
```

How do you get ctags binary:

compiled from source

kumarstack55:
Closes #3570

### DIFF
--- a/parsers/powershell.c
+++ b/parsers/powershell.c
@@ -41,6 +41,7 @@ typedef enum {
 	K_VARIABLE,
 	K_CLASS,
 	K_FILTER,
+	K_ENUM,
 	COUNT_KIND
 } powerShellKind;
 
@@ -49,6 +50,7 @@ static kindDefinition PowerShellKinds[COUNT_KIND] = {
 	{ true, 'v', "variable",	"variables" },
 	{ true, 'c', "class",		"classes" },
 	{ true, 'i', "filter",		"filter" },
+	{ true, 'g', "enum",		"enum names" },
 };
 
 
@@ -77,6 +79,7 @@ enum {
 	KEYWORD_function,
 	KEYWORD_filter,
 	KEYWORD_class,
+	KEYWORD_enum,
 };
 
 /* We need an integer that is not an unsigned to allow KEYWORD_NONE. */
@@ -86,6 +89,7 @@ static const keywordTable PowerShellKeywordTable[] = {
 	{ "function",	KEYWORD_function },
 	{ "filter",		KEYWORD_filter },
 	{ "class",		KEYWORD_class },
+	{ "enum",		KEYWORD_enum },
 };
 
 typedef struct {
@@ -170,6 +174,18 @@ static void makeClassTag (const tokenInfo *const token)
 		tagEntryInfo e;
 
 		initPowerShellEntry (&e, token, K_CLASS, NULL);
+
+		makeTagEntry (&e);
+	}
+}
+
+static void makeEnumTag (const tokenInfo *const token)
+{
+	if (PowerShellKinds[K_ENUM].enabled)
+	{
+		tagEntryInfo e;
+
+		initPowerShellEntry (&e, token, K_ENUM, NULL);
 
 		makeTagEntry (&e);
 	}
@@ -600,6 +616,39 @@ static bool parseClass (tokenInfo *const token)
 	return readNext;
 }
 
+/* parse a enum
+ *
+ * 	enum EnumName {}
+ */
+static bool parseEnum (tokenInfo *const token)
+{
+	bool readNext = true;
+	vString *nameFree = NULL;
+
+	readToken (token);
+
+	if (token->type != TOKEN_IDENTIFIER)
+		return false;
+
+	makeEnumTag (token);
+	nameFree = vStringNewCopy (token->string);
+	readToken (token);
+
+	while (token->type != TOKEN_OPEN_CURLY && token->type != TOKEN_EOF)
+	{
+		readToken (token);
+	}
+
+	if (token->type == TOKEN_OPEN_CURLY)
+		enterScope (token, nameFree, K_ENUM);
+	else
+		readNext = false;
+
+	vStringDelete (nameFree);
+
+	return readNext;
+}
+
 /* parses declarations of the form
  * 	$var = VALUE
  */
@@ -673,6 +722,10 @@ static void enterScope (tokenInfo *const parentToken,
 
 					case KEYWORD_class:
 						readNext = parseClass (token);
+						break;
+
+					case KEYWORD_enum:
+						readNext = parseEnum (token);
 						break;
 
 					default: break;
--- a/parsers/powershell.c
+++ b/parsers/powershell.c
@@ -574,18 +574,16 @@ static bool parseFunction (tokenInfo *const token, int kind)
 static bool parseClass (tokenInfo *const token)
 {
 	bool readNext = true;
-	tokenInfo *nameFree = NULL;
+	vString *nameFree = NULL;
 
 	readToken (token);
 
 	if (token->type != TOKEN_IDENTIFIER)
 		return false;
 
-	nameFree = newToken ();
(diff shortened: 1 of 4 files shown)


parsers/powershell.c
```diff
@@ -602,6 +618,41 @@ static bool parseClass (tokenInfo *const token)
...
+	readToken (token);
+
+	if (token->type != TOKEN_IDENTIFIER)
+		return false;
+
+	nameFree = newToken ();
+	copyToken (nameFree, token, true);
+	readToken (token);
+
+	makeEnumTag (nameFree);
```
masatake, line 639: Do we need `nameFree` token?

At line 635 and 636, you make a backup of `token`: `nameFree`.
Then, at line 637, read new data to `token`.
Finally, you did make a tag for `nameFree`.
I think we can do it without `nameFree`.

```
635:  makeEnumTag(token);
636:  readToken(token);
```

parsers/powershell.c
```diff
@@ -602,6 +618,41 @@ static bool parseClass (tokenInfo *const token)
...
+
+	while (token->type != TOKEN_OPEN_CURLY && token->type != TOKEN_EOF)
+	{
+		readToken (token);
+	}
+
+	if (token->type == TOKEN_OPEN_CURLY)
+		enterScope (token, nameFree->string, K_CLASS);
```
masatake, line 647: Oh, I see. You need the string in this step.
In that case, you need only a vstring.
See main/vstring.h.

See (*1).

parsers/powershell.c
masatake: (*1)

```
635: makeEnumTag(token);
636: vString *nameFree = vStringCopyNewCopy(token->string);
637: readToken(token);
```

See (*2).

parsers/powershell.c
masatake: (*2)
```
647: enterScope (token, nameFree, K_CLASS);
```

parsers/powershell.c
```diff
@@ -602,6 +618,41 @@ static bool parseClass (tokenInfo *const token)
...
+	else
+		readNext = false;
+
+	deleteToken (nameFree);
```
masatake, line 651: If you will use vString, you can use `vStringDelete (nameFree);` here.

parsers/powershell.c
masatake: I wonder why K_CLASS. Don't you use K_ENUM here?

parsers/powershell.c
```diff
@@ -49,6 +50,7 @@ static kindDefinition PowerShellKinds[COUNT_KIND] = {
 	{ true, 'v', "variable",	"variables" },
 	{ true, 'c', "class",		"classes" },
 	{ true, 'i', "filter",		"filter" },
+	{ true, 'e', "enum",		"enums" },
```
masatake, line 53: `g` can be another candidate for `enum`.

You may want to extract items inside enum like `Label11` and `Label12`.
If that case, what kind name will you give to the items?

I don't know the terms of PowerShell. However, we can find hints in C parser:

```
$ ./ctags --list-kinds-full=C
#LETTER NAME       ENABLED REFONLY NROLES MASTER DESCRIPTION
D       macroparam no      no      0      C      parameters inside macro definitions
L       label      no      no      0      C      goto labels
Y       unknown    no      yes     1      C      unknown identifier
d       macro      yes     no      2      C      macro definitions
e       enumerator yes     no      0      C      enumerators (values inside an enumeration)
f       function   yes     no      0      C      function definitions
g       enum       yes     no      0      C      enumeration names
...
```

The C parser calls items inside an enum "enumerator".
For enum, the C parser uses `g` as its kind letter.
For enumerator, the C parser uses `e` as its kind letter.

You can make the PowerShell parser similar to the C parser.
Of course, the parsers in ctags are independent each other, So what I wrote here is nothing more than a comment.

kumarstack55:
@masatake : Thanks for your comment.

I implemented parseEnum() based on parseClass() .
However, I realized that I had not given enough consideration to the implementation.
Let me think about it for a moment.

parsers/powershell.c
kumarstack55: LETTER for kinds in PowerShell What should we do?

As for ctags, I think it is better to make them in a form similar to the C and C++ languages.

Since my use of ctags is not so deep, I have not come up with a good way to choose a LETTER.
In such a case, I think it is better to choose the LETTER in conjunction with its neighbors.
In view of ctags, PowerShell is similar to C and C++ languages, so I think it is better to make it similar to the C parser.

For the kind for the enum name, let us pick `g` to match the C language.

In the future, when choosing labels for enums and properties for classes, the C parser may be helpful as well.

parsers/powershell.c
kumarstack55: I understand your comment.
I noticed that there is an unwanted token.
I would like to delete those tokens.

I created parseEnum() based on parseClass().
It appears that parseClass() has an unwanted token as well.
I hope to change this as well in the near future.

parsers/powershell.c
masatake: Good.

parsers/powershell.c
kumarstack55: I understood that in this pull request, you want me to include a commit to fix parseClass() first, and then a commit to add parseEnum().

If you mean that you want me to create a new pull request, please let me know. I will do so.

parsers/powershell.c
kumarstack55: Thank you for teaching me about vString.

We should use K_ENUM.

Please let me fix it.

parsers/powershell.c
masatake: YES. I want you to create a new pull request for fixing parseClass().
I will merge it quickly. Then you can create another pull request adding parseEnum().

masatake: Doing these things in a pull request is also acceptable as far as commits are separated.

masatake (approved):

masatake:
Looks good. Thank you for updating your pull request.

I have one very minor request.
Could you add "PowerShell: " as the prefix to the commit headers?

"PowerShell: introduce enum kind", this one is o.k.

"Change the variable type in parseClass()"; this missed the prefix. This should be 
"PowerShell: change the variable type in parseClass()".

masatake (changes requested):
Please add "PowerShell: " to the log header of the first commit.

kumarstack55:
Sorry. I forgot that there is a commit message rule in this project.

I modified the commit message and pushed.

Please check it out.

masatake (approved):

masatake:
Thank you.

masatake:
I remember writing about our rules for commit headers in this project.
https://docs.ctags.io/en/latest/contributions.html#title-of-commit-log-and-pull-request

About the first commit, "PowerShell,refactor: " is better.

Issue #3551, opened 2022-11-29
Add --verison (and --list-features) option to readtags


masatake:

masatake:
--version-library is not needed.
Just --version is enough for the first step.
The option must appear on --help output.

leleliu008:
maybe make a new subcommand is better.

following is a example:

![Screenshot from 2022-12-07 10-44-53](https://user-images.githubusercontent.com/3040852/206075353-819840fc-837b-46f4-afbb-72157bddf617.png)

masatake:
Adding a subcommand to readtags?

I cannot imagine a way to add a subcommand to readtags without breaking its command line interface.

leleliu008:
`readtags --env`

masatake:
Oh, I see.

PR #3578, opened 2022-12-04
readtags: print the input fields literally even if -E option is given

### LINKED ISSUE #3559: readtags: improper handling of escape sequences in input field
The manpage tags(5) defines some escaping rules, including:

- When a value contains a `\t`, this stands for a `<Tab>`.
- When a value contains a `\\`, this stands for a single `\` character.

When using readtags to read regular tags, we can confirm this works well.

``` console
$ cat tags
!_PTAG  value   /comment/;"
regular_tag     /path/to/file   /^pattern$/;"   kind:\
$ readtags -Ee -l
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
```

The `\` is translated to a `\\` to make it machine readable. And

``` console
$ cat tags
!_PTAG  value   /comment/;"
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
$ readtags -Ee -l
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
```

Readtags knows `\\` is an escape sequence and doesn't escape it more.

But things are different in pseudo tags:

``` console
$ cat tags
!_PTAG  value\  /comment/;"
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
$ readtags -Ee -D
!_PTAG  value\\ /comment/
```

This is expected. But

``` console
$ cat tags
!_PTAG  value\\ /comment/;"
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
$ readtags -Ee -D
!_PTAG  value\\\\       /comment/
```

and

``` console
$ cat tags
!_PTAG  value\t /comment/;"
regular_tag     /path/to/file   /^pattern$/;"   kind:\\
$ readtags -Ee -D
!_PTAG  value\\t        /comment/
```

Now it seems readtags always reads pseudo tags literally, and doesn't deal with escape sequences in it.

If this is the case, it may prevent us from encoding arbitrary information in pseudo tags, which may influence https://github.com/universal-ctags/ctags/issues/3085.

This is related to https://github.com/universal-ctags/citre/issues/143 but you can ignore it for now, as it seems can be fixed on the Citre side.

masatake:
Close #3559.

### DIFF
--- a/Tmain/readtags-escaping-input-no-roundtrip.d/output.tags
+++ b/Tmain/readtags-escaping-input-no-roundtrip.d/output.tags
@@ -1,0 +1,1 @@
+x	a\tb\nc/f.c	/^int x;$/;"	kind:v	typeref:typename:int
--- a/Tmain/readtags-escaping-input-no-roundtrip.d/run.sh
+++ b/Tmain/readtags-escaping-input-no-roundtrip.d/run.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Copyright: 2022 Masatake YAMATO
+# License: GPL-2
+
+READTAGS=$3
+
+. ../utils.sh
+
+if ! [ -x "${READTAGS}" ]; then
+	skip "no readtags"
+fi
+
+${READTAGS} -eE -t output.tags -l
+${READTAGS} -e  -t output.tags -l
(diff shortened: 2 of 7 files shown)


masatake:
Does testing of the "roundtrip" way doesn't work on windows?

masatake:
@AmaiKinono Thank you for reviewing.

> I think we still need to write about how {tagfile} is printed in the manual, according to how it's actually implemented, so a client tool knows how to get the actual path.

I will write it in "EXCEPTIONS IN UNIVERSAL CTAGS" or somewhere in tags(5).
I would like you to review the text I will add.

> Another problem is how could a client tool know if readtags has this patch. It's hard to do without a version number.


I agree with you. I opened #3551.

masatake:
I wonder whether we can use the same version number in ctags and readtags.

masatake:
The extra issues pointed out by @AmaiKinono are traced by #3584 and #3551.

PR #3579, opened 2022-12-05
Python:  extract reference tags

masatake:


masatake:
I must redefine the semantics of the scope field:
```
def foo():
    a.b()
```

In this case, for the reference tag `b`, how the scope filed should be: `scope:foo` or `scope:a`.  I got the hint about this from @pragmaware ago. I must re-read his comment, and I must document it in "the reference tag design guide" or something.

As @AmaiKinono suggested, there are two types of reference tags: dependency reference tags and generic reference tags. The former one should be enabled by default. The latter one would be disabled by default. This also should be written in the guide.

masatake:
Versioning and updating the man page for Python are needed for merging.

PR #3609, opened 2022-12-25
GitHub actions: fix testing-mac.yml python problem

leleliu008:


leleliu008:
I touched other commit in this pr, I file a new one(#3610), so close this one.

PR #3611, opened 2022-12-27
Org:  optimize

masatake:
The Org parser is incredibly slow.

The original:
```
[yamato@dev64]~/var/ctags-github% ./ctags --totals /home/yamato/var/emacs/doc/misc/org.org
1 file, 22606 lines (802 kB) scanned in 98.8 seconds (8 kB/s)
625 tags added to tag file
625 tags sorted in 0.00 seconds
longest tag line = 127
```

With this pull request:
```
[yamato@dev64]~/var/ctags-github% ./ctags --totals /home/yamato/var/emacs/doc/misc/org.org
1 file, 22606 lines (802 kB) scanned in 0.6 seconds (1291 kB/s)
624 tags added to tag file
624 tags sorted in 0.00 seconds
longest tag line = 127
```

It becomes 150% faster with this pull request.

BTW, this parser doesn't fill scope fields.


masatake:
Howeever,... one tag is lost with this change.

masatake:
I found the Org parser extracts many noises. It may be rewritten more line-oriented way.

PR #3619, opened 2023-01-04
KernelDoc: new parser

masatake:
```
/** 
 * getData() - get data from containers 
 * 
 */
void * getData() {...
```

The first data can be extracted as a tag with "documentation" role of "function" kind in C language or a tag with "function" kind of KenrelDoc language.

My choice in this pull request is the latter one.


parsers/kerneldoc.c
```diff
@@ -0,0 +1,314 @@
...
+	if (e)
+		e->extensionFields.endLine = getInputLineNumber ();
+}
+
+static void initialize (const langType language)
+{
+	Lang_kerneldoc = language;
+}
+
+static void inputStart (subparser *s)
```
masatake, line 271: Why is this line not covered?

Units/parser-kerneldoc.r/simple.d/input.c
```diff
@@ -0,0 +1,84 @@
+/* Taken from linux/kernel/cpu.c */
```
masatake, line 1: enum and function() must be added to input-0.

PR #3640, opened 2023-02-06
GitHubActions: fix testing-openbsd.yml automake version broken issue

leleliu008:


leleliu008:
It seems that the ubuntu's offical APT server is being updated. Iet's wait a while then rerun failed jobs.

masatake:
@leleliu008 Thank you.

PR #3644, opened 2023-02-14
regcomp: Increase pointer assignment check

likunyur:
When DEBUG macro takes effect, Store data when re_str pointer is allocated successfully.


leleliu008:
following place needs to do such checking too:
https://github.com/universal-ctags/ctags/blob/master/dsl/es.c#L2675

masatake:
@likunyur thank you, but this should be reported to gnulib project.

@leleliu008 I will look into return value of *alloc functions.

PR #3654, opened 2023-03-02
SQL: Skip PL/SQL selection directives and add sanity check for inquiry directive size

### LINKED ISSUE #3647: SQL: slow parsing of certain SQL files
We got a bug report here https://github.com/geany/geany-osx/issues/42#issuecomment-1431195766 in Geany where Geany stopped responding for very long time for not so large SQL file (~16000 LOCs). I bisected this problem to this commit:

https://github.com/universal-ctags/ctags/commit/4dbda622c699e1ab602c0c6d4ec12750a4a7af69

Basically the problem isn't the commit itself but rather that `string` which is passed to this function and which should contain just the current token seems to contain a huge amount of the SQL code and not just a single token (I haven't checked the size but it certainly is hundreds of characters long). As this string is getting bigger, `lookupCaseKeyword()` is called every time a charactger is added in

```C
if (empty_tag
    && KEYWORD_inquiry_directive == lookupCaseKeyword (vStringValue (string), Lang_sql))
```

which consumes a lot of time. A simple workaround which reduces the parsing time significantly is to check the size of the string against some constant corresponding to the longest keyword:

```C
if (empty_tag && string->length < 20
    && KEYWORD_inquiry_directive == lookupCaseKeyword (vStringValue (string), Lang_sql))
```

Still, it would be good to investigate what the reason behind the long token is and fix it properly.

I assume that @cloudis-ild could send the file causing this issue privately or I could send it as well if he agrees.

techee:
PLSQL selection directives are of the form
```
$IF boolean_static_expression $THEN
   text
[ $ELSIF boolean_static_expression $THEN
   text
]...
[ $ELSE
   text
]
$END
```
In addition, boolean_static_expression conditions like
```
$$MY_CUSTOM_VAR > 3
```
can appear which confuses the current parser because it
thinks `$$` is the start of dollar-quoted string and drops
the rest of the code. In addition, as the double-quoted
string is getting bigger and bigger and tested against
inquiry directives by
```
lookupCaseKeyword (vStringValue (string), Lang_sql)
```
it takes huge amount of time to perform the repeated
lookups in the hash table.

This patch skips all the code between `$IF` and `$END` because
we can't know which of the branches to take and this is the
easiest way to avoid invalid code. In addition, this patch
also drops `$` from all identifiers starting with `$$` between
`$IF` and `$END` as these may be conditional variables in the
`$IF` block.

---

Another issue is that
`$$` typed anywhere in the code, e.g. by accident, makes the rest of the
code appear to the parser as a dollar-quoted string which can be thousands
of bytes long. In this case `lookupCaseKeyword()` is called repeatedly on
this ever increasing string which consumes a lot of time and makes the
parser appear completely unresponsive for large files.

This patch adds a sanity check to perform `lookupCaseKeyword()` only for
strings of length smaller than 30 (currently the longest inquiry directive
keyword is 21 characters long so there should be some safe extra margin
even for longer keywords if added in the future).

Fixes #3647.

### DIFF
--- a/parsers/sql.c
+++ b/parsers/sql.c
@@ -87,6 +87,7 @@ enum eKeywordId {
 	KEYWORD_drop,
 	KEYWORD_else,
 	KEYWORD_elseif,
+	KEYWORD_elsif,
 	KEYWORD_end,
 	KEYWORD_endif,
 	KEYWORD_event,
@@ -177,7 +178,12 @@ typedef enum eTokenType {
 	TOKEN_CLOSE_SQUARE,
 	TOKEN_TILDE,
 	TOKEN_FORWARD_SLASH,
-	TOKEN_EQUAL
+	TOKEN_EQUAL,
+	TOKEN_PREPROC_IF,
+	TOKEN_PREPROC_ELSIF,
+	TOKEN_PREPROC_ELSE,
+	TOKEN_PREPROC_THEN,
+	TOKEN_PREPROC_END,
 } tokenType;
 
 typedef struct sTokenInfoSQL {
@@ -294,6 +300,7 @@ static const keywordTable SqlKeywordTable [] = {
 	{ "drop",							KEYWORD_drop			      },
 	{ "else",							KEYWORD_else			      },
 	{ "elseif",							KEYWORD_elseif			      },
+	{ "elsif",							KEYWORD_elsif			      },
 	{ "end",							KEYWORD_end				      },
 	{ "endif",							KEYWORD_endif			      },
 	{ "event",							KEYWORD_event			      },
@@ -431,6 +438,7 @@ static struct SqlReservedWord SqlReservedWord [SQLKEYWORD_COUNT] = {
 	[KEYWORD_drop]          = {1 & 0&1&1&1 & 1&1 & 1},
 	[KEYWORD_else]          = {1 & 1&1&1&1 & 1&1 & 1},
 	[KEYWORD_elseif]        = {1 & 0&0&0&0 & 0&0 & 1},
+	[KEYWORD_elsif]         = {0 & 0&0&0&0 & 0&1 & 0},
 	[KEYWORD_end]           = {0 & 1&1&1&1 & 0&1 & 1},
 	[KEYWORD_endif]         = {0 & 0&0&0&0 & 0&0 & 1},
 	[KEYWORD_event]         = {0 & 0&0&0&0 & 0&0 & 0},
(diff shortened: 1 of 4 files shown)


b4n:
> This patch skips all the code between `$IF` and `$END` because
> we can't know which of the branches to take and this is the
> easiest way to avoid invalid code.

Shouldn't it rather take one branch, any branch?  I don't know the syntax (and have been too lazy to look it up any closely), but if it's anything similar to e.g. C preprocessor taking *no* branch is the best way to get invalid syntax.  I'd then take the first branch (because it's likely the easiest solution), and if it's not what's wanted, well… at least it gives something.

techee:
> Shouldn't it rather take one branch, any branch?

Yeah, you are right, I'll have a look at it when I have time.

techee:
@b4n Done.

techee:
I've created #3664 which might be a better and more universal fix for the original "slow parser" problem as it fixes it at the level of the keyword table so individual parsers don't have to worry about it any more.

techee:
> I've created https://github.com/universal-ctags/ctags/pull/3664 which might be a better and more universal fix for the original "slow parser" problem as it fixes it at the level of the keyword table so individual parsers don't have to worry about it any more.

Just to be clear, it replaces the patch performing the string length check
```C
if ((string->length < 30 && KEYWORD_inquiry_directive == lookupCaseKeyword (vStringValue (string), Lang_sql))
```
but the PL/SQL selection directive is still worth implementing.

masatake:
(I'm catching up now.)

parsers/sql.c
```diff
@@ -294,6 +300,7 @@ static const keywordTable SqlKeywordTable [] = {
 	{ "drop",							KEYWORD_drop			      },
 	{ "else",							KEYWORD_else			      },
 	{ "elseif",							KEYWORD_elseif			      },
+	{ "elsif",							KEYWORD_elsif			      },
```
masatake, line 303: Could you add an entry for `KEYWORD_elsif` to 
```
static struct SqlReservedWord SqlReservedWord [SQLKEYWORD_COUNT] = {
	/*
	 * RESERVED_BIT: MYSQL & POSTGRESQL&SQL2016&SQL2011&SQL92 & ORACLE11g&PLSQL & SQLANYWERE
	 *
	 * {  0  } means we have not inspect whether the keyword is reserved or not.
	 */
```

masatake (changes requested):

parsers/sql.c
```diff
@@ -797,8 +797,8 @@ static tokenType parseDollarQuote (vString *const string, const int delimiter, i
 		{
 			vStringPut (string, c);
 			if (empty_tag
-				&& (KEYWORD_inquiry_directive == lookupCaseKeyword (vStringValue (string),
-																	Lang_sql)
+				&& ((string->length < 30 && KEYWORD_inquiry_directive == lookupCaseKeyword (vStringValue (string),
```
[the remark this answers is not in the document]
techee: The patch checking string->length here can be dropped completely now. It is the reason why I created #3664 which is more universal and works for every parser automatically.

parsers/sql.c
techee: Done, thanks for noticing. I'll then squash the commits when you think there's not more work to do.

masatake: Yes, please. Could you remove [c3b0324](https://github.com/universal-ctags/ctags/pull/3654/commits/c3b0324af1b9db57dc2c5a913bb8d7a650328654) and [9f532d3](https://github.com/universal-ctags/ctags/pull/3654/commits/9f532d3f72d358be7ecf0d932f098f5f72b13c10)?

parsers/sql.c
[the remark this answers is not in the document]
techee: Done.

parsers/sql.c
techee: Done.

PR #3655, opened 2023-03-03
doc: Verilog, add new kind "define"

my2817:
refer to https://github.com/universal-ctags/ctags/pull/3653


k-takata:
Could you also update the files in `docs/man/` by running `make -C man`?

I think that this kind of mismatch had been detected by the CI, but the CI is green now. Not sure why.

masatake:
It is checked on CircleCI. These days, sometimes CircleCI doesn't work.

masatake:
Thank you.

PR #3657, opened 2023-03-04
GitHub Actions: change github actions checkout from v2 to v3

leleliu008:
https://github.com/actions/checkout@v2 use node12 runtime by default 
https://github.com/actions/checkout@v3 use node16 runtime by default

Reference: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


masatake:
Thank you.

PR #3664, opened 2023-03-14
main: add quick path for looking up too long strings in the keyword table

techee:
Parser code like
```C
vString *str = vStringNew();
while (someCondition)
{
    int c = getcFromInputFile();
    vStringPut(str, c);
    if (lookupCaseKeyword (vStringValue(str), some_lang))
    {
        do_stuff();
        vStringClear(str);
    }
}
```
is prone to quadratic complexity because when `someCondition` isn't satisfied in the parsed file, the string str grows by one in each iteration and in each iteration `lookupCaseKeyword()` has to go through `strlen(str)` characters to compute the hash.

Since we know the maximum length of the strings inside the keyword hash table, we can store this value and if the queried string is longer than this value, we can be sure it isn't present in the hash table and return quickly without having to compute the full hash of the string.

### DIFF
--- a/main/keyword.c
+++ b/main/keyword.c
@@ -36,6 +36,7 @@ typedef struct sHashEntry {
 */
 static const unsigned int TableSize = 2039;  /* prime */
 static hashEntry **HashTable = NULL;
+static unsigned int MaxEntryLen = 0;
 
 /*
 *   FUNCTION DEFINITIONS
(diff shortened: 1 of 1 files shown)


masatake:
Thank you,

masatake:
Now we can add

```
int lookupKeywordVString (const vString *string, langType language);
```

:-)

PR #3671, opened 2023-03-20
Fortran: add "linkName" extra

### LINKED ISSUE #3668: Improve handling of Fortran case insensitivity
Fortran is a case insensitive language and as a result it would be great if ctags took this into account.

For example the following code defines a subroutine as `FOO` but can happily be called by `foo`, and similarly uses `myvar` in multiple ways, but the tags file generated from it is not as useful as I'd like.

```Fortran
      subroutine FOO()
      integer myVar
      myvar = 3
      print *, "myVar is ", MyVar
      end subroutine

      program my_program
      call foo()
      end program my_program
```
```
$ ./ctags --kinds-Fortran=+L -o - test.f
FOo     test.f  /^      subroutine FOo(/;"      s
myVar   test.f  /^      integer myVar$/;"       L       subroutine:FOo  file:
my_program      test.f  /^      program my_program$/;"  p
```

Additionally Fortran symbols are appended with an underscore by most compilers, meaning that if in a multi language codebase I want to go to the definition of a Fortran function called in c as `foo_` I similarly wont find it.

I think it makes sense for ctags to lowercase symbols, and potentially make sense for it to append an underscore to subroutine and function names. However because the latter may not play well with other tooling (my usecase is with opengrok), it might make the most sense to generate two tags, with and without the underscore, probably gated behind a flag.

Looks like it should be easy enough to implement this in the Fortran parser but as I'm not all that familiar with ctags I'd appreciate some input to this problem before raising a PR.

masatake:
Close #3668.

The test input is taken from
https://gist.github.com/RaoulHC/a241fb714f191b5fc3a7c790d7b23523.


RaoulHC (approved):
LGTM, thanks so much for this!

PR #2952, opened 2021-04-11
Terraform (HCL) (*.tf): new parser

antonysouthworth-halter:
Add support for Terraform (HCL) files.

Related:
- https://github.com/preservim/tagbar/issues/757

# TODO

- [ ] add unit tests
- [ ] update documentation
  - [ ] manpages
  - [ ] codebase

PR #3684, opened 2023-03-29
Terraform: new parser

masatake:
Close #2952.
Close #3683.


optlib/terraform.ctags
```diff
@@ -0,0 +1,42 @@
...
+
+--regex-Terraform=/^resource[[:space:]]"([^"]+)"[[:space:]]"([^"]+)"/\2/r/
+--regex-Terraform=/^data[[:space:]]"([^"]+)"[[:space:]]"([^"]+)"/\2/d/
+--regex-Terraform=/^variable[[:space:]]"([^"]+)"/\1/v/
+--regex-Terraform=/^provider[[:space:]]"([^"]+)"/\1/p/
+--regex-Terraform=/^module[[:space:]]"([^"]+)"/\1/m/
+--regex-Terraform=/^output[[:space:]]"([^"]+)"/\1/o/
+
+# ???
+--regex-Terraform=/^([a-z0-9_]+)[[:space:]]*=/\1/f,TFVar/
```
masatake, line 42: @Haggus What do you think about this?

It seems that this is not a variable definition.
Definitions of variables are extracted with `v`.

Is `TFVar` a different thing from `v`?

Haggus: So in Terraform, you can define variables (`variable` keyword, or `v` in this case), but then you can use "*.tfvars" file to overwrite them.

For example, let's say you have a virtual machine instance which is named "user-service":
```hcl
variable "user_service_name" {
  type = string
}

resource "aws_instance" "user-service" {
  name = var.user_service_name
}
```

Then for different environments, you can overwrite this variable. For example with "dev.tfvars" file:
```hcl
user_service_name = "dev-user-service"
```

In this case, when searching for tags in "dev.tfvars", `user_service_name` should point the `variable "user_service_name" {` line.

I hope this makes sense. I'm new to the internals of ctags, so I any changes I can make, just let me know. I'm open to suggestions.

Haggus: I should've added a test case for this

masatake: Thank you. Interesting. As far as reading your comment .tf and .tfvars files are not the same things. I think we should introduce just another parser: TFVar parser, which defines only one kind.

```
--langdef=TFVar
--map-TFVar=.tfvars
--kinddef-TFVar=o,overwritevars,the names for overwriting variables
--regex-TFVar=/^([a-z0-9_]+)[[:space:]]*=/\1/o/
```

We we can remove `--regex-Terraform=/^([a-z0-9_]+)[[:space:]]*=/\1/f,TFVar/` from the Terraform parser.

masatake: The ideal way is to introduce the TerraformVariables parser and let the parser emit Terraform's variable kind tags with "overwritten" role.
A bit complicated but clean. I will try to implement the TerraformVariables parser. Give me some time.

Haggus:
I'm trying out your branch on a project I'm building for work, and I can't get this build of ctags to pick up my .tfvars files. I can see in `tags` that Terraform parser shows up: 
```
!_TAG_ROLE_DESCRIPTION!Terraform!variable	assigned	/assigned in Variable Definitions (.tfvars) files/
```

However, `tags` file does not contain any other references to tfvars.

Also "assigned" for variables in tfvars sounds good to me.

masatake:
As you might find, you have to add `--extras=+r` to let the TerraformVariables parser emit "variable" kind tags with the "assigned" role.

In addition, use `--fields=+r` to print the role ("assigned") for the tags.

See also https://docs.ctags.io/en/latest/output-tags.html?highlight=reference#reference-tags .

I'm working on TODO items in

https://github.com/universal-ctags/ctags/pull/3684/commits/c27b6ec099ba11530769c60e1fe6321f4db17af5

Haggus:
Interesting. Thank you for the explanation. I'm slowly discovering all the available options.

Regarding this pull request, I've tested your changes with multiple terraform projects and it seems to work correctly.

masatake:
The man page has been added.

Haggus:
@masatake Thank you for your help and finishing it!

PR #3689, opened 2023-04-02
AutoIt: Slightly optimize parsing #region

b4n:
This should marginally improve performance parsing `#region`s, and (marginally, again) improve the code.
It's unlikely to make any significant change, though.  Just something that I saw looking at the code.


masatake (approved):

masatake:
Thank you.

PR #3690, opened 2023-04-02
vstring: Avoid int -> char truncation warnings

b4n:
Either avoid or silence expected casts from int to char.

I'm not entirely sure whether vStringPut() shouldn't actually take a char argument instead of an int, as it will cast it in all cases. It is however slightly more convenient as the APIs to retrieve characters uses int only for edge cases (e.g. EOF), so the common case is a safe cast and it's more convenient not to explicit it everywhere.

### DIFF
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -244,11 +244,7 @@ extern void vStringCopyToLower (vString *const dest, const vString *const src)
 		vStringResize (dest, src->size);
 	d = dest->buffer;
 	for (i = 0  ;  i < length  ;  ++i)
-	{
-		int c = s [i];
-
-		d [i] = tolower (c);
-	}
+		d [i] = (char) tolower (s [i]);
 	d [i] = '\0';
 }
 
@@ -294,7 +290,7 @@ extern char    *vStringStrdup (const vString *const string)
 	return str;
 }
 
-static char valueToXDigit (int v)
+static int valueToXDigit (int v)
 {
 	Assert (v >= 0 && v <= 0xF);
 
--- a/main/vstring.c
+++ b/main/vstring.c
@@ -304,7 +304,7 @@ extern void vStringCatSWithEscaping (vString* b, const char *s)
 {
 	for(; *s; s++)
 	{
-		int c = *s;
+		unsigned char c = (unsigned char) *s;
 
 		/* escape control characters (incl. \t) */
 		if ((c > 0x00 && c <= 0x1F) || c == 0x7F || c == '\\')
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -102,7 +102,7 @@ CTAGS_INLINE void vStringPutNewlinAgainUnsafe (vString *const string)
 	string->buffer [string->length++] = '\n';
 }
 
-CTAGS_INLINE void vStringPut (vString *const string, const int c)
+CTAGS_INLINE void vStringPutImpl (vString *const string, const int c)
 {
 	/* verify the given character is an unsigned char value */
 	Assert (c >= 0 && c <= 0xff);
@@ -115,8 +115,12 @@ CTAGS_INLINE void vStringPut (vString *const string, const int c)
 		string->buffer [++string->length] = '\0';
 }
 
-CTAGS_INLINE bool vStringPutWithLimit (vString *const string, const int c,
-									   unsigned int maxlen)
+#define vStringPut(s, c) (sizeof(c) == sizeof(char) \
+						  ? vStringPutImpl((s), (unsigned char) (c)) \
+						  : vStringPutImpl((s), (c)))
+
+CTAGS_INLINE bool vStringPutWithLimitImpl (vString *const string, const int c,
+										   unsigned int maxlen)
 {
 	if (vStringLength (string) < maxlen || maxlen == 0)
 	{
@@ -125,5 +129,10 @@ CTAGS_INLINE bool vStringPutWithLimit (vString *const string, const int c,
 	}
 	return false;
 }
+
+#define vStringPutWithLimit(s, c, l) \
+	(sizeof(c) == sizeof(char) \
+	 ? vStringPutWithLimitImpl((s), (unsigned char) (c), (l)) \
+	 : vStringPutWithLimitImpl((s), (c), (l)))
 
 #endif  /* CTAGS_MAIN_VSTRING_H */
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -18,6 +18,7 @@
 
 #include <stdio.h>
 
+#include "debug.h"
 #include "inline.h"
 
 /*
@@ -103,6 +104,9 @@ CTAGS_INLINE void vStringPutNewlinAgainUnsafe (vString *const string)
 
 CTAGS_INLINE void vStringPut (vString *const string, const int c)
 {
+	/* verify the given character is an unsigned char value */
+	Assert (c >= 0 && c <= 0xff);
+
 	if (string->length + 1 == string->size)  /*  check for buffer overflow */
 		vStringResize (string, string->size * 2);
 
--- a/main/vstring.h
+++ b/main/vstring.h
@@ -106,7 +106,7 @@ CTAGS_INLINE void vStringPut (vString *const string, const int c)
 	if (string->length + 1 == string->size)  /*  check for buffer overflow */
 		vStringResize (string, string->size * 2);
 
-	string->buffer [string->length] = c;
+	string->buffer [string->length] = (char) c;
 	if (c != '\0')
 		string->buffer [++string->length] = '\0';
 }
--- a/dsl/optscript.c
+++ b/dsl/optscript.c
@@ -4184,7 +4184,7 @@ op__put_str (OptVM *vm, EsObject *name,
 		for (size_t i = 0; i < d; i++)
 			vStringPut (vstr, ' ');
 		if (c != 0)
-			vStringPut (vstr, (char)c);
+			vStringPut (vstr, c);
 	}
 
 	ptrArrayDeleteLastInBatch (vm->ostack, 3);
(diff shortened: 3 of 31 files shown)


main/vstring.h
```diff
@@ -106,7 +106,7 @@ CTAGS_INLINE void vStringPut (vString *const string, const int c)
 	if (string->length + 1 == string->size)  /*  check for buffer overflow */
 		vStringResize (string, string->size * 2);
 
-	string->buffer [string->length] = c;
+	string->buffer [string->length] = (char) c;
```
b4n, line 109: maybe we should make debug builds with something like
```c
assert(string->buffer [string->length] == c)
```

to catch the unlikely case where a parser passes something that would get truncated?
That could make sense in a CI build of some kind, and could possibly find subtle edge case bugs.

masatake:
Sorry for the late response. I will look into this one next.

main/vstring.h
masatake: I agree with you.
Could you add the change you proposed to this pull request?

masatake:
@b4n, sorry for being late.

main/vstring.h
masatake: The assertion fails if the 8th bit of `c` is set.

```
$ cat foo.c
#include <stdio.h>

int
main(void)
{
  int i = 0b10000000;
  char c = i;
  fprintf(stderr, "%d -> %d\n", i, c);
  return 0;
}
$ ./a.out 
128 -> -128
```

I think the assertion should be `assert (c >= 0 && c <= 0xff)`.
Putting the assertion at the beginning of the inline function may be more readable.

masatake:
@b4n Thank you. Please, delete my commit appended to your branch (if you need).

main/vstring.h
b4n: > I think the assertion should be `assert (c >= 0 && c <= 0xff)`.

I makes sense, but it actually depends on what we want the API to be. If we want it to take the return of `fgetc()` it's indeed prefect. (FTR, `fgetc()` returns the character as `unsigned int` cast to `int` (e.g. `0 <= c <= 0xff`), or `EOF` (which is `< 0`)).

However, it gets a bit trickier if we want to allow passing a `signed char` (which is the default `char` in many platforms), in cases like e.g. `int c = getcFromInputFile(); char v = c; /* assume it's in a branch where c is indeed a proper char */ vStringPut(v);`. In this case, we'd actually get negative values for `(unsigned char) c >= 0x80`, and we'd have to allow this.  But then, we couldn't detect passing EOF inadvertently, as it's usually `-1`. That check would be `assert (c >= -0x80 && c <= 0xff)`, but again, is a lot less useful.

Unfortunately, currently there are a few tests that don't pass the `Assert (c >= 0 && c <= 0xff)` check.  I'll see how hard it'd be to make them work.

Maybe we should make all characters API actually use `unsigned char`?  That would be in line with `fgetc()` and alike, and in most situations would probably be just as easy (to be verified, though), any thoughts?

> Putting the assertion at the beginning of the inline function may be more readable.

Yeah indeed, will do.

b4n:
OK, this blew widely out of proportions, but I verified and fixed all[^1] calls to `vStringPut()` to pass values that fir the `unsigned char`-as-`int` scheme.

[^1]: Unfortunately I caught that `cppGetc()` can return weird values (> 0xff) late (when a Java and a D test failed in a weird way), so I didn't check all calls that stream from this.  More checks might be needed, but putting this in a vString is exactly the kind of bugs this could catch :slightly_smiling_face:

main/vstring.c
```diff
@@ -244,11 +244,7 @@ extern void vStringCopyToLower (vString *const dest, const vString *const src)
 		vStringResize (dest, src->size);
 	d = dest->buffer;
 	for (i = 0  ;  i < length  ;  ++i)
-	{
-		int c = s [i];
-
-		d [i] = tolower (c);
-	}
+		d [i] = (char) tolower (s [i]);
```
masatake, line 247: Your change itself looks good to me.

Following the consideration that I made during reviewing is not directly related to this pull request. If I wrote incorrect things, please, let me know.

`s [i]` can be a negative if `char` on this platform means `signed char`.
In `tolower(3)` man page:

        If c is neither an unsigned char value nor EOF, the behavior of these functions is undefined. 


So `tolower (s [i])` can be a trouble. Am I wrong?

If we declare s as `unsigned char *`, and d as `unsigned char *`, 
we can avoid the trouble.

main/vstring.h
masatake: > Maybe we should make all characters API actually use unsigned char? 

I agree with you.

b4n:
Yes, i have a changeset almost ready going over most of the ctype calls to clean them up. Hopefully I'll submit it today still, no guarantee :)

Le 15 mai 2023 18:59:35 GMT+02:00, Masatake YAMATO ***@***.***> a écrit :

masatake:
The commit, [vstring: Avoid int -> char truncation warnings](https://github.com/universal-ctags/ctags/pull/3690/commits/0ef46b3a9cb1e949ae2b02fa136c78488e2cb6ac) should be merged.

The rest commits are optional.
I'm looking for a simpler way to call vStringPut; I'm afraid I forget to put (unsigned char).
vStrigPut is used so frequently.

@b4n, have you ever thought about _Generic of C11?
```
#define vStringPut (string, c) _Generic((c),				\
					char: vStringPutChar,		\
					const char: vStringPutChar,	\					
					unsigned char: vStringPutUChar,	\
					const unsigned char: vStringPutUChar, \
					int: vStringPutInt,		\
					const int: vStringPutInt,	\
					default: vStringPutInt) (string, c)



extern void vStringPutChar (vString *const string, const char c)
{
	...
	/* buffer has char[] type. So assigning directly. */		
}

extern void vStringPutUChar (vString *const string, const unsigned char c)
{
	...
	string->buffer [string->length] = (char)c; 
		
}

extern void vStringPutInt (vString *const string, const int c)
{
	Assert (0 <= c && c <= 0xff);
	...
}
```

This looks fit on our demand.

The biggest issue is we have to move to C11.
However, _Generic looks useful.
If you are not familiar with _Generic, see http://www.robertgamble.net/2012/01/c11-generic-selections.html .

I also wrote a short program to understand _Generic:
```
#include <stdio.h>

#define string(X) _Generic((X),						\
			   signed char: "signed char",			\
			   const signed char: "const signed char",	\
			   unsigned char: "unsigned char",		\
			   const unsigned char: "const unsigned char",	\
			   char: "char",				\
			   const char: "const char",			\
			   int: "int",					\
			   const int: "const int",			\
			   unsigned int: "unsigned int",		\
			   const unsigned int: "const unsigned int",	\
			   default: "unknown")

#define test_val(X) do { X v = 0; printf("val: %s -> %s\n", #X, string(v)); } while(0)
#define test_addr(X) do { X y = 0; X *v = &y; printf("addr: %s -> %s\n", #X, string(*v)); } while(0)
#define test_array0(X) do { X v[1] = {0,}; printf("array0: %s -> %s\n", #X, string(v[0])); } while(0)
#define test_array1(X) do { X v[1] = {0,}; printf("array1: %s -> %s\n", #X, string(*v)); } while(0)

#define test(X) test_val(X); test_addr(X); test_array0(X); test_array1(X); 
int
main(void)
{
  test (char);
  test (signed char);
  test (unsigned char);
  test (const char);
  test (const signed char);
  test (const unsigned char);  
  test (int);
  test (unsigned int);
  test (const int);
  test (const unsigned int);
  return 0;
}



```

b4n:
> @b4n, have you ever thought about _Generic of C11?
> […]

No, but admittedly I never took time to get into C11 much, mostly because it was too recent for project I worked on.

But indeed, it looks interesting here *if we can depend on it* (e.g. not if it's conditional, because then callers either have to be correct anyway, or some builds will be broken and some will not).

In practice it could be as simple as something like this (with `vStringPutImpl()` being the proposed int-based variant performing the `Assert`):
```c
#define vStringPut(s, c) _Generic((c), \
                                  char: vStringPutImpl((s), (unsigned char) (c)), \
                                  signed char: vStringPutImpl((s), (unsigned char) (c)), \
                                  default: vStringPutImpl((s), (c)))
```

`const` doesn't seem to play, which makes sense as it's not taking part of type conversion, but `signed` seem to for `char`, but not for `int` (probably due to the fact `char` is not defined as either signed or unsigned).

This said, I didn't test this and I don't know if it's reasonable to depend on `_Generic()`.  It's more than 10 years old, and I'd think most compilers have it for at least as long though, so it probably depends on our target users.

In our case though, I wonder if we couldn't simply do a tiny hack without C11:
```c
#define vStringPut(s, c) (sizeof(c) == sizeof(char) \
                          ? vStringPutImpl((s), (unsigned char) (c)) \
                          : vStringPutImpl((s), (c)))
```

What do you think?

masatake:
@b4n looks nice!
 I wonder why I could not find the condition `sizeof(c) == sizeof(char)` when considering _Generic.

_Generic is interesting but we don't need now.

b4n:
I updated the patchset to move the commit making `vStringPut()` more generic earlier in the chain, and removed the now unnecessary casts from the commit fixing the calls.  There was still a few calls to fix (that were incorrectly casting `int` to `char` explicitly, or `char` to `int`, both of which are incorrect and prevent the generic handling in `vStringPut()` from working), but quite a lot fewer.

> @b4n looks nice! I wonder why I could not find the condition `sizeof(c) == sizeof(char)` when considering _Generic.

Well, to be fair I didn't think about it right away, but playing a little with `_Generic()` made me realize for the case at hand there was another trick up C's sleeves :)

b4n:
Fixed, I removed too many things from the "fix calls" commit, sorry.  Should be all good now, and I reviewed them more carefully again.

masatake (approved):

masatake:
LGTM. Thank you very much.

PR #3694, opened 2023-04-07
C++: accept prototypes starting from :: operator

### LINKED ISSUE #3693: C++: fully qualified return type breaks parsing prototypes
### The name of the parser:

C++

### The command line you used to run ctags:

```
$ ctags --options=NONE --C++-kinds=+p -o- foo.cxx
```

### The content of input file:

```C++
#include <string>

::std::string bar();
::std::string baz() { return "not a prototype"; }
```

### The tags output you are not satisfied with:

```
baz	foo.cxx	/^::std::string baz() { return "not a prototype"; }$/;"	f	typeref:typename:::std::string
```

### The tags output you expect:

```
bar	foo.cxx	/^::std::string bar();$/;"	p	typeref:typename:::std::string	file:
baz	foo.cxx	/^::std::string baz() { return "not a prototype"; }$/;"	f	typeref:typename:::std::string
```

### The version of ctags:

```
$ ./ctags --version
Universal Ctags 6.0.0(c260d9e8), Copyright (C) 2015-2022 Universal Ctags Team
  Compiled: Apr  6 2023, 15:34:13
  URL: https://ctags.io/
  Output version: 0.0
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +yaml, +packcc, +optscript, +pcre2
```

### How do you get ctags binary:

Built it locally under Debian GNU/Linux Testing.

### Details:

If you remove the leading `::` from the return type (e.g. remove the FQ property) for the prototype that is missing here, it is properly recognized:

```C++
#include <string>

std::string bar();
::std::string baz() { return "not a prototype"; }
```

```shell
$ ctags --options=NONE --C++-kinds=+p -o- foo.cxx
bar	foo.cxx	/^std::string bar();$/;"	p	typeref:typename:std::string	file:
baz	foo.cxx	/^::std::string baz() { return "not a prototype"; }$/;"	f	typeref:typename:::std::string
```

This seems to happen in any context (e.g. it's the same inside a class or such, and same with a pure virtual method using a FQ return type).

masatake:
Close #3693.

The original code could not extract "bar" in
```
  ::std::string bar();
```

### DIFF
--- a/Units/parser-cxx.r/prototype-starting-from-scope-op.d/args.ctags
+++ b/Units/parser-cxx.r/prototype-starting-from-scope-op.d/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--C++-kinds=+p
--- a/Units/parser-cxx.r/prototype-starting-from-scope-op.d/expected.tags
+++ b/Units/parser-cxx.r/prototype-starting-from-scope-op.d/expected.tags
@@ -1,0 +1,2 @@
+bar	input.cpp	/^::std::string bar();$/;"	p	typeref:typename:::std::string	file:
+baz	input.cpp	/^::std::string baz() { return "not a prototype"; }$/;"	f	typeref:typename:::std::string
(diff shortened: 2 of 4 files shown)


b4n:
Sorry for not having commented earlier here, but I've been using this the past week and it worked without visible side effects on my end 👍

masatake:
@b4n  Thank you for reporting.

PR #3697, opened 2023-04-16
autoit: Drop $ from variable names

techee:
At least PHP, which also declares variables using `$`, such as
```
$x = 5;
```
outputs pure variable name without the `$` prefix and this patch unifies this behavior.

### DIFF
--- a/Units/parser-autoit.r/simple-au3.d/expected.tags
+++ b/Units/parser-autoit.r/simple-au3.d/expected.tags
@@ -1,8 +1,3 @@
-$iDoubled	input.au3	/^Local $iDoubled = 0$/;"	l	roles:def
-$iFirst	input.au3	/^    LOCAL STATIC $iFirst = 0$/;"	l	func:All functions.MyDummy	roles:def	properties:static
-$iNumber	input.au3	/^Local $iNumber = 10$/;"	l	roles:def
-$iSecond	input.au3	/^    STATIC LOCAL $iSecond = 1$/;"	l	func:All functions.MyDummy	roles:def	properties:static
-$iSomething	input.au3	/^    Local $iSomething = 42$/;"	l	func:All functions.MyDouble0	roles:def
 All functions	input.au3	/^#Region All functions$/;"	r	roles:def	end:54
 Constants.au3	input.au3	/^#include <Constants.au3>$/;"	S	roles:system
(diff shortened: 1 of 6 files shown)


techee:
Autocompletion doesn't work for variables in Geany because of this, see https://github.com/geany/geany/pull/3460. We could add a workaround if having the `$` is really the desired behavior though I think as a rule, ctags doesn't add these prefix characters to generated tags.

masatake:
Thank you. This makes sense. However, the impact is not small; there can be a client tool that expects names start from `$`.

Since version 6.0.0, I introduced per-parser versioning. This is a good time to utilize it.

```
struct sParserDefinition {
	/* defined by parser */

	/* The concept of CURRENT and AGE is taken from libtool.
	 * However, we deleted REVISION in libtool when importing
	 * the concept of versioning from libtool.
	 *
	 * If kinds, roles, fields, and/or extras have been added,
	 * removed or changed since last release, increment CURRENT.
	 * If they have been added since last release, increment AGE.
	 * If they have been removed since last release, set AGE to 0
	 *
	 * From the command line of ctags, you can see the version
	 * information with --version=<LANG>.
	 *
	 * In the tags file, !_TAGS_PARSER_VERSION!<LANG> shows the
	 * information for <LANG>.
	 */
	unsigned int    versionCurrent;
	unsigned int    versionAge;
```
Though your change doesn't match the criteria I wrote as the comment, obviously this is the
time to increment the current and set the age to 0.


Could you add the following change to your commit?
```
diff --git a/parsers/autoit.c b/parsers/autoit.c
index 8cc9b6386..24b2ee116 100644
--- a/parsers/autoit.c
+++ b/parsers/autoit.c
@@ -313,5 +313,7 @@ parserDefinition *AutoItParser (void)
 	def->extensions = extensions;
 	def->parser     = findAutoItTags;
 	def->useCork    = CORK_QUEUE;
+	def->versionCurrent= 1;
+	def->versionAge    = 0;
 	return def;
 }
```

I will add a man page for the parser to explain this change, like https://docs.ctags.io/en/latest/man/ctags-lang-kconfig.7.html, later.

masatake:
I'll take over this.

techee:
> I'll take over this.

OK, great.

masatake:
Thank you.

PR #3699, opened 2023-04-25
main: add missing const modifiers

masatake:


leleliu008:
ftp.openbsd.org is down. I think it will be up later.

masatake:
@leleliu008 Thank you for the information.

PR #3707, opened 2023-05-01
D: set template members parent name

ntrel:
From Geany.
So:
* Template has children TemplateAlias and new symbol memb.
* new template `each` has children tmethod and child.

Add templates.d test, move 5 template declarations here from simple.d test. Commented out templates AT, ET have also been moved.

I'm not sure if the build system needs updating for the new test, please let me know.


masatake:
> I'm not sure if the build system needs updating for the new test, please let me know.

No, you don't have to update any build scripts.


Non-sorted output is easier for me to compare expected.tags and the actual output (FILTERED.out). So could you add Units/parser-d.r/templates.d/args.ctags with the following content:
```
--sort=no
```
. See https://docs.ctags.io/en/latest/testing-parser.html#how-to-write-a-test-case about args.ctags file. 


You may have to update the expected.tags. misc/review script may be helpful.

```
$ echo "--sort=no" > Units/parser-d.r/templates.d/args.ctags
$ make units
...
$ bash misc/review
[jet@living]~/var/ctags-github% misc/review 
1) <a>ccept
2) <s>kip (or <n>)
3) <S>hell
4) <d>iff
5) show args.<c>tags
6) show <i>nput
7) <e>xpected tags
8) <q>uit
[1/2 [ Units/parser-d.r/templates.d ]]? a
...[Y/n]: Y
...[Y/n]: Y
```

Feel free to do "git commit --amend; git push --force".

ntrel:
> --sort=no

Great!

> You may have to update the expected.tags. misc/review script may be helpful

That's useful, thanks.

masatake (approved):

PR #3709, opened 2023-05-01
D: parse const(T), immutable, inout and shared type qualifiers

ntrel:
https://dlang.org/spec/const3.html
https://dlang.org/spec/const3.html#inout

`inout` is not working yet (missing test symbol).


masatake:
Thank you.

FYI, ".b" test case can be used to gather "FIXME" input.
See https://docs.ctags.io/en/latest/testing-parser.html#gathering-test-cases-for-known-bugs .

PR #3710, opened 2023-05-03
D: remove `overload`, not a keyword

ntrel:
Presumably this was mistakenly added instead of `override` (which is also there).
https://dlang.org/spec/lex.html#keywords

`override` only appears before a function declaration. 
https://dlang.org/spec/function.html#virtual-functions


masatake:
Thank you.

PR #3711, opened 2023-05-03
C#, D: add ref keyword

ntrel:
https://dlang.org/spec/declaration.html#ref-storage

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref


masatake:
I wonder what I should do with this.
Even without your patch for c-base.c, ctags can generate tags:
```
fref	input.d	/^ref int fref(return ref int x) { return x; }$/;"	f	file:
```
from
```
ref int fref(return ref int x) { return x; }
```
.

Regarding the test case, the change for .c has no effect.

My rules not documented:
1. If adding a keyword improves tags output, I should accept it.
2. The improvement should be explained with test cases.

This one violates the rules. 
You did two independent things: adding a keyword and adding a test case.

The test case itself is valuable. So if you send a pull request only for the part, I am happy to merge it.

PR #3716, opened 2023-05-10
D: parse template instance types

ntrel:
2 commits to parse them in either a function parameter list, or for general tokens.

https://dlang.org/spec/template.html#explicit_tmp_instantiation


masatake (approved):

masatake:
Thank you.

PR #3720, opened 2023-05-12
Verilog: support virtual interface variables

hirooih:
`virtual interface` is a variable. But the Verilog parser mistakenly declared it as an interface.
This PR fixes the issue.

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1,27 +1,27 @@
 /*
- *   Copyright (c) 2003, Darren Hiebert
- *   Copyright (c) 2017, Vitor Antunes
- *   Copyright (c) 2020, Hiroo Hayashi
+ *	 Copyright (c) 2003, Darren Hiebert
+ *	 Copyright (c) 2017, Vitor Antunes
+ *	 Copyright (c) 2020, Hiroo Hayashi
  *
- *   This source code is released for free distribution under the terms of the
- *   GNU General Public License version 2 or (at your option) any later version.
+ *	 This source code is released for free distribution under the terms of the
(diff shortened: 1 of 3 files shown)


parsers/verilog.c
```diff
@@ -132,6 +133,7 @@ typedef struct sTokenInfo {
 	bool                classScope;    /* Context is local to the current sub-context */
 	bool				parameter;	   /* parameter which can be overridden */
 	bool				hasParamList;  /* module definition has a parameter port list */
+	bool		    virtual;       /* has virtual */
```
masatake, line 136: I don't care about whitespaces in code that I don't maintain directly.
However, you can care:-).

image

If you use Emacs, M-x whitespace-mode helps you see how the sequence of whitespace is made.

masatake (approved):

parsers/verilog.c
hirooih: @masatake san,

> However, you can care:-).

Sorry.  I have been caring about it.

The indentation rules for this project are different from those I usually use.
I worked on a new PC that had not yet been configured for this project. I wrongly indented the code by hand and committed it.

I have corrected the original commit.

I was also concerned about the inconsistent use of spaces in the indentation. I took this opportunity to add a commit that fixes that as well.

Could you review it again?

masatake: About the style of spacing(?), I don't have a strong preference.
If the file's maintainer feels comfortable, that's good.


As a reference, we have .editorconfig globally used in our code base.
If you have not introduced editorconfig on your new PC, this will be a good time to introduce it.

masatake (approved):

hirooih:
Thank you for your reviewing.

PR #3725, opened 2023-05-15
colprint: Fix column width truncation

b4n:
This is very theoretical though, but pleases -Wconversion.

### DIFF
--- a/main/colprint.c
+++ b/main/colprint.c
@@ -27,7 +27,7 @@ enum colprintJustification {
 struct colprintHeaderColumn {
 	vString *value;
 	enum colprintJustification justification;
-	unsigned int maxWidth;
+	size_t maxWidth;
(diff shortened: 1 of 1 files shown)


masatake (approved):

masatake:
Should we specify -Wconversion in our build scripts?

b4n:
Not yet I think, there are way too may warnings with it for the moment, and I'm afraid of less skilled C developers trying to "fix them", as I've seen my share of wrong fixes with those, mostly because it's often a bit tricky to fix property yet easy to think a partial or incorrect fix is ok.

So I'd leave it for now for people that actually know enough to enable it and carefully check each one.

Le 15 mai 2023 20:03:38 GMT+02:00, Masatake YAMATO ***@***.***> a écrit :

PR #3726, opened 2023-05-15
erlang: Fix crash parsing directives longer than 31 characters

b4n:
Even if they might not exist, fix the code not to crash on them.

The issue was that the pointer to the vString buffer holding the directive name was cached at its value right after the vString creation, yet that buffer might get realloc()ated if it grows past the initial length of 32, possibly making the initial pointer invalid.

Fix this by simply getting the pointer to the value later on.


masatake (approved):

masatake:
Thank you.

PR #3737, opened 2023-05-29
Refactor a function in Perl 6 parser for readability

dtikhonov:
I was never proud of this particular small function I wrote. Looking at it again, I realized that I can change it to use a Duff's Device (which is cool!) and make it more readable at the same time.


dtikhonov:
```
parsers\perl6.c(273) : warning C4144: '!' : relational expression as switch expression
```

Hmm...

PR #3739, opened 2023-05-30
htable: use precalculated list of primes

dtikhonov:
This seems like it should work, @masatake -- what's a good way to test this code?  There seem to be no unit tests for this.

### DIFF
--- a/extra-cmds/utiltest.c
+++ b/extra-cmds/utiltest.c
@@ -215,6 +215,24 @@ static void test_htable_update(void)
 	hashTableDelete(htable);
 }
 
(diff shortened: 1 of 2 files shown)


masatake:
See https://github.com/universal-ctags/ctags/blob/master/extra-cmds/utiltest.c.

"make tutil" is the way to run the test.

masatake (approved):

masatake (approved):

masatake:
Thank you.

dtikhonov:
You're welcome! 🙂

PR #3740, opened 2023-05-31
main,refactor: move field null string literals to header.

dtikhonov:


masatake:
Thank you. you've been active. So I would like to write some rules I introduced during you were absent.

Your change is not about a specific parser. In that case, could you add  "main" to the prefix of the commit header? It implies the change has an impact on the whole of the ctags.

If the change is just for code transformation, and if not intended to change the command behavior, use "refactor" (as you already did). You can combine some of the with a comma character.

So the header of your commit log should be: `main,refactor: move field null string literals to header`. 

If your change is about a parser, use the name of parser as the prefix.

See also https://docs.ctags.io/en/latest/contributions.html#committing-and-submitting-a-pull-request

main/field.h
```diff
@@ -79,6 +79,9 @@ typedef enum eFieldDataType {
 	FIELDTYPE_END_MARKER = 1 << 3,
 } fieldDataType;
 
+#define FIELD_NULL_LETTER_CHAR '-'
+#define FIELD_NULL_LETTER_STRING "-"
```
masatake, line 83: field.h is visible to parsers.

fmt.c is a part of the `main` part. So let's put the definition to field_p.h.
The _p.h files are private to the `main` part.

masatake (changes requested):

masatake:
Let's squash the two commits into one and update the commit log.
IF you don't know how to do it, I can explain. Let me know.

dtikhonov:
I always squash merges to master (see, for example, #3739), but here you go.

masatake (approved):

masatake:
Thank you.

PR #3744, opened 2023-05-31
parsers,perl: allow tokens following curly bracket

dtikhonov:


masatake:
Looks fine.
About the prefix of the commit log, "Perl: " is enough. "parsers," is not needed.

masatake:
Thank you.

Issue #3748, opened 2023-06-24
Markdown: tagging a chapter unexpectedly


masatake:
foo.md:

```
  #included
```
```
$ ~/var/ctags-github/ctags --quiet --options=NONE -o - /tmp/foo.md
#included	/tmp/foo.md	/^  #included$/;"	c
```

              This comment is nothing to do with this pull request.
`#included` at line 15 is tagged as `c`hapter. This may be a bug of the original code.

_Originally posted by @masatake in https://github.com/universal-ctags/ctags/pull/3747#discussion_r1240648900_

jiangyinzuo:
I believe I can fix this bug. :smile:

masatake:
Markdown test:

  # section started from spaces

PR #3749, opened 2023-06-24
Markdown: bug fix for tagging a chapter unexpectedly

jiangyinzuo:
The previous commit had an error in handling examples such as
`   # title2`, `  #not-a-title`.

See also: #3748


masatake:
The original issue was closed with #3749.

PR #3755, opened 2023-06-27
refact github/workflows/tagging.yml

### LINKED ISSUE #3754: Last tags are identical?
At https://github.com/universal-ctags/ctags/tags I see that last tags are identical.

Screenshot 2023-06-27 at 01 39 17

leleliu008:
1. use `fetch-depth: 0` with `actions/checkout@v3`, beacuse default is `fetch-depth: 1` which means only one commit is fetched and no tags and branches are fetched.
close #3754
    
2. It looks like someone tries to use `USER_NAME` and `USER_EMAIL` to configure git `user.name` and `user.email`, this didn't take effect. so I add following:
 
```bash
git config user.name  'Universal-ctags CI'
git config user.email 'Universal-ctags@users.noreply.github.com'
```

3. I removed some unsed code in this file.

### DIFF
--- a/.github/workflows/tagging.yml
+++ b/.github/workflows/tagging.yml
@@ -1,26 +1,19 @@
-name: Create a periodical tag
+name: Create a weekly tag
 
 on:
   schedule:
     - cron:  '0 0 * * 0'
-
-env:
-  # Account for creating a tag
-  USER_NAME: "Universal-ctags CI"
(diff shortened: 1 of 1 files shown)


masatake (approved):

masatake:
Thank you. I understand why misc/git-tag-maybe.sh script has not work.

PR #3758, opened 2023-07-04
buildsys: drop the support for MacOS

masatake:
Building ctags on MacOS on Github Actions fails.  Till we find a person who can fix it, turn off the testing on the platform.


jafl:
@masatake  Is this because of the homebrew/github issue?  I use ctags exclusively on macOS, so I want to fix this!

masatake:
> @masatake Is this because of the homebrew/github issue? Was this closed because https://github.com/universal-ctags/ctags/pull/3759 fixed the issue?

Basically yes. #3759 enough mitigates the original issue.

PR #3770, opened 2023-07-07
Javascript: multiple prototype assignments

jafl:


parsers/jscript.c
```diff
@@ -922,6 +918,16 @@ static void parseRegExp (void)
 /* Read a C identifier beginning with "first_char" and places it into
  * "name".
  */
+
+static bool include_period_in_identifier = false;
```
masatake, line 922: How about making this an integer?

If this variable is used in a nested context (unexpectedly), debugging related to this variable may be hard.

If we use int for the variable, we may need the following functions.

```
static int include_period_in_identifier = 0;

static void accept_period_in_identifier()
{
    include_period_in_identifier++;
}

static void reject_period_in_identifier()
{
    include_period_in_identifier--;
}
```

Just an idea.

parsers/jscript.c
```diff
@@ -2385,6 +2392,16 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
 
 			token->scope = state->indexForName;
 
+			/*
+			 * We could make prototype_tokens a dynamic array,
+			 * but it doesn't seem worth the trouble.  How many
+			 * aliases does one need for one function?
+			 */
+			tokenInfo *const identifier_token = newToken ();
+			tokenInfo * prototype_tokens[] = { NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL };
```
masatake, line 2401: Putting just one NULL is enough. https://stackoverflow.com/questions/2589749/how-to-initialize-array-to-0-in-c 
```
tokenInfo * prototype_tokens[10] = { NULL };
```

masatake: When thinking about performance, a raw C array on the stack is a good choice.
However, to make the code simple, I think ptrArray is better; you can forget about prototype_token_count if you use ptrArray.

parsers/jscript.c
```diff
@@ -2398,10 +2415,47 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
 					skipArgumentList(method_body_token, false,
 									 vStringLength (signature) == 0 ? signature : NULL);
 				else
+				{
+					if ( isType (method_body_token, TOKEN_EQUAL_SIGN) &&
+						! isType (identifier_token, TOKEN_UNDEFINED) &&
+						prototype_token_count < sizeof(prototype_tokens)/sizeof(tokenInfo*) - 1)
```
masatake, line 2421: You can use ARRAY_SIZE macro.

parsers/jscript.c
masatake: Another idea is adding `include_period_in_identifier` field to tokenInfo.
The code can be;

```
token.include_period_in_identifier++;
readToken(token);
token.include_period_in_identifier--;
```

parsers/jscript.c
```diff
@@ -2398,10 +2415,47 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
...
+
+			tokenInfo ** temp_token = prototype_tokens;
+			while (*temp_token != NULL)
+			{
+				char * s1 = vStringValue((*temp_token)->string);
+				char * s2 = strstr(s1, ".prototype.");
+				if (s2 != NULL)
+				{
+					s1[ s2-s1 ] = 0;	// terminate before "prototype"
+					vStringSetLength((*temp_token)->string);
```
masatake, line 2444: See vStringTruncate. It may be what you want.

masatake:
Instead of adjusting a parser to vString API, you can add a function to vString API.
extra-cmds/utiltest.c is for writing a test case.

parsers/jscript.c
jafl: Switched to integer.

parsers/jscript.c
jafl: Switched to ptrArray

parsers/jscript.c
```diff
@@ -2398,10 +2415,47 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
...
+					(*temp_token)->scope = makeClassTag (*temp_token, NULL, NULL);
+
+					vString *const n = vStringNewInit (s2 + 11);	// after ".prototype."
```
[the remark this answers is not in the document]
jafl: Yes, that worked.

parsers/jscript.c
jafl: No longer relevant after switching to ptrArray

parsers/jscript.c
jafl: I don't want to lose the text after `.prototype.`

jafl:
@masatake Thanks for all the feedback!  I figured I was missing some useful stuff, e.g., ptrArray.

parsers/jscript.c
```diff
@@ -922,6 +918,28 @@ static void parseRegExp (void)
 /* Read a C identifier beginning with "first_char" and places it into
  * "name".
  */
+
+static bool include_period_in_identifier = 0;
```
masatake, line 922: bool?

masatake:
I don't know JavaScript well but...

```
$ /tmp/foo.js 
B.prototype.m5 = () => {};
$ node /tmp/foo.js 
/tmp/foo.js:1
B.prototype.m5 = () => {};
^

ReferenceError: B is not defined
    at Object.<anonymous> (/tmp/foo.js:1:1)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.16.0
```

The error tells us `B` in `B.prototype.m5 = () => {};` is not a definition but it is a reference to its definition. In such case, ctags should not make a definition tag for `B`.

This is very unlike variables in a shell script:
```
source something.sh
X=0
```
It is difficult to know whether X is a definition or reference (with left side role).

However, you want to make a tag for B. You need it to make a scope for m5.
Understandable.

The `nameChain` I introduced in 19f653c46e provides a solution.
With it, you can make reference tags for `a`, and `b` in `a.b.c` easily.
The `nameChain` helps you to make a cork Index for `b`. So you can use the index for filling the scope field of `c`.

However, we must add some new code for handling `prototype` in `nameChain` as a special case.

Here is my proposal:

* Delete B from the expected.tags.
* Pass deleteToken to the ptrArrayNew. The passed function is called to free elements in the array. So you can reduce the code for deleting tokens in the array manually.
* Don't make a definition tag for a class in a name chain. As the side effect, I deleted some calls of vString APIs.
* Skip "prototype" in makeJsRefTagsForNameChain().

Comparing a given token with "prototype" twice is a quirk. However, making parsePrototype short is more important.

Your code does strstr while iterating the tokens in the array.
However, the strstr can be done in the stage adding tokens to the array.



```
index 82724eb97..a5d13ced4 100644
--- a/Units/parser-javascript.r/js-scope.d/expected.tags
+++ b/Units/parser-javascript.r/js-scope.d/expected.tags
@@ -1,5 +1,4 @@
 A      input.js        /^function A() {$/;"    c
-B      input.js        /^B.prototype.m4 =$/;"  c
 m1     input.js        /^    m1 : function() {$/;"     m       class:A
 m2     input.js        /^    m2: function() {$/;"      m       class:A
 m3     input.js        /^A.prototype.m3 =$/;"  m       class:A
index 4a59759d8..a880feac0 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -145,6 +145,7 @@ static tokenInfo *NextToken;
 static langType Lang_js;
 
 static objPool *TokenPool = NULL;
+static void deleteTokenFn(void *token) { deleteToken(token); }
 
 #ifdef HAVE_ICONV
 static iconv_t JSUnicodeConverter = (iconv_t) -2;
@@ -448,7 +449,16 @@ static int makeJsRefTagsForNameChain (char *name_chain, const tokenInfo *token,
        char *next = strchr(name_chain, '.');
        if (next)
                *next = '\0';
-       int index = bestJSEntryInScope (scope, name);
+
+       int index;
+       if (strcmp(name, "prototype") == 0)
+       {
+               /* Let's skip this link in the chain. */
+               index = scope;
+               goto out;
+       }
+       else
+               index = bestJSEntryInScope (scope, name);
 
        if (index == CORK_NIL)
        {
@@ -478,6 +488,7 @@ static int makeJsRefTagsForNameChain (char *name_chain, const tokenInfo *token,
                        registerEntry (index);
        }
 
+ out:
        return next
                ? makeJsRefTagsForNameChain (next + 1, token, leaf_kind, index)
                : index;
@@ -919,7 +930,7 @@ static void parseRegExp (void)
  * "name".
  */
 
-static bool include_period_in_identifier = 0;
+static int include_period_in_identifier = 0;
 
 static void accept_period_in_identifier(bool incl)
 {
@@ -2426,7 +2437,7 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
                                                ! isType (identifier_token, TOKEN_UNDEFINED))
                                        {
                                                if (prototype_tokens == NULL)
-                                                       prototype_tokens = ptrArrayNew (NULL);
+                                                       prototype_tokens = ptrArrayNew (deleteTokenFn);
 
                                                tokenInfo *const save_token = newToken ();
                                                copyToken (save_token, identifier_token, true);
@@ -2450,20 +2461,8 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
                                {
                                        identifier_token = ptrArrayItem (prototype_tokens, i);
                                        char * s1 = vStringValue (identifier_token->string);
-                                       char * s2 = strstr(s1, ".prototype.");
-                                       if (s2 != NULL)
-                                       {
-                                               s1[ s2-s1 ] = 0;        // terminate before "prototype"
-                                               vStringSetLength (identifier_token->string);
-                                               identifier_token->scope = makeClassTag (identifier_token, NULL, NULL);
-
-                                               s1[ s2-s1 ] = '.';
-                                               vStringSetLength (identifier_token->string);
-                                               vStringTruncateLeading (identifier_token->string, s2+11-s1);    // after ".prototype."
-
+                                       if (strstr(s1, ".prototype."))
                                                makeJsTag (identifier_token, JSTAG_METHOD, signature, NULL);
-                                       }
-                                       deleteToken (identifier_token);
                                }
                                ptrArrayUnref (prototype_tokens);
                        }
```

parsers/jscript.c
jafl: Doh!  Fixed.

jafl:
Yes!  Much better.  However, I can't pass `deleteTokenFn` to `ptrArrayNew` because it's a macro, not a function.

masatake:
`deleteToken` is a macro but `deleteTokenFn` is a function.
```
+static void deleteTokenFn(void *token) { deleteToken(token); }
```
So you can pass it to `ptrArrayNew`.

parsers/jscript.c
```diff
@@ -2398,10 +2431,42 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
...
 									 vStringLength (signature) == 0 ? signature : NULL);
 				else
+				{
+					if ( isType (method_body_token, TOKEN_EQUAL_SIGN) &&
+						! isType (identifier_token, TOKEN_UNDEFINED))
+					{
+						if (prototype_tokens == NULL)
+							prototype_tokens = ptrArrayNew (NULL);
+
+						tokenInfo *const save_token = newToken ();
```
masatake, line 2441: I think you can move `strstr(vStringValue (identifier_token->string), ".prototype.")` here.

```C
if (is_prototype(identifier_token))
{
   tokenInfo *const save_token = newToken ();
   copyToken (save_token, identifier_token, true);
   ptrArrayAdd (prototype_tokens, save_token);
}
identifier_token->type = TOKEN_UNDEFINED;
```

So you can reduce the number of token allocations.

parsers/jscript.c
```diff
@@ -2398,10 +2431,42 @@ static bool parsePrototype (tokenInfo *const name, tokenInfo *const token, state
...
+			deleteToken (identifier_token);
+			accept_period_in_identifier(false);
 
 			int index = makeJsTag (token, JSTAG_METHOD, signature, NULL);
+
+			if (prototype_tokens != NULL)
+			{
+				for (int i=0; i<ptrArrayCount (prototype_tokens); i++)
+				{
+					identifier_token = ptrArrayItem (prototype_tokens, i);
```
masatake, line 2461: If you check ".prototype." in the gathering stage, you can make this making-tag stage simpler.

masatake:
Thank you for updating. After re-reading your change, I found we can reuse your original idea of the string manipulation earlier.

```C
					if ( isType (method_body_token, TOKEN_EQUAL_SIGN) &&
						! isType (identifier_token, TOKEN_UNDEFINED))
					{
						char *str = vStringValue (identifier_token->string);
						char *needle = strstr(str, ".prototype.")

						if (prototype_tokens == NULL)
							prototype_tokens = ptrArrayNew (deleteTokenFn);

						tokenInfo *const save_token = newToken ();
						copyToken (save_token, identifier_token, true);
						/* Keep \1 of #/(.*)\.prototype\.(.*)/. */
						vStringTruncate (save_token->strng, needle - str);
						vStringPut (save_token->strng, '.');
						/* Append \2 of #/(.*)\.prototype\.(.*)/. */						
						vStringNCatS (save_token->strng, needle + 11);
						ptrArrayAdd (prototype_tokens, save_token);
						identifier_token->type = TOKEN_UNDEFINED;
					}

```

With this code, we no longer need the change for makeJsRefTagsForNameChain() I proposed.
We can reduce redundant string comparisons. However, the new optimized code makes parsePrototype() larger. Please, choose the combination (modifying makeJsRefTagsForNameChain and modifying parsePrototype) you like.

jafl:
Definitely better to do the string processing in `parsePrototype`.  I found an even simpler way to do it.

masatake (approved):

masatake:
Thank you.

PR #3772, opened 2023-07-14
Update matlab.c

portalgun:
Update function regexps for 'get\.*' and 'set\.*' methods. For example:
     function get.myproperty(obj)
     ...
Otherwise function tag returned is just 'get' instead of 'get.myproperty'.

### DIFF
--- a/parsers/matlab.c
+++ b/parsers/matlab.c
@@ -19,13 +19,13 @@
 
 static tagRegexTable matlabTagRegexTable [] = {
 	/* function [x,y,z] = asdf */
-	{ "^[ \t]*function[ \t]*\\[.*\\][ \t]*=[ \t]*([a-zA-Z0-9_]+)",
+	{ "^[ \t]*function[ \t]*\\[.*\\][ \t]*=[ \t]*([.a-zA-Z0-9_]+)",
 	  "\\1", "f,function", NULL},
 	/* function x = asdf */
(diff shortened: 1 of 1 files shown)


parsers/matlab.c
```diff
@@ -1,53 +1,53 @@
...
+	static selectLanguage selectors [] = { selectByObjectiveCAndMatLabKeywords,
+					       NULL };
+	parserDefinition* const def = parserNew ("MatLab");
+	def->extensions = extensions;
+	def->tagRegexTable = matlabTagRegexTable;
+	def->tagRegexCount = ARRAY_SIZE (matlabTagRegexTable);
+	def->method     = METHOD_NOT_CRAFTED|METHOD_REGEX;
+	def->selectLanguage = selectors;
+	return def;
+}
```
masatake, line 53: I'm surprised that you changed all lines in the file.
I think you may have to change just lines 22, 25, and 28.

We need a test case.
See https://docs.ctags.io/en/latest/testing-parser.html .

masatake (changes requested):

masatake:
BTW, you use your master branch for this pull request.
I strongly recommend you use a topic branch.

masatake:
Could you introduce documentation about the 'get.' and 'set.' ?

portalgun:
> BTW, you use your master branch for this pull request. I strongly recommend you use a topic branch.

Ok. I've created another pull request. See #3773

PR #3773, opened 2023-07-14
MatLab: return appropriate tags for set get methods + unit 

portalgun:
Update function regexps for 'get.' and 'set.' methods. For example:
function get.myproperty(obj)
...
Otherwise function tag returned is just 'get' instead of 'get.myproperty'.

Replaces #3772


masatake:
I'll merge this. Let me rearrange some commits before merging.

masatake:
After thinking, I dropped the man page.

The VERSIONS section is for describing the change adding a kind, a role, a field, or an extra.
The original change in the pull request doesn't fit VERSIONS.

masatake:
Thank you!

PR #3779, opened 2023-07-22
main: reserve errno in external sorting

masatake:
This change fixes that ctags reports the following strange error message:
    
      $ /usr/bin/ctags --quiet --options=NONE -o - main/*.c | no-such-command
      zsh: no-such-command: command not found...
      ctags: cannot sort tag file : Success

Success?
    
With this change, ctags reports:
    
      ./ctags --quiet --options=NONE -o - main/*.c | no-such-command
      zsh: no-such-command: command not found...
      ctags: cannot sort tag file

### DIFF
--- a/main/sort.c
+++ b/main/sort.c
@@ -12,6 +12,7 @@
 */
 #include "general.h"  /* must always come first */
 
+#include <errno.h>
 #if defined (HAVE_IO_H)
 # include <io.h>
 #endif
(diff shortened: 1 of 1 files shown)


leleliu008:
why use system(3) to execute the external command? for portable reason? As we all known system(3) is dangerous and have limitations.

References:
https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/system.html
https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152177

masatake:
exuberant-ctags used system(3).

PR #3787, opened 2023-07-29
Verilog,unit: lint-check unit tests

hirooih:
- fix issues reported by [svlint](https://github.com/dalance/svlint)

- Intended errors still remain.


masatake:
If you are interested in keeping the input files for your language valid, you can use a validator. See https://docs.ctags.io/en/latest/testing-parser.html?highlight=validator#validator-file

PR #3790, opened 2023-08-02
update Mac installation instructions

euclio:
There's an official Homebrew formula now, no need to use a custom tap.


masatake:
@KazuakiM Do you have more things to do in universal-ctags/homebrew-universal-ctags repository?

masatake:
@euclio Thank you.

PR #3791, opened 2023-08-05
Verilog validator

hirooih:
@masatake san,

From #3787

> If you are interested in keeping the input files for your language valid, you can use a validator.

Please review the validator functionality to ensure that it is being used as you intend.


misc/validators/validator-svlint
```diff
@@ -0,0 +1,35 @@
+# -*- sh -*-
+# validator-svlint - validating Verilog/SystemVerilog input files with svlint
+#   https://github.com/dalance/svlint
+#
+#  Copyright (c) 2023 Masatake YAMATO
```
masatake, line 5: I think your name should be here.

masatake (approved):
Other than the copyright notice, it looks good for me.
I will integrate this into the test running on CircleCI::fedora38.

hirooih:
@masatake san,

Thank you for your review.

> I think your name should be here.

I also thought about this if I should fix this.
I just changed the command name in your script. So I did not intentionally change the copyright.

I fixed it and merged this PR.

PR #3793, opened 2023-08-12
input-validate: add CATEGORIES make variable

hirooih:
This PR lets *validate-input* make target support the CATEGORIES variable as *units* make target does.

Some minor fixes  of input validation are also included.

### DIFF
--- a/docs/testing-parser.rst
+++ b/docs/testing-parser.rst
@@ -477,7 +477,7 @@ Here is an example validating an input file for JSON.
 
   $ make validate-input VALIDATORS=jq
   ...
-  Category: ROOT
+  Category: parser-json.r
   ------------------------------------------------------------
   simple-json.d/input.json with jq                                 valid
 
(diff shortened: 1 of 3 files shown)


masatake (approved):

hirooih:
@masatake san,

Thank you for your approval.

BTW I have a question after reading the section, [Input validation for Units](https://docs.ctags.io/en/latest/testing-parser.html#toc-entry-17).

> With VALIDATORS variable passed via command-line, you can specify validators to run. Multiple validators can be specified using a comma-separated list. If you don’t give VALIDATORS, the make target tries to use all available validators.

What is a use-case of the VALIDATORS variable?
`make validate-input` works fine. On what situation this variable is needed?

masatake:
@hirooih As you implemented, CATEGORIES may cover most of all use cases.
VALIDATORS is mainly designed for C++.
Some languages like C++ have generations of standards like C++-11 and C++-23.
Even if an input is valid as C++11, the input can be invalid as C++-23.
For validating only C++-23 input but not C++-11 ([or vice versa]), the VALIDATORS may help you.

PR #3795, opened 2023-08-13
[WIP] V: new parser

edam:
New parser for V (http://vlang.io/).

Includes fairly extensive unit tests and parses entire V compiler/runtime sources (written in V).


parsers/v.c
```diff
@@ -0,0 +1,2892 @@
...
+
+// for: list? [';' expr]* block
+static void parseFor (tokenInfo *const token, int scope)
+{
+	Assert (isKeyword (token, KEYWORD_for, KEYWORD_Sfor));
+	PARSER_PROLOGUE ("for");
+
+	vString *access = NULL;
+	readToken (token);
+	skipAccessAndReadToken (token, &access);
```
masatake, line 1691: The buffer assigned to &acccess can be never freed.

masatake: Install valgrind, and run `make units LANGUAGES=V VG=1`
```
  #valgrind-error:                        3
	parser-v.r/v-import
	parser-v.r/v-match
	parser-v.r/v-statements

```

See Units/parser-v.r/v-*.d/VALGRIND.tmp for an example.

parsers/v.c
```diff
@@ -0,0 +1,2892 @@
...
+		}
+		deleteToken (identToken);
+		vStringDelete (idents);
+		if (!access)
+			vStringDelete (accesses);
+	}
+	else
+		parseExpression (token, scope, NULL);
+
+    while (true)
```
masatake, line 2229: Strange depth of indentation.

image

masatake (changes requested):

masatake:
```
$ cat /tmp/foo.v        
\"
$ ./ctags --language-force=V /tmp/foo.v
ctags: parsers/v.c:588: readTokenFull: Assertion `LastTokenType != TOKEN_NONE' failed.
ctags: parsers/v.c:588: parsing /tmp/foo.v:1 as V
zsh: IOT instruction (core dumped)  ./ctags --language-force=V /tmp/foo.v
```

masatake:
noise, fuzz, chop, and slap targets help you break your parser.
See https://docs.ctags.io/en/latest/testing-parser.html .

parsers/v.c
```diff
@@ -0,0 +1,2892 @@
...
+// match: 'match' access* fqident cont '{' [
+//     [[vtype | expr] [',' [vtype | expr]]* | 'else'] block
+// ]* '}'
+static void parseMatch (tokenInfo *const token, int scope)
+{
+	Assert (isKeyword (token, KEYWORD_match));
+	PARSER_PROLOGUE ("match");
+
+	vString *access = NULL;
+	readToken (token);
```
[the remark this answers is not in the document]
edam: My very sincere apologies! I use emacs and I have no idea how it did that. Fixed.

parsers/v.c
```diff
@@ -0,0 +1,2892 @@
...
+		tokenInfo *moduleToken = dupToken (token);
+		vString *moduleName = vStringNew ();
+		parseFQIdent (token, moduleName);
+		readToken (token);
+		if (isKeyword (token, KEYWORD_as))
+		{
+			readToken (token);
+			if (expectToken (token, TOKEN_IDENT))
+			{
+				moduleName = token->string;
```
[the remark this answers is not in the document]
edam: Fixed.

parsers/v.c
```diff
@@ -0,0 +1,2892 @@
...
+//     [[vtype | expr] [',' [vtype | expr]]* | 'else'] block
+// ]* '}'
+static void parseMatch (tokenInfo *const token, int scope)
+{
+	Assert (isKeyword (token, KEYWORD_match));
+	PARSER_PROLOGUE ("match");
+
+	vString *access = NULL;
+	readToken (token);
+    skipAccessAndReadToken (token, &access);
```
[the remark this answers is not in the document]
edam: Fixed.

parsers/v.c
edam: Fixed. Thanks for running valgrind checks! 😊  I did try to run it on my Mac, but it didn't work...

parsers/v.c
edam: Yeah, `whitespace-mode` shows something interesting. If I hit TAB at the start of this line:
Screenshot 2023-08-15 at 11 22 34 pm
Emacs converts the tabs before `TOKEN_CLOSE_CURLY` to spaces! I presume this is to stop the indentation depending on tab-width. I've converted all spaces to tabs now though anyway.

edam:
> ```
> $ cat /tmp/foo.v        
> \"
> $ ./ctags --language-force=V /tmp/foo.v
> ctags: parsers/v.c:588: readTokenFull: Assertion `LastTokenType != TOKEN_NONE' failed.
> ctags: parsers/v.c:588: parsing /tmp/foo.v:1 as V
> zsh: IOT instruction (core dumped)  ./ctags --language-force=V /tmp/foo.v
> ```

Fixed.

edam:
> noise, fuzz, chop, and slap targets help you break your parser. See https://docs.ctags.io/en/latest/testing-parser.html .

Will take a look tomorrow. Thanks!

edam:
Worse than that... I'll take a look this eve.
Screenshot 2023-08-16 at 8 42 05 am

edam:
Update: I have windows and msvc 64-bit builds working and have reproduced the problem... 😅

edam:
`cl : Command line warning D9002 : ignoring unknown option '/experimental:preprocessor'`

I think the compiler in AppVeyor is too old @masatake! 😆 This should fix the issue. (Actually, in later compilers, the option is no longer experimental: `/Zc:preprocessor`). The build issue is that the preprocessor incorrectly expands `__VA_ARGS__` in macros when those args have come from `...`. I can reproduced it on windows (in a VM) and this fixes it with a later MSVC compiler.

Can we update the AppVeyor compiler? Or is that not really possible?

masatake:
> Can we update the AppVeyor compiler? Or is that not really possible?

Do you know how to do it? If you know, could you make a pull request?
I don't know Windows at all. So @k-takata may evaluate the change.

@k-takata could you help us?

leleliu008:
@masatake 

It seems that @k-takata has no time to update it.

I can help you. I wonder why not migrate to GitHub Actions?

`win32/appveyor.bat` do building for `msbuild` `nmake` `msys2` `mingw` `cygwin` respectively. 

`msys2` `mingw` `cygwin` already have tested with GitHub Actions.  

appveyor could be entirely superseded by GitHub Actions  by adding GitHub Actions workflow for `msbuild` and `nmake` https://github.com/leleliu008/ctags-manually-build/blob/master/.github/workflows/manually-build-for-windows-nmake.yml

`GitHub Actions` only supports `Visual Studio Enterprise 2022` and `Visual Studio Enterprise 2019`, do not support older versions, but I think they are enough.

https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md

`appveyor` supports even `Visual Studio 2008`, but I think they are out of touch with reality.

 https://www.appveyor.com/docs/windows-images-software/

masatake:
@leleliu008 Thank you.

I don't care about the platform we will use for testing ctags on Windows.
If things work fine, it's o.k. We can run the test on Appveyor and GitHub Actions simultaneously. So using Appveyor may be a good idea. This is the reason I'm using CircleCI for testing ctags on Fedora.

About the version of the build tool, I have no comment. We need a tool supporting C99.

Units/parser-v.r/torture.d/expected.tags
```diff
@@ -0,0 +1,4 @@
+main	input.v	/^fn main() {$/;"	f	line:1	typeref:typename:	signature:()	roles:def	end:17
+a	input.v	/^	a := 1$/;"	v	line:3	fn:main	roles:def
+C.SYSTEM_INFO	input-1.v	/^struct C.SYSTEM_INFO {$/;"	s	line:1	roles:def	end:2
+C.foo	input-1.v	/^fn C.foo(bar string) i64$/;"	f	line:3	typeref:typename:i64	signature:(bar string)	roles:def
```
masatake, line 4: Though we don't have consistent policy, but in some important parsers in ctags choose `foo` instead of `C.foo` when tagging this kind of item.

Let me show an example from C++:

```
$ cat /tmp/Input.cpp 
class C {
	void f(void);
};

void C::f(void)
{
	return;
}
$ ctags -o - Input.cpp
C	Input.cpp	/^class C {$/;"	c	file:
f	Input.cpp	/^void C::f(void)$/;"	f	class:C	typeref:typename:void

$ ctags -o - --extras=+q Input.cpp
C	Input.cpp	/^class C {$/;"	c	file:
C::f	Input.cpp	/^void C::f(void)$/;"	f	class:C	typeref:typename:void
f	Input.cpp	/^void C::f(void)$/;"	f	class:C	typeref:typename:void
```

The parser doesn't make a tag for `C::f` by default.
`C::f` is tagged only when `--extras=+q` option is passed.

Units/parser-v.r/v-enum.d/expected.tags
```diff
@@ -0,0 +1,15 @@
+Foo	input.v	/^enum Foo {$/;"	g	line:3	roles:def	end:10
+.aaa	input.v	/^    aaa = 5$/;"	e	line:4	enum:Foo	roles:def
```
masatake, line 2: I wonder why these start from `.`.

Units/parser-v.r/torture.d/expected.tags
edam: @masatake Thanks! I'll look in to making making the V parser respect the `q` extra. 👍  For some modules/objects, I may even be able to work out the type of the qualifying symbol from CORK, I think?

Units/parser-v.r/v-enum.d/expected.tags
edam: Because enum values in V always start with a dot when they are referred to. [See here](https://github.com/vlang/v/blob/master/doc/docs.md#enums). Do you think they should be tagged without the dot then? (I'll defer to your wisdom -- I have less experience with how editors/external software will use the tags!)

Units/parser-v.r/torture.d/expected.tags
masatake: I will explain in a language independent way.

Consider a parser reads "C.f()" in input.foo.

We recognize this as "a function f is defined in C".
How about ctags?

<D>If C is defined in the input.foo like: "class C {...}", ctags may know the corkId for C.
When making a tag for `f`, put the cork id of C into the scope field of `f`. See `
scopeIndex` in sTagEntryInfo in main/entry.h. You may know this.


<R> What I would like to write is about the case when C is not defined in the input.foo.
(Though I don't know there is such a case in V language.)

In such a case, make a reference tag for C when parsing `C.f`, and get a cork ID for `C`.
Once, you get a cork ID for C, what you have to do is the same as <D>.
In some languages, we can know the kind for C in "C.f()". e.g. namespace, class, struct, etc. For the case, you reserved "Y":
```
{true, 'Y', "unknown", "unknown, imported symbols",
	 .referenceOnly = false, ATTACH_ROLES (VUnknownRoles)},
```
.

I used this technique in parsers/python.c and parsers/jscript.c.
See also 19f653c46e and https://docs.ctags.io/en/latest/man/ctags-lang-python.7.html.

Units/parser-v.r/v-enum.d/expected.tags
masatake: I remembered this af4901def63546ae5ca4d48dae74bbf8d5fa2682 when I read your reply.

Language processors of Fortran emits `foo_` for the name `foo`. `foo` appears in Fortran input file. The language processor emits `foo_` in the object file. `foo_` can be used in the other language like C.

If you think `.aaa` is useful, you can do the same as the Fortran parser.

1. Make a tag for `aaa` by default.
2. Define V specific extra like... `handyEnum` (just an idea)
3. Make a tag for `.aaa` in addition to `aaa` when `handyEnum` extra is enabled.
4. Turn on the `handyEnum` extra by default if you want.

See also #3668.

Units/parser-v.r/torture.d/expected.tags
edam: I think this is fixed now.

`torture.d/expected.tags`
```
main	input.v	/^fn main() {$/;"	f	line:1	typeref:typename:	signature:()	roles:def	end:17
a	input.v	/^	a := 1$/;"	v	line:3	fn:main	roles:def
SYSTEM_INFO	input-1.v	/^struct C.SYSTEM_INFO {$/;"	s	line:1	external:C	roles:def	end:2
foo	input-1.v	/^fn C.foo(bar string) i64$/;"	f	line:3	external:C	typeref:typename:i64	signature:(bar string)	roles:def
```

Units/parser-v.r/v-enum.d/expected.tags
edam: Fixed.

masatake:
(I'm slowly reviewing this).

masatake:
@edam for resolving the conflicts, can you forth-update your branch? or could you solve the conflicts?

edam:
Tests still failing because VS<2019 (on windows) has a non-conformant preprocessor. VS2019 fixes this.

masatake:
input.v:
```
const test_init_func = {;
```

`ctags input.v` doesn't stop.

masatake:
Could you rebase your branch on the latest master branch?

masatake:
```
[yamato@dev64]~/var/ctags-github% cat > /tmp/input.v <<EOF
heredoc> c'
heredoc> EOF
[yamato@dev64]~/var/ctags-github% cat /tmp/input.v 
c'
[yamato@dev64]~/var/ctags-github% ./ctags -o - /tmp/input.v 
^C
```

For the input, the V parser never stops.

edam:
@masatake rebased, and lockup fixed. You are truly amazing at inf. loops! 👍
CI tests passing!

masatake:
``
$ cat /tmp/input.v                                                                                                                        
fn f()x.                                                                                                                                                                  
4 ./ctags -o - /tmp/input.v                                                                                                               
f       /tmp/input.v    /^fn f()x.$/;"  f       typeref:typename:                                                                                                         
free(): invalid next size (fast)                                                                                                                                          
zsh: IOT instruction (core dumped)  ./ctags -o - /tmp/input.v                                                                                                             
$ valgrind ./ctags -o - /tmp/input.v                                                                                                      
==1336387== Memcheck, a memory error detector                                                                                                                             
==1336387== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.                                                                                               
==1336387== Using Valgrind-3.21.0 and LibVEX; rerun with -h for copyright info                                                                                            
==1336387== Command: ./ctags -o - /tmp/input.v                                                                                                                            
==1336387==                                                                                                                                                               
==1336387== Invalid write of size 8                                                                                                                                       
==1336387==    at 0x47F492: unreadTokenFull (v.c:885)                                                                                                                     
==1336387==    by 0x481559: parseFunction (v.c:1612)                                                                                                                      
==1336387==    by 0x485496: parseFile (v.c:2927)         
...                                                                                                                 
```

File: parsers/v.c [hunk 8]
```diff
@@ -0,0 +1,3011 @@
...
+			{
+				int end = getcFromInputFile ();
+				captureChar (capture, token, end);
+				do
+				{
+					c = getcFromInputFile ();
+					captureChar (capture, token, c);
+				}
+				while (c != end && c != EOF);
+				vStringDelete (token->string);
```
masatake, line 837: Of course, this is correct C code. However, I don't like this. Allocating and Deleting token->string doesn't follow the Ctags' code pattern.

In many MODERN parsers, the life cycles of `token` and `token->string` are the same.
This is not a must. However, this makes the life of people (who?) maintaining multiple parsers easier. 

"Keeping a token as-is but replacing its `vstring`" looks tricky to me.

Please compare clearPoolToken() of v.c and clearPoolToken() in python.c.
v.c uses vStringDelete. However, pythons.c v.c uses vStringClear.

edam: The V parser does not have a string for many tokens, so I defaulted it to NULL. Happy to change though.

masatake:
```
$ cat /tmp/input.v         
fn f()x.!
$ ./ctags -o - /tmp/input.v
f       /tmp/input.v    /^fn f()x.!$/;" f       typeref:typename:
munmap_chunk(): invalid pointer
zsh: IOT instruction (core dumped)  ./ctags -o - /tmp/input.v
```

masatake:
`make noise LANGUAGES=V` is passed. I'm reviewing the parser slowly. Give me time.

masatake:
I'm thinking about 
```
struct C.SYSTEM_INFO {
}
fn C.foo(bar string) i64
```
Give me more time.

masatake:
(I myself have the will to implement What I wrote here.)

```
fn C.f(voidptr)

struct C.s {
}
```

I'm thinking again and again about what f is and what s is.

I think `f` is a name for C language object called function.
It is declared in a foreign function in a language (v) other than C.

I think `s` is a name for C language object called struct.
It is declared in a foreign struct in a language (V) other than C.

So I expect the following tags output:
```
f	foo.v	/^fn C.f(voidptr)$/;"	kind:function	line:1	language:C	roles:foreigndel	end:1
s	foo.v	/^struct C.s {$/;"	kind:struct	line:3	language:C	roles:foreigndel	end:4
```

If you see f in your editor like (pseudo C code):

```
      x = C.f(NULL)
            ^
```

`^` represents the cursor of your editor.
Press `[Jump-TO-Definition]`. Where does your **IDEAL** editor bring you to?

A) foo.v
```
fn C.f(voidptr)
     ^
```

B) foo.c
```
void f (void) { ...
     ^
```


I think `[Jump-TO-Definition]` should bring the user to B).
A) is the place that `[Jump-TO-Declaration]` should bring the user to.

The idea may be applicable to `JS.`.

edam:
The C.* and JS.* declarations are used to link against C symbols, (re)define C data structures and use JS functions (when C is transpiled into JS).

A) is a V declaration of an external symbol.
B) is the actual C function.

But B) could be anywhere! There is no way of knowing whether it is something you can jump to or not. If it is in a library (very likely) then you can probably only jump to a C declaration in a C header (which is not better). Often, you may not be able to jump to anything at all.

I will defer to your preference, of course. :) I'm just making sure that you're sure this is how it should work. :)

> The idea may be applicable to `JS.`.

I think so, yes.

masatake:
> But B) could be anywhere! 

The declaration itself is tagged like this:

```
f	foo.v	/^fn C.f(voidptr)$/;"	kind:function	line:1	language:C	roles:foreigndel	end:1
```

So if a tags file records the position of the definition of `f` in a C source file, a client tool like vim brings the user there.

However, if the position for the definition is not recorded in the tags file, the client tool may bring the user to a line matching `/^fn C.f(voidptr)$/` in foo.v as a fallback.

Univeral Ctags provides such flexibility to client tools:-)

edam:
Ahh, ok. I understand. So if the user has C source for the function/struct AS WELL AS the V source, then the tag generated from the V sources should reference the C symbols, ideally. I understand. That's nice.

So, for:
```v
struct C.bar {}
fn C.foo()
```
would you want to see BOTH the C and V tags?

E.g.
```
bar	foo.v	/^struct C.bar {$/;"	s	extern:C
foo	foo.v	/^fn C.foo()$/;"	f	extern:C	typeref:typename:
bar	foo.v	/^struct C.bar {$/;"	kind:struct	line:1	language:C	roles:foreigndel	end:1
foo	foo.v	/^fn C.foo()$/;"	kind:function	line:2	language:C	roles:foreigndel	end:2
```

masatake:
> bar	test.v	/^struct C.bar {$/;"	kind:struct	line:1	language:V	roles:def	extern:C	end:1

I think more about this. If we make a tag for `C`, things become consistent.

```
C	test.v	/^struct C.bar {$/;"	kind:module	line:1	language:V	roles:foreignlang	end:1
bar	test.v	/^struct C.bar {$/;"	kind:struct	line:1	language:V	roles:def	module:C	end:1
C	test.v	/^fn C.foo()$/;"	kind:module	line:2	language:V	roles:foreignlang	end:2
foo	test.v	/^fn C.foo()$/;"	kind:fn	line:2	language:V	roles:def	module:C	typeref:typename:
```

`C` is not defined here. So it is tagged as a reference tag having a role named "foreignlang".
Another candidate for the name is "foreign" or "lang".

The scope fields for `foo` and `bar` look better: `module:C`.

You may think, a tags file can have too many tag entries for `C` and `JS`.
It doesn't become an issue; we can make 'enabled = false' as the default value for the `foreignlang` role.

masatake:
Sorry for making you wait. I'm struggling with the bug of libxml2 now.
I would like to release 6.1 in this year. Till then I will merge this with minor adjustments related to the naming.

edam:
> Sorry for making you wait.

No worries!

> I would like to release 6.1 in this year.

Ok, I have some updates and I would like to make the change you proposed above. How long do I have? Is it better to merge this and make changes in next version?

masatake:
@edam, see https://github.com/edam/ctags/pull/3.

What I did can apply to `JS.` Like `C.`, is `JS.` part of V language, or is it just a convention?

parsers/v.c
```diff
@@ -0,0 +1,2993 @@
...
+		copyToken (token, PS->replays[PS->numReplays].token);
+		deleteToken (PS->replays[PS->numReplays].token);
+		vDebugParserPrintf (
+			"˅%s%s%s%s%s ", token->id > 1? "[" : "", tokenNames[token->type],
+			!vStringIsEmpty (token->string)? ":" : "",
+			vStringValue (token->string), token->id > 1? "]" : "");
+		if (capture)
+		{
+			Assert (PS->replays[PS->numReplays].capture);
+			vStringCat (capture, PS->replays[PS->numReplays].capture);
```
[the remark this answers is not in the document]
edam: Concatenate. :)

```c
#define vStringAccumulate(a, s) \
	do{ if (a && s) { vStringCat (a, s); vStringClear (s); } }while(0)
```

I used it so often I made a macro for it. Would you prefer I did this another way?

masatake:
@edam I will start working on releasing 6.1.0 on December 26th. It is the soft deadline.

My primary concern is the parser interface: the quadruplet of kinds, roles, fields, and extras. I don't want to change them after the release.

masatake:
When merging, I will squash all the changes into one or some.

edam:
I am working on adding JS functions as `foreigndecl` kinds, as well.

Units/parser-v.r/torture.d/expected.tags
```diff
@@ -2,8 +2,10 @@ main	input.v	/^fn main() {$/;"	f	line:1	language:V	typeref:typename:	signature:(
 a	input.v	/^	a := 1$/;"	v	line:3	language:V	fn:main	roles:def
 C	input-1.v	/^struct C.SYSTEM_INFO {$/;"	p	line:1	language:V	roles:foreignlang	extras:reference
 SYSTEM_INFO	input-1.v	/^struct C.SYSTEM_INFO {$/;"	s	line:1	language:V	module:C	roles:def	end:2
+SYSTEM_INFO	input-1.v	/^struct C.SYSTEM_INFO {$/;"	s	line:1	language:C	roles:foreigndecl	extras:reference
 C	input-1.v	/^fn C.foo(bar string) i64$/;"	p	line:3	language:V	roles:foreignlang	extras:reference
 foo	input-1.v	/^fn C.foo(bar string) i64$/;"	f	line:3	language:V	module:C	typeref:typename:i64	signature:(bar string)	roles:def
+C.foo	input-1.v	/^fn C.foo(bar string) i64$/;"	f	line:3	language:C	roles:foreigndecl	extras:reference
```
masatake, line 8: I made a mistake. This should be "foo". "C." is not needed. I'll make a pull request.

masatake:
@edam I added some commits. Please, see https://github.com/edam/ctags/pull/4 .

masatake:
Could you rebase on the latest master branch?

Units/parser-v.r/torture.d/expected.tags
edam: I caught this in the JS changes. :)

masatake:
image

image

I introduced these.

masatake:
gcc accepts `int f[] = {};' with -std=c99.

masatake:
I updated https://github.com/edam/ctags/pull/4 .

parsers/v.c
```diff
@@ -0,0 +1,2993 @@
...
+
+static void clearPoolToken (void *data)
+{
+	tokenInfo *token = data;
+
+	token->type = TOKEN_NONE;
+	token->keyword = KEYWORD_NONE;
+	vStringClear (token->string);
+	token->fullyQualified = false;
+	token->lineNumber = getInputLineNumber ();
```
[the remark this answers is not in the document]
edam: No, removed.

parsers/v.c
```diff
@@ -0,0 +1,2993 @@
...
+	token->filePosition = getInputFilePosition ();
```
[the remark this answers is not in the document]
edam: No, removed.

masatake:
I wonder we can utilize /Ze described in https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2466?view=msvc-170 .

masatake:
@edam I will squash some commits after fixing the zero-length array issue (See https://github.com/universal-ctags/ctags/pull/3870).

I will split the commits into three groups:

* Group 1: C related changes
* Group 2: JS related changes
* Group 3: only 1 change, the V parser

masatake:
I'll merge this via #3871.
#3871 includes up to [3c70f08](https://github.com/universal-ctags/ctags/pull/3795/commits/3c70f08a6ac57d2a9bf8f25876c166a5f73dfa42).

To avoid confusion, I will close this.

PR #3797, opened 2023-08-16
[TEMP] V parser+debug for msvc/base

masatake:
I reduced the environments for running test case.
We can use this as the base for debugging.


edam:
@masatake the issue with MSVC turned out to be the preprocessor incorrectly expanding `__VA_ARGS__` in `...` in macro definitions. I think I fixed it in the other PR.

PR #3801, opened 2023-08-17
Appveyor: run nmake with defining DEBUG

masatake:
- [ ] msbuild
- [x] msvc
- [ ] mingw


masatake:
I should verify the other runtime?/compiler combinations.
However, this change may be needed in the development of the V parser, so I merge this.

PR #3804, opened 2023-08-19
C-based:  put markers to signature

masatake:
C-based: put literal markers (' and ") to signatures

If a default-string literal or a default-char literal is included in a signature, the marker of the literal should be recorded as part of the signature.

The original code didn't record the markers.


jafl (approved):

masatake:
I didn't consider vera.c. Maybe we must update it, too.

PR #3806, opened 2023-08-21
GitHub Actions: add testing for Windows Vistual Studio / nmake

leleliu008:


masatake:
@k-takata Could you review this?

k-takata (approved):
I think it is okay for the first step.

For the next step, we should run `misc/units.py` on msys2 as we already to in AppVeyor.

I'm thinking if we can use `vcpkg` in the future to set up external libraries like iconv.

leleliu008:
`libyaml` `libxml2` `jansson` all support cmake build system, cmake is supported as of Virsual Studio 2017 
https://devblogs.microsoft.com/cppblog/cmake-support-in-visual-studio-the-visual-studio-2017-rc-update/

These packages could be easily built with cmake even if do not use any package managers.

k-takata:
MS document about C11 and C17 is here:
https://learn.microsoft.com/en-us/cpp/overview/install-c17-support?view=msvc-170
Also related:
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170

> I frequently want to use anonymous union and struct.

This has been supported in MSVC for more than 25 years, I think.

leleliu008:
per https://learn.microsoft.com/en-us/cpp/overview/install-c17-support?view=msvc-170

```
Support for C11 and C17 standards is available in Visual Studio 2019 version 16.8 and later. 

Support requires an updated Universal C Runtime (UCRT) and Windows SDK version to work properly with the conforming preprocessor ([/Zc:preprocessor](https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-170)).
```

If we do support the latest two releases of Vistual Studio (`Vistual Studio 2022` and `Vistual Studio 2019` at this point),  using C11 should have no problem.

masatake:
@leleliu008, let's merge this.

masatake:
Thank you. I'll make a pull request that uses the "anonymous struct and union" of C11 and see what happens.

k-takata:
> You mean `anonymous union and struct` had been suppoted by Vistual Studio prior to C11?

Yes.

masatake:
@leleliu008 @k-takata Can we remove the combination
```
    - compiler: msvc
      ARCH: x64
```
from appveyor.yml now?

I would like to work on #3795 after making the pull request pass all the testing.

leleliu008:
this PR is implemented as https://github.com/universal-ctags/ctags/blob/master/win32/appveyor.bat#L63-L105

https://github.com/universal-ctags/ctags/blob/master/win32/appveyor.bat#L107-L109 is left, I don't known if https://github.com/universal-ctags/ctags/blob/master/.github/workflows/testing-msys2.yml#L56 is covered these tests. if not, we should add these tests to testing-msys2.yml

leleliu008:
If you just want to pass the workflows on appveyor, I think applying following patch is the easiest way:

```patch
diff --git a/appveyor.yml b/appveyor.yml
index edd8c5ae..f8a38380 100644
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,7 @@ environment:
     #   MSYSTEM: MINGW32
     - compiler: msvc
       ARCH: x64
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
     #- compiler: msvc
     #  ARCH: x86
     - compiler: mingw
```

https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds

leleliu008:
If the build configuration does not specify build worker image then `Visual Studio 2015` image is used.

ohter `appveyor.yml` doesn't specify build worker image, so `Visual Studio 2015` image is used. We could explicitly tell appveyor that we want to use `Visual Studio 2019` or `Visual Studio 2022`

leleliu008:
@masatake 
I was gonna tweak `appveyor.yml` and `win32/appveyor.bat` via PR (https://github.com/universal-ctags/ctags/pull/3815), but it fails due to some compiler warnings. I have no idea what's going on,  GitHub Actions use `Visual Studio 2019 Enterprise` has no warnings. AppVeyor use `Visual Studio 2019 Community` reports some warnings, due to `/WX` compiler option is used, these warrnings are treated as errors and abort the build process. 



> @leleliu008 @k-takata Can we remove the combination
> 
> ```
>     - compiler: msvc
>       ARCH: x64
> ```
> 
> from appveyor.yml now?
> 
> I would like to work on #3795 after making the pull request pass all the testing.

I think `msvc-x64` environment for appveyor can be removed.

masatake:
> I was gonna tweak appveyor.yml and win32/appveyor.bat via PR (https://github.com/universal-ctags/ctags/pull/3815), but it fails due to some compiler warnings. I have no idea what's going on, GitHub Actions use Visual Studio 2019 Enterprise has no warnings.

It seems that the code generated by packcc causes the warnings. I'll look into it.

leleliu008:
> > I was gonna tweak appveyor.yml and win32/appveyor.bat via PR (#3815), but it fails due to some compiler warnings. I have no idea what's going on, GitHub Actions use Visual Studio 2019 Enterprise has no warnings.
> 
> It seems that the code generated by packcc causes the warnings. I'll look into it.

What I'm confused is that GitHub Actions and AppVeyor both use `Visual Studio 2019`, the only difference is that the former is  `Enterprise Edition`, the later is `Community Edition`

k-takata:
> What I'm confused is that GitHub Actions and AppVeyor both use `Visual Studio 2019`, the only difference is that the former is `Enterprise Edition`, the later is `Community Edition`

The AppVeyor job is compiled with `-DDEBUG` and the GHA job is compiled without `-DDEBUG`.
Line 133 of peg_common.h is compiled only when `DEBUG` is defined.

leleliu008:
@k-takata Thank you very much for the hint.

leleliu008:
> It seems that the code generated by packcc causes the warnings. I'll look into it.

It seems that `peg\peg_common.h` is not generated by packcc.

PR #3809, opened 2023-09-09
adding forth ctags

farvardin:
this PR is adding the ctags for FORTH. It's very simple, it only detects the ":" for the beginning of a word:

image


masatake:
Thank you.

We have to work more to integrate your code into the source code repository of Universal Ctags.

1. The glue code. What you show is just the parser code. The parser code is not registered to the main part of Universal Ctags, so there is no way to use the parser from the command line of ctags.
2. The test case. We need a test case.
3. Updating docs/news.rst
4. The reference for the forth language.

You don't have to use C for the parser. Optlib may be enough.
See 
https://github.com/universal-ctags/ctags/blob/master/optlib/passwd.ctags as an example.

Universal ctags has a translator from .ctags to .c.

In addition, you use your `master` branch to make a pull request. Bad idea. I think you should use a topic branch.


For merging, I expect you to do 2 and 4. I just wanted to let you know that the rest can be done on my side.

parsers/forth.c
```diff
@@ -0,0 +1,78 @@
...
+
+static void findForthTags (void)
+{
+	vString *name = vStringNew();
+	const unsigned char *line;
+
+	while ((line = readLineFromInputFile()) != NULL)
+	{
+		if (line[0] == ':') {
+			vStringCatS(name, " ");
```
masatake, line 51: Do you really want to put a white space at the beginning of the tag name?

If the whitespace is for decorating the name rendered on your IDE or editor, it should be done on the IDE or editor; it is not the job of ctags.

parsers/forth.c
```diff
@@ -0,0 +1,78 @@
+/*
+*   Copyright (c) 2023, Eric Forgeot
+*
+*   Based on work by Jon Strait
```
masatake, line 4: Any references (e.g. URL) for the original work?

parsers/forth.c
```diff
@@ -0,0 +1,78 @@
...
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your opinion) any later version.
+*
+*   This module contains functions for generating tags for Forth files
```
masatake, line 9: Any reference for the language?

parsers/forth.c
```diff
@@ -0,0 +1,78 @@
...
+	{
+		if (line[0] == ':') {
+			vStringCatS(name, " ");
+			vStringCatS(name, (const char *) line);
+//			makeSimpleTag(name, K_SECTION);
+			makeSimpleTag(name, K_FUNCTION);
+		}
+		else {
+			vStringClear (name);
+			if (! isspace(*line))
```
masatake, line 58: I don't understand when this branch is needed.
A test case is needed to understand this.

farvardin: it seems to be necessary otherwise it will repeat catching the tags, like this:

image

parsers/forth.c
farvardin: It doesn't seem necessary, I've removed it

parsers/forth.c
farvardin: I've put it in the newest commit (it was for markdown ctags in geany, it seems to have been replaced later but I've found the original code again)

parsers/forth.c
farvardin: Forth is a very old language from the 70's. It seems to have some kind of revival now. I've put a reference URL in the newest commit

farvardin:
1/ The glue code should work now? I've included references to forth.c into source.mak and parsers_p.h, I don't know if it was what you had expected. Now forth is listed with the --list-languages option
2/ I've included a test case, which is passing.
3/ I've included the mention to forth into  docs/news.rst
4/ And I've also include the reference manual URL (on forth.com) for the forth language.

I've changed .f to .fth. Both are used (along with .ft), but .f was maybe clashing with fortran.

> You don't have to use C for the parser. Optlib may be enough.

It looks interesting. I'm using ctags with geany, I don't see any optlib in the geany source code but since there is a ctags to c transpiler maybe it should work too. After the integration of forth into ctags, I'll post the information on some forth discussion group, so maybe some people could help improving this (it could also list variable and constant words, into a different section, if it's possible).

parsers/forth.c
masatake: I don't understand the exmple.

Event with the following change, the test case you prepared is passed.
```
diff --git a/parsers/forth.c b/parsers/forth.c
index d6d7b8cd5..582888d4c 100644
--- a/parsers/forth.c
+++ b/parsers/forth.c
@@ -54,11 +54,7 @@ static void findForthTags (void)
 			vStringCatS(name, (const char *) line);
 //			makeSimpleTag(name, K_SECTION);
 			makeSimpleTag(name, K_FUNCTION);
-		}
-		else {
 			vStringClear (name);
-			if (! isspace(*line))
-				vStringCatS(name, (const char*) line);
 		}
 	}
 	vStringDelete (name);
@@ -77,4 +73,3 @@ extern parserDefinition* ForthParser (void)
 	def->parser = findForthTags;
 	return def;
 }
 ```

Could you prepare a test case that requires the branch?

File: Units/simple-forth.d/expected.tags [hunk 5]
```diff
@@ -0,0 +1,4 @@
+: fizz?  	input.fth	/^: fizz?  $/;"	s
```
masatake, line 1: I think `: `should not be part of the tag name.
Am I wrong?

parsers/forth.c
```diff
@@ -0,0 +1,78 @@
...
+*   DATA DEFINITIONS
+*/
+
+typedef enum {
+//   K_SECTION,
+   K_FUNCTION,
+} ForthKind;
+
+static kindDefinition ForthKinds[] = {
+	{ true, 's', "word", "words" },
```
masatake, line 39: word?
It seems that this should be "function".
Am I wrong?

masatake:
See #3810. #3810 is based on this pull request. I reflected the all review comments of mine to the pull request.

parsers/forth.c
farvardin: In Forth there is no "function", the equivalent is called "word"

masatake: In that case, `K_FUNCTION` is not a good name.
Should we use `s` as the letter for "word"? `w` looks better to me.

parsers/forth.c
farvardin: I don't undestand. With the same file example loaded into geany (and without this part), I got this, which is erroneous:

image

parsers/forth.c
farvardin: I don't know how to label it "words", the result in geany is called "section", but maybe it should be changed into the geany's code instead

Units/simple-forth.d/expected.tags
farvardin: probably not, you're right. But since I didn't find a way to label it "word" instead of "section", it might be safer, at the moment, to keep the ``: ``. Later it should be great to be able to add variables and constant and to have them sorted in their own subsections.

farvardin:
> See #3810. #3810 is based on this pull request. I reflected the all review comments of mine to the pull request.

Thank you for having made this conversion.

I've seen the page about using the translator from ctags into C:

https://docs.ctags.io/en/latest/optlib.html#translating-an-option-file-into-c-source-code-optlib2c

It's funny to see optscript is inspired by postscript which was itself inspired by forth ;) (they are all postfix stack langages)

parsers/forth.c
farvardin: > In that case, `K_FUNCTION` is not a good name. Should we use `s` as the letter for "word"? `w` looks better to me.

sure, why not. You can change it, but it looks like this code won't be used since you've made the forth.ctags in replacement? 

btw I've tested the generated optlib/forth.c into the geany code, and it didn't compile, I got a:
> parsers/forth.c:44:12: error: ‘parserDefinition’ {aka ‘struct sParserDefinition’} has no m
> ember named ‘versionCurrent’               
>   44 |         def->versionCurrent= 0;                                                   
>      |            ^~
> parsers/forth.c:45:12: error: ‘parserDefinition’ {aka ‘struct sParserDefinition’} has no m
> ember named ‘versionAge’                                                                     45 |         def->versionAge    = 0;

parsers/forth.c
masatake: I don't use Geany. So I don't understand the trouble well.
Could you show the trouble as a test case under Units/simple-forth.d?

Units/simple-forth.d/expected.tags
masatake: Could you write a .b test case (https://docs.ctags.io/en/latest/testing-parser.html#gathering-test-cases-for-known-bugs) to show the ideal behavior of forth parser?

If we implement word, section, variable, and constant, what kind of input can we test the parser? I would like to see the input and imaginary tags output.

parsers/forth.c
masatake: Fixing the way to integrate ctags with Geany is the next step.
We should focus on making the forth parser work from CLI is the first.

parsers/forth.c
farvardin: > I don't use Geany. So I don't understand the trouble well. Could you show the trouble as a test case under Units/simple-forth.d?

Sorry, I didn't read well your diff.  I didn't notice you had kept the ``vStringClear (name);``. Removing it was causing the side effect shown above. Testing this in ctags gave me the same result as in geany.

Using vStringClear in the way you did is correct, and even better it works now if the ``: word ;`` definitions don't have a blank line between them.

I'll update it this way. (sorry, I'm not used to work with branches in git)

Units/simple-forth.d/expected.tags
farvardin: You can have such a sample input there:
https://github.com/universal-ctags/ctags/blob/b97acf060ddb84c13ccd78d4eaea5d614db506c5/Units/simple-forth.d/input.fth

```
10 constant DUMMY1
12 constant DUMMY2

variable dummy3
variable dummy4

11 dummy3 !
13 dummy4 !

: fizz?  
3 mod 0 = dup if ." Fizz" then ;
: buzz?  
5 mod 0 = dup if ." Buzz" then ;

: fizz-buzz?  
dup fizz? swap buzz? or invert ;

: do-fizz-buzz  
25 1 do cr i fizz-buzz? if i . then loop ;
```

And the tags output could be something like:

```
10 constant DUMMY1 	input.fth	/^DUMMY1  $/;"	s
12 constant DUMMY2 	input.fth	/^DUMMY2  $/;"	s
variable dummy3 	input.fth	/^dummy3  $/;"	s
variable dummy4 	input.fth	/^dummy3  $/;"	s
```

(with w, c and v instead of s I suppose)

parsers/forth.c
farvardin: > Fixing the way to integrate ctags with Geany is the next step. We should focus on making the forth parser work from CLI is the first.

They've said they don't import the optlib from ctags, it doesn't look compatible I'm afraid:
https://github.com/geany/geany/pull/3377#issuecomment-1713642517

Units/simple-forth.d/expected.tags
farvardin: to be more precise, "variable" and "constant" are also some special words in forth (in forth something is either a word or a value, there is nothing else), and therefore some forth could use CONSTANT and VARIABLE while others would use constant and variable (historically they were in capital letters). It's also possible to use either small or capital letters for the constants and variable names, but the usage tends to use capital letters for constant names and small letters for variable names, it's the reason I've written them this way in the sample

masatake:
> It's funny to see optscript is inspired by postscript which was itself inspired by forth ;) (they are all postfix stack languages)

Therefore, more funny thing is ctags has no parser for postscript:-)
The language is too flexible to write an indexer for it.

masatake:
#3812 is the successor of this pull request.

PR #3810, opened 2023-09-10
adding forth ctags (optlib based)

masatake:
(@masatake converted to the original forth.c proposed in #3809 to forth.ctags.)

PR #3812, opened 2023-09-11
Forth: new parser

farvardin:
see https://github.com/universal-ctags/ctags/pull/3810#


docs/news.rst
```diff
@@ -413,6 +413,7 @@ The following parsers have been added:
 * Elixir *optlib*
 * Elm *peg/packcc*
 * Falcon
+* Forth
```
masatake, line 416: `*Forth *optlib*`

optlib/forth.ctags
```diff
@@ -0,0 +1,23 @@
...
+--map-Forth=+.fth
+--map-Forth=+.forth
+# TODO: a selector arbitrating Fortran and Forth
+# --map-Forth=+.f
+
+--kinddef-Forth=w,word,words
+--kinddef-Forth=v,variable,variables
+--kinddef-Forth=c,constant,constants
+
+--regex-Forth=/^:[[:space:]]*([^[:space:]]+)/\1/w/
```
masatake, line 21: `{exclusive}` may improve the performance:
```
--regex-Forth=/^:[[:space:]]*([^[:space:]]+)/\1/w/{exclusive}
```

optlib/forth.ctags
```diff
@@ -0,0 +1,23 @@
...
+--regex-Forth=/^(variable|VARIABLE)[[:space:]]*([^[:space:]]+)/\2/v/
```
masatake, line 22: You can use a flag for case-insensitive.

```
--regex-Forth=/^variable[[:space:]]*([^[:space:]]+)/\2/v/{exclusive}{icase}
```

optlib/forth.ctags
```diff
@@ -0,0 +1,23 @@
...
+--regex-Forth=/^([[:alnum:]]+)[[:space:]](constant|CONSTANT)[[:space:]]*([^[:space:]]+)/\3/c/
```
masatake, line 23: We can do the same as the `variable`:

```
--regex-Forth=/^([[:alnum:]]+)[[:space:]]constant[[:space:]]*([^[:space:]]+)/\2/c/{exclusive}{iscase}
```

Do you know if we need the first group?

```
--regex-Forth=/^[[:alnum:]]+[[:space:]]constant[[:space:]]*([^[:space:]]+)/\1/c/{exclusive}{iscase}
```

parsers/forth.c
```diff
@@ -0,0 +1,76 @@
+/*
```
masatake, line 1: We don't need this file anymore.

optlib/forth.ctags
```diff
@@ -0,0 +1,23 @@
...
+--langdef=Forth
+--map-Forth=+.fth
+--map-Forth=+.forth
+# TODO: a selector arbitrating Fortran and Forth
+# --map-Forth=+.f
+
+--kinddef-Forth=w,word,words
+--kinddef-Forth=v,variable,variables
+--kinddef-Forth=c,constant,constants
+
```
masatake, line 20: If you add a `--regex-Forth` for a comment with `{exclusive}` flag, we can improve the performance of the parser a bit.

```
```

optlib/forth.ctags
```diff
@@ -0,0 +1,23 @@
...
+#  This source code is released for free distribution under the terms of the
+#  GNU General Public License version 2 or (at your opinion) any later version.
+#
+#  This module contains functions for generating tags for Forth files
+#  https://www.forth.com/starting-forth/1-forth-stacks-dictionary/
+#
+--langdef=Forth
+--map-Forth=+.fth
+--map-Forth=+.forth
+# TODO: a selector arbitrating Fortran and Forth
```
masatake, line 14: This is my job.

How can we distinguish a Fortran source file and a Forth source file?
I need heuristics.

Fortran:
- a line starts from `program
- a line starts from `!`

Forth:
- a line starts from `\`
- a line starts from `: `
- a line starts from a digit

What do you think?

optlib/forth.ctags
farvardin: I haven't kept this one, it doesn't catch the constants.

The first alnum is necessary because a number is always needed to initialise a constant. All the cases can't be catched anyway, because instead of:

 ```
10 constant DUMMY1
```

it would also be possible to write


 ```
10 
constant DUMMY1
```

but it doesn't really matter

parsers/forth.c
farvardin: yes, I understand. But couldn't it be possible to skip it anyway? Geany told me they wanted it to be pushed upstream before integrating it into their code, and on the other hand they can't handle the .ctags / optlib tags...

optlib/forth.ctags
farvardin: it should do the job. 
I don't know Fortran, but Forth gives much freedom for naming words, so it could also be legal to name a word ``!word``. ``!`` is even a command (for storing a value into a variable). As it's stack based, it would be possible to write:

```
50
SomeVariable
!
```

but most people would probably write instead ``50 SomeVariable !``

There are also many other commands which can start a line, but a file with a high number of ``:`` and ``\`` at the beginning of lines would probably be forth

optlib/forth.ctags
```diff
@@ -0,0 +1,24 @@
...
+# --map-Forth=+.f
+
+--kinddef-Forth=w,word,words
+--kinddef-Forth=v,variable,variables
+--kinddef-Forth=c,constant,constants
+
+--regex-Forth=/^[[:space:]]*\\.*//{exclusive}
+--regex-Forth=/^:[[:space:]]+([^[:space:]]+)/\1/w/{exclusive}
+--regex-Forth=/^variable[[:space:]]*([^[:space:]]+)/\1/v/{exclusive}{icase}
```
masatake, line 23: I forgot to write this one.

```
--regex-Forth=/^variable[[:space:]]+([^[:space:]]+)/\1/v/{exclusive}{icase}
```

optlib/forth.ctags
masatake: Thank you for the input. I will implement the selector.

parsers/forth.c
masatake: See https://github.com/geany/geany/issues/3557 . 

The .ctags is translated to C. So technically, it is possible to integrate the .ctags based Forth parser into Geany.

Let's see how Geany people say.

Units/simple-forth.d/expected.tags
```diff
@@ -0,0 +1,9 @@
+DUMMY1	input.fth	/^10 constant DUMMY1$/;"	c
```
masatake, line 1: I have a plan to add another test case.
So could you rename `Units/simple-forth.d` to `Units/parser-forth.r/simple-forth.d`?
So I can add a case like `Units/parser-forth.r/masatake.d`.

optlib/forth.ctags
farvardin: it's done, I've just pushed them, and the new test case folder

masatake:
I push two commits to your branch.

farvardin:
> Do you know any large (open) source code bases written in Forth.
> 
> I would like to add them to https://github.com/universal-ctags/codebase. So we can evaluate the performance of the parser.

good idea.
Most forth systems are written in forth itself. Here are some open source samples and codes:

https://git.savannah.gnu.org/cgit/gforth.git/tree/
https://git.sr.ht/~vdupras/duskos/tree/master/item/fs
https://gitlab.com/b2495/uf/-/blob/master/uf.f?ref_type=heads
https://sourceforge.net/p/forth-4th/code/HEAD/tree/trunk/4th.src/demo/
https://github.com/flagxor/ueforth/tree/main/examples
https://github.com/flagxor/ueforth/tree/main/common

some files are quite small, but others might be good candidates

masatake:
Added Forth based source trees to the our codebase.
See https://github.com/universal-ctags/codebase/commit/df322155227e2604b0f913102469efe83b701e02

parsers/forth.c
masatake: @farvardin we got a positive answer from Geany developers introducing an optlib-based parser to Geany. 

Could you remove parsers/forth.c ?

If we recognize the optlib-based implementation is not enough, you can open a pull request implementing the parser with crafted-C-code in the future. We can re-implement the parser safely because we have the test case.

farvardin: @masatake ok, I've done it. There was a conflict merge, I hope I haven't deleted some of your code, but it seems fine from my side

masatake:
You removed `.forth` from the map. I guess you mistakenly removed it.
So I will add `.forth` again. If you really want to remove the file extension, please, make a pull request after I merge this one.

masatake:
Thank you.

farvardin:
@masatake thank you for your help and for this inclusion in ctags!

PR #3820, opened 2023-09-15
GitHub Actions: cleanup code testing-mac.yml

leleliu008:
`/usr/bin/cc` `/usr/bin/gcc` `/usr/bin/clang` are actually the same.

`libxml2` is a keg_only package, which means it was not linked into `/usr/local`. I removed it, because it was pre-installed on macOS.


masatake:
(About MacOS, I don't know anything. So you don't need my ack.)

PR #3821, opened 2023-09-15
using %zu for printing size_t values

leleliu008:
References:

https://man7.org/linux/man-pages/man3/ssize_t.3type.html

https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-160#size-prefixes-for-printf-and-wprintf-format-type-specifiers


masatake:
Thank you.

PR #3825, opened 2023-09-20
Doc: update docs/windows.rst

leleliu008:


docs/windows.rst
```diff
@@ -46,7 +46,46 @@ Building ctags from the command line
...
+
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"
+
+For Microsoft Visual Studio 2019 Enterprise::
+
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"
+
+For Microsoft Visual Studio 2019 Community::
+
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"
```
k-takata, line 75:
```suggestion
        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat"
```

docs/windows.rst
```diff
@@ -60,29 +101,33 @@ If you want to create PDB files for debugging even for a release version, you mu
 
         nmake -f mk_mvc.mak PDB=1
 
-GCC
-.............................................................................
+Building ctags with MSBuild
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**General**
+This supports only Microsoft Visual Studio 2013.
```
k-takata, line 107: I've just tried with Microsoft Visual Studio Express 2017 and it seems that the project files for 2013 still works. (I haven't tried with 2019 or 2022 yet.)

I had to copy some files before running the `msbuild` command.

```
copy win32\config_mvc.h config.h
copy win32\gnulib_h\langinfo.h gnulib
copy win32\gnulib_h\fnmatch.h gnulib
```

leleliu008: Using VS2019 reported following:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(439,5): error MSB8020: The build too 
ls for Visual Studio 2013 (Platform Toolset = 'v120') cannot be found. To build using the v120 build tools, please install Visual Studio 2013 build tools. 
  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarge 
t solution". [C:\Users\vagrant\ctags\win32\ctags_vs2013.vcxproj]
```

I think the file name `ctags_vs2013.*` should rename to other name if it really support VS2015 and VS2017.

k-takata: Ah, this PC has both VS2013 and VS2017 installed. So, even VS2017's msbuild is used, toolset from VS2013 is used.

If `<PlatformToolset>v120</PlatformToolset>` lines in the .vcxproj file are changed to `<PlatformToolset>v141</PlatformToolset>`, it might work on the newer VS versions.

docs/windows.rst
```diff
@@ -60,29 +101,33 @@ If you want to create PDB files for debugging even for a release version, you mu
...
 
-All the GCC's come with installers or with zipped archives. Install or extract them in a directory without spaces.
+The simplest build instructions like below::
 
-GNU Make builds for Win32 are available as well, and sometimes are included with the compilers. Make sure it is in your path, for instance by copying the make.exe in the bin directory of your compiler.
+        msbuild win32\ctags_vs2013.sln
```
k-takata, line 111: As I said in another comment, I think that some files need to be copied:

```suggestion
        copy win32\config_mvc.h config.h
        copy win32\gnulib_h\langinfo.h gnulib
        copy win32\gnulib_h\fnmatch.h gnulib

        msbuild win32\ctags_vs2013.sln
```

docs/windows.rst
leleliu008: As per https://learn.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=msvc-170#platform-toolset

```
Visual Studio 2010: v100
Visual Studio 2013: v120
Visual Studio 2015: v140
Visual Studio 2017: v141
Visual Studio 2019: v142
Visual Studio 2022: v143
```

PlatformToolset can't be changed via msbuild options.

It seems that we need `ctags_vs2013.vcxproj` `ctags_vs2015.vcxproj` `ctags_vs2017.vcxproj` `ctags_vs2019.vcxproj` `ctags_vs2022.vcxproj`.

k-takata: It seems that there is a way to support multiple versions of VS in one vcxporj file:
https://qiita.com/yumetodo/items/a8324efaf83c9c08d168 (Japanese)
The basic idea is to switch `PlatformToolset` based on `VisualStudioVersion`.
(This page suggests using CMake instead of this method, though.)

leleliu008: Yes, I think we should use `cmake` instead of `msbuild`

masatake:
@leleliu008 Thank you.

PR #3832, opened 2023-10-20
README.md: make installation commands directly runnable

rhythm16:
Allow users to directly copy and run the installation commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,12 @@ code from GitHub.
 
 To build with Autotools (Autoconf and Automake) on GNU/Linux, OSX, or Windows 10 WSL,
 ```
-    $ git clone https://github.com/universal-ctags/ctags.git
-    $ cd ctags
-    $ ./autogen.sh
-    $ ./configure --prefix=/where/you/want # defaults to /usr/local
-    $ make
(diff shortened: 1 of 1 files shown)


masatake:
Thank you for the suggestion. Such a user should install a binary package on the system one's using or install a prebuilt executable from https://github.com/universal-ctags/ctags-nightly-build.

I added `$` to the lines to represent you can build ctags without root privilege.

A change like `s/\$ make install/# make install/' makes sense.

masatake:
```
+ ./configure  # use --prefix=/where/you/want to override installation directory, defaults to /usr/local
```

This change is helpful. Could you make a pull request for this line?

rhythm16:
> I added `$` to the lines to represent you can build ctags without root privilege.

Ah, I see.

> A change like `s/\$ make install/# make install/` makes sense.

After your explanation I feel it is better to leave this unchanged, since some users won't have root access, and it's obvious one needs root privileges to install in system directories.

> ```
> + ./configure  # use --prefix=/where/you/want to override installation directory, defaults to /usr/local
> ```
> 
> This change is helpful. Could you make a pull request for this line?

Sure, see #3835 .

Thanks for the feedback.

PR #3834, opened 2023-10-20
main,tlib: fix wording

masatake:
Close #3880

Suggested by @dataf3l.


leleliu008:
The macos runner would become green after https://github.com/universal-ctags/ctags/pull/3836 is merged.

PR #3836, opened 2023-10-21
GitHub Actions: do not try to install python3 package via brew, it has already been included in macos-xx runner images, install docutils package via pip3 rather than brew

leleliu008:


masatake (approved):

masatake:
@leleliu008 Thank you.

PR #3840, opened 2023-10-26
Various warnings fixes

b4n:
This fixes various mostly harmless warnings from GCC.  This is not everything yet, and most of it is trivial, but it still gets rid of some and helps tidy code a little.

### DIFF
--- a/dsl/dsl.c
+++ b/dsl/dsl.c
@@ -36,6 +36,9 @@ typedef struct sDSLEngine DSLEngine;
  * MACROS
  */
 
+/* dummy definition to allow/require an extra semicolon */
+#define END_DEF(sfx) typedef int ctags_dummy_int_type_ignore_me_##sfx
+
 #define DECLARE_VALUE_FN(N)									\
 static EsObject* value_##N (EsObject *args, DSLEnv *env)
 
@@ -43,7 +46,7 @@ static EsObject* value_##N (EsObject *args, DSLEnv *env)
 static EsObject* value_##N (EsObject *args, DSLEnv *env)	\
 {															\
 	return dsl_entry_##N (env->entry);						\
-}
+} END_DEF(value_##N)
 
 /*
  * FUNCTION DECLARATIONS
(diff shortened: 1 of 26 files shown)


masatake:
The extra semicolons help the indentation engine of editors and source code indexes like ctags.
How about changing the definition of the macro for defining something like:
```
+#define END_DEF extern int ctags_nosuch_name_this_is_just_dummy_ignore_me
 #define DECLARE_ALT_VALUE_FN(N)									\
 static EsObject* alt_value_##N (EsObject *args, DSLEnv *env)
 
@@ -31,7 +32,7 @@ static EsObject* alt_value_##N (EsObject *args, DSLEnv *env)			\
 	if (!env->alt_entry)												\
 		dsl_throw (NO_ALT_ENTRY, es_symbol_intern ("&" #N)); 			\
 	return dsl_entry_##N (env->alt_entry);								\
-}
+} END_DEF
```

b4n:
@masatake Hum, OK… so maybe it's easiest not to care about this specific issue?  It's not like it's likely to cause an issue, and not everybody is building with `-Wpedantic`.  But OTOH it's an actual non-ISO usage, which could (very theoretically) be an issue somewhere.

If you like your solution enough, we could indeed do it, but use `#define END_DEF typedef int ctags_dummy_int_type_ignore_me`, because redundant declarations would we warned about by `-Wredundant-decls` while redundant typedefs wouldn't.

If you do like it, I can update the patch to include this.

masatake:
@b4n, your `typedef` version is better than mine. Could you update the patch?

b4n:
@masatake done.  For now I added a new commit, but it can be merged.

I'm however a bit puzzled, in *parsers/asm.c* I see the warning from `-Wredundant-decls` even with a typedef, although I don't see it in a dummy C source, or in *dsl/\**… anyway, I switched to use a specific suffix which actually make the typedefs unique.

masatake:
@b4n Thank you. You developed an interesting technique. I would like to merge this pull request as is because the commit log for the commit introducing the technique is written well. People in another project can refer to the commit and this pull request.

PR #3841, opened 2023-10-31
Update kotlin parser

dolik-rce:
There has been a few small improvements to the [Kotlin language specification](https://github.com/Kotlin/kotlin-grammar-gpl2) this year, so it's time to update the grammar for the kotlin parser.


masatake:
@dolik-rce Thank you. Don't we need a test case for the change?

dolik-rce:
@masatake: I don't think so. The changes are really minor and the existing tests only check basic kotlin syntax.

It might be good idea to try to compose a kotlin file that would cover all the rules, but it would be quite big and tedious to construct, so it wouldn't fit into this PR.

masatake:
@dolik-rce  I see. Thank you.

PR #3846, opened 2023-11-16
JavaScript: handle spread syntax

masatake:
Related to #3435.

---

    JavaScript: handle spread syntax
    
    With this change, the parser skip "..." and the subsequent expression.
    
    ref. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax


masatake:
I must add a reference to a page describing the spread operator.

leleliu008:
@masatake our GitHub Actions BSD workflows alwasy fails these days. I have no idea how to fix it. The only way is waiting GitHub Actions Team's fix. Reference: https://github.com/actions/runner-images/issues/8730

masatake:
@leleliu008 Thank you. Could you make a "temporarily disabling" pull request and merge it?

leleliu008:
> @leleliu008 Thank you. Could you make a "temporarily disabling" pull request and merge it?

Done via https://github.com/universal-ctags/ctags/pull/3848

jafl (approved):

masatake:
It seems that "spread operator" is a term in TypeScript.
After reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax, "spread syntax" is the term I should use here.

PR #3848, opened 2023-11-16
GitHub Actions: temporarily disable BSD workflows

leleliu008:
temporarily disable them, watting for GitHub Actions Team's fix

Reference: https://github.com/actions/runner-images/issues/8730


masatake:
Thank you.

masatake:
@leleliu008 Could you re-apply this?

BSD workflows become a serious bottleneck of this project.

leleliu008:
@masatake 

The BSD workflows fails sometimes, but not that frequent. The reason is 

@leleliu008, @dmotte and @hummeltech, there is some amount of macOS-12 agents with the old image, it's being tracked in other issue and we are working on resolving it. https://github.com/actions/runner-images/issues/8642
About 'Error Type: Disconnect'. It is almost always about available resources boottleneck and we can recommend only to try optimise workflow. But I'll take a look at the available information about your runs just in case I may find something to fix.

_Originally posted by @erik-bershel in https://github.com/actions/runner-images/issues/8730#issuecomment-1834391700_

They uaually will become green after rerunning it. If you cann't accept it. let's disable them again.

leleliu008:
The BSD workflows are running in Virtualbox on macOS-12 for free, that's the only way running BSD workflows with GitHub Actions for free. GitHub Team encourage us to use [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners), but they're not free, we should pay at least `4$/month`

PR #3850, opened 2023-11-20
 SQL: extact views in "create view if not exists VIEW ..."

masatake:
At least, SQLite supprots the syntax: https://www.sqlite.org/lang_createview.html


b4n (approved):
Not that I know SQL nearly enough, but LGTM -- simple, clean, works.

k-takata:
It seems that there was a typo in the title of this PR. I've updated it to "extract", but is it correct?

masatake:
YES!

Issue #3851, opened 2023-11-21
Compilation error with MSYS2 CLANG{64,32} environments


k-takata:
https://github.com/universal-ctags/ctags-win32/actions/runs/6924537331/job/18833909212
https://github.com/universal-ctags/ctags-win32/actions/runs/6924537331/job/18833909365

```
main/lxpath.c:166:3: error: call to undeclared function 'xmlLineNumbersDefault'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  166 |                 xmlLineNumbersDefault (1);
      |                 ^
main/lxpath.c:167:9: error: call to undeclared function 'xmlParseMemory'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  167 |                 doc = xmlParseMemory((const char*)data, size);
      |                       ^
main/lxpath.c:167:7: error: incompatible integer to pointer conversion assigning to 'xmlDocPtr' (aka 'struct _xmlDoc *') from 'int' [-Wint-conversion]
  167 |                 doc = xmlParseMemory((const char*)data, size);
      |                     ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3 errors generated.
```

It seems that `xmlLineNumbersDefault` and `xmlParseMemory` are deprecated.
https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-parser.html
(Not sure why these errors don't occur with MSYS2 MINGW{32,64} environments, though. Both use libxml2-2.12.0.)

k-takata:
These error didn't occur when libxml2-2.11.6 was used:
https://github.com/universal-ctags/ctags-win32/actions/runs/6911068107/job/18805112817

k-takata:
It seems that MINGW{32,64} have another errors with libxml2-2.12.0.

MINGW64:
https://github.com/universal-ctags/ctags-win32/actions/runs/6917682193/job/18819132209
```
	--- ./Tmain/selector-xml-root-elelement.d/stdout-expected.txt	2023-11-19 01:59:33.486877700 +0000
	+++ D:/a/ctags-win32/ctags-win32/ctags/Tmain/selector-xml-root-elelement.d/stdout-actual.txt	2023-11-19 02:05:51.950974900 +0000
	@@ -1 +0,0 @@
	-input.xml: DBusIntrospect
```

MINGW32:
[...]

PR #3853, opened 2023-11-22
xmlLineNumbersDefault and xmlParseMemory functions are declared in libxml/parser.h, which in turn was not included in main/selectors.c, adding it to prevent compiler warnings and errors.

leleliu008:
https://github.com/universal-ctags/ctags/issues/3851


masatake (approved):

k-takata (approved):
I've confirmed that xmlLineNumbersDefault and xmlParseMemory are used only in lxpath.c and selectors.c.

PR #3854, opened 2023-11-22
GitHub Actions: update testing-msys2.yml

leleliu008:
keep testing-msys2.yml in sync with https://github.com/universal-ctags/ctags-win32/blob/master/.github/workflows/build-ctags.yaml

### DIFF
--- a/.github/workflows/testing-msys2.yml
+++ b/.github/workflows/testing-msys2.yml
@@ -11,34 +11,33 @@ jobs:
     runs-on: windows-latest
 
     strategy:
+      fail-fast: false
       matrix:
-        msystem: [MSYS, MINGW32, MINGW64]
+        msystem: [MSYS, MINGW32, MINGW64, CLANG32, CLANG64]
 
     defaults:
       run:
         shell: msys2 {0}
 
     steps:
-    - uses: msys2/setup-msys2@v2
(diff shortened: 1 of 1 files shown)


leleliu008:
Although successfully compiled with `libxml2-2.12.0`, but running unit tests fails on MSYS2.

as motioned in https://github.com/universal-ctags/ctags/issues/3851#issuecomment-1820141019

masatake:
Sorry I took a mistake.

.github/workflows/testing-msys2.yml
```diff
@@ -11,8 +11,9 @@ jobs:
     runs-on: windows-latest
 
     strategy:
+      fail-fast: false
       matrix:
-        msystem: [MSYS, MINGW32, MINGW64]
+        msystem: [MINGW32, MINGW64, CLANG32, CLANG64]
```
k-takata, line 16: I removed `MSYS` in ctags-win32 because it is not an environment for building Win32 apps.
However, I think we should keep this in this repository. (Even it is effectively the same as Cygwin.)

leleliu008: Well, I will add it back after the issue is fixed.

masatake:
Merged via #3858. Thank you.

PR #3858, opened 2023-11-22
build-sys: don't use libxml-2 if its CRLF handling is broken

masatake:
See #3851 .


masatake:
??? Why the errors of the test case is not reproduced?

masatake:
It seems that I must cherry-pick #3854.

masatake:
@leleliu008 I added AC_RUN_IFELSE to configure.ac.
What do you think about this? I'm afraid of breaking the cross-compiling infrastructure you have polished.

masatake:
As I am afraid, running configure on s390x has failed.

masatake:
@leleliu008 thank you.

```
 -- Macro: AC_RUN_IFELSE (INPUT, [ACTION-IF-TRUE], [ACTION-IF-FALSE],
          [ACTION-IF-CROSS-COMPILING = ‘AC_MSG_FAILURE’])
 ```

O.K. I will fill the 4th parameter.

leleliu008:
About the macos-13 workflow fails, please rebase master branch, I have merge https://github.com/universal-ctags/ctags/pull/3861 to branch, it should be fixed.

masatake:
@leleliu008 Thank you.

masatake:
What I studied while developing this puu request is that `pkg-config` doesn't support `&&` operator.

masatake:
I won!

PR #3859, opened 2023-11-24
GitHub Actions: enable BSD workflows

leleliu008:
We disabled BSD workflows due to the issue
https://github.com/actions/runner-images/issues/8730, which in turn had been fixed, so let's enable these BSD workflows


masatake:
About this area, if you think your PR is o.k., you can merge it. You don't have to wait o my ack.
I may have no comment on PRs in this area.

If you want to get my comment, please specify me as a reviewer explicitly.

leleliu008:
OK, I got it.

PR #3861, opened 2023-11-26
GitHub Actions: DO NOT install libxml2 package via brew, because macOS already provides it

leleliu008:


masatake:
LGTM. Interesting approach.

PR #3894, opened 2023-12-22
Revise the release process

### LINKED ISSUE #3767: create annotation tags instead of lightweight tag
in the first version of .github/workflows/tagging.yml (https://github.com/universal-ctags/ctags/commit/e9baeff49) 

```
env:
  # Account for creating a tag
  USER_NAME: "Universal-ctags CI"
  USER_EMAIL: "Universal-ctags@users.noreply.github.com"
```

It seems that you want to create annotation tags, beacuse only annotation tag can store metadata including `tagger's name`,  `tagger's email`, `tagged timestamp`, `tagged commit hash`, etc. the lightweight tag store nothing but a commit hash.

Reference: https://git-scm.com/book/en/v2/Git-Basics-Tagging

It is using commiter's name, but actually you want to use `Universal-ctags CI`, right?

There is no account named `Universal-ctags CI`, we can use the name `github-actions`, looks like following

||
|-|
||

# some screenshots show you what look like now

||
|-|
||
||
||

# 

If you think current is good enough, I think we should remove the `Universal-ctags CI` and `Universal-ctags@users.noreply.github.com` or add more comments.

masatake:
Close #3767.


masatake:
@leleliu008  could you review [f94bdfa](https://github.com/universal-ctags/ctags/pull/3894/commits/f94bdfaeb8f18cde1d9b4208fba8f04e4b14c4ba) ?
See #3767.

leleliu008 (commented):
1. we usually say `creating an annotation tag`, I never saw someone said `putting an annotation tag`, so I think it would be better to change the word `putting` to `creating` 
2. `NOTE: for putting an annotation tag, we need a GitHub account for the user.`  I,m not clear what you mean. we use GitHub Actions to create a new tag, I myself hope our project's tag page could display a line like `github-actions tagged this `,  rather than the HEAD commit's commiter.

leleliu008:
If we want to display something like `github-actions release this`, we need to use [github-cli](https://cli.github.com/) , if we use `git` command, github would display `universal-ctags release this`, because tagger's email `Universal-ctags@users.noreply.github.com` belogs to `universal-ctags` account.

masatake:
@leleliu008 Thank you.

I don't want to make an annotation tag for a snapshot. So displaying the HEAD commit's commiter is ok. I don't want to increase the number of commits just for making snapshots.

At the initial comment in #3767, you wrote: 
> If you think current is good enough, I think we should remove the Universal-ctags CI and Universal-ctags@users.noreply.github.com or add more comments.

I implemented what I wrote here in [f94bdfa](https://github.com/universal-ctags/ctags/pull/3894/commits/f94bdfaeb8f18cde1d9b4208fba8f04e4b14c4ba). Do I misunderstand what you wrote?

leleliu008:
@masatake There is no misunderstandings.

masatake:
@leleliu008 Thank you for confirmation. I will remove the comments newly added in this pull request.

PR #3897, opened 2023-12-28
XML Based Resource System (XRC): new parser

masatake:
https://docs.wxwidgets.org/3.0/overview_xrc.html


masatake:
Till 7.0, I will convert this short C program into:
```
--xpath-langdef=XRC{base=XML}{dedicated}
--map-XRC=+.xrc

--kinddef=o,object,objects

--xpath-tabledef:main
--xpath-XRC=main:o,def:/resource//object/@name
```

PR #3908, opened 2024-01-07
Ruby: extract m in "define_method(:m" as a method

masatake:


parsers/ruby.c
```diff
@@ -1307,8 +1333,11 @@ static void findRubyTags (void)
 		{
 			NestingLevel *nl = nestingLevelsGetCurrent (nesting);
 			tagEntryInfo *e_scope  = getEntryOfNestingLevel (nl);
+
 			if (e_scope && (e_scope->kindIndex == K_CLASS
-							|| e_scope->kindIndex == K_MODULE))
+							|| e_scope->kindIndex == K_MODULE
+							|| ((e_scope->kindIndex == K_METHOD) &&
```
masatake, line 1339: Comments should be added.

PR #3937, opened 2024-02-13
lex: add support for implementation blocks of conditions

jafl:
In some lexers, all the rules for a condition are grouped together as: <COND>{
...
}

It is useful to be able to jump to these blocks.


masatake:
I wonder whether we should capture it as a definition or reference tag.

Tagging B_ENVIRONMENT as `grouping` role of `condition` kind is my alternative idea.

masatake:
I wonder whether we should capture it as a definition or reference tag.

jafl:
> I wonder whether we should capture it as a definition or reference tag.

How would that change the code and the output?

masatake:
> How would that change the code and the output?

I'll make a pull request based on your pull request to demonstrate tagging '<FOO> {..}` with a reference tag. Please, give me time. I have trouble with my health.

jafl:
Closing in favor of #3939

jafl:
And I hope you feel better soon!

PR #3939, opened 2024-02-14
LEX: make reference tags for the conditions used for making groups

masatake:
In some lexers, all the rules for a condition are grouped together as: 
```
<COND>{
...
}
```
It is useful to be able to jump to these blocks.

Conditions used for grouping are tagged as reference tags having "grouping" role of "cond" kind. So you must add --extras=+r to your ctags cmdline for enabling this support.

Update the parser version to 1.1.

This is derrived from #3937.

### DIFF
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -71,6 +71,7 @@ Julia          n/module          imported           on      loaded by "import"
 Julia          n/module          namespace          on      only some symbols in it are imported
 Julia          n/module          used               on      loaded by "using"
 Kconfig        k/kconfig         source             on      kconfig file loaded with source directive
+LEX            c/cond            grouping           on       conditions used for grouping of start or exclusive condition rules
 LdScript       i/inputSection    destination        on      specified as the destination of code and data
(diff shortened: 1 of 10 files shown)


Tmain/list-roles.d/stdout-expected.txt
```diff
@@ -208,6 +209,7 @@ Julia          n/module          imported           on      loaded by "import"
 Julia          n/module          namespace          on      only some symbols in it are imported
 Julia          n/module          used               on      loaded by "using"
 Kconfig        k/kconfig         source             on      kconfig file loaded with source directive
+LEX            c/cond            grouping           on       condtions used for grouping of start or exclusive condition rules
```
jafl, line 212: condtions -> conditions

jafl (approved):
This is much better than my solution!  Please merge!

Tmain/list-roles.d/stdout-expected.txt
masatake: Fixed. Thank you.

masatake:
https://sources.debian.org/patche/exuberant-ctags/1:5.9~svn20110310-11/python-disable-imports.patch/

From the comment in the patch, You will know why I am concerned about whether a language object is extracted as a definition or reference tag.

PR #3941, opened 2024-02-17
[WIP] C++ : support C++20 modules

### LINKED ISSUE #3932: C++: support modules
Hello

Can ctags skip the new module import statements in C++ 20 please ?

For example the following line produces a definition tag for symbol 'std':

```C++
import std;
```

The resulting tag should be a reference and not a definition. The corresponding `export module std;` line would be expected definition, plus multiple `module std;` lines if present in other files

But just skipping the import line entirely would still fix the wrong definition.

----
(@masatake added)
- https://cpprefjp.github.io/lang/cpp20/modules.html

masatake:
Close #3932.

- [X] "export module M;"
- [X] "export something_other_than_module_and_import ..."
- [X] "module M;"
- [X] "module M:P;"
- [ ] "export {...}"
- [ ] "import ..."
- [ ] "module;"
- [ ] "module :private;"


masatake:
We can use the same technique as cxxKeyworkdEnableFinal for handling the keywords "module" and "import".

PR #3942, opened 2024-02-20
GitHub Actions: add macos-14 runner

leleliu008:


leleliu008:
Let's wait for Android NDK r26c is included in macos runner images.

PR #3954, opened 2024-03-13
GitHub Actions: set HOMEBREW_NO_INSTALL_CLEANUP

masatake:


leleliu008:
You do not need to set that, that already has been done for you. https://github.com/actions/runner-images/issues/2466

masatake:
@leleliu008 Thank you. How can we avoid the failure on "run units target on macOS / testing (macos-13) (pull_request)" ?

leleliu008:
```
brew install --overwrite python@3.12
```

masatake:
However, I found the following comment that you wrote in testing-mac.yml:
```
    # NOTE: DO NOT try to install python3 package via brew, it has already been included in macos-xx runner images
```

Should I delete the comment and add `brew install --overwrite python@3.12` instead?

leleliu008:
There are many python packages like python@3.10 python@3.11 python@3.12 in brew, they always conflict with each other.

masatake:
??? So what should we do? 
Should we specify the version of python explicitly though though a version of python has already been included in macos-xx runner images?

leleliu008:
> However, I found the following comment that you wrote in testing-mac.yml:
> 
> ```
>     # NOTE: DO NOT try to install python3 package via brew, it has already been included in macos-xx runner images
> ```
> 
> Should I delete the comment and add `brew install --overwrite python@3.12` instead?


we do not install a python package though, but many packages may depend on different version of python packages. HomeBrew Team update their package formula frequently, it's very unstable, but it usually canbe fix via `brew install --overwrite python@x.y`

leleliu008:
> ??? So what should we do? Should we specify the version of Python explicitly, though a version of Python has already been included in macos-xx runner images?

Yes, no good solution. It is not a stable package manager after all.

leleliu008:
let me fix it. I will file a new PR to fix it.

leleliu008:
fixed as https://github.com/universal-ctags/ctags/pull/3955

PR #3958, opened 2024-03-14
add support for loongarch

wuruilong01:
The package failed to compile at loong64 for the following detailed reasons:[https://buildd.debian.org/status/fetch.php?pkg=universal-ctags&arch=loong64&ver=5.9.20210829.0-1&stamp= 1709418146&raw=0](url)

### DIFF
--- a/main/seccomp.c
+++ b/main/seccomp.c
@@ -41,7 +41,9 @@ int installSyscallFilter (void)
 	seccomp_rule_add (ctx, SCMP_ACT_ALLOW, SCMP_SYS (exit_group), 0);
 
 	// The bowels of stdio want to know the size of a file, even for stdout.
+#if (defined(__SNR_fstat) && __SNR_fstat)
 	seccomp_rule_add (ctx, SCMP_ACT_ALLOW, SCMP_SYS (fstat), 0);
+#endif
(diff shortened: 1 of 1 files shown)


main/seccomp.c
```diff
@@ -41,7 +41,9 @@ int installSyscallFilter (void)
 	seccomp_rule_add (ctx, SCMP_ACT_ALLOW, SCMP_SYS (exit_group), 0);
 
 	// The bowels of stdio want to know the size of a file, even for stdout.
+#if (defined(__SNR_fstat) && __SNR_fstat)
```
masatake, line 44: Just a question.

Is
```
#ifdef __SNR_fstat
```
not enough?

wuruilong01: Simply defining __SNR_fstat is not enough, because there are architectures that lack __NR_fstat and __PNR_fstat, where __SNR_fstat is an empty macro definition, and therefore results in the error

masatake: Should we do for __SNR_newfstatat the same as you did for __SNR_fstat?

wuruilong01: I think it's needed.

masatake: Thank you. I will update the other #ifdefs.

PR #3971, opened 2024-03-27
DOC: parser-iPythonCell.r / input.py: Note about the percent format

westurner:
Possibly also worth noting somewhere that the linked docs have the percent format grammar as:

```
# %% Optional title [cell type] key="value"
```

https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-percent-format


masatake:
I don't understand why adding such a comment to a test input is worth it.
The *worth* is unclear to me. Users have no chance to see the comment.

westurner:
Fair point.

Hadn't realized there's an .ipynb ctags parser. Now I'll remember.

PR #3972, opened 2024-03-27
honor XDG_CONFIG_HOME environment variable on Windows as well

### LINKED ISSUE #3969: Support XDG specification on Windows
I am forced to work on windows, but I would like to build a unix-style development environment, without the overhead of WSL, or even Cygwin, MSYS2 (I want to work in PowerShell). For such a proper development environment, universal-ctags is a must.

I want to distinguish the distinguish the separate configuration of my environment via following XDG directory specifications. However, I recently found out that universal-ctags on windows does not look for option files in
- $ENV:XDG_CONFIG_HOME\\ctags\\
- $HOME\\.config\ctags\\
- $HOME\\.ctags.d\\
directories.
So I was forced to put an option file into $HOMEV\\.ctags.d\\ . This however furthers the dotfile creep in the home directory.

Why the XDG directory specification (and $HOME\\.ctags.d\\) is not supported in the peculiar case of universal-ctags? Many modern version of unix-style tools, that are cross-platform, supports XDG (and $HOME\\.ctags.d\\) on Windows as much as reasonably achievable. A good example for that is Neovim .

It should not be much difficulty to implement, right?

leleliu008:
MS Windows users could benefit from honoring XDG_CONFIG_HOME environment variable as well.

Although some people might thought it was not commonly used on MS Windows.

Actually, some well-known softwares honor this environment variable, as far as I known, git honors this environment variable on all platforms including MS Windows. git introduced this feature as earlier as 2015 (git/git@ea19289) .

Many users host their personal dotfiles on a GitHub repository, they just clone their repository to XDG_CONFIG_HOME, if software do not honor this environment variable, users should do more steps.

fixed #3969

### DIFF
--- a/Tmain/load-conf-files-under-default-xdg-config-home.d/run.sh
+++ b/Tmain/load-conf-files-under-default-xdg-config-home.d/run.sh
@@ -5,8 +5,6 @@ CTAGS=$1
 
 . ../utils.sh
 
-exit_if_win32 "$CTAGS"
-
 # $HOME/.config/ctags/*.ctags are loaded
 export HOME=./myhome
 export XDG_CONFIG_HOME=
--- a/Tmain/load-conf-files-under-xdg-config-home.d/run.sh
+++ b/Tmain/load-conf-files-under-xdg-config-home.d/run.sh
@@ -5,8 +5,6 @@ CTAGS=$1
 
 . ../utils.sh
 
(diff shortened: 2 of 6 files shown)


masatake (approved):

masatake:
Fine. Thank you.

PR #3974, opened 2024-03-31
Containerfile: new parser

### LINKED ISSUE #3970: Containerfile/Dockerfile parser
STORY: Users can parse Dockerfile and Containerfile with universal-ctags in order to navigate and review with tool support.

- `man Containerfile`:
  https://github.com/containers/common/blob/main/docs/Containerfile.5.md
- List of Dockerfile instructions; e.g. 1 or more `FROM`, `RUN`, `ARG`, `ENV`,: https://docs.docker.com/reference/dockerfile/
-
- There's a grammar for parsing docker image URLs:
  https://pkg.go.dev/github.com/distribution/reference#pkg-overview
- There's an OCI Image manifest (JSON) spec:
  https://docs.docker.com/reference/dockerfile/
- FWICS there is not (yet?) an PEG grammar for Dockerfile / OCI Containerfile
- FWIU there needn't be a PEG grammar to add something to `parsers/`
- It's been awhile since I've written any C; I would have to llm it and only then write tests for my rusty c code.
  - https://github.com/universal-ctags/ctags/blob/master/Units/parser-iPythonCell.r/default-formats.d/input.py

masatake:
Close #3970

### DIFF
--- a/Tmain/list-map-patterns.d/stdout-expected.txt
+++ b/Tmain/list-map-patterns.d/stdout-expected.txt
@@ -1,7 +1,7 @@
 ## all|grep LdScript
-#LANGUAGE     PATTERN
-LdScript      *.lds.S
-LdScript      ld.script
+#LANGUAGE      PATTERN
+LdScript       *.lds.S
+LdScript       ld.script
 ## LdScript
 #PATTERN
 *.lds.S
--- a/Tmain/list-roles.d/stdout-expected.txt
+++ b/Tmain/list-roles.d/stdout-expected.txt
@@ -49,6 +49,7 @@ CUDA           h/header          system             on      system header
 Cobol          S/sourcefile      copied             on      copied in source file
 CobolFree      S/sourcefile      copied             on      copied in source file
 CobolVariable  S/sourcefile      copied             on      copied in source file
+Containerfile  i/image           base               on      referenced as a base image in FROM directive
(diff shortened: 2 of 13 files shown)


masatake:
I used CPreprocessor as the base parser.
Using CPreprocessor as a base is a good choice for Dockerfile.in/Containerfile.in processed by podman. However, the choice can be troublesome for Dockerfile/Containerfile.

Supporting both  Dockerfile.in/Containerfile.in and Dockerfile/Containerfile in a parser is not easy.

We may have to have Containerfile parser and ContainerfileIn parser.
Sigh.

The behavior of podman for processing `\` is very different from what I expected.

```
$ cat Containerfile
From ubi8:latest
RUN echo hello \
                        

    world
ENV A \

ENV X Y
$ podman build -f Containerfile
STEP 1/3: FROM ubi8:latest
STEP 2/3: RUN echo hello     world
--> Using cache 49aca30d51520057b113038edd4ac22596325950f9f9e9095c2c796962dc9ce6
--> 49aca30d5152
STEP 3/3: ENV A ENV X Y
--> Using cache 510da2bbd3ec0dea46afcb4efc938ceed23ec435e904962722db4f858faff363
--> 510da2bbd3ec
510da2bbd3ec0dea46afcb4efc938ceed23ec435e904962722db4f858faff363
```

I withdraw this pull request.

PR #3977, opened 2024-04-03
build-sys: fix checking Windows platform with _WIN32 macro

Biswa96:
This replaces WIN32 with _WIN32 macro to check Windows platform.
    The correct one is _WIN32 as documented in the official documentation.
    https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -75,7 +75,7 @@
 
 /*  Hack for ridiculous practice of Microsoft Visual C++.
  */
-#if defined (WIN32) && defined (_MSC_VER)
+#if defined (_WIN32) && defined (_MSC_VER)
 # define chsize         _chsize
 # define open           _open
 # define close          _close
@@ -1739,7 +1739,7 @@ static void writeTagEntry (const tagEntryInfo *const tag)
 
 	DebugStatement ( debugEntry (tag); )
 
-#ifdef WIN32
+#ifdef _WIN32
 	if (getFilenameSeparator(Option.useSlashAsFilenameSeparator) == FILENAME_SEP_USE_SLASH)
 	{
 		Assert (((const tagEntryInfo *)tag)->inputFileName);
--- a/main/general.h
+++ b/main/general.h
@@ -14,7 +14,7 @@
 */
 #if defined (HAVE_CONFIG_H)
 # include <config.h>
-#elif defined (WIN32)
+#elif defined (_WIN32)
 # include "e_msoft.h"
 #endif
 
--- a/main/main.c
+++ b/main/main.c
@@ -552,7 +552,7 @@ extern int ctags_cli_main (int argc CTAGS_ATTR_UNUSED, char **argv)
 {
 	cookedArgs *args;
 
-#if defined(WIN32) && defined(HAVE_MKSTEMP)
+#if defined(_WIN32) && defined(HAVE_MKSTEMP)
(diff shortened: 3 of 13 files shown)


masatake:
I don't know Windows well. However, our own `WIN32` is defined in mk_mingw.mak and mk_mvc.mak.
So I guess our WIN32 has nothing to do with `_WIN32`. Just replacing it may cause trouble.

leleliu008:
@Biswa96 `WIN32` is a user defined macro https://github.com/universal-ctags/ctags/blob/master/mk_mingw.mak#L10, `_WIN32` is a predefined macro which is defined by MSVC or MINGW. what problem have you encountered?

Biswa96:
The issue was seen in geany project, here https://github.com/geany/geany/pull/3660

> `WIN32` is a user defined macro

It is not required at all and contradicts with official and well-known `_WIN32` macro.

Units/parser-lex.r/simple-lex.d/input.l
```diff
@@ -123,7 +123,7 @@
 #include <stdio.h>
 #include <string.h>
 
-#ifdef WIN32
+#ifdef _WIN32
```
masatake, line 126: This is a part of test input. Please don't touch this.

masatake:
Can we remove -DWIN32 from *.mak files?

Units/parser-lex.r/simple-lex.d/input.l
Biswa96: Would not the test fail if it is kept unchanged?

Biswa96:
> Can we remove -DWIN32 from *.mak files?

Yes, those can be removed also. I did not notice them at first.

Units/parser-lex.r/simple-lex.d/input.l
masatake: No. But this is a part of test input. Please don't touch this.
This change makes your original intent of this pull request unclear.

masatake:
Could you refer to https://github.com/geany/geany/pull/3660 in the commit log?

Biswa96:
I have added the geany pull request in commit message.

Units/parser-lex.r/simple-lex.d/input.l
Biswa96: I have removed that change from the commit.

Biswa96:
Here are some example of the compiler and linker error in geany which are fixed by this change.

```
../ctags/gnu_regex/regex_internal.h:424:11: fatal error: alloca.h: No such file or directory
  424 | # include <alloca.h>
      |           ^~~~~~~~~~

../ctags/main/routines.c:463:21: warning: implicit declaration of function 'lstat'; did you mean 'wstat'? [-Wimplicit-function-declaration]
  463 |                 if (lstat (file.name, &status) != 0)
      |                     ^~~~~
      |                     wstat

ld.exe: libctags.a.p/ctags_main_routines.c.obj:routines.c:(.text+0x7cc): undefined reference to `lstat'
```

masatake (approved):

masatake:
Can I ask you to change the prefix of the header of the commit log from "build.main,lex" to "build-sys" ?

```
build-sys: fix checking Windows platform with _WIN32 macro
```

k-takata (approved):

PR #3978, opened 2024-04-03
configure.ac: add a new judgment condition for the use of flto.

iLeeWell:
In addition to non-cross-compilation and compiler support for flto, add the condition that flto can only be enabled by actively using --enable-lto option.

I found a segment fault when ctags generates tags against parser_overflow.c (https://github.com/llvm/llvm-project/blob/main/clang/test/Parser/parser_overflow.c). The reason is that there are 16384 pairs of curly braces in the file (rare, deliberate and even tricky, but not wrong), and ctags, with flto optimization turned on, inflates the stack frame size with a large number of IPOs, which easily exceeds the default system specified stack size. The fastest solution is to disable flto, or to expand the stack size by the user (ulimit -s unlimted). However, from the point of view of prioritizing stability over performance, the flto optimization should not be used by default (even if it is supported by the compiler), and the choice of whether or not to turn it on should be left to the user, although there may be better solutions.

This is my rough opinion so far.


leleliu008:
I think you want to change to `"x$enable_lto" = "xyes"`

leleliu008 (commented):
if user pass --disable-lto, then enable_lto_option=yes also would be run

leleliu008:
`"x$enable_lto" != "xno"` means `--disable-lto` isn't specified, in other words, if neither `--enable-lto` nor `disable-lto` is specified, then it would check if the comipler support LTO.

iLeeWell:
Thanks for your comments, learned a lot, I will follow your suggestions to make changes.

iLeeWell:
You are right, thanks for your comments, i will make changes immediately.

leleliu008:
you need to run `git rebase -i ` command to delete first one commit and leave last one commit.

iLeeWell:
i tried, but it seems not work as expected..., when i use git rebase -i and then push, my fork ctags reject so i add -f to force this. I'll study it further.

masatake:
The change looks good.

I would like you to add the document about `--enable-lto` to docs/autotools.rst.

```
diff --git a/docs/autotools.rst b/docs/autotools.rst
index 0c0693105..2fdbf3639 100644
--- a/docs/autotools.rst
+++ b/docs/autotools.rst
@@ -71,6 +71,11 @@ To completely change the program's name run the following:
 
 Please remember there is also an 'etags' installed alongside 'ctags' which you may also want to rename as shown above.
 
+
+Link time optimization
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+...TBW...
+
 Cross-compilation
 ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 
```

iLeeWell:
Thank you so much. @masatake @leleliu008

iLeeWell:
> The change looks good.
> 
> I would like you to add the document about `--enable-lto` to docs/autotools.rst.
> 
> ```
> diff --git a/docs/autotools.rst b/docs/autotools.rst
> index 0c0693105..2fdbf3639 100644
> --- a/docs/autotools.rst
> +++ b/docs/autotools.rst
> @@ -71,6 +71,11 @@ To completely change the program's name run the following:
>  
>  Please remember there is also an 'etags' installed alongside 'ctags' which you may also want to rename as shown above.
>  
> +
> +Link time optimization
> +,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
> +...TBW...
> +
>  Cross-compilation
>  ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
>  
> ```

ok, i will do this tomorrow.

iLeeWell:
> > The change looks good.
> > I would like you to add the document about `--enable-lto` to docs/autotools.rst.
> > ```
> > diff --git a/docs/autotools.rst b/docs/autotools.rst
> > index 0c0693105..2fdbf3639 100644
> > --- a/docs/autotools.rst
> > +++ b/docs/autotools.rst
> > @@ -71,6 +71,11 @@ To completely change the program's name run the following:
> >  
> >  Please remember there is also an 'etags' installed alongside 'ctags' which you may also want to rename as shown above.
> >  
> > +
> > +Link time optimization
> > +,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
> > +...TBW...
> > +
> >  Cross-compilation
> >  ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
> >  
> > ```
> 
> ok, i will do this tomorrow.

Supplementary description of Link time optimization. @masatake

masatake (approved):

masatake:
@iLeeWell Thank you very much.

Could you change the subject of the commit log to "build-sys: make lto optional"  (or something better) ?

iLeeWell:
> @iLeeWell Thank you very much.
> 
> Could you change the subject of the commit log to "build-sys: make lto optional" (or something better) ?

Of course.

masatake:
Thank you.

PR #3984, opened 2024-04-14
dsl: allow to specify a default value in $ and & operators

masatake:
$ and & return #f if a specified field is not found in a tag. With this change, you can make the operators return a value other than #f in the situation.

You can write

	($ field default-value)

as a shorthand of

	(or ($ field) default-value)
.


masatake:
Reporting the coverage to coverall failed.
Does anyone know the way to fix it?

leleliu008:
codecov-cli report following:

```
{'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

That's to say codecov invoke the Github Actions API, but Github Actions API has the rate limitation. To remove the rate limitation, you should configure a `Codecov upload token`.

**how to get a  Codecov upload token:**

https://github.com/codecov/codecov-cli?tab=readme-ov-file#how-to-get-an-upload-token

Another solution is to use old version of  `codecov/codecov-action`, but I am not try it:

https://github.com/codecov/codecov-action/issues/598#issuecomment-1030074427

leleliu008:
I found `CODECOV_TOKEN` environment variable already has been configured, is it expired?

masatake:
> I found CODECOV_TOKEN environment variable already has been configured, is it expired?

I updated a few days ago.

k-takata:
`CODECOV_TOKEN` was updated 3 days ago. (Unintentionally?)
I've updated it again. Let's see if it works.

Currently we are using the codecov CLI directly. Is it better to using `codecov/codecov-action`?

k-takata:
Hmm, still failing.

k-takata:
https://github.com/universal-ctags/ctags/actions/runs/8680720408/job/23960447081#step:18:22
> ```
> env:
>   CC: gcc
>   BUILDDIR: ubuntu-20.04-gcc
>   SANDBOX_CASES: sandbox,sandbox-crash,sandbox-default-req,sandbox-unknown-submode
>   DEBIAN_FRONTEND: noninteractive
>   CODECOV_TOKEN: 
> ```

I think it is strange that `CODECOV_TOKEN` is shown as empty.
In my understanding, it is normally shown masked: `CODECOV_TOKEN: ***`.

leleliu008:
https://github.com/universal-ctags/ctags/settings/secrets/actions/CODECOV_TOKEN has no value filled

k-takata:
In my understanding, it will be always shown empty for security reasons when you updating the secrets.

leleliu008:
That's true.

k-takata:
It looks fine with codecov/codecov-actions: #3986
https://github.com/universal-ctags/ctags/actions/runs/8733393588/job/23961993899

masatake:
The coverall comes back! Thank you.

masatake:
@AmaiKinono I expect the feature I added to readtags via this pull request can simplify readtags filter expressions in citre. You can expect this feature to be available if the version of readtags is greater or equal to 6.2.0.

PR #3989, opened 2024-04-22
Update pascal.c

HoTschir:
support for inline or one line comments added

### DIFF
--- a/parsers/pascal.c
+++ b/parsers/pascal.c
@@ -8,6 +8,8 @@
 *   including some extensions for Object Pascal.
 */
 
+// HTs 2024-04-21 handle OneLineComments
+
 /*
 *   INCLUDE FILES
 */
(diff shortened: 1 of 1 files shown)


parsers/pascal.c
```diff
@@ -178,6 +180,10 @@ static void findPascalTags (void)
 
 		if (c == '\0')  /* if end of line */
 		{
+			if (incomment && comment_char == '/') 
+				{
```
masatake, line 184: Is it possible to align the `if,` `{, `and `}` ?

masatake:
Thank you.

Could you change the commit header to `Pascal: add support for inline or one line comments` (or something better) ? 

`Update pascal.c` doesn't help people understand your change. 


I would like you to add a test case under Units/parser-pascal.r/.
Here is the document explaining the way to write a test case:
https://docs.ctags.io/en/latest/testing-parser.html#id1

masatake:
You use your `master` branch for improving ctags. I recommend you to make a topic branch.
See https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows .

masatake:
@HoTschir Have you lost interest? 
I'm looking forward to your update.

masatake:
I'll take over this.

PR #3993, opened 2024-05-07
JavaScript: Improve support for contextual keywords as identifiers

b4n:
As the title says, better handle contextual keywords as function, property and method identifiers.

All these are probably fairly rare, but some did appear in e.g. a minified jQuery, leading to emitting null tags.  Now, only the actually empty-named property it contains (`""://…`) leads to emitting a null tag.

This mostly builds up the `get` and `set` handling for all contextual keywords valid in these situations.

### DIFF
--- a/Units/parser-javascript.r/contextual.d/expected.tags
+++ b/Units/parser-javascript.r/contextual.d/expected.tags
@@ -1,0 +1,95 @@
+Function	input.js	/^  Function() { return 0x32 }$/;"	m	class:Syn3
+Function	input.js	/^  Function() { return Syn2; },$/;"	m	class:Syn2
+Function	input.js	/^  Function:function() { return Syn1; },$/;"	m	class:Syn1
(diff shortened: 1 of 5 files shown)


masatake:
About the new test case, adding `args.ctags` file with contents `--sort=no` makes `expected.tags` easier to read.

b4n:
> About the new test case, adding `args.ctags` file with contents `--sort=no` makes `expected.tags` easier to read.

I would agree in general, but I think that in the case of *contextual.d* it's pretty handy that it's sorted so you can easily see that there is basically 4 identical symbols of each name (but for the instances and class names, and the `length` property I could just get rid of as well).
If you prefer sorted though I'll add it not problem.

For the other test case sure, yet it's already sorted in the source so won't change a thing for now :)

jafl (approved):

masatake (approved):

masatake:
@jafl I'm going to merge this as you approved. So we can work on the topic discussed at #3363.

PR #3999, opened 2024-05-13
matlab: A couple fixes for corner cases

b4n:
Fix for corner cases the parser previously used in Geany was tested against that the uctags one fails at.

:warning: here again, I have no clue about MatLab, so I'm just trusting the input makes sense, which isn't necessarily true.  @portalgun maybe?


masatake (approved):

masatake:
Thank you.

PR #4008, opened 2024-05-19
jscript: Fix representation of held tokens

### LINKED ISSUE #4005: javascript: segmentation fault
The name of the parser: jscript

```
$ ctags --options=NONE -R
```

The content of input file:

```js
test = ({}
) => {}
```

Error response:

```
segmentation fault  ctags --options=NONE -R
```

It works for:

```js
test = ({}) => {}
```

The version of ctags:

```
$ ctags --version
Universal Ctags 6.1.0, Copyright (C) 2015-2023 Universal Ctags Team
  Compiled: May 12 2024, 01:01:34
  URL: https://ctags.io/
  Output version: 0.0
  Optional compiled features: +wildcards, +regex, +gnulib_fnmatch, +gnulib_regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +case-insensitive-filenames, +packcc, +optscript, +pcre2
```

How do you get ctags binary:
```
$ brew install universal-ctags
```

b4n:
Also fill in representation when emitting a held token, e.g. after an implicit semicolon.

Fixes #4005.


jafl (approved):

b4n:
@masatake I kind of knew that, but if I was actually more or less waiting for you was because there was no particular rush and you often have useful review comments to make; so I wanted to give you all the time you could need to have a look at this if you wanted.
If you'd like e.g. me to merge stuff without giving you some time to look at it, I'm OK with it as well -- just tell me you don't actually want to review *everything* and are happier if things happen without burden on your shoulders :slightly_smiling_face: 

Or less seriously, just stop making useful comments so I stop valuing your review as much :laughing:

masatake:
I don't have a strong idea about how long we should wait to merge after getting approval from a parser maintainer. I added the comment and merged this pull request in a bit of a hurry because we have received some issues with the JavaScript parser these days; I'm afraid that the fixes that you (@b4n and @jafl) are working on conflict. In my experience, merging quickly and serializing commits helps developers. Now I understand your expectation; I should review pull requests and show the result, including approval, explicitly.

b4n:
> I don't have a strong idea about how long we should wait to merge after getting approval from a parser maintainer.

Arbitrary times don't really work in my experience, as I for example usually don't have the same amount of time to devote to a particular project every week.  It might be different for you though, but the thing is I don't expect an absolute value.

Applying our own judgment is probably important, and indeed if 2 PRs conflict it's best not to wait too long, and remember we can actually always fix things if they got overlooked (although I'd obviously rather avoid introducing as many bugs as possible, but it's never gonna be perfect).

> I should review pull requests and show the result, including approval, explicitly.

Well, no, you don't have to, but I would like to know if you don't *want* to indeed :slightly_smiling_face:.  And if you did look at it and found it good, approval helps indeed (but I can also easily understand that you didn't to a thorough enough review to warrant actual approval).

In the end, I think we can have a "feel" that e.g. here indeed if @jafl and I were happy about the change and you didn't complain it's fine.
BTW, it's also a possibility: if you'd really want to review something you say so and we wait, otherwise we use our judgment.  And actually now I think about it I believe you're already doing it :)

PR #4020, opened 2024-06-14
Add nix language parser

bsima:
This is a work in progress.


Units/simple-nix.d/args.ctags
masatake, line 1: Put 

```
--sort=no
```

at leasat.

Units/simple-nix.d/expected.tags
masatake, line 1:
```
make units LANGUAGES=nix
```

or

```
make units LANGUAGES=Nix
```

may generate FILTERED.tmp. It is a good starting point.

source.mak
masatake, line 1: Build scripts for Windows are needed. I will write how to generate it.

masatake:
To docs/news/HEAD.rst, please add:
```
* Nix *optlib* by YOURNAME 
```

masatake:
Thank you.

source.mak
masatake: At `win32` directory, run `make -B`. The command updates the scripts. Add them to your commit.

Units/simple-nix.d/expected.tags
bsima: hm I wasn't able to get a FILTERED.tmp with any content

bsima: I was able to create a simple expected.tags with

    ./ctags -o - --options=optlib/nix.ctags --sort=no  Units/simple-nix.d/input.nix  > Units/simple-nix.d/expected.tags

But this is *very* simple, I should expand the `input.nix` to have more real-world code.

bsima: Oh I see this in STDERR.tmp

    ignoring ./Units/simple-nix.d/input.nix (unknown language/language disabled)

probably this is why my FILTERED.tmp won't populate with content.

bsima:
I updated the commit with your comments. Next I'll work on expanding the test case to actually test real-world nix code: I'll make a small library of functions and a couple builders with different formatting.

optlib/nix.ctags
```diff
@@ -0,0 +1,40 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301,
+# USA.
+#
```
masatake, line 21: If there is a web page explaining the Nix language well, put the URL here like: https://github.com/universal-ctags/ctags/blob/8976ec3d2c48ec862593910388ae3e55f0e20a3f/optlib/iPythonCell.ctags#L10

masatake (commented):
You may have to put "NixParser" here: https://github.com/universal-ctags/ctags/blob/8976ec3d2c48ec862593910388ae3e55f0e20a3f/main/parsers_p.h#L135 .

After doing "make", you can run the test cases with "make units LANGUAGES=Nix".

Units/simple-nix.d/input.nix
```diff
@@ -0,0 +1,9 @@
+{ stdenv, fetchgit }:
+
+stdenv.mkDerivation {
+  name = "hello";
+  src = fetchgit {
+    url = "https://example.com";
+    hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
+  };
+}
```
bsima, line 9: still TODO: expand on this to have something more like real-world nix code

bsima:
I added some better test nix code, I tried to look at some of my personal nix code and copy a few formatting patterns that I use, which is similar to the style you see in upstream nixpkgs afaik. So I would expect this optlib to work reasonably well everywhere.

I thought I would need a multiline regex for the `multilineFunc` and `multiline_attrset` but I don't? It just worked? Maybe I don't understand regex as well as I thought?

Anyway sorry for taking so long to finish this. But I've been using this optlib in my personal code, which runs over nixpkgs and indeed finds most definitions I need, and my work code, which has thousands of lines of custom nix and it seems to work more often than not.

masatake:
Could you rebase your chnages on the latest code?

optlib/nix.ctags
```diff
@@ -0,0 +1,42 @@
...
+# References:
+#
+# - https://nix.dev/manual/nix/latest/language/
+#
+--langdef=Nix
+--map-Nix=+.nix
+
+# Index packages when we see "name = <tag>" or "pname = <tag>"
+--kinddef-Nix=p,package,package definition
```
masatake, line 29: As far as reading the test case, `package` is not tested at all.

optlib/nix.ctags
```diff
@@ -0,0 +1,42 @@
...
+--regex-Nix=/[p?]name\s*=\s*"(\w+)"/\1/p/
+
+# Functions have args, so look for a : right of the =
+# (This should probably be a multiline regex.)
+--kinddef-Nix=f,function,function definition
+--regex-Nix=/(\S+)\s*=\s+\w+:/\1/f/
+
+# Attrs definitions just have =, but only index if they have >=4 chars,
```
masatake, line 37: The reason for rejecting short names looks weak.

One of the mottos of Universal Ctags design is to "prepare raw information to ctags client tools like vim when ctags doesn't know what's incorrect."

Rejecting can be down in later.

masatake (changes requested):

masatake:
The `multiline regex` meta parser is a toy in many situations.
The multi-table byte-oriented meta parser (--_tabledef-<LANG>, and --_mtable-regex-<LANG>) is the way to go.

I don't know the Nix language at all. However, I guess you may want to fill out the scope fields for tag entries. Without it, a client tool like Vim cannot draw a tree, which is useful for navigation.

```c
struct point {
   int x;
   int y;
};
```

For the input of C language, ctags can emit the following tags:
```
point	input.c	/^struct point {$/;"	s	file:
x	input.c	/^  int x;$/;"	m	struct:point	typeref:typename:int	file:
y	input.c	/^  int y;$/;"	m	struct:point	typeref:typename:int	file:
```

From the tags output, `x` is a part of `point`.


```
  attrset = {
    foo = "bar";
  };
```

The current nix parser doesn't extract `foo` because the name is too short.
I guess an ideal Nix parser may emit:
```
attrest	input.nix	/^  attrset = {$/;"	a
foo	input.nix		;"	a	attr:attrest
```

Too many tags may not be a problem if the scope fields are filled out well.

However, these improvements can be made after merging this pull request.

optlib/nix.ctags
bsima: doh, good catch, i fixed the regex here

optlib/nix.ctags
bsima: ok, removed that part

bsima:
Ah, scoped tags is a very good idea. It's something I would need if I want to
continue using tags in nixpkgs. Currently I just scan through the list of
duplicated tags for the correct one, but scopes is a much better solution. I
will read about scopes tomorrow.

PR #4045, opened 2024-08-02
Fix quotes & option syntax in manpage

JaSpa:
* use backticks instead of single quotes
* use double dashes for long option name


masatake:
Thank you.

PR #4052, opened 2024-08-16
iniconf: Allow dot and dash for ini keys

techee:
The iniconf parser can be used as a poor man's TOML parser which can contain '.' and '-' inside keys. Also, since there's no official ini specification, I can imagine some ini dialects allow this too.

I'm now aware of https://github.com/universal-ctags/ctags/pull/3509 as mentioned in https://github.com/geany/geany/pull/3934 but right now we don't use PEG parsers in Geany and we'd have to evaluate its performance so we'll probably stick with this parser for TOML for now.


masatake:
Thank you.

Issue #4056, opened 2024-08-22
SystemVerilog typedef not recognized if `include macro used


dawpud:
The name of the parser:
SystemVerilog

The command line you used to run ctags:

```
$ ctags --options=NONE -f - test.sv
```

The content of input file:

```verilog
typedef enum {
  `include "test.txt"
  OTHER_VAL
} my_enum;

my_enum e;
```

The tags output you are not satisfied with:

```
`include	test.sv	/^  `include "test.txt"$/;"	T
`include	test.sv	/^  `include "test.txt"$/;"	c	typedef:`include
e	test.sv	/^my_enum e;$/;"	r
```

The tags output you expect:

```
`include	test.sv	/^  `include "test.txt"$/;"	T
`include	test.sv	/^  `include "test.txt"$/;"	c	typedef:`include
e	test.sv	/^my_enum e;$/;"	r
OTHER_VAL	test.sv	/^  OTHER_VAL$/;"	c	typedef:my_enum
my_enum	test.sv	/^} my_enum;$/;"	T
```

The version of ctags:

```
$ ctags --version
Universal Ctags 6.1.0(e4cb2a0), Copyright (C) 2015-2023 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Dec 29 2023, 05:50:45
  URL: https://ctags.io/
  Output version: 0.0
  Optional compiled features: +wildcards, +regex, +gnulib_regex, +iconv, +option-directory, +xpath, +json, +interactive, +yaml, +packcc, +optscript
```

How do you get ctags binary:

Unix binary from nightly builds:
uctags-2023.12.29-linux-x86_64.tar.xz

---

As seen in the results above, as soon as the `include macro is used within the typedef enum, the rest of the enum values stop getting parsed, and the enum itself is not listed in the output tags file. I appreciate any help you can provide! Thank you very much!

masatake:
Just for the input, the following patch solves the issue;

```
diff --git a/parsers/verilog.c b/parsers/verilog.c
index 0b107687b..e7555eb60 100644
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1612,6 +1612,7 @@ static int pushEnumNames (tokenInfo* token, int c)
 	if (c == '{')
 	{
 		c = skipWhite (vGetc ());
+		c = skipMacro (c, token);
[...]

PR #4058, opened 2024-08-24
verilog: skip compiler directives in enum definition (#4056)

hirooih:
This is a fix for #4056.


masatake:
In addition to pushEnumNames, what do you think about fixing pushMembers now?

```
diff --git a/parsers/verilog.c b/parsers/verilog.c
index 0b107687b..c33cd6221 100644
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1652,6 +1652,7 @@ static int pushMembers (tokenInfo* token, int c)
 		{
 			verilogKind kind = K_UNDEFINED;	// set kind of context for processType()
 			bool not_used;
+			c = skipMacro(c, token);
 			if (!isWordToken (c))
 			{
 				VERBOSE ("Unexpected input: %c\n", c);
```

As far as reading processType, skipPastMatch can be another candidate for fixing. But I'm not sure.

hirooih:
> In addition to pushEnumNames, what do you think about fixing pushMembers now?

I also thought about it. However unlike enum, struct/union has few use-cases that require `` `include``.
However, I decided to modify it this time from you comment.

Unlike enum, we need to do one more process when there is an `` `include`` at the end of a block.

masatake (approved):

masatake:
Thank you.

PR #4063, opened 2024-08-29
GitHub Actions: fix syntax error of building-with-pegof.yml

leleliu008:


.github/workflows/building-with-pegof.yml
```diff
@@ -43,8 +43,8 @@ jobs:
       - run: make -j2
 
       - run: ./ctags --list-features | grep pegof
+      - run: ./ctags --list-languages=_packcc | tr '\n' ','
```
masatake, line 46: The prefix "Targets:" is needed to help the developers (us) remember the meaning of the message.
How about adding the following line before the line you added?
```
      - run: print "Target: "
```

masatake: But I wonder why the original caused a syntax error.

leleliu008: because `Target: ` ':' follow a empty char. 

```
 - run: print "Target: "
```

This line has two map, one is `run:`, one is `Target:`. This is not allowed in YAML.

masatake: Oh, I see. Today I hate computers even more!

masatake:
Thank you.

PR #4066, opened 2024-08-31
C++: add unicode identifier name support

popkc:
Ctags currently does not support Unicode identifiers in c++, but I found that it only requires modifying one line of code to make it support them.

### DIFF
--- a/parsers/cxx/cxx_parser_tokenizer.c
+++ b/parsers/cxx/cxx_parser_tokenizer.c
@@ -24,7 +24,7 @@
 
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


masatake:
The C++ related test cases failed on the may platforms.
If you use autotools to build ctags, the failures may be reproduced with following command line:
```
$ ./configure --enable-debugging
$ make units LANGUAGES=C++
```

See https://docs.ctags.io/en/latest/testing-parser.html#units-test-facility about the test harness used in ctags.

masatake:
I'll close this because the tests failed with this pull request, which has not been updated for more than three months.

PR #4074, opened 2024-09-12
readtags: refactor for searching multiple tag files

masatake:


masatake:
I should add more test cases.

PR #4078, opened 2024-09-21
 build-sys: eliminate READTAGS_DSL condition 

masatake:
Just for testing with GitHub Actions


masatake:
It was easier than I was afraid of.

PR #4080, opened 2024-09-25
Update libreadtags

masatake:


masatake:
before:
```
$ /bin/time ./readtags -t ~/.citre/kernel-9_1.tags  -l > /dev/null
13.30user 0.43system 0:13.82elapsed 99%CPU (0avgtext+0avgdata 3870520maxresident)k
0inputs+0outputs (0major+60587minor)pagefaults 0swaps
```

after:
```
/bin/time ./readtags -t ~/.citre/kernel-9_1.tags  -l > /dev/null
10.64user 0.51system 0:11.21elapsed 99%CPU (0avgtext+0avgdata 3870536maxresident)k
0inputs+0outputs (0major+60571minor)pagefaults 0swaps
```

masatake:
@leleliu008 Can we do something the failure on "MSYS2 / testing (CLANG32)" ?

leleliu008:
The package name is not changed, https://packages.msys2.org/base/mingw-w64-jansson

It might be lost the download url. I will report to MSYS2 TEAM.

I think you can ignore this. because other environmtent all green.

leleliu008:
as per https://www.msys2.org/news/ `MSYS2/CLANG32` environment packages are being removed as of 2024-09-23 , `MSYS2/CLANG32` environment is not supported any more.  we should remove it as well.

masatake:
Oh, I see.

Issue #3727, opened 2023-05-15
[RFC] ban fpos_t


masatake:
fpos_t has caused making our code complicated.

Will ctags deal with files larger than sizeof(long)?
I think it will not.

So I would like to use only ftell/fseek, not together with fsetpos/fgetpos.

Anjuta-ctags took the way I like.

https://gitlab.gnome.org/Archive/anjuta/-/commit/91f5e7fe0a6e4d548f4888711cdaaf7bedab6f85
https://gitlab.gnome.org/Archive/anjuta/-/commit/0adbb5451148027bc6cb283acbfd98c5dc79aad0

What do you think?

k-takata:
You may need to consider that `long` is 32 bits on Windows.
Windows has the 64-bit version of fseek:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=msvc-170
If we don't need to consider the files that are larger than 2 GB, simply using `long` should be okay.
Otherwise, the source code of Vim might be helpful:
https://github.com/vim/vim/blob/c0da540466c89e388e7a15a12bab2f9fc42d9671/src/vim.h#L384-L416

masatake:
@k-takata, thank you for the suggestion. Only about input, I think 2GB is large enough.
For output, 2GB is too small.

k-takata:
Oh, I remember that we've already done this in readtag:
https://github.com/universal-ctags/ctags/blob/3547021a009d3c4b89f17dfb7e6ffd4cacacfa0c/libreadtags/readtags.c#L111-L135

masatake:
Nice! We should reuse it.

PR #4097, opened 2024-10-08
Fix promises at EOF

b4n:
`getInputLineOffset()` was giving very wrong results at EOF, leading to overlong promises, leading to crashes in building the narrowed input stream (see e.g. newly introduced test case, and https://github.com/geany/geany/issues/3939).

I'm unsure if this is a good way to fix the issue, but I couldn't find a better one; but admittedly I didn't poke around this code for a long time, and it got a fair bit more complicated with its newer features, so I might have missed something.


masatake:
Thank you.

> it got a fair bit more complicated with its newer features


I agree with you. Some issues I have opened these days are related to this area.
Eventually, I must revise the whole area after merging this pull request.

I will review this.

Units/parser-php.r/wp-guest.d/expected.tags
```diff
@@ -0,0 +1,14 @@
+top	input.php	/^<div id="top">$/;"	I	line:3	language:HTML	roles:def
+inner	input.php	/^ <div class="inner">$/;"	c	line:4	language:HTML	roles:attribute
+logo	input.php	/^<div id="logo">$/;"	I	line:6	language:HTML	roles:def
+pages	input.php	/^<div id="pages">$/;"	I	line:12	language:HTML	roles:def
+menubar	input.php	/^<ul class="menubar" id="menubar">$/;"	c	line:13	language:HTML	roles:attribute
+menubar	input.php	/^<ul class="menubar" id="menubar">$/;"	I	line:13	language:HTML	roles:def
+body	input.php	/^<div id="body">$/;"	I	line:26	language:HTML	roles:def
+content	input.php	/^  <main id="content">$/;"	I	line:27	language:HTML	roles:def
+post	input.php	/^  <article class="post" id="post-<?p/;"	c	line:31	language:HTML	roles:attribute
+post-<?p	input.php	/^  <article class="post" id="post-<?p/;"	I	line:31	language:HTML	roles:def
```
b4n, line 10: I have a follow-up to fix the HTML promise range from the PHP lexer to fix this spurious `<?p`, but it currently depends on these changes so I'll post it after this is sorted out (no pressure, just saying that indeed, this tag is not exactly as it should be).

masatake: Could you paste this comment into the commit log for this commit?
So people can understand this commit from its log only with 'git log'.

masatake:
I'm very sorry to be late to respond.
I cannot find enough time to work on this item.

Though I introduced the original bug, let me report a case where this pull request doesn't work this time.

input.html
```
<%=
```
makes enter an infinite loop:

```
~/bin/ctags --options=NONE --verbose -o - --extras='*' input.html
```

b4n:
> I'm very sorry to be late to respond.
> I cannot find enough time to work on this item.

Don't worry, I understand only too well :)

> Though I introduced the original bug, let me report a case where this pull request doesn't work this time. […]

I don't think it's a bug in this PR, but rather an existing issue in the HTML parser, and in the fact that the HTML and PHP parser both make promises for each other.  I've added a commit that fixes this, but the issue is a bit deeper: with both parser triggering each other, there is a risk that such an infinite loop arises if both parsers don't agree on what's theirs to deal with.  Here, the issue was that the HTML parser fed the PHP one JSP/ASP tags it doesn't accept, and so the PHP parser gave them back to the HTML parser, so on and so forth.
Maybe a more robust solution would be for a parser never to trigger the triggering parser at offset 0, because that's bound to create this problem.  E.g. there might need to be a safeguard at some point that say "if I got called by parser X, I can't ask it to parse this from offset 0".  This might need to be even generalized as "if I got called by a parser, I **must** handle at least one character before triggering another parser" to avoid more complex loops where e.g. parser A triggers parser B that triggers parser C that triggers A again.

However, as this is likely a pretty rare issue limited to possibly the PHP/HTML couple, maybe a more localized fix would be fine -- if there's a way to know the input is from a promise, it might be easy.  Or at the promise layer, refuse 2 promises starting at the same offset?

b4n:
Maybe even something like this?
```diff
diff --git a/main/promise.c b/main/promise.c
index 01e84f5ca..428742c73 100644
--- a/main/promise.c
+++ b/main/promise.c
@@ -96,6 +96,17 @@ int  makePromise   (const char *parser,
 			return -1;
 	}
 
+	for (int i = 0; i < promise_count; i++)
+	{
+		p = promises + i;
+		if (p->startLine == startLine && p->startCharOffset == startCharOffset)
+		{
+			error (WARNING, "Ignoring promise at line %lu offset %lu because there is already an existing one at this location\n",
+			       startLine, startCharOffset);
+			return -1;
+		}
+	}
+
 	if ( promise_count == promise_allocated)
 	{
 		size_t c = promise_allocated? (promise_allocated * 2): 8;
```
**beware** that I know very little about this and didn't dig deep, it might drop legitimate promises e.g. if the offset is relative to the reduced input rather than the input file itself.  But this illustrates a potential solution anyway.

masatake:
(I have not read the comments yet. I'm currently running ctags on the entire source code in Fedora now. So I have many chances to find bugs.)

input.php
```
<?=
```
makes enter an infinite loop:

```
~/bin/ctags --options=NONE --verbose -o - --extras=g input.php
```

masatake:
Instead of fcab90a, the following change also passes the test cases (parser-php.r/wp-guest.d)
```
commit d3ea3a8c17861e2e7695ffd4d057533362993318
Date:   Tue Nov 12 18:42:44 2024 +0900

    TEMP

diff --git a/main/read.c b/main/read.c
index 67586051d..f975ba7d7 100644
--- a/main/read.c
+++ b/main/read.c
@@ -160,12 +160,12 @@ extern int getInputLineOffset (void)
         * So the way to calculate the offset at the end of file is tricky.
         */
        ret = (mio_tell (File.mio) - (File.bomFound? 3: 0))
-           - getInputFileOffsetForLine(File.input.lineNumber);
+           - getInputFileOffsetForLine(File.input.lineNumber) - File.ungetchIdx;
    }
    else
    {
        /* At the first line of file. */
-       ret = mio_tell (File.mio) - (File.bomFound? 3: 0);
+       ret = mio_tell (File.mio) - (File.bomFound? 3: 0) - File.ungetchIdx;
    }
 
    return ret >= 0 ? ret : 0;
```

When PEG-based parsers run as guests, they don't work at all.
However, it has nothing to do with the bug initially pointed out by this pull request.

masatake:
I have inspected the issue behind this pull request.

Finally, I found a way to fix it.
This is a prompt report.

Prerequisite knowledge
A. In getInputLineOffset(), we use mio_tell() and getInputFileOffsetForLine().
B. We use 2 Mio objects when running a promise (File.mio and BackupFile.mio)

Consider input.html, `<HTML><script>var x =  f();</script></HTML>`.
The HTML parser makes a promise that runs the JavaScript parser.
Before ctags runs the HTML parser, ctags loads the contents into File.mio.
After ctags runs the HTML parser on File.mio, ctags doesn't close File.mio.
Instead, it moves the file objects pointed by File.mio to BackupFile.mio.
The ctags ask Mio to make a memory type stream with the contents `var x =  f();`.
The ctags sets the memory stream to File.mio.

At this time, File.mio opens `var x =  f();` and BackupFile.mio opens input.html.

Then ctags runs the JavaScript parser on File.mio.
Finally, ctags closes all mio.


Consider we are at the time when ctags runs JavaScript parser as a guest.
getInputLineOffset() assumes BackupFile.mio as the context of input.
mio_tell() assumes File.mio as the context of input.

image

There was a gap between the functions.

So the offset calculation failed when running a guest parser.

The fix on my local PC is complicated because we use fpos_t behind MIOPos.
We cannot use + and - operators on the data of fpos_t.
We have to use mio_seel and mio_tell to convert fpos_t to off_t.
We can use + and - operators on the data of off_t.

Though the fix on my local PC is complicated, I will make a pull request based on it.
However, eventually, I will throw fpos_t away as I opened #3727.

@b4n, I would like to take over this pull request.

b4n:
@masatake thanks for taking the time to dive deeper in this! And sure, if you have a better fix go ahead :)

Issue #4096, opened 2024-10-08
TOML: infinite loop


masatake:
```
$ curl -LO https://github.com/BurntSushi/toml/raw/refs/heads/master/internal/toml-test/tests/invalid/encoding/bad-utf8-at-end.toml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   242  100   242    0     0   1901      0 --:--:-- --:--:-- --:--:--  1901
$ timeout --verbose 10 ~/bin/ctags -o - ./bad-utf8-at-end.toml 
timeout: sending signal TERM to command ‘/home/yamato/bin/ctags’
$
```

masatake:
Putting `--exclude=bad-utf8-at-end.toml` to your .ctags.d/foo.ctags is a workaround.

masatake:
The parser is broken. I replaced `|` in ABNF with `/` in PEG without enough consideration.
Things are not so simple.

I will disable TOML and Cargo subparser in the next release.

masatake:
TOML and Cargo parsers are disabled in v6.2.0.

PR #4099, opened 2024-10-13
iniconf: some more adjustments for parsing TOML

techee:
These two patches:
- improve parsing of TOML `[[arrays_of_tables]]`
- support `'single quoted keys' = true` and `"double quoted keys" = true`

See the individual commits for more details.

### DIFF
--- a/Units/parser-iniconf.r/toml-support-in-geany.d/args.ctags
+++ b/Units/parser-iniconf.r/toml-support-in-geany.d/args.ctags
@@ -1,0 +1,3 @@
+--fields=+lK
+--langmap=Iniconf:.toml
+--sort=no
--- a/parsers/iniconf.c
+++ b/parsers/iniconf.c
@@ -32,10 +32,13 @@
 #include "subparser.h"
 #include "vstring.h"
 
+#define TOML_QUOTED_KEY_CHAR(c) ( (c) == '"' || (c) == '\'' )
+
 static bool isIdentifier (int c)
 {
-    /* allow whitespace within keys and sections */
-    return (bool)(isalnum (c) || isspace (c) || c == '_' || c == '-' || c == '.');
+	/* allow whitespace within keys and sections */
+	return (bool)(isalnum (c) || isspace (c) || c == '_' || c == '-' || c == '.'
+		|| TOML_QUOTED_KEY_CHAR(c));
 }
 
 static bool isValue (int c)
--- a/parsers/iniconf.c
+++ b/parsers/iniconf.c
@@ -130,7 +130,8 @@ static void findIniconfTags (void)
 		if (*cp == '[')
 		{
 			++cp;
-			while (*cp != '\0' && *cp != ']')
+			/* look for the final ] to support TOML [[arrays.of.tables]] */
+			while (*cp != '\0' && (*cp != ']' || *(cp+1) == ']'))
 			{
 				vStringPut (name, *cp);
 				++cp;
--- a/Units/parser-iniconf.r/toml-support-in-geany.d/expected.tags
+++ b/Units/parser-iniconf.r/toml-support-in-geany.d/expected.tags
@@ -1,0 +1,41 @@
+table	input.toml	/^[table]$/;"	section	language:Iniconf
+key	input.toml	/^key = "value"$/;"	key	language:Iniconf	section:table
+bare_key	input.toml	/^bare_key = "value"$/;"	key	language:Iniconf	section:table
+bare-key	input.toml	/^bare-key = "value"$/;"	key	language:Iniconf	section:table
+1234	input.toml	/^1234 = "value"$/;"	key	language:Iniconf	section:table
+"127.0.0.1"	input.toml	/^"127.0.0.1" = "value"$/;"	key	language:Iniconf	section:table
+"character encoding"	input.toml	/^"character encoding" = "value"$/;"	key	language:Iniconf	section:table
+'key2'	input.toml	/^'key2' = "value"$/;"	key	language:Iniconf	section:table
+'quoted "value"'	input.toml	/^'quoted "value"' = "value"$/;"	key	language:Iniconf	section:table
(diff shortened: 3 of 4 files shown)


techee:
@masatake Any opinion on this pull request?

masatake:
Sorry for the late reply.

My primary question is whether this pull request intends a bug fix or a hack for adjusting the Iniconf parser to TOML.

We provide a PEG-based TOML parser. So modifying the Iniconf parser for supporting TOML is not a standard practice. This is my baseline to evaluate this pull request.

On the other hand, the TOML parser doesn't work well (#4096). As you were afraid, I'm not familiar with PEG enough so I have not fixed the simple bug yet. So I don't ask you to use the TOMP parser.  Hacking Iniconf for supporting TOML makes sense.

But this is still a hack. So I have some requests. Please, see my review comments.

Units/parser-iniconf.r/simple-pythonLoggingConfig.d/expected.tags
```diff
@@ -1,3 +1,6 @@
+"key in double quotes"	input.conf	/^"key in double quotes" = 1$/;"	key	language:Iniconf	section:[toml_array_of_tables]
+'key in single quotes'	input.conf	/^'key in single quotes' = 2$/;"	key	language:Iniconf	section:[toml_array_of_tables]
+[toml_array_of_tables]	input.conf	/^[[toml_array_of_tables]]$/;"	section	language:Iniconf
```
masatake, line 3: This is about the hack.
Instead of extending an existing test case, add a new one.
e.g. adding:

 Units/parser-iniconf.r/toml-support-in-geany.d/{args.ctags,input.toml,exected.tags}

Putting `--langmap=Iniconf:.toml` to args.ctags can make ctags use the Initconf to parser input.toml.

Units/parser-iniconf.r/simple-pythonLoggingConfig.d/input.conf
```diff
@@ -35,3 +35,7 @@ class=logging.Formatter
 [section.with.dots]
 key.with.dots=value1
 key-with-dashes=value2
+
+[[toml_array_of_tables]]
+"key in double quotes" = 1
+'key in single quotes' = 2
```
masatake, line 41: See my comment for expected.tags.

parsers/iniconf.c
```diff
@@ -34,8 +34,9 @@
 
 static bool isIdentifier (int c)
 {
-    /* allow whitespace within keys and sections */
-    return (bool)(isalnum (c) || isspace (c) || c == '_' || c == '-' || c == '.');
+	/* allow whitespace within keys and sections */
+	return (bool)(isalnum (c) || isspace (c) || c == '_' || c == '-' ||
+		c == '.' || c == '"' || c == '\'');  /* for TOML */
```
masatake, line 39: I want you to be clear the range of text that `/* for TOML */` qualifies more.
What do you think about using macro like:
```C
#define TOML_ID_CHAR(C) ( (c) == '"' || (c) == '\'' )
return (bool)(isalnum (c) || isspace (c) || c == '_' || c == '-' || c == '.'
    || (TOML_ID_CHAR(c));
```

parsers/iniconf.c
```diff
@@ -130,7 +131,7 @@ static void findIniconfTags (void)
 		if (*cp == '[')
 		{
 			++cp;
-			while (*cp != '\0' && *cp != ']')
+			while (*cp != '\0' && (*cp != ']' || *(cp+1) == ']'))
 			{
```
masatake, line 135: Could you comment something like:
```
/* the extra ] is for supporting TOML */
```

masatake (changes requested):

techee:
> My primary question is whether this pull request intends a bug fix or a hack for adjusting the Iniconf parser to TOML.

You are right, I want to use this parser for parsing TOML in Geany so it's kind of hack. On the other hand, it's not such a hack - there's no formal ini specification so I think what I propose is still a valid ini. The way I understand ini files intuitively is:
1. `[section name]\n` for sections
2. `whatever=whatever\n` for value assignments

So in fact, I believe the key name it could be any character sequence (with stripped leading and trailing spaces). And for the section name it should be the first `[` and the last `]` that count as section delimiters, the rest is the section name.

In any case, no problem to add the changes you propose.

Units/parser-iniconf.r/simple-pythonLoggingConfig.d/expected.tags
techee: Done.

Units/parser-iniconf.r/simple-pythonLoggingConfig.d/input.conf
techee: Done.

parsers/iniconf.c
techee: Done.

parsers/iniconf.c
techee: Done.

Units/parser-iniconf.r/toml-support-in-geany.d/args.ctags
```diff
@@ -0,0 +1,2 @@
+--fields=+lK
+--langmap=Iniconf:.toml
```
masatake, line 2: Could you add -`-sort=no` to this file and update `expected.tags` file?

masatake (approved):

masatake:
Thank you for updating. I have one more request to change. See my comment.

> On the other hand, it's not such a hack - there's no formal ini specification so I think what I propose is still a valid ini. 

I see. However, the updated version of this pull request looks more understandable.

Units/parser-iniconf.r/toml-support-in-geany.d/args.ctags
techee: Done.

masatake:
Thank you.

PR #4102, opened 2024-10-25
GitHub Actions: fix netbsd pkgin error.

leleliu008:
Don't try to install libxml2. libxml2 has already been pre-installed, if we try to install it, it will result in upgrading curl and it's dependencies. we do not use curl here.


masatake:
Thank you.

PR #4103, opened 2024-10-25
GitHub Actions: change s390x/ubuntu:latest to s390x/ubuntu:23.04 due to s390x/ubuntu:latest is not available recently

leleliu008:


masatake:
Thank you very much.

PR #4110, opened 2024-11-09
verilog.c: keyword "unsigned1" should be "unsigned", without the 1

cousteaulecommandant:
The list of Verilog keywords fetched from Appendix B of the Verilog standard accidentally captured the footnote "1" in keyword `unsigned¹`:

> ### List of keywords
> ... triand trior trireg **unsigned¹** use uwire vectored ...
>> ¹ **unsigned** is reserved for possible future usage

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -324,7 +324,7 @@ static const struct keywordGroup verilogKeywords = {
 		"scalared", "showcancelled", "signed", "small", "specify",
 		"specparam", "strong0", "strong1", "supply0", "supply1", "table",
 		"task", "time", "tran", "tranif0", "tranif1", "tri", "tri0", "tri1",
-		"triand", "trior", "trireg", "unsigned1", "use", "uwire", "vectored",
+		"triand", "trior", "trireg", "unsigned", "use", "uwire", "vectored",
 		"wait", "wand", "weak0", "weak1", "while", "wire", "wor", "xnor", "xor",
 		NULL
 	},


roccomao (approved):

b4n (approved):
LGTM

masatake (approved):

hirooih:
Thank you for your report.

The cause is exactly as you described. It was overlooked for a long time.

cousteaulecommandant:
> The cause is exactly as you described. It was overlooked for a long time.

Well, that's not surprising since `unsigned` is not a valid Verilog keyword (as the footnote points out), so no one writing valid Verilog would have come across this issue (unless they happened to name a variable `unsigned1`).  I only found out by chance, because I was comparing the list of keywords in ctags with those from another program.

roccomao:
In fact, I also found this typo right after you submitted this PR.

Because according to the definition of the formal syntax in LRM, almost all places where data types are required also support implicit data types. Take `data_declaration` as an example, it supports `implicit_data_type`:

```
implicit_data_type ::= [ signed | unsigned ] { packed_dimension }
```

> An implicit data type declaration uses the `implicit_data_type` syntax and includes only a signedness keyword and/or packed dimensions. **An implicit data type declaration implies a net unless the `var` keyword is used.**

So should we give him a kind, such as `K_NET` or `K_REGISTER` ? Then I searched in the source code, so the `unsigned1` typo was found.

Then it is found that `signed` and `unsigned` are both kind of `K_IGNORE`. If so, the `data_declaration` like `signed [7:0] test_var;` will cause that the tag `test_var` is ignored. Although the syntax supports declaring data this way, when I actually tested it in compiler, such as QuestaSim, I got an error. So their kind is `K_IGNORE` which is acceptable and reasonable, even though we may lose the `test_var` tag here.

In addition, if you use it in other places where the `implicit_data_type` is supported, for example, `function unsigned [7:0] test_func;`, or in the port, etc. the compiler like QuestaSim will not report an error at this time, and of course ctags will not have any problems when parsing it. So finally this issue was passed.

hirooih:
Good suggestion. Thanks!
I will fix it.

roccomao:
> I will fix it.

The current implementation seems to be OK, So I'm not sure what you mean.

cousteaulecommandant:
> ```
> implicit_data_type ::= [ signed | unsigned ] { packed_dimension }
> ```

> Then it is found that `signed` and `unsigned` are both kind of `K_IGNORE`. If so, the `data_declaration` like `signed [7:0] test_var;` will cause that the tag `test_var` is ignored. Although the syntax supports declaring data this way, when I actually tested it in compiler, such as QuestaSim, I got an error. So their kind is `K_IGNORE` which is acceptable and reasonable, even though we may lose the `test_var` tag here.

I'm not sure this is what the standard means.  If I understand correctly, the `[ signed | unsigned ]` part is optional, so something as simple as
```sv
my_signal;
```
would be considered a data_declaration.  The use of signed/unsigned doesn't change this, or transforms something from "not a declaration" to "a declaration".

This makes sense to me in e.g. function arguments, where the default for everything is as if you had used `input`, but for standalone declarations like in the `my_signal;` example it seems odd.

roccomao:
> If I understand correctly, the [ signed | unsigned ] part is optional

Having said that, it also needs to be supported by the compiler. Even if `signed signal;` is used to declare variables, the compiler will report an error.

> This makes sense to me in e.g. function arguments

Your understanding is right. Except for the `data_declaration` statement, it is much loose in other places, such as function parameters, like `function signed [10:0] test(x);` is acceptable.

hirooih:
@roccomao 

> The current implementation seems to be OK, So I'm not sure what you mean.

I've been out of town. I read and responded to only the first half of your message on my smart phone. Sorry to make you confuse.

BTW I'm glad to see that there are 5 participants in this MR. This is the first time on the Verilog / SystemVerilog parser and I am really encouraged.

roccomao:
Just in time to give a detailed reply to @cousteaulecommandant :

For `data_declaration`, if `[ signed | unsigned ] { packed_dimension } identifier` is used alone, the compiler will report an error. Actually, we have to add `wire`, `var` or `int` etc before the `signed` / `unsigned`. So it is very reasonable that `signed` and `unsigned` are both kind of `K_IGNORE`. If not, I think the analysis in many places may become slightly more complicated.

Secondly, it is much more relaxed in other places, just like the function arguments you mentioned, these definitions will be passed by the compiler. Of course, our ctags will also all **correctly** parse the *port* `test_arg` tag.

```systemverilog
function signed [10:0] test_a(test_arg);
  return test_arg;
endfunction
function unsigned test_b(input test_arg);
  return test_arg;
endfunction
function unsigned test_c(signed test_arg);
  return test_arg;
endfunction
function unsigned test_d([1:0] test_arg);
  return test_arg;
endfunction
function signed test_e(signed [1:0] test_arg);
  return test_arg;
endfunction
function signed test_f(input signed [1:0] test_arg);
  return test_arg;
endfunction
```

roccomao:
@hirooih Thanks for your reply. Of course, thank you very much for your work and maintenance on the Verilog / SystemVerilog parser, which makes it very convenient to read and view source code and do tag-based completion, etc. Wish you a pleasant journey! 😄

Issue #4109, opened 2024-11-09
SystemVerilog: Parsing error when the subroutine's return type contains class resolution `::`


roccomao:
When the return type of a subroutine contains the scope resolution of a class, the current and subsequent subroutines will have an additional class identifier, which is not expected. This bug was found when parsing file [UVM-2020.3.1 src/base/uvm\_component.svh](https://github.com/accellera-official/uvm-core/blob/78c06547a2a0a29b3dc9dcafae62b75b2ff61544/src/base/uvm_component.svh), Let's make a minimal reproduction:

```systemverilog
virtual class uvm_component extends uvm_report_object;
  typedef bit [1:0] config_mode_t;
  extern virtual function void test_ok();
  // ------------------------------------------------------
  extern virtual function config_mode_t return_scope_res();
  // ------------------------------------------------------
  extern function void m_set_cl_msg_args();
  extern function void m_set_cl_verb;
  extern function void m_set_cl_action;
  extern function void m_set_cl_sev;
endclass

function void uvm_component::test_ok();
  // function body, this function is parsed OK
endfunction

// From here on, there will be one more class scope, not expected
// ---------------------------------------------------------------------
function uvm_component::config_mode_t uvm_component::return_scope_res();
  return 2'b00;
endfunction
// ---------------------------------------------------------------------

// If you use option `ctags --kinds-SystemVerilog=+Q`, you will see that prototype
// `uvm_cmdline_processor` will also have the WRONG scope: `class:uvm_component`
// In fact, the scope of the prototype `uvm_cmdline_processor` should be empty.
typedef class uvm_cmdline_processor;

function void uvm_component::m_set_cl_msg_args();
  // function body, the following parser will also go WRONG
endfunction
function void uvm_component::m_set_cl_verb; return; endfunction
function void uvm_component::m_set_cl_action; return; endfunction
function void uvm_component::m_set_cl_sev; return; endfunction
```

The tags output as follow:

```
config_mode_t	test.sv	/^  typedef bit [1:0] config_mode_t;$/;"	T	class:uvm_component
m_set_cl_action	test.sv	/^function void uvm_component::m_set_cl_action; return; endfunction$/;"	f	class:uvm_component.uvm_component
m_set_cl_msg_args	test.sv	/^function void uvm_component::m_set_cl_msg_args();$/;"	f	class:uvm_component.uvm_component
[...]

PR #4111, opened 2024-11-12
Verilog: Fix function parse when return type contains `::`

roccomao:
This is a fix for #4109, and this TODO list (#2674) now can be checked:

- `fixme` in [`systemverilog-task-function.d/inputs.sv`](https://github.com/universal-ctags/ctags/blob/master/Units/parser-verilog.r/systemverilog-task-function.d/input.sv)
  - [x] `function void foo::bar();`
  - [x] `function foo #(IF) bar();`

We should ignore any return type of a `function`. We only need to parse the `class_scope` before the `function` or `task` name.


parsers/verilog.c
```diff
@@ -1171,27 +1172,66 @@ static int skipParameterAssignment (int c)
...
 {
 	verilogKind kind = token->kind;	// K_FUNCTION or K_TASK
+#define TOKEN_NAME_SIZE 16
+	vString *tokenNameSaved[TOKEN_NAME_SIZE];
+	int doubleColonPos[TOKEN_NAME_SIZE];
+	int functionNameStart = 0;
+	int i, j;
+
+	for (i = 0; i < TOKEN_NAME_SIZE; i++)
+		tokenNameSaved[i] = vStringNew ();
```
masatake, line 1182: With my idea, you can remove this initialization step.

parsers/verilog.c
```diff
@@ -1171,27 +1172,66 @@ static int skipParameterAssignment (int c)
...
 	 * Last identifier found before a '(' or a ';' is the function name */
+	i = 0;
+	j = 0;
 	while (c != '(' && c != ';' && c != EOF)
 	{
 		if (isWordToken (c))
+		{
 			c = readWordToken (token, c);
+			if (i < TOKEN_NAME_SIZE)
+				vStringCopy (tokenNameSaved[i++], token->name);
```
masatake, line 1194: With my idea, using the dynamic arrays, you can write:
```
ptrArrayAdd (tokenNameSaved, vStringNewCopy(token->name));
```

parsers/verilog.c
```diff
@@ -1171,27 +1172,66 @@ static int skipParameterAssignment (int c)
...
+	{
+		for (int k = j - 1; k >= 1; k--)
+			if (doubleColonPos[k] - doubleColonPos[k-1] > 1)
+			{
+				functionNameStart = doubleColonPos[k];
+				break;
+			}
+		if (functionNameStart == 0 && (i - doubleColonPos[j-1]) == 1)
+			functionNameStart = doubleColonPos[0];
+	}
```
masatake, line 1220: These line for deciding functionNameStart.
What do you think about converting these lines to a function that returns funtionNameStart like:
```
functionNameStart = calcFunctionNameStart(tokenNameSaved, doubleColonPos);
```

parsers/verilog.c
```diff
@@ -1171,27 +1172,66 @@ static int skipParameterAssignment (int c)
...
-			verbose ("Found function declaration with class type %s\n", vStringValue (token->name));
-			createContext (K_CLASS, token->name);
+			verbose ("Found function declaration with class type %s\n", vStringValue (tokenNameSaved[k-1]));
+			createContext (K_CLASS, tokenNameSaved[k-1]);
 			currentContext->classScope = true;
 		}
 	}
+
+	for (i = TOKEN_NAME_SIZE - 1; i >= 0; i--)
+		vStringDelete (tokenNameSaved[i]);
```
masatake, line 1232: ptrArrayDelete(...);
intArrayDelete(...);

parsers/verilog.c
```diff
@@ -1171,27 +1172,66 @@ static int skipParameterAssignment (int c)
 static int processFunction (tokenInfo *const token, int c)
 {
 	verilogKind kind = token->kind;	// K_FUNCTION or K_TASK
+#define TOKEN_NAME_SIZE 16
+	vString *tokenNameSaved[TOKEN_NAME_SIZE];
+	int doubleColonPos[TOKEN_NAME_SIZE];
```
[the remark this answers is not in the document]
roccomao: This is a really good idea. I really need to dynamically grow the array, otherwise I need to make a good trade-off with the array length. I don't know if there is already an implementation. I just searched in the project, so this form of writing is a copy from `./main/parse.c`

parsers/verilog.c
roccomao: Okay, but according to the Verilog language LRM, this function is unlikely to be used in other situations. I can do this and add some comments to this.

parsers/verilog.c
roccomao: Thank you very much for your suggestion, I will update later.

masatake:
Much easier to read!

roccomao:
Aha! Thanks for your detailed review and guidance, it helped me a lot. 😄

hirooih:
@roccomao,

Thanks a lot!
I will review your fix this evening.

hirooih (changes requested):
@roccomao,

Thank you for your great fix.
You also fix the bug of `currentContext->virtual` in `dropContex()`.

I've added a note on your comment. Please fix it.

I still don't understand how `getFirstClassNameIndex()` works well.  I may be able to simply it some more, but not sure yet.  Let me some more time to think about it.

parsers/verilog.c
```diff
@@ -1163,6 +1165,34 @@ static int skipParameterAssignment (int c)
 	return c;
 }
 
+// Using LRM Annex A Formal syntax, The declaration of function and task can be combined
+// to describe as follows (ignore parameter_value_assignment in the class_scope):
+// -------------------------------------------------------------------------------------
+//                       |  class_scope  |               |  class_scope  |
+// [ function | task ] [ { class_name :: } return_type ] { class_name :: } function_name
```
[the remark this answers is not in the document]
roccomao: You are right. In fact, it is not just that. If there is something like `logic [LOW:HIGH]`, the token `LOW` and `HIGH` will also be detected, but it will be skipped too.

This `[lifetime]` does not affect the execution of the algorithm. Considering that you have commented the complete function body below, I have omitted it here for simplicity. If added, it will be a duplicate of your complete comment below. However, I will add a supplementary comment later.

> Please add on the comment and add static or automatic on some complex pattens in your input.sv.

I will reply you at the location of the test code.

Units/parser-verilog.r/systemverilog-github4109.d/input.sv
```diff
@@ -0,0 +1,209 @@
...
+  endtask
+
+  // 6. Corner TEST
+  // -----------------------------------------------------------------
+  localparam LOW = 8;
+  localparam HIGH = 15;
+  function automatic logic [ LOW : HIGH ] test_func_V ();
+  endfunction
+  function automatic int signed [LOW-1:HIGH-1] test_func_W;
+  endfunction
```
roccomao, line 106: @hirooih Continue to reply to you here.

In fact, I just added a lot of irrelevant tokens such as `automatic` and data types, etc. here to test their non-influence.

In addition, The reason why the `automatic` for class is not tested is that, In LRM 8.6 Object methods says :

> The lifetime of methods declared as part of a class type shall be automatic. It shall be illegal to declare a class method with a static lifetime.

That is to say, the lifetime of a class method is always `automatic`, and a `static method` is different from a lifetime declaration. It is described as `static [function|task]` in LRM 8.10 Static methods, with the `static` keyword in front. Considering that the algorithm will not be affected and generally `automatic` will not be actively declared, so no additional test cases are made for the declaration of `automatic` *lifetime* class methods.

I could add a few methods containing `::` with `automatic` keyword here later, mainly to test the stability of the current implementation.

roccomao:
At first, I just added as detailed a description and test cases as possible to submit an issue, hoping to help developers analyze the problem. Suddenly, I wanted to see how it was implemented. The clear implementation code structure makes it understandable to newbies, so I raised this PR and received detailed review and guidance. Thanks a lot!

> You also fix the bug of `currentContext->virtual` in `dropContex()`.

This is mentioned in [6de290bc](https://github.com/roccomao/ctags/commit/6de290bc477962f10d7a0bba67f2283dd605acb8) detailed commit messages.

> I still don't understand how `getFirstClassNameIndex()` works well. I may be able to simply it some more, but not sure yet. Let me some more time to think about it.

In fact, we only add the class scope to the token before `::`, and `::` can only appear before the return type and function name. So how to separate the return type and function name? If they both contain `::`, then there are at least two adjacent tokens between the two `::` that is exactly the split point. This is my implementation idea.

parsers/verilog.c
```diff
@@ -1163,6 +1165,35 @@ static int skipParameterAssignment (int c)
...
+		return tokensCount - 2;
+	else if (colonsCount >= 2 && intArrayItem(doubleColonPos, 0) >= 1)
+	{
+		for (int i = colonsCount - 1; i >= 1; i--)
+			if (intArrayItem(doubleColonPos, i) - intArrayItem(doubleColonPos, i - 1) > 1)
+				return intArrayItem(doubleColonPos, i) - 1;
+		if (intArrayLast(doubleColonPos) == (tokensCount - 1))
+			return intArrayItem(doubleColonPos, 0) - 1;
+	}
+	return -1;
```
hirooih, line 1194: > In fact, we only add the class scope to the token before ::, and :: can only appear before the return type and function name. So how to separate the return type and function name? If they both contain ::, then there are at least two adjacent tokens between the two :: that is exactly the split point. This is my implementation idea.

Here is my interpretation of your explanation.

```c
	// if there is no `::`, then no class_name is found
	if (colonsCount == 0 || intArrayItem (doubleColonPos, colonsCount - 1) != (tokensCount - 1))
		return -1;
	int startPos = intArrayItem (doubleColonPos, colonsCount - 1) - 1;
	for (int i = colonsCount - 1; i > 0; i--) {
		if (intArrayItem (doubleColonPos, i - 1) == startPos)
			// if there is no gap between two `::`, then the next class_name is found
			startPos--;
		else	// if there is a gap between two `::`, then the first class_name is found
			break;
	}
	return startPos;
```

Your version has 4 `if`s.  And mine is 2.  And same for `return`s.
How do you think?

parsers/verilog.c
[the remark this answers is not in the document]
hirooih: Thank you.

Your test code was very easy to understand because the variable names were carefully chosen. I learned a lot.

Units/parser-verilog.r/systemverilog-github4109.d/input.sv
hirooih: > The lifetime of methods declared as part of a class type shall be automatic. It shall be illegal to declare a class method with a static lifetime.

You are right. I had completely forgotten this rule.

I have been relying on this BNF rule:-) It is a simplified version for this project.

https://docs.google.com/document/d/1Sb0-iO9HofB7BR52XUarhGYPph6mFMx5oIFcU8eeFj0/edit?tab=t.0

roccomao: Thanks for sharing, I always check from the original PDF document 😄

parsers/verilog.c
roccomao: My original logic was to first assume that I was looking for a sequence of `name::name::...`, and then determine whether this sequence was in the scope of the *function name field*. Your logic was to first limit this sequence of `name::name::...` to the *function name field*, and then search forward. The essence is the same. It seems that your revised writing is more concise. I am preparing another PR, and then I will refactor it to your current logic, which may seem easier to understand. Then I will ping you.

roccomao:
Refer to PR #4116, the class now supports `final_specifier`. Similarly, the class method now supports `dynamic_override_specifiers`. The current implementation of this PR is actually unaffected. Anyway, let's update the comments again and add a few test cases.

> latest [IEEE Std 1800-2023](https://standards.ieee.org/ieee/1800/7743/) 8.3 Syntax:
>
> ```
> dynamic_override_specifiers ::= [ initial_or_extends_specifier ] [ final_specifier ]
>     initial_or_extends_specifier ::= : initial | : extends
>     final_specifier ::= : final
> ```

In addition, the implementation of the `getFirstClassNameIndex` function was refactored.

hirooih (approved):

parsers/verilog.c
hirooih: Very nice!

I was also thinking after I went to bed last night that we could use a while statement :-).

Units/parser-verilog.r/systemverilog-github4109.d/input.sv
hirooih: The gray part is from the original LRM.  We have to check many pages for each construct.  I made this to see it at a glance.

hirooih:
Again thank you very much for your great work!

Units/parser-verilog.r/systemverilog-github4109.d/input.sv
roccomao: Yes, very good notes, and indeed much more concise and clear.

roccomao:
Fixed #4109

❤️ Many Thanks to the reviewers, you helped me a lot. Now let's close the associated issue.

PR #4116, opened 2024-11-15
Verilog: Add final_specifier support for class parse

roccomao:
The syntax for class declaration is outdated and has been updated in the latest version of the LRM.

Please refer to latest [IEEE Std 1800-2023](https://standards.ieee.org/ieee/1800/7743/) 8.3 Syntax:

> ```
> [ virtual ] class [ final_specifier ] class_identifier [ parameter_port_list ]
>     final_specifier ::= : final
> ```

In any case, this PR just makes a very minor change to adapt it, making it as harmless as possible.

@hirooih Please Take a Look, Thanks a lot!

### DIFF
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1355,18 +1355,24 @@ static int processParameterList (tokenInfo *token, int c)
 	return c;
 }
 
-// [ virtual ] class [ static | automatic ] class_identifier [ parameter_port_list ]
+// LRM IEEE Std 1800-2017 and previous versions
+//  [ virtual ] class [ static | automatic ] class_identifier [ parameter_port_list ]
+// LRM IEEE Std 1800-2023 ( Fixed )
+//  [ virtual ] class [ : final ] class_identifier [ parameter_port_list ]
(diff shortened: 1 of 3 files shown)


parsers/verilog.c
```diff
@@ -1355,13 +1355,16 @@ static int processParameterList (tokenInfo *token, int c)
 	return c;
 }
 
-// [ virtual ] class [ static | automatic ] class_identifier [ parameter_port_list ]
+// [ virtual ] class [ : final ] class_identifier [ parameter_port_list ]
```
masatake, line 1358: Generally, ctags should accept older syntax, too.
(Your change doesn't reject the older syntax. This is good.)
Keeping the comment explaining older syntax is better.
Just adding the comment explaining the new syntax is enough:
```
// LRM 19??
//  [ virtual ] class [ static | automatic ] class_identifier [ parameter_port_list ]
// LRM 20??
//  [ virtual ] class [ : final ] class_identifier [ parameter_port_list ]
```

roccomao: Thanks for your suggestion, I will fix it.

hirooih:
LGTM.

Thanks!

Issue #4123, opened 2024-11-20
Clojure improvements


aartaka:
The name of the parser: clojure

Ctags only supports functions and namespaces in Clojure, lacking e.g. types and variables.

Find the examples of [def](https://clojuredocs.org/clojure.core/def) and [defonce](https://clojuredocs.org/clojure.core/defonce) for variables; [defmacro](https://clojuredocs.org/clojure.core/defmacro) for macros; [defmulti](https://clojuredocs.org/clojure.core/defmulti) for multimethods and [defmethod](https://clojuredocs.org/clojure.core/defmethod) for methods; [deftype](https://clojuredocs.org/clojure.core/deftype) and [defrecord](https://clojuredocs.org/clojure.core/defrecord) for types; [defprotocol](https://clojuredocs.org/clojure.core/defprotocol) and [definterface](https://clojuredocs.org/clojure.core/definterface) for Clojure and Java interfaces respectively.

I cannot provide examples beyond these posts because most of the Clojure I do it private/proprietary :see_no_evil: 

The version of ctags:

```
$ ctags --version
Universal Ctags 6.1.0(p6.1.20241117.0), Copyright (C) 2015-2023 Universal Ctags Team
Universal Ctags is derived from Exuberant Ctags.
Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
  Compiled: Nov 18 2024, 17:31:59
  URL: https://ctags.io/
  Output version: 0.0
  Optional compiled features: +wildcards, +regex, +iconv, +option-directory, +xpath, +json, +interactive, +sandbox, +yaml, +packcc, +optscript, +pcre2
```

How do you get ctags binary: pacman on Arch Linux

masatake:
I will accept pull requests to solve this issue.
However, I will not work on this issue directly; I would like to solve it using a Sexp-based meta parser.

HOWEVER, '.b' test cases are welcome.
People requesting to add parsers supporting new languages and/or to enhance parsers expect ctags developers to know the target languages. The expectation is wrong. The developers know only a tiny subset of the languages that ctags supports. Generally, the developers postpone the tasks related to unknown languages.

Preparing test cases is one of the ways to drive the developers. The test cases help the developers by reducing the time to learn the target language. See https://github.com/universal-ctags/ctags/issues/4065.

The value of test cases is immeasurable. We can add code to extract `def*` from the current implementation. I will call the code a direct solution. I want to avoid the direct solution.
[...]

PR #4126, opened 2024-11-21
Tests for more Clojure tags

aartaka:
Hi! Here's some Clojure tests for #4123. That's not all, but it's the most frequently used things—vars and multimethods.

### DIFF
--- a/Units/parser-clojure.r/clojure-methods.b/input.clj
+++ b/Units/parser-clojure.r/clojure-methods.b/input.clj
@@ -1,0 +1,19 @@
+(ns multimethod.test)
+
+(defmulti test
+  (fn [type] type))
+
+(defmulti documented-multimethod "Documentation"
+  (fn [type] type))
+
+(defmethod test nil
+  [& _]
+  nil)
+
+(defmethod test :test
+  [& _]
+  nil)
+
+(defmethod test :test2 named-method
+  [& _]
+  nil)
--- a/Units/parser-clojure.r/clojure-methods.b/README
+++ b/Units/parser-clojure.r/clojure-methods.b/README
@@ -1,0 +1,1 @@
+About methodId kind, see #4126.
--- a/Units/parser-clojure.r/clojure-methods.b/args.ctags
+++ b/Units/parser-clojure.r/clojure-methods.b/args.ctags
@@ -1,0 +1,2 @@
+--sort=no
+--fields=+KzZ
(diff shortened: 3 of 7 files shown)


masatake:
Thank you for making this pull request. I will merge this with some revising.

Units/parser-clojure.r/clojure-methods.b/input.clj
```diff
@@ -0,0 +1,19 @@
...
+
+(defmethod test nil
+  [& _]
+  nil)
+
+(defmethod test :test
+  [& _]
+  nil)
+
+(defmethod test :test2 named-method
```
masatake, line 17: Could you explain this line?
I wonder why `named-method` should be extracted because my knowledge of Clojure is limited.

aartaka: It's a method-specific name that's added to multimethod function name. See https://clojuredocs.org/clojure.core/defmethod#example-542692c7c026201cdc3269cd

aartaka: No, it doesn't work as a self-sufficient function. It's merely a piece of metadata identifying the method. It's only useful in debugging and (given ctags support) in location search.

masatake: I see. Thank you. We should introduce a new kind for such a language object. What do Clojure hackers call it? I think of `method-identifier/i` or `method-id/i`.

The ideal tags for `named-method` looks like:
```
named-method	input.clj	/^(defmethod test :test2 named-method$/;"	i	multimethod.test	method:test
test	input.clj	/^(defmethod test :test2 named-method$/;"	m	multimethod.test	identifier:named-method
```

Anyway, the current Clojure parser cannot extract caddr.
So, to satisfy these new test cases, we must rewrite the parser.

aartaka: Sorry for disappearing :pensive: Why not named method? But yes, `method-id(endifier)` is fine by me too!

masatake: If `named-method` is tagged as a method, a client tool like Vim may show it as a candidate in input completion.

> No, it doesn't work as a self-sufficient function.

However, it is not callable, as you wrote. So, `named-method` should not be shown on the candidate list. If `named-method` is tagged as a method-id, Vim may not show as far as Vim shows only `method` kind tags.

aartaka: Yes, all fair!

masatake:
@aartaka Thank you for contributing.

PR #4129, opened 2024-11-25
(System)Verilog: escaped identifiers (LRM 5.6.1)

cousteaulecommandant:
Add support for escaped identifiers in Verilog and SystemVerilog: `\` + zero or more non-whitespace characters + whitespace.

Note that the `\` itself isn't part of the identifier, and that `\foo` is the same as just `foo` (unlike in VHDL), but that identifiers identical to keywords such as `\begin` are also allowed.

This definition also theoretically allows an empty identifier, `\`, which ctags doesn't support.
~~To avoid issues, empty identifiers (`\`) or identifiers that start with (a second) `\` keep the leading `\` even if it's technically not part of the name.~~

Escaped identifiers are defined in the standard in section 3.7.1 (Verilog) / 5.6.1 (SystemVerilog).

Support for escaped identifiers was listed in the Verilog TODO list in #2674.


cousteaulecommandant:
By the way, I also added a unit test for the feature.  I don't know what's the protocol here, if I add the unit test myself to demonstrate the feature or if someone *else* should add it to verify that what I did works.

masatake:
The error looks strange. I will look into this.

cousteaulecommandant:
> The error looks strange. I will look into this.

Which error do you mean?

parsers/verilog.c
```diff
@@ -873,12 +873,26 @@ static int _readWordToken (tokenInfo *const token, int c, bool skip)
...
-		c = vGetc ();
-	} while (isIdentifierCharacter (c));
-	_updateKind (token);
+		// Escaped identifier (may be empty)
+		c = vGetc (); // skip leading '\'
+		while (isgraph (c))
+		{
+			vStringPut (token->name, c);
+			c = vGetc ();
+		}
```
roccomao, line 884: As your comment says, the escaped identifier may be empty, so we need to do special handling at the end :

```c
		if (vStringIsEmpty (token->name) || vStringChar (token->name, 0) == '\\')
		{
			vString *tempName = vStringNewInit ("\\");
			vStringCat (tempName, token->name);
			vStringCopy (token->name, tempName);
			vStringDelete (tempName);
		}
```

Otherwise we will get a prompt `ctags: Notice: ignoring null tag in input.sv(line: 3, language: SystemVerilog)` and miss the `\` tag, consider adding the following test case :

```sv
int \ = 100;
int \\ = 200;
int \\test = 300;
int \\\test = 400;
```

cousteaulecommandant: That makes sense.  So if the name is empty or starts with `\`, add an additional `\` (even if it's not technically the true name of the variable, but at least it'll be unique).
I'll add that as soon as possible (and the test).

cousteaulecommandant: Done.  I've also added a test to verify that and `\end` identifier isn't mistaken for an `end` keyword.

roccomao: Nice, your implementation is so concise, LGTM. :thumbsup:

> even if it's not technically the true name of the variable, but at least it'll be unique

In fact, for any non-normal identifier, we need to add an extra `\` when accessing it, unless it can be used as an identifier itself, in which case we can omit `\`. Our current solution is mainly just to implement support for parsing single `\`.

For example, the following variables must be preceded by `\` when used.

```verilog
module test;
  initial begin
    static int \ = 100;
    static int \\ = 200;
    static int \\test = 300;
    static int \\\test = 400;
    static int \r\n = 500;
    static int \test*var = 600;
    static int \**test** = 700;
    static int \begin = 800;
    #1 $display("%d", \ );
    #1 $display("%d", \\ );
    #1 $display("%d", \\test );
    #1 $display("%d", \\\test );
    #1 $display("%d", \r\n );
    #1 $display("%d", \test*var );
    #1 $display("%d", \**test** );
    #1 $display("%d", \begin );
  end
endmodule
```

Only if it is like `\test` or `\var_name`, we can omit `\` in subsequent use.

cousteaulecommandant: > Nice, your implementation is so concise, LGTM. 👍

Thanks!

> In fact, for any non-normal identifier, we need to add an extra `\` when accessing it, unless it can be used as an identifier itself

I was referring to what the Verilog standard says.  *"Neither the leading backslash character nor the terminating white space is considered to be part of the identifier."*  I.e., `\foo+bar` creates a variable whose "official name" is "foo+bar", without the `\`.  But we probably don't care what Verilog considers to be the "official name" of the variable, and just want a symbol to be able to represent the variable.
(And, as you mentioned, in this case what ctags would give us is what you need to type to access the variable, in case it starts with `\`, so I agree that it's "justified" to use it.)

roccomao: Yes, I totally agree with you. I just made some additional explanations from the compiler's perspective in my previous comment. Anyway, as you said, *"we just want a symbol to be able to represent the variable"*.

In fact, from the perspective of LRM standard alone, there is still an exception. For example, in the latest *IEEE Std 1800-2023*, an example is given for *escaped keyword*.

> `net` is a keyword. `\net` is a user-defined identifier.

That is to say, `\net` creates a variable whose "_official name_" is `\net`, But we don't care.

leleliu008:
The CI failure has been fixed in master branch. please rebase master.

cousteaulecommandant:
Rebased and all CI checks passed 😃 Thanks!

parsers/verilog.c
```diff
@@ -873,12 +873,30 @@ static int _readWordToken (tokenInfo *const token, int c, bool skip)
...
 	{
-		vStringPut (token->name, c);
-		c = vGetc ();
-	} while (isIdentifierCharacter (c));
-	_updateKind (token);
+		// Escaped identifier
+		c = vGetc (); // skip leading '\'
+		// A single `\` would result in an empty identifier, which is unsupported.
+		// Add the initial `\` in that case, and also in case it starts with `\\`.
+		if (!isgraph (c) || c == '\\')
```
masatake, line 882: I wonder if C is EOF (== -1).
isgraph(c) may return false. Then `\` will be appended to `token->name`. It's not so harmful that I am afraid of. But putting something like `goto end;` will be better.

cousteaulecommandant: I think the behavior is correct.

We treat the sequence `\<EOF>` the same as `\<whitespace>`, i.e. the token finishes right there so it's just a single `"\"`.  This would generate an empty identifier which ctags doesn't like, so as suggested by @roccomao we add the leading `\` to that corner case, and also to all identifier names starting with `\`.  Note that this is only done for the first character.

If I'm interpreting [the C standard, §7.4](https://www.iso-9899.info/n1570.html#7.4) correctly, passing EOF to `isgraph()` et al is valid and would return false, since EOF is not a graph character.  So EOF will be treated the same as space, whitespace, control characters, etc., which is what we want.

(PS: Note that I'm using `while` and not `do while` to handle this case, since the sequence of characters after the `\` may be empty, whereas in the case of regular identifiers there's one or more characters, hence `do while`)

masatake: I see.

masatake (dismissed):

hirooih:
First, thank you for your great contribution.

> To avoid issues, empty identifiers (\) or identifiers that start with (a second) \ keep the leading \ even if it's technically not part of the name.

I doubt that this specification is correct.

***

> This definition also theoretically allows an empty identifier, 

LRM just says:

> Neither the leading backslash character nor the terminating white space is considered to be part of the identifier. Therefore, an escaped identifier `\cpu3` is treated the same as a nonescaped identifier `cpu3`.

It is not clear an empty identifier is allowed.
Other other hand, "`\\` should be treated the same as a nonescaped identifier `\`" from this definition..

[tags (5)](https://github.com/universal-ctags/ctags/blob/master/docs/man/tags.5.rst)

> The tags file is a list of lines, each line in the format:
>
> `{tagname}<Tab>{tagfile}<Tab>{tagaddress}`
>
> `{tagname}`
>
> Any identifier, not containing white space..
>
> EXCEPTION: Universal Ctags violates this item of the proposal; tagname may contain spaces. However, tabs are not allowed.

This format can support an empty identifier "theoretically". Even if it were not currently supported, it would not be worth the bother to support it, I think.

cousteaulecommandant:
> > To avoid issues, empty identifiers (`\`) or identifiers that start with (a second) `\` keep the leading `\` even if it's technically not part of the name.
> 
> I doubt that this specification is correct.

Well, it's what the standard says; the `\` is not considered to be part of the identifier, so it's there just to indicate that an escaped identifier follows.  But we're just discussing semantics at this point.
(Although you may argue that ctags *should* explicitly indicate "dangerous" identifiers such as `\end`, `\1+1`, or `\\x` by explicitly adding the leading `\` always, to remind the user that they need to include the `\` to refer to the identifier, with the exception being identifiers that CAN be written without the `\`, such as `\foo`.  That behavior could be implemented; I think ctags provides all the tools for that; but I think it's overcomplicating things.)

I can remove that part from the commit message if you prefer.

> It is not clear an empty identifier is allowed.

Not in that paragraph, but from the LRM,[^1] section A.9.3:
```
escaped_identifier ::= \ {any_printable_ASCII_character_except_white_space} white_space
```
with { } meaning zero or more.  If they meant one or more, they would've put an extra character explicitly (they do so e.g. in the definition of `system_task_identifier`).

> On the other hand, "`\\` should be treated the same as a nonescaped identifier `\`" from this definition.

But `\` is not a (valid) *nonescaped* identifier, so that statement doesn't apply.  Same for things like `\1+1` or `\#CS`.  It does apply to `cpu3` which is a valid nonescaped identifier though.

> This format can support an empty identifier "theoretically". Even if it were not currently supported, it would not be worth the bother to support it, I think.

Yeah, definitely not.  As I said, I don't think ctags must encode exactly what the standard says the identifier is, and even if it could, I imagine it would cause all sorts of trouble.  So adding the `\` is fine in my opinion.

[^1]: The Verilog-2005 standard says it's "any character except white space", and doesn't specify that it has to be printable, but I'm ignoring that detail and assuming they didn't mean to include all the ASCII control characters so `isgraph()` will do.

hirooih:
> `escaped_identifier ::= \ {any_printable_ASCII_character_except_white_space} white_space`

OK. It is clear an empty identifier is allowed.

> But \ is not a (valid) nonescaped identifier, so that statement doesn't apply.

I don't understand what you mean, here...

> On the other hand, "`\\` should be treated the same as a nonescaped identifier `\`" from this definition.

I meant;

```verilog
localparam \\ = \ ;
```

this line should emit;

```
\\	input.v	/^localparam \\\\ = \\ ;$/;"	c	module:themodule#(x=42)
```

instead of (the line 6 of `expected.tags`);

```
\\\\	input.v	/^localparam \\\\ = \\ ;$/;"	c	module:themodule#(x=42)
```

How can we explain that the latter follows the LRM?

masatake:
You can make a tag with an empty string when you set `true` to `ParserDefinition::allowNullTag`.
If you don't want to touch the parser-wide configuration, we can add a new member `allowNullTag` to `tagEntryInfo`.

roccomao:
> How can we explain that the latter follows the LRM?

This is indeed **not** in accordance with LRM. But let me explain the previous suggestion in detail and sort out the different opinions at first. Then determine how should we make a trade-off?

Because of the ambiguity of LRM, the code in my [previous comment](https://github.com/universal-ctags/ctags/pull/4129#discussion_r1864974834) was actually verified in the *QuestaSim* and it works fine. I haven't tested other compilers, but this should give us a reference. Because we don't care what its *official name* is, we just give each identifier a *tag name*, too.

Firstly, This is an example given by the LRM 5.6.1 chapter. The document only describes that *escaped keyword* is an exception.

```
| declaration            | LRM identifier        |
| :--------------------- | :-------------------- |
| \busa+index            | busa+index            |
| \-clock                | -clock                |
| \***error-condition*** | ***error-condition*** |
| \net1/\net2            | net1/\net2            |
| \{a,b}                 | {a,b}                 |
| \a*(b+c)               | a*(b+c)               |
| \net                   | \net                  |
```

The compiler does the opposite. It does not consider keywords. It only looks at whether there are special characters in the identifier to determine whether there should have a leading `\` character.

```
| declaration            | compiler identifier    |
| :--------------------- | :--------------------- |
| \busa+index            | \busa+index            |
| \-clock                | \-clock                |
| \***error-condition*** | \***error-condition*** |
| \net1/\net2            | \net1/\net2            |
| \{a,b}                 | \{a,b}                 |
| \a*(b+c)               | \a*(b+c)               |
| \net                   | net                    |
```

Next, let's look at the empty identifier (a single `\`), again using my previous example to test:

image

1. The compiler supports the empty identifier (a single `\` followed by space).
2. If there are special characters at any position, we should add `\` at the beginning (like `\r\n`, `\test*var`).
3. Normal identifiers, even keywords, ignore the leading `\` (like `\begin`, `\test`).

--------------------

So, finally, how do we create tag names?

1. We can follow LRM completely, that is, remove the leading `\` from all identifiers except `\keyword` (*escaped keyword*). The example is as follows:

   ```
   (identifier) ->  (tag name)
   \busa+index  ->  busa+index
   \net1/\net2  ->  net1/\net2
   \*           ->  *
   \\           ->  \
   \\\          ->  \\
   \test        ->  test
   \begin       ->  \begin
   \            ->  ERROR
   ```

   But this will require comparing all keywords, and will cause problems if a single `\`, the empty identifier, is parsed.

2. We can also follow the compiler's parsing ideas and only consider whether there are special characters in any position, regardless of keywords. The example is as follows:

   ```
   (identifier) ->  (tag name)
   \busa+index  ->  \busa+index
   \net1/\net2  ->  \net1/\net2
   \*           ->  \*
   \\           ->  \\
   \\\          ->  \\\
   \test        ->  test
   \begin       ->  begin
   \            ->  \
   ```

   This solves the problem of empty identifiers (a single `\`), allowing us to create a tag name `\` for empty identifier, but this implementation requires us to scan the entire identifier to see if it contains special characters to decide whether to add a leading `\` to the tag name.

3. So my personal idea and previous suggestions combine the above two points, **mainly to make the parsing of tags easier**. The example is as follows:

   ```
   (identifier) ->  (tag name)
   \busa+index  ->  busa+index
   \net1/\net2  ->  net1/\net2
   \*           ->  *
   \\           ->  \\
   \\\          ->  \\\
   \test        ->  test
   \begin       ->  begin
   \            ->  \
   ```

   That is, by default, we do not consider keywords, and then follow LRM: *"Neither the leading backslash character nor the terminating white space is considered to be part of the identifier"*. But in order to accept the empty identifier (a single `\`) and give it the tag name `\`, we also need to add an extra `\` to all identifiers that start with `\`, otherwise, for example, `\` and `\\` will have the same tag name `\`.

   The advantage of this idea is that it accepts empty identifiers (a single `\`). At the same time, there is no need to compare whether the entire identifier is a keyword (*escaped keyword*), and there is no need to scan the entire identifier to see if it contains special characters. It only needs to determine whether the character after the beginning `\` character is also `\`. This is also what the current latest PR implements, and it is probably the view that @cousteaulecommandant agrees with. @hirooih what do you think?

cousteaulecommandant:
> How can we explain that the latter follows the LRM?

My bad; I misunderstood what you meant, but I get you now.

As @roccomao already mentioned, the addition of the extra `\` was added following their suggestion to do so; the original commit didn't have it and just removed the `\` from ALL escaped identifiers, even the empty one.  Ctags didn't like this and complained.  So I thought that the suggestion to add the `\` to handle that case was a good idea (and to any tag starting with `\`, because otherwise `\` and `\\` would be indistinguishable).

The options I see to handle escaped identifiers are:
1. Follow LRM strictly and always remove the `\`, even for empty identifiers.  We should modify the behavior so that empty identifiers are allowed.
2. Follow LRM strictly but ignore the `\` identifier; if someone decides to use it, don't list it as a tag.
3. Add a leading `\` only to identifiers starting with `\` to disambiguate (@roccomao's suggestion).
4. Add a leading `\` to all identifiers that aren't valid nonescaped identifiers by checking their characters, but not to identifiers that look like keywords (e.g. `\end` -> `end`).  (Note that my current code will still interpret these as identifiers and not keywords because I'm assigning the token kind explicitly.)
5. Add a leading `\` to all identifiers that need to be escaped, either because they contain special characters or because they're keywords.  This approach matches "what you would need to write" to use the identifier.

Option 1 is the most strictly compliant with the LRM "semantic" definition of an identifier, and option 5 is technically also what the LRM defines as an `escaped_identifier` (but this is from a *syntax* point of view).  The rest are less "compliant" compromise solutions.

If, as @masatake said, setting allowNullTag is an option (and doesn't involve changing anything outside of the Verilog parser and can be enabled for Verilog only), then I'd say that (1.) is the best approach.

masatake:
If you want, I will add this change.

```
commit d78b623bee0d963f39274ce28cfd2d5318e14fbf (HEAD -> master)
Date:   Wed Dec 4 22:05:10 2024 +0900

    main: add allowNullTag flag to tagEntryInfo
    
    We have allowNullTag in parserDefinition already but enabling the new
    flag has a much smaller impact.
    
index 6795c1b82..bdd509271 100644
--- a/main/entry.c
+++ b/main/entry.c
@@ -1938,7 +1938,7 @@ extern int makeTagEntry (const tagEntryInfo *const tag)
 
    if (tag->name [0] == '\0' && (!tag->placeholder))
    {
-       if (!doesInputLanguageAllowNullTag())
+       if (! (doesInputLanguageAllowNullTag() || tag->allowNullTag))
            error (NOTICE, "ignoring null tag in %s(line: %lu, language: %s)",
                   getInputFileName (), tag->lineNumber,
                   getLanguageName (tag->langType));
index 19ac0865d..f600fa652 100644
--- a/main/entry.h
+++ b/main/entry.h
@@ -76,6 +76,7 @@ struct sTagEntryInfo {
                                             * Set in the cork queue; don't touch this.*/
    unsigned int boundaryInfo: 2; /* info about nested input stream */
    unsigned int inIntevalTab:1;
+   unsigned int allowNullTag:1;    /* allow a tag with an empty string. */
 
    unsigned long lineNumber;     /* line number of tag;
                                     use updateTagLine() for updating this member. */
```

cousteaulecommandant:
> @cousteaulecommandant
> 
> Could you try "1." with the following patch?

I tested it with a null tag and it didn't create an entry.

cousteaulecommandant:
In fact, is `allowNullTag` tested at all?  The only other language that uses it is JSON and the unit tests don't include empty labels.
```json
{
  "": "empty",
  "x": "nonempty"
}
```
only yields:
```
x	input.json	/^  "x": "nonempty"$/;"	s
```

masatake:
I misunderstood. Ctags cannot make an empty tag.
The `allowNulTag = true` suppresses warnings when a parser tries to make an empty tag. Sorry for making you confused.

cousteaulecommandant:
Don't worry :)

So what do we do then?  Option 1 is not available.  Shall we go for option 2?  That means not creating tags for the null identifier, i.e., there is one identifier ctags will ignore.  This can be done via the allowNullTag option (the name is misleading; should be "tagMayBeNull"), explicitly in the parsing code (if identifier is empty, return without creating a tag) or simply trigger a warning because we want to inform the user that a tag was dropped.

If we do want to include a tag for `\`, that leaves us on options 3 to 5 – we have to give it a made up name as a replacement for the empty string. It doesn't even have to start with `\`; it could be `(null)` or `(null identifier)` (the latter contains a space so it's impossible that it is mistaken for a literal `\(null)`).

hirooih:
> That means not creating tags for the null identifier, i.e., there is one identifier ctags will ignore.

I think this is the way to go.

> This can be done via the allowNullTag option (the name is misleading; should be "tagMayBeNull"), explicitly in the parsing code (if identifier is empty, return without creating a tag)

I think `ignoreNullTag` is a better name.

> or simply trigger a warning because we want to inform the user that a tag was dropped.

I think this is better, but just ignoring is enough.

***

And I hope someone will implement the true `allowNullTag`.  Don't you? Masatake-san :-)

cousteaulecommandant:
> I think `ignoreNullTag` is a better name.

Perhaps, yeah.

> > or simply trigger a warning because we want to inform the user that a tag was dropped.
> 
> I think this is better, but just ignoring is enough.

I mentioned this because triggering a warning was the behavior you got in my original commit, since the default behavior for ctags is to display the message "Notice: ignoring null tag".  If I remove the `def->allowNullTag = true;` I have added (but not pushed yet) we'll get this behavior.
Then again, keeping it enabled "just in case support for null tags is added in a future" might be a good idea.
Whichever you decide I'll do :)

> And I hope someone will implement the true `allowNullTag`. Don't you? Masatake-san :-)

This could be useful for a variety of languages.  JSON is currently the only one that enables this, but e.g. Tcl allows empty variable/array/function names.  (No idea how this would be handled internally, but I'll let the gurus of ctags figure out.) :)

hirooih:
> If I remove the def->allowNullTag = true; I have added (but not pushed yet) we'll get this behavior.

I see.  Let's remove it!

cousteaulecommandant:
Do I remove the `localparam \` from the unit test as well?

masatake:
I have two concerns:

1. The current implementation of libreadtags skips lines with empty names.
```
[yamato@dev64]~/var/ctags-github% cat tags | tail -3
x	/tmp/foo.sh	/^function x$/;"	f
	/tmp/foo.sh	/^function ""$/;"	f
y	/tmp/foo.sh	/^function y$/;"	f
[yamato@dev64]~/var/ctags-github% ./readtags -t tags -l
x	/tmp/foo.sh	/^function x$/
y	/tmp/foo.sh	/^function y$/
```

libreadtags skips such lines at
```
static int readTagLine (tagFile *const file, int *err)
{
	int result;
	do
	{
		result = readTagLineRaw (file, err);
	} while (result && *file->name.buffer == '\0');
	return result;
}
```

The line comparing with `'\0'` comes from the exuberant ctags. 
Our ancestors assume an empty tag is invalid.

2. the other file formats

u-ctags supports etags and xref. Do they consider null tags?

Could you open a separate issue to support a null tag?

cousteaulecommandant:
> I have two concerns:

This is regarding the idea of handling null tags as a future addition and not something blocking this PR I need to handle right now, right?

> Could you open a separate issue to support a null tag?

I suppose this was addressed at @hirooih, right?

hirooih (approved):
LGTM

Thanks a lot!

parsers/verilog.c
```diff
@@ -873,12 +873,30 @@ static int _readWordToken (tokenInfo *const token, int c, bool skip)
...
+			vStringPut (token->name, '\\');
+		while (isgraph (c))
```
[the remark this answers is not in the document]
hirooih: When I tried to Merge this PR, I remembered this.

@cousteaulecommandant,

How do you think?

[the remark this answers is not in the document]
cousteaulecommandant: > Considering locales excepting C, I think `isgraph()` can not be used here.

`isalpha()`, `isalnum()`, and `isspace()` are used throughout the Verilog parser, and are also affected by locales.  I don't see why `isgraph()` would be any different, and would have assumed it is preferred for consistency with the rest of the parser, which uses the `is*()` functions rather than checking the values explicitly.

> The isalpha function tests for any character for which isupper or islower is true, or any character that is one of a **locale-specific** set of alphabetic characters for which none of iscntrl, isdigit, ispunct, or isspace is true.

> The isspace function tests for any character that is a standard white-space character or is one of a **locale-specific** set of characters for which isalnum is false.

If you think this family of functions shouldn't be used at all just in case non-ASCII characters or messy locales are present, then I think it should be fixed *for the whole parser*, in a separate PR.  The purpose of this PR was to introduce a new feature; I don't think it's within its scope to fix a different problem that was already there.

So, for consistency with the rest of the coding style in the project, I think it's best to use `isgraph()` here, since it's the approach the rest of the parser seems to use; and perhaps fix it along with the rest in a future commit.  But I can make my piece of code an exception and make it use explicit ASCII numeric boundaries if you prefer (which, in my opinion, is less readable and maintainable).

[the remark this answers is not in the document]
hirooih: I totally agree with you on each of your comments.

I got curious about locale after reading the C specification you cited. Sorry for bothering you.


I've also checked the LRM about locale.  What I could only found was;

> A.9.2 Comments
> ...
> comment_text ::= { Any_ASCII_character }

The LRM does not consider locale at all even in comments.
If we would consider locale, we should rather use `isgraph()` here.

hirooih:
I've merged this PR.

@cousteaulecommandant,

Again thank you very much!

parsers/verilog.c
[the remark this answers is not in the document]
cousteaulecommandant: OR, we set the locale to C for parsing.  Is that feasible?  Parsing should be locale-independent anyway, and not be affected by the user's locale.
(Or more specifically, some languages such as Python do support Unicode characters, and others such as Verilog do not, so we could control that.)

cousteaulecommandant:
Nice!  And thank you for all the feedback :)

PR #4157, opened 2024-12-14
Units: Add unit tests for null tags

cousteaulecommandant:
(This is related to #4151.)

Add unit tests in various languages (JSON, Tcl, Verilog/SystemVerilog) to test if null tags are properly implemented.

Note that Tcl doesn't support empty namespace names using `::` syntax since `foo::::bar` is synonymous of `foo::bar`, so we're not testing empty namespace names there, only empty proc names.

Note that the current Tcl parser doesn't seem to support {braced} identifiers at all (including the empty identifier), so perhaps adding support for null tags in Tcl is trickier.
(Maybe I should remove Tcl entirely from the test since it currently doesn't seem to support empty identifiers due to this, and these tests are for testing empty identifiers in general and not a language in particular.)


cousteaulecommandant:
As discussed in #4151, this PR would fail the CI tests (and it indeed did).  This is expected.

(On a side note, I need to configure these CI tests so that I don't receive a dozen email notifications every time I push a PR that doesn't pass the tests.)

masatake:
Thank you. I will merge/cherry-pick incrementally.

masatake:
All changes were merged via #4167, #4262, and #4257.

Issue #4175, opened 2025-01-11
Kotlin: not detecting a function in a specific case


amgdev9:
The name of the parser: Kotlin

The command line you used to run ctags:

```
$ ctags -R --languages=kotlin .
```

The content of input file:

```kotlin
package com.example.myapplication.ui.theme

import android.os.Build
import androidx.compose.foundation.isSystemInDarkTheme
import androidx.compose.material3.MaterialTheme
import androidx.compose.material3.darkColorScheme
import androidx.compose.material3.dynamicDarkColorScheme
import androidx.compose.material3.dynamicLightColorScheme
import androidx.compose.material3.lightColorScheme
import androidx.compose.runtime.Composable
import androidx.compose.ui.platform.LocalContext

private val DarkColorScheme =
    darkColorScheme(primary = Purple80, secondary = PurpleGrey80, tertiary = Pink80)

private val LightColorScheme =
    lightColorScheme(
        primary = Purple40,
        secondary = PurpleGrey40,
        tertiary = Pink40,

        /* Other default colors to override
        background = Color(0xFFFFFBFE),
        surface = Color(0xFFFFFBFE),
        onPrimary = Color.White,
        onSecondary = Color.White,
        onTertiary = Color.White,
        onBackground = Color(0xFF1C1B1F),
        onSurface = Color(0xFF1C1B1F),
        */
    )

@Composable
fun MyApplicationTheme(
    darkTheme: Boolean = isSystemInDarkTheme(),
    // Dynamic color is available on Android 12+
    dynamicColor: Boolean = true,
    content: @Composable () -> Unit,
) {
    val colorScheme =
        when {
            dynamicColor && Build.VERSION.SDK_INT >= Build.VERSION_CODES.S -> {
                val context = LocalContext.current
                if (darkTheme) dynamicDarkColorScheme(context) else dynamicLightColorScheme(context)
            }

            darkTheme -> DarkColorScheme
            else -> LightColorScheme
        }

    MaterialTheme(colorScheme = colorScheme, typography = Typography, content = content)
}
```

The tags output you are not satisfied with:

```
(no output for MyApplicationTheme)
...
```

The tags output you expect:

```
MyApplicationTheme	./app/src/main/java/com/example/myapplication/ui/theme/Theme.kt	/^fun MyApplicationTheme($/;"	m	package:com.example.myapplication.ui.theme
...
```

The version of ctags:

```
[...]

PR #4176, opened 2025-01-13
Kotlin: fix annotation parsing

dolik-rce:
This should fix the problem from #4175.


masatake:
Could you add a test case for this fix?

masatake (approved):

masatake:
Thank you. I will merge this.

PR #4191, opened 2025-02-15
JavaScript,HTML: skip JSX elements

masatake:
This may fix the issue reported at
https://stackoverflow.com/questions/79395369/why-does-universal-ctags-fail-to-record-some-functions-in-the-tags-file-for-a-pr

TODO:
* add more test cases
* extract objects in JavaScript area in HTML area in JavaScript area


jafl (approved):

masatake:
@jafl Thank you for reviewing.
I will polish my change and add comments to the code before merging.

masatake:
We should merge #4193 first.

PR #4193, opened 2025-02-17
JavaScript: Fix held tokens breaking implicit semicolon insertion

### LINKED ISSUE #4192: JavaScript: a bug in handling newline(?)
input-0.js:
```
f = () => {
    return 0
}
var x;
```

ctags extracts only `f`. It doesn't extract `x` though `node` accepts the input.

input-1.js:
```
f = () => {
    return 0;
}
var x;
```
If I added `;`, ctags extracts both `f` and `x`.

b4n:
When emitting a held token, the previous token type was not properly updated, breaking the implicit semicolon emission logic for the next token.

Fixes #4192.


masatake:
Thank you very much!

jafl (approved):

masatake:
Thank you.

PR #4197, opened 2025-02-17
JavaScript: don't extract local constants and variables defined in arrow functions

### LINKED ISSUE #4194: JavaScript:  extract local vars unexpectedly
f.js
```
const f = () => {
    let x = 1
    return x
}

function g() {
    let y = 1
    return y
}
```

```
$ ~/bin/ctags --options=NONE --sort=no -o - /tmp/f.js
f	/tmp/f.js	/^const f = () => {$/;"	f
x	/tmp/f.js	/^    let x = 1$/;"	v
g	/tmp/f.js	/^function g() {$/;"	f
```

ctags extracts `x` though it doesn't extract `y`.
`node` accepts the input.

masatake:
Fix #4194.


masatake:
Surprisingly, even with this change, ctags extracts `x`, `y`, and `z` in

```
const f = () => {
    let x = () => 1
    return x
}

const g = () => {
    let y = function () { return 1 }
    return y
}

function h() {
    let z = function () { return 1 }
    return z
}

```

Do we need to treat callables in any special way?

jafl (approved):

parsers/jscript.c
```diff
@@ -2908,6 +2908,13 @@ static bool parseStatementRHS (tokenInfo *const name, tokenInfo *const token, st
 				state->indexForName = makeFunctionTag (name, sig, false);
 			else
 				convertToFunction (state->indexForName, vStringValue (sig));
+
+			readToken (token);
+			if (isType (token, TOKEN_OPEN_CURLY))
+				parseBlock (token, state->indexForName);
+
+			if (isType (token, TOKEN_CLOSE_CURLY))
+				state->isTerminated = false;
```
b4n, line 2917: This is incorrect: `parseBlock()` will leave `token` on the *block's* close curly brace, which has nothing to do with the current statement's end.  See review

parsers/jscript.c
```diff
@@ -2942,6 +2949,9 @@ static bool parseStatementRHS (tokenInfo *const name, tokenInfo *const token, st
 
 				vStringDelete (sig);
 				sig = NULL;
+				readToken (token);
+				if (isType (token, TOKEN_OPEN_CURLY))
+					parseBlock (token, state->indexForName);
```
b4n, line 2954: Same here, the (weird?) check for the close curly brace is gonna break nesting.

parsers/jscript.c
b4n, line 2917:
```suggestion
			else if (! isType (token, TOKEN_SEMICOLON))
				state->isTerminated = false;
```

parsers/jscript.c
b4n, line 2954:
```suggestion
				if (isType (token, TOKEN_OPEN_CURLY))
				{
					parseBlock (token, state->indexForName);
					/* here we're at the close curly but it's part of the arrow
					 * function body, so skip over not to confuse further code */
					readTokenFull(token, true, NULL);
					state->isTerminated = isType (token, TOKEN_SEMICOLON);
				}
```

b4n (changes requested):
This breaks scope after the end of the arrow function's body.  Suggestion:

```diff
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 6c09af3e3..a6d8e9fde 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -2912,8 +2912,7 @@ static bool parseStatementRHS (tokenInfo *const name, tokenInfo *const token, st
 			readToken (token);
 			if (isType (token, TOKEN_OPEN_CURLY))
 				parseBlock (token, state->indexForName);
-
-			if (isType (token, TOKEN_CLOSE_CURLY))
+			else if (! isType (token, TOKEN_SEMICOLON))
 				state->isTerminated = false;
 		}
 		vStringDelete (sig);
@@ -2951,7 +2950,13 @@ static bool parseStatementRHS (tokenInfo *const name, tokenInfo *const token, st
 				sig = NULL;
 				readToken (token);
 				if (isType (token, TOKEN_OPEN_CURLY))
+				{
 					parseBlock (token, state->indexForName);
+					/* here we're at the close curly but it's part of the arrow
+					 * function body, so skip over not to confuse further code */
+					readTokenFull(token, true, NULL);
+					state->isTerminated = isType (token, TOKEN_SEMICOLON);
+				}
 			}
 		}
 		if (isType (token, TOKEN_CLOSE_CURLY))
```

And to test this:
```console
$ cat /tmp/a.js 
function a() {
  function b() {
    let x = 42;
    return x;
  }
  let y = b();
  return y
}

function c() {
  let d = () => {
    return 42;
  }
  
  function d2() {
    return 1;
  }
}

let e = () => {
  function f() {
    
  }
}

let g = () => {
  let h = () => {
    return 42;
  }
  let i = x => {
    return x * 42;
  }
  let j = () => {
    return 42;
  }
}

function k() {}
$ ./ctags -o- /tmp/a.js 
a	/tmp/a.js	/^function a() {$/;"	f
b	/tmp/a.js	/^  function b() {$/;"	f	function:a
c	/tmp/a.js	/^function c() {$/;"	f
d	/tmp/a.js	/^  let d = () => {$/;"	f	function:c
d2	/tmp/a.js	/^  function d2() {$/;"	f	function:c
e	/tmp/a.js	/^let e = () => {$/;"	f
f	/tmp/a.js	/^  function f() {$/;"	f	function:e
g	/tmp/a.js	/^let g = () => {$/;"	f
h	/tmp/a.js	/^  let h = () => {$/;"	f	function:g
i	/tmp/a.js	/^  let i = x => {$/;"	f	function:g
j	/tmp/a.js	/^  let j = () => {$/;"	f	function:g
k	/tmp/a.js	/^function k() {}$/;"	f
```

masatake:
@b4n Thank you very much. I should be strict when writing a test case.
Will you make a pull request? Or should I include your change to my commit and put a 'Co-Author' mark on the commit?

b4n:
@masatake as you want, but you can integrate these changes to yours so we get one single PR, I'm happy with that.

> […] Do we need to treat callables in any special way?

I don't think we need to.  ATM this is the same with usual nested functions, and so long as the scope is correct I think it's fine, and even useful when using the tags as a TOC.

b4n (changes requested):

parsers/jscript.c
```diff
@@ -2908,6 +2908,12 @@ static bool parseStatementRHS (tokenInfo *const name, tokenInfo *const token, st
 				state->indexForName = makeFunctionTag (name, sig, false);
 			else
 				convertToFunction (state->indexForName, vStringValue (sig));
+
+			readToken (token);
+			if (isType (token, TOKEN_OPEN_CURLY))
+				parseBlock (token, state->indexForName);
+			else if (isType (token, TOKEN_SEMICOLON))
```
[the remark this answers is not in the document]
masatake: Sorry for making too poor a mistake. Updated as your original suggestion.

Units/parser-javascript.r/js-arrow-funcs.d/expected.tags
```diff
@@ -5,7 +5,21 @@ f3	input.js	/^let f3 = () => {$/;"	function	signature:()
 f4	input.js	/^let f4 = (...x) => {$/;"	function	signature:(...x)
 f5	input.js	/^let f5 = (x,...y) => {$/;"	function	signature:(x,...y)
 f6	input.js	/^let f6 = (x,y,...z) => {$/;"	function	signature:(x,y,...z)
+a	input.js	/^var a = {}$/;"	variable
+b	input.js	/^var b = {}$/;"	variable
 f7	input.js	/^a.f7 = x => {$/;"	function	variable:a	signature:(x)
 f8	input.js	/^a.f8 = async () => {$/;"	function	variable:a	signature:()
 f9	input.js	/^b.f9 = (async () => {$/;"	function	variable:b	signature:()
 f10	input.js	/^b.f10 = (async x => {$/;"	function	variable:b	signature:(x)
+a	input-0.js	/^function a() {$/;"	function	signature:()
```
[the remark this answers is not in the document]
masatake: Yes. I extended the units test harness a few years ago.
See https://docs.ctags.io/en/latest/testing-parser.html#how-to-write-a-test-case

(However, it is a quick hack. It is not integrated with misc/review script.)

masatake:
@b4n, thank you again.

b4n:
My pleasure :)

PR #4216, opened 2025-03-22
Asm: extract section from .pushsection directive

masatake:


masatake:
This is unacceptable because `.previous` is not considered.

PR #4231, opened 2025-04-21
main: (bugfix) don't reset lineFposMap even in the 2nd pass when the parser runs as a guest

masatake:


Units/parser-markdown.r/c-guest.d/README
```diff
@@ -0,0 +1 @@
+This triaged an assertion failure in mio_setpos.
```
b4n, line 1:
```suggestion
This triggered an assertion failure in mio_setpos().
```

masatake: Thank you!

PR #4236, opened 2025-04-27
V: don't use multi-bytes chars in debug print

masatake:
This change avoids the trouble reported in #4235.


b4n:
Looks like a good way not to trigger the issue if it does indeee work 👍

PR #4240, opened 2025-05-05
SQL: make foreign tags fo 'interenal' functions

masatake:


masatake:
Now we have version information for each kind, role, field, and extra. So the Changes section of the man pages is redundant.

PR #4246, opened 2025-05-09
circleci: switch to use Fedoa42

masatake:


masatake:
Supprisingly awk is not installed by default.

PR #4257, opened 2025-05-28
SystemVerilog,Verilog: accept empty names for any kind of language objects

masatake:
See #4151, #4152, and #4157.

The first patch makes the SystemVerilog parser accept empty names ~for modules, structs, and members. (You can add more.)~ .

The case derived from #4157 tests the first patch.

### DIFF
--- a/Units/parser-verilog.r/systemverilog-nulltags.d/args.ctags
+++ b/Units/parser-verilog.r/systemverilog-nulltags.d/args.ctags
@@ -1,0 +1,3 @@
+--sort=no
+--extras=+{nulltag}
+--_xformat=%-16N %-10K %4n %-16F %C %{scopeKind}:%{scope}
--- a/Units/parser-verilog.r/systemverilog-nulltags.d/expected.tags-x
+++ b/Units/parser-verilog.r/systemverilog-nulltags.d/expected.tags-x
@@ -1,0 +1,14 @@
+                 module        6 input.sv module \ ; :
+                 struct       16 input.sv } \ ; module:
+                 member       11 input.sv } \ ; struct:.
+een              member       15 input.sv } \een ; struct:.
+en               struct       26 input.sv } \en ; module:
+                 member       21 input.sv } \ ; struct:.en
+enn              member       25 input.sv } \enn ; struct:.en
+n                module       29 input.sv module \n ; :
+                 struct       39 input.sv } \ ; module:n
+                 member       34 input.sv } \ ; struct:n.
+nen              member       38 input.sv } \nen ; struct:n.
+nn               struct       49 input.sv } \nn ; module:n
+                 member       44 input.sv } \ ; struct:n.nn
+nnn              member       48 input.sv } \nnn ; struct:n.nn
--- a/Units/parser-verilog.r/systemverilog-nulltags.d/input.sv
+++ b/Units/parser-verilog.r/systemverilog-nulltags.d/input.sv
@@ -1,0 +1,51 @@
+/* Legend:
+ * <e> = empty (null)
+ * <n> = non-empty
+ */
+
+module \ ;
+	struct {
+		struct {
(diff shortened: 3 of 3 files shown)


hirooih:
Hi,

I think the following fix is enough and better, if I understand correctly.

```diff
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -2207,6 +2207,7 @@ extern parserDefinition* VerilogParser (void)
        def->parser     = findVerilogTags;
        def->initialize = initializeVerilog;
        def->selectLanguage  = selectors;
+       def->allowNullTags = true;
        return def;
 }
 
@@ -2223,5 +2224,6 @@ extern parserDefinition* SystemVerilogParser (void)
        def->extensions = extensions;
        def->parser     = findVerilogTags;
        def->initialize = initializeSystemVerilog;
+       def->allowNullTags = true;
        return def;
 }
```

> ... accept empty names for modules, structs, and members.

How did you choose them?

hirooih:
Masatake-san,

>> How did you choose them?

> Based on the test input ...

I see.

I don't see any good reason to allow null-tag on only certain TAGs.

It seems that I don't understand the difference between `parserDefinition::allowNulltag` and `tagEntryInfo::allowNulltag`.

> However, we will lose the way to get the report.

Do you mean the following is better than my previous diff?

```diff
--- a/parsers/verilog.c
+++ b/parsers/verilog.c
@@ -1043,6 +1043,7 @@ static void createTagFull (tokenInfo *const token, verilogKind kind, int role, t
        else
                initRefTagEntry (&tag, vStringValue (token->name), kind, role);
        updateTagLine (&tag, token->lineNumber, token->filePosition);
+       tag.allowNullTag = true;
 
        verbose ("Adding tag %s (kind %d)", vStringValue (token->name), kind);
        if (currentContext->kind != K_UNDEFINED
```

masatake:
> Do you mean the following is better than my previous diff?

I don't mean so. Conceptually, both your versions are the same; the main part allows SystemVerilog and Verilog parsers to emit null tags for ANY language objects.

> I don't see any good reason to allow null-tag on only certain TAGs.

SystemVerilog supports many kinds:
```
~/bin/ctags --list-kinds=SystemVerilog | head
c  constants (parameter, specparam, enum values)
d  text macros
e  events
f  functions
m  modules
n  net data types
p  ports
r  variable data types
t  tasks
b  blocks (begin, fork)
```

Does the SystemVerilog language specification allow null-name for these kinds?
I don't know SystemVerilog well. However, a text macro may need a name.

If the SystemVerilog parser emits a null tag for a text macro kind, it implies a bug in the parser.
If you don't set parserDefinition::allowNulltag in the parser, the main part can warn about such unexpected null tag emission. We have a chance to know the potential bug.

hirooih:
> Does the SystemVerilog language specification allow null-name for these kinds?

Now I understand what you mean.

I will check LRM how `escaped_identifier` (See. https://github.com/universal-ctags/ctags/pull/4129#issuecomment-2515437734) is used.

Give me some days.

hirooih:
@masatake san

I've checked LRM.

> Does the SystemVerilog language specification allow null-name for these kinds?

The answer is "yes".

A.9.3 Identifiers

```text
escaped_identifier ::= \ {any_printable_ASCII_character_except_white_space} white_space
...
identifier ::=
  simple_identifier
  | escaped_identifier
...
simple_identifier ::= [ a-zA-Z_ ] { [ a-zA-Z0-9_$ ] }
```

All kinds are classified as identifiers, including text macros.

```text
text_macro_definition ::=
  `define text_macro_name macro_text
text_macro_name ::=
  text_macro_identifier [ ( list_of_formal_arguments ) ]
...
text_macro_identifier ::=
  identifier
```

hirooih (approved):
Thank you!

PR #4266, opened 2025-06-02
Various fixes related to makePromise() pointing behind EOF

techee:
For more info, see the individual commit messages and the discussion in https://github.com/geany/geany/pull/4303

To reproduce the crash, use https://github.com/geany/geany/pull/4303/commits/79fb41516f746c588a52b1b01aa805cdf31da852 from https://github.com/geany/geany/pull/4303. Note that to reproduce this issue, it has to use CRLF line endings (might get lost when copy/pasting from the github web interface).

The issue can be reproduced using `ctags --extras=+rg -o- wp-guest.php`.

### DIFF
--- a/main/mio.c
+++ b/main/mio.c
@@ -1082,8 +1082,13 @@ int mio_seek (MIO *mio, long offset, int whence)
 		switch (whence)
 		{
 			case SEEK_SET:
-				if (offset < 0 || (size_t)offset > mio->impl.mem.size)
+				if (offset < 0)
 					errno = EINVAL;
+				else if ((size_t)offset > mio->impl.mem.size)
+				{
+					errno = EINVAL;
+					mio->impl.mem.pos = mio->impl.mem.size;
+				}
 				else
 				{
 					mio->impl.mem.pos = (size_t)offset;
@@ -1092,10 +1097,12 @@ int mio_seek (MIO *mio, long offset, int whence)
 				break;
 
 			case SEEK_CUR:
-				if ((offset < 0 && (size_t)-offset > mio->impl.mem.pos) ||
-					mio->impl.mem.pos + (size_t)offset > mio->impl.mem.size)
+				if (offset < 0 && (size_t)-offset > mio->impl.mem.pos)
+					errno = EINVAL;
+				else if (mio->impl.mem.pos + (size_t)offset > mio->impl.mem.size)
 				{
 					errno = EINVAL;
+					mio->impl.mem.pos = mio->impl.mem.size;
 				}
 				else
 				{
@@ -1282,7 +1289,10 @@ int mio_setpos (MIO *mio, MIOPos *pos)
 		rv = -1;
 
 		if (pos->impl.mem > mio->impl.mem.size)
+		{
 			errno = EINVAL;
+			mio->impl.mem.pos = mio->impl.mem.size;
+		}
 		else
 		{
 			mio->impl.mem.ungetch = EOF;
--- a/main/mio.c
+++ b/main/mio.c
@@ -1082,13 +1082,8 @@ int mio_seek (MIO *mio, long offset, int whence)
 		switch (whence)
 		{
 			case SEEK_SET:
-				if (offset < 0)
+				if (offset < 0 || (size_t)offset > mio->impl.mem.size)
 					errno = EINVAL;
-				else if ((size_t)offset > mio->impl.mem.size)
-				{
-					errno = EINVAL;
-					mio->impl.mem.pos = mio->impl.mem.size;
-				}
 				else
 				{
 					mio->impl.mem.pos = (size_t)offset;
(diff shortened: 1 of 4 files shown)


techee:
Note that the MIO patches are more or less a "bonus" and not really essential. I think, regardless of what the exact problem of the offset miscalculation is, `pushArea()` should detect such situations and if something invalid is detected, discard the promise.

b4n:
See also #4267 that adds a simpler version of the test case mentioned above.  It seems that any CRLF that triggers a sub-parser will reveal the issue.

main/mio.c
b4n, line 1: IMO this change is not a good idea.  I see a few issues, which basically stem from the C standard.

First, the C standard does *not* specify this behavior.  Quoting [C17 N2176](https://web.archive.org/web/20181230041359/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf):
   > ## Description
   > The **`fseek`** function sets the file position indicator for the stream pointed to by `stream`. If a read or
write error occurs, the error indicator for the stream is set and **`fseek`** fails.
   >
   > For a binary stream, the new position, measured in characters from the beginning of the file, is
obtained by adding `offset` to the position specified by `whence`. The specified position is the
beginning of the file if `whence` is **`SEEK_SET`**, the current value of the file position indicator if
**`SEEK_CUR`**, or end-of-file if **`SEEK_END`**. A binary stream need not meaningfully support **`fseek`** calls
with a `whence` value of **`SEEK_END`**.
   >
   > For a text stream, either `offset` shall be zero, or `offset` shall be a value returned by an earlier
successful call to the **`ftell`** function on a stream associated with the same file and `whence` shall be
**`SEEK_SET`**.
   >
   > After determining the new position, a successful call to the **`fseek`** function undoes any effects of the
**`ungetc`** function on the stream, clears the end-of-file indicator for the stream, and then establishes
the new position. After a successful **`fseek`** call, the next operation on an update stream may be
either input or output.
   
With *After determining the new position, a successful call to the **`fseek`** function […] then establishes
the new position*, one could even read that it *should not* change the position if the call is not successful (and can it be if it's not actually honored?).

But at any rate, as there is nothing specifying this behavior, it can't be relied upon for FILE-based streams, and thus I don't see any value of changing the behavior for memory streams, it just adds a more or less random behavior for a specific case.  If it's not standard, the caller shouldn't rely on it, or it's gonna be brittle.

Additionally, it's *not* something my libc does. I doesn't behave like you suggest (here I have glibc 2.36):
Using the very crude
```c
#include <stdlib.h>
#include <stdio.h>

int main(int argc, char **argv)
{
  FILE *fp = fopen(argv[1], "rb");
  
  if (! fp)
    return 1;

  long off = atol(argv[2]);

  fprintf(stderr, "initial position: %ld\n", ftell(fp));
  fprintf(stderr, "seeking to %ld: %d\n", off, fseek(fp, off, SEEK_SET));
  fprintf(stderr, "ftell(): %ld\n", ftell(fp));
  fprintf(stderr, "seeking to +32: %d\n", fseek(fp, 32, SEEK_CUR));
  fprintf(stderr, "ftell(): %ld\n", ftell(fp));
  
  fprintf(stderr, "seeking to end: %d\n", fseek(fp, 0, SEEK_END));
  fprintf(stderr, "ftell(): %ld\n", ftell(fp));

  fclose(fp);
  
  return 0;
}
```
and running it:
```console
$ ./seek.out seek.c 1000
initial position: 0
seeking to 1000: 0
ftell(): 1000
seeking to +32: 0
ftell(): 1032
seeking to end: 0
ftell(): 613
```
You can clearly see that it sought (hehe, `seek()`ed) past the end: when seeking to `SEEK_END`, I get 613, but if I seek to 1000 I get that 1000 (which is then past the end), and I can even further `SEEK_CUR` happily.

---

So in the end I'd rather not have this change any internal on failure, and leave it to the caller.  Originally MIO was designed to mimic as closely as possible the equivalent C library calls, not adding any fanciness anywhere it, and I don't think it should deviate to what can be expected from the C library.

At some point it had tests to compare against the C library itself, maybe we should re-introduce this in uctags -- not that it would catch anything here, as the test would have not to depend on unspecified behavior, but we could verify it's not diverging on specified behavior.

b4n:
Apart from the remark above, looks reasonable.  I'll have to test it properly with a clean head though, it's 1:20am here and I'm not fresh anymore -- but what's left of my brain like the other changes :slightly_smiling_face:

main/mio.c
techee: OK, it was just an idea inspired by the fact that ctags uses mio_seek() and mio_setpos() without any checks of the returned value and that I thought that in the error cases it would be better to be "closer" to the provided value. The right thing of course is to properly check the return values and recover accordingly.

Let's discard this patch.

techee:
I've added a few more checks inside pushArea() - I believe all the `mio_setpos()` and `mio_setpos()` return values should be checked because the offsets coming from parsers could be wrong.

masatake:
I'm sorry for not getting back to you sooner. I'm looking for bulk-free time to work on these pull requests submitted from the Geany project. I had a hard time resolving https://github.com/universal-ctags/ctags/pull/4097, so I will have a hard time again. However, these are at the head of my run queue in my mind. Please wait for a while.

techee:
@masatake No worries. I think for Geany itself, if we apply just this PR alone without any other ctags patches that were made in the meantime, we'll be fine for the upcoming Geany release.

masatake:
I started working on this one.
Before reading your change, I would like to fix the issue myself. After finding a fix, I would like to compare it with the change in this pull request. Please, give me time.

NOTE: The way to reproduce.

1. Save wp-guest.php in https://github.com/geany/geany/pull/4303/commits/79fb41516f746c588a52b1b01aa805cdf31da852 .
2. Convert the line ending with CRLF with M-x set-buffer-file-coding-system => dos on Emacs.
3. Run `ctags --options=NONE --extras=+g -o - wp-guest.php` .

What I got:
```
ctags --options=NONE --extras=+g -o - /tmp/input.php
ctags: out of memory
```

techee:
We did some more things related to invalid promises in Geany in https://github.com/geany/geany/pull/4330. Just let me know if I should port those changes to ctags too.

PR #4267, opened 2025-06-02
Fix guest parser offset on the last line with CRLF line endings

b4n:
This is related to #4266, and is basically the test part of #4097 that is actually still relevant, only the CRLF got lost in translation.
This is just a test case showing the issue, so we have something to work on.

FWIW, here's the output:
```console:
$ ./ctags -o- $(cat Units/parser-php.r/subparser-crlf.b/args.ctags) Units/parser-php.r/subparser-crlf.b/input.php
ctags: memory for mio may be exhausted
```

If we trick *mini-geany* to run it, it's worse:
```console
$ git diff
diff --git a/main/mini-geany.c b/main/mini-geany.c
index e02b3c19a..06d2f8be6 100644
--- a/main/mini-geany.c
+++ b/main/mini-geany.c
@@ -370,7 +370,7 @@ extern int main (int argc, char **argv)
         {
             printf("\nParsing %s:\n", argv[i]);
             /* parseRawBuffer() is called repeatadly during Geany execution */
-            parseRawBuffer(argv[i], NULL, 0, getNamedLanguage("C", 0), tagArray);
+            parseRawBuffer(argv[i], NULL, 0, getNamedLanguage("PHP", 0), tagArray);
 
             processCollectedTags(tagArray);
         }
$ ./mini-geany Units/parser-php.r/subparser-crlf.b/input.php
This parser only parses C files - provide them as arguments on the command line or get a hard-coded buffer parsed when no arguments are provided


Parsing Units/parser-php.r/subparser-crlf.b/input.php:
Error: memory for mio may be exhausted
Segmentation fault
```
But that's only because with it the `fatal()` call does not call `exit()`, and thus does not prevent the use of an `NULL` `submio`.


b4n:
@masatake I have a fix that mostly look reasonable, but I'm still not sure I get everything around this, so please review carefully and don't trust the logic :)  Results seem solid though.

PR #4269, opened 2025-06-06
JavaScript: # indicates a private class field or method

jafl:


jafl:
@masatake How can I indicate that a private field is also static?

masatake:
> @masatake How can I indicate that a private field is also static?

Interesting topic.

The main part doesn't provide a standard way for representing static methods (or class methods).

There are two approaches:

1. Introducing a parser-specific field for recording "this method is static".
2. Introducing a specialized kind for representing static methods.

Let's see some parsers.

Python uses a parser-specific field named `decorators:`.
```
$ cat foo.py
class C:
    @classmethod
    def m():
        pass
$ ~/bin/ctags -o - --fields-Python='*' /tmp/foo.py 
C	/tmp/foo.py	/^class C:$/;"	c
m	/tmp/foo.py	/^    def m():$/;"	m	class:C	decorators:classmethod
```

C++ also uses a parser-specific field named `properties:`.
```
$ cat /tmp/foo.cpp 
class C {
  static void m(void) {}
};
$ ~/bin/ctags -o - --fields-C++='*' /tmp/foo.cpp
C	/tmp/foo.cpp	/^class C {$/;"	c	file:
m	/tmp/foo.cpp	/^  static void m(void) {}$/;"	f	class:C	typeref:typename:void	file:	properties:static
```

Ruby uses a specialized kind named `singletonMethod` for the purpose:
```
$ cat /tmp/foo.rb
class C
  def self.m
    
  end
end
$ ~/bin/ctags -o - --fields=+Kz  --fields-Ruby='*' /tmp/foo.rb
C	/tmp/foo.rb	/^class C$/;"	kind:class
m	/tmp/foo.rb	/^  def self.m$/;"	kind:singletonMethod	class:C
```

jafl:
@masatake I am hoping to do it the C++ way.  From studying the code, however, I couldn't figure out how it was done.  I found `cxxTagSetProperties` in `cxx_tag.c`, which ultimately calls `attachParserField` in `entry.c`.  What `fieldType` should I specify (from `field.h`)?

masatake:
I recommend you read the code `parser/python.c` and the comment about `attachParserField()` in `main/entry.h`.

```diff
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 168dfb533..869daac0a 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -41,6 +41,7 @@
 #include "debug.h"
 #include "dependency.h"
 #include "entry.h"
+#include "field.h"
 #include "keyword.h"
 #include "numarray.h"
 #include "parse.h"
@@ -251,6 +252,10 @@ typedef enum {
 	JS_CLASS_CHAINELT,
 } jsClassRole;
 
+typedef enum {
+	F_THEFIELD,
+} jsField;
+
 static roleDefinition JsFunctionRoles [] = {
 	/* Currently V parser wants this items. */
 	{ true, "foreigndecl", "declared in foreign languages" },
@@ -280,6 +285,14 @@ static kindDefinition JsKinds [] = {
 	{ true,  'M', "field",		  "fields"           },
 };
 
+static fieldDefinition JsFields[] = {
+	{
+		.name = "theNameOfField",
+		.description = "...description...",
+		.enabled = true,
+	},
+};
+
 static const keywordTable JsKeywordTable [] = {
 	/* keyword		keyword ID */
 	{ "function",	KEYWORD_function			},
@@ -527,6 +540,14 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 	updateTagLine (&e, token->lineNumber, token->filePosition);
 	e.extensionFields.scopeIndex = scope;
 
+	/* There are two ways (A and B) to attach a value to field.
+	 *
+	 * This is the way A. A. is suitable for a static value.
+	 * Attaching after initTagEntry() before makeTagEntry():  */
+	if (is_static)
+		attachParserField (&e, JsFields[F_THEFIELD].ftype,
+						   "value");
+
 	if (is_private)
 		e.extensionFields.access = "private";
 
@@ -565,6 +586,21 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 		e.allowNullTag = 1;
 
 	index = makeTagEntry (&e);
+
+	/*
+	 * This is the way B. B. is suitable for dynamic value.
+	 * Attaching after makeTagEntry():
+	 *
+	 * Choose A or B.
+	 */
+	{
+		vString *tmp = vStringNew();
+		/* build the value for the field */
+		attachParserFieldToCorkEntry (index, JsFields[F_THEFIELD].ftype,
+									  vStringValue (tmp));
+		vStrngDelete (tmp);
+	}
+
 	/* We shold remove This condition. We should fix the callers passing
 	 * an empty name instead. makeTagEntry() returns CORK_NIL if the tag
 	 * name is empty. */
@@ -3797,6 +3833,8 @@ extern parserDefinition* JavaScriptParser (void)
 	 */
 	def->kindTable	= JsKinds;
 	def->kindCount	= ARRAY_SIZE (JsKinds);
+	def->fieldTable = JsFields;
+	def->fieldCount = ARRAY_SIZE (JsFields);
 	def->parser		= findJsTags;
 	def->initialize = initialize;
 	def->finalize   = finalize;
```

parsers/jscript.c
```diff
@@ -2352,7 +2369,16 @@ static bool parseMethods (tokenInfo *const token, int class_index,
 						else if (is_setter)
 							kind = JSTAG_SETTER;
 
-						index_for_name = makeJsTag (name, kind, signature, NULL);
+						if (is_private)
+						{
+							vString * s = vStringNew ();
+							vStringPut (s, '#');
```
masatake: You can unify the two function calls into one with:
```
vString * s = vStringNewInit ("#");
```

I don't say this is efficient:-).

parsers/jscript.c
```diff
@@ -2447,8 +2473,17 @@ static bool parseMethods (tokenInfo *const token, int class_index,
 			}
 			else
 			{
+				if (is_private)
+				{
+					vString * s = vStringNew ();
+					vStringPut (s, '#');
```
masatake: vStringNewInit() can be used here.

parsers/jscript.c
```diff
@@ -2352,7 +2369,16 @@ static bool parseMethods (tokenInfo *const token, int class_index,
...
 							kind = JSTAG_SETTER;
 
-						index_for_name = makeJsTag (name, kind, signature, NULL);
+						if (is_private)
+						{
+							vString * s = vStringNew ();
+							vStringPut (s, '#');
+							vStringCat (s, name->string);
+							vStringCopy (name->string, s);
+							vStringDelete (s);
```
masatake: Instead of updating the `string` member of the `name` token, we can make the `#` prefixed private name in makeJsTagCommon.

```diff
diff --git a/parsers/jscript.c b/parsers/jscript.c
index 168dfb533..e88799c2e 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -486,12 +499,21 @@ static int makeJsRefTagsForNameChain (char *name_chain, const tokenInfo *token,
 		: index;
 }
 
+static vString *makPrivateName (vString *orignal)
+{
+	vString * s = vStringNewInit ("#");
+	vStringCopy (s, orignal);
+	return s;
+}
+
 static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 							vString *const signature, vString *const inheritance,
 							bool anonymous, bool is_private, bool nulltag)
 {
 	int index = CORK_NIL;
-	const char *name = vStringValue (token->string);
+	vString *vname = is_private? makePrivateName (vname): token->string;
+	const char *name = vStringValue (vname);
+
 
 	const char *p;
 	char *name_chain = NULL;
@@ -501,7 +523,11 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 			name_chain = eStrndup (name, (size_t) (p - name));
 		name = p + 1;
 		if (name[0] == '\0')
+		{
+			if (is_private)
+				vStringDelete (vname);
 			return CORK_NIL;
+		}
 	}
 
 	int scope = token->scope;
@@ -565,8 +599,11 @@ static int makeJsTagCommon (const tokenInfo *const token, const jsKind kind,
 	 * an empty name instead. makeTagEntry() returns CORK_NIL if the tag
 	 * name is empty. */
 	if (index != CORK_NIL)
 		registerEntry (index);
 
+	if (is_private)
+		vStringDelete (vname);
+
 	return index;
 }
 
```

Just an idea.

masatake:
Whether `#` is part of the method name is a question you/we must ponder.
These days, I discuss these topics with ChatGPT:-P

Thinking about shell scripts, do you think `$` is a part of the name of a variable?

Even if a name doesn't include `#`, a user can know the method is private from `acccess:private` field.


These are just inputs to you. I have no answer for you. Please, make JavaScript people and developers of client tools comfortable!

jafl:
> Thinking about shell scripts, do you think `$` is a part of the name of a variable?

Absolutely.  In JavaScript, a class can have both a public function `foo` and a private function `#foo`.  And in Perl, you can have `$foo` and `@foo` and `%foo`.

jafl:
@masatake Thanks for the help.  I have updated the PR.

masatake:
Unlike C++, JavaScript itself has the concept named properties. I'm afraid that using the name `properties` as the new parser-specific field name makes ctags' users programming with JavaScript confused. Is it O.K.?  

----

The test is failed because you added a new field.
After running `make tmain` locally, run `misc/review`.
You can reflect the actual output to expectext output interactively.

```
... make tmain ...
Failed tests
ptag-in-optlib-parser/stdout-compare
list-roles/stdout-compare
list-fields/stdout-compare

make: *** [Makefile:8166: tmain] Error 1
[yamato@dev64]~/var/ctags-github% misc/review 
1) <n>ext
2) <d>iff
3) <a>ccept
4) <S>hell
5) <R>un
6) <q>uit
[1/3 [ Tmain/list-fields.d ]]? =
###  stdout
--- ./Tmain/list-fields.d/stdout-expected.txt	2025-06-10 15:45:33.851066553 +0900
+++ /home/yamato/var/ctags-github/Tmain/list-fields.d/stdout-actual.txt	2025-06-10 15:45:43.336692476 +0900
@@ -34,0 +35,2 @@
+-	macrodef	no	C++	s--	no	--	macro definition
+-	name	yes	C++	s--	no	--	aliased names
[1/3 [ Tmain/list-fields.d ]]? a
1) <n>ext
2) <d>iff
3) <a>ccept
4) <S>hell
5) <R>un
6) <q>uit
[2/3 [ Tmain/list-roles.d ]]? 
...
```

----

When adding/removing kinds, roles, parser-specific fields, and/or parser-specific extras (KRFX or KRFE), I would like you to update versionCurrent and versionAge of the parser like:
```
index d9cb6d07d..d7ff2d44e 100644
--- a/parsers/jscript.c
+++ b/parsers/jscript.c
@@ -3773,8 +3773,8 @@ extern parserDefinition* JavaScriptParser (void)
 	def->dependencies = dependencies;
 	def->dependencyCount = ARRAY_SIZE (dependencies);
 
-	def->versionCurrent = 1;
-	def->versionAge = 1;
+	def->versionCurrent = 2;
+	def->versionAge = 2;
 
 	return def;
 }
```

See main/parse.h when we should increment the members.
See also a planed change about parser versioning, https://github.com/universal-ctags/ctags/pull/4271 .

Units/parser-javascript.r/js-es6-class-private.d/args.ctags
```diff
@@ -0,0 +1,2 @@
+--fields=+a
```
masatake, line 1: Add --sort=no to make review the test case easier.

jafl:
> Unlike C++, JavaScript itself has the concept named properties. I'm afraid that using the name `properties` as the new parser-specific field name makes ctags' users programming with JavaScript confused. Is it O.K.?

I think it's good to be consistent with the rest of the languages that use "static".  I suppose we could use "attributes", though that is also a keyword.

masatake:
My alternative ideas are `static:1`, `static:yes`, `static:true`, or `static:`. 
These are just ideas. If you, a person who uses JavaScript, think it is comfortable, it's good.

masatake:
Sorry to be late in saying the following:

* Write about the new field to "Parser related changes" in docs/news/HEAD.rst
* Enabling the field by default is not a bad idea.

jafl:
> Sorry to be late in saying the following

No problem!  Updated.

masatake:
`misc/review` may help you.

jafl:
"properties" is off by default for C++, so I will leave it that way for JavaScript, too.

masatake (approved):
~LGTM. Please, merge this.~

I will merge this.
I remember we have to update man/ctags-lang-javascript.7.rst.in till I merge the KRFX versioning. I will update the man page after merging this pull request.

masatake:
Thank you.

PR #3287, opened 2022-01-27
[SELF-INCOMPATIBLE] RpmMacros: remove (macros.*) from the map

masatake:
With this map (macros.\*), ctags could not choose C++ parser
when macros.h is specified as input.
(macros.\*) is too generic as a pattern.

Close #3286.

PR #4270, opened 2025-06-08
main: using regex for choosing a paser for given file name

masatake:
main: using regex for choosing a parser for the given file name
    
This change extends --map-<LANG> option to support regular
expression matching with the full file name.
    
The original --map-<LANG> option supports glob based matching
and extension comparison with the file basename.
However, two methods are not enough if the file names are too
generic. See https://github.com/universal-ctags/ctags/pull/3287 .
    
The regular expression passed to --map-<LANG> must be surround
by % character like

   --map-RpmMacros='%(.*/)?macros\.d/macros\.([^/]+)$%'

If you want to match in a case-insensitive way, append `i' after the second % like
    
   --map-RpmMacros='%(.*/)?macros\.d/macros\.([^/]+)$%i'
    
If you want to use % as part of an expression, put \ before % for escaping.
    
TODO:
    
- [ ] reconsider name regex, rxpr, or something
- [ ] update ctags.1
- [ ] add Tmain test cases
- [ ] add description to --help
- [X] extend optlib2c
- [X] add --list-map-regex
- [X] add --list-maps
- [ ] add pcre backend

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -59,6 +59,9 @@
 /*  The following separators are permitted for list options.
  */
 #define EXTENSION_SEPARATOR '.'
+#define REXPR_START '%'
+#define REXPR_STOP '%'
+#define REXPR_ICASE 'i'
 #define PATTERN_START '('
 #define PATTERN_STOP  ')'
 #define IGNORE_SEPARATORS   ", \t\n"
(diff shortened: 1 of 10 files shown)


masatake:
Updating the man page is the last task.

PR #4275, opened 2025-06-12
Meson: extract config vars defined inside configuration_data({...})

masatake:
I should pull this request into two.

(
This random should be removed from here:
```
input : "\n" L3781
match :'\n'            main[ 0] / ^([a-zA-Z_][a-zA-Z_0-9]*)[ \t\n]*=([^=]|$)/
match :'\n'            main[ 1] / ^project[ \t\n]*\\([ \t\n]*'([^']*[^\\\\])'[ \t\n]*/
match :'\n'            main[ 2] / ^subdir[ \t\n]*\\([ \t\n]*'([^']*[^\\\\])'[ \t\n]*/
match :'\n'            main[ 3] / ^test[ \t\n]*\\([ \t\n]*'([^']*[^\\\\])'[ \t\n]*/
match :'\n'            main[ 4] / ^[ \t\n]+/

...                   TABLE[ N] ...
result: matched 1 bytes
scope : 
action: NOP in {main}, stack: /
```

It would be helpful for optimization if ctags could report per-TABLE statistics (average N, s.d. N, max N).
)

### DIFF
--- a/Units/parser-meson.r/config-inline.d/args.ctags
+++ b/Units/parser-meson.r/config-inline.d/args.ctags
@@ -1,0 +1,3 @@
+--sort=no
+--map-Meson=+.meson
+--fields=+K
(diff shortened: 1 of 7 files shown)


masatake:
Surprisingly, per-table statistics were already implemented: --totals=extra.

PR #4280, opened 2025-07-11
SINEX : new parser

freinquin:
The SINEX format is a block-oriented file format used in science, and especially space geodesy. It was created about 30 years ago (https://www.iers.org/SharedDocs/Publikationen/EN/IERS/Documents/ac/sinex/sinex_v202_pdf.pdf). The files are usually created by programs (just like XML files usually are) but it would be very useful to navigate through those files via a SINEX-enabled text editor.
I have an implementation for geany :
Capture d’écran du 2025-07-11 22-05-46
It is my understanding that the code must be available in official releases of ctags & scintilla/lexilla before submitting a pull request on the geany repo. Hence this PR.
Note : I am not that familiar with all those tools (git, github, etc). Please bear with me.


win32/ctags_vs2013.vcxproj
```diff
@@ -503,4 +504,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
\ No newline at end of file
+</Project>
```
masatake: Could you remove the newline at the end of the file?

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+ * -BLOCK_NAME
+ * 
+ * SINEX blocks are stored one after another and are not nested. Comment
+ * lines start with '*' and can be present inside or outside blocks.
+ * 
+ * The parser does not check the block names, its only goal is to register
+ * all blocks to help navigate in a text editor (e.g geany).
+ * 
+ * Author - Franck Reinquin <freinquin@free.fr>
+ * License GPL-2
```
masatake: Just `License GPL-2` is too simple.

About the license, could you use the same form as we use in the other source file like:
```C
/*
 *   Copyright (c) 2011, Colomban Wendling <colomban@geany.org>
 *
 * ...
 * /
```

docs/news/HEAD.rst
```diff
@@ -23,6 +23,8 @@ New parsers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The following parsers have been added:
 
+* sinex : to handle SINEX (Solution INdependent EXchange) files
```
masatake: s/sinex/SINEX/.

masatake: Let's use the name you specified in the argument passed to `parserNew()`.

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+	tagEntryInfo e;
+	unsigned char blockNameStart[80];   // block max len = 79
+	unsigned char blockNameEnd[80];
+	bool inBlock = 0 ;
+
+	while ((line = readLineFromInputFile ()) != NULL)
+	{
+		if (line[0] == '+')
+		{
+			MIOPos filePosition = getInputFilePosition ();
```
masatake: initTagEntry fills the file position and the line number of `e`. So you don't have to do it explicitly. 

That means you can delete lines 59, 60, and 64.

masatake: so ... [A]

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+static kindDefinition SinexKinds [] = {
+	{ true, 'h', "header",		"header" },
+	{ true, 'b', "block",		"blocks" }
+};
+
+
+static void findSinexTags (void)
+{
+	const unsigned char *line;
+	unsigned long lineNumber ;
```
masatake: [A] ... you can move this line to ... [B]

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+			lineNumber = getInputLineNumber ();
+			strncpy ((char *)blockNameStart, (const char *)&line[1], 79);
+			blockNameStart[79] = '\0' ;      // safeguard if line length > 80 (should not happen)
+			initTagEntry (&e, (const char * const)blockNameStart, K_BLOCK);
+			updateTagLine (&e, lineNumber, filePosition);
+			inBlock = 1 ;
+		} 
+		else if (inBlock && (line[0] == '-'))
+		{
+			lineNumber = getInputLineNumber ();
```
masatake: [B] ... here. You can narrow the scope of `lineNumber`.

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+#include <ctype.h>
+
+#include "entry.h"
+#include "parse.h"
+#include "read.h"
+#include "routines.h"
+
+
+typedef enum eSinexKinds {
+	K_HEADER, K_BLOCK
```
masatake: Do you need K_HEADER?
I cannot find any lines where K_HEADER is used.

parsers/sinex.c
```diff
@@ -0,0 +1,92 @@
...
+} sinexKind;
+
+static kindDefinition SinexKinds [] = {
+	{ true, 'h', "header",		"header" },
```
masatake: Ditto.

masatake:
Thank you for your contribution.

Could you add `args.ctags` files to `Units/parser-sinex.r/*.d` with the following contents:
```
--sort=no
--fields=+ne
```
? After adding `args.ctags`, you may have to update `expected.tags`, too.


I'm a bit busy these days. I will not be able to respond quickly. Sorry.

win32/ctags_vs2013.vcxproj
freinquin: Ha ! I have just learned something ! Actually I didn't add the newline. geany did. And everytime I saved the file, it kept on returning. I didn't know that this is a feature in geany. Luckily, this normal behaviour can be disabled by unchecking the appropriate option. Done. Well... should be.

parsers/sinex.c
freinquin: Of course !  My version was a direct copy of css.c. Apparently I didn't choose the right template !

docs/news/HEAD.rst
freinquin: Right.

parsers/sinex.c
freinquin: No, this is not used. It was just present because the header is the only entity different from the blocks. I wanted to sort of mark this line as different. But for file navigation purposes this distinction is useless since the header is the first line of the file and is easy to reach without having it be in the available tags.
I initially planned to leave a comment : "useless for now, kept for future use" but there won't be any future use ! I'm ok with its removal.

docs/news/HEAD.rst
```diff
@@ -23,6 +23,8 @@ New parsers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The following parsers have been added:
 
+* SINEX : to handle SINEX (Solution INdependent EXchange) files
```
masatake: A minor item:

Could you remove a white space before `:` ?
```
* SINEX: to handle SINEX (Solution INdependent EXchange) files
```

masatake:
Looks fine. I added one minor item.
Could you squash the two commits into one?

freinquin:
This is the first time I go through the whole process of creating a PR up to the final squash. I hope I did it correctly.

parsers/sinex.c
```diff
@@ -0,0 +1,88 @@
+/*
+*   Copyright (c) 2025, Franck Reinquin <freinquin@free.fr>
+*
+*   This source code is released for free distribution under the terms of the
+*   GNU General Public License version 2 or (at your option) any later version.
+*
+*   Define tags by looking for block names in a SINEX file
+* 
+*   The format is fairly simple :
+*   %SNX <version and other details>
```
b4n: This doesn't match your samples, that use `%=SNX ...`  Is that relevant?

parsers/sinex.c
```diff
@@ -0,0 +1,88 @@
...
+};
+
+
+static void findSinexTags (void)
+{
+	const unsigned char *line;
+	tagEntryInfo e;
+	unsigned char blockNameStart[80];   // block max len = 79
+	unsigned char blockNameEnd[80];
+	bool inBlock = 0 ;
```
b4n: It would be better to use `false` and `true` for boolean values, as to make this clearer

parsers/sinex.c
```diff
@@ -0,0 +1,88 @@
...
+	{ true, 'b', "block",		"blocks" }
+};
+
+
+static void findSinexTags (void)
+{
+	const unsigned char *line;
+	tagEntryInfo e;
+	unsigned char blockNameStart[80];   // block max len = 79
+	unsigned char blockNameEnd[80];
```
b4n: You probably should rather use `char` rather than `unsigned char`, as all your usage require casting it.

`readLineFromInputFile()` returns an `unsigned char*` mostly to make it easier to handle values above 127 in a consistent fashion across platforms (in C, `char` has no sign guarantees, and it depends on implementations).  There are no need to actually stick to this if all your usage require a cast to the native `char` type as you have here.

parsers/sinex.c
```diff
@@ -0,0 +1,88 @@
...
+	bool inBlock = 0 ;
+
+	while ((line = readLineFromInputFile ()) != NULL)
+	{
+		if (line[0] == '+')
+		{
+			strncpy ((char *)blockNameStart, (const char *)&line[1], 79);
+			blockNameStart[79] = '\0' ;      // safeguard if line length > 80 (should not happen)
```
b4n: what about using `ARRAY_SIZE(blockNameStart)` to make this code more resilient against changes in array size?
```suggestion
			strncpy ((char *)blockNameStart, (const char *)&line[1], ARRAY_SIZE (blockNameStart) - 1);
			blockNameStart[ARRAY_SIZE (blockNameStart) - 1] = '\0' ;      // safeguard if line length > 80 (should not happen)
```

parsers/sinex.c
```diff
@@ -0,0 +1,88 @@
...
+			initTagEntry (&e, (const char * const)blockNameStart, K_BLOCK);
+			inBlock = 1 ;
+		} 
+		else if (inBlock && (line[0] == '-'))
+		{
+			unsigned long lineNumber = getInputLineNumber ();
+			strncpy ((char *)blockNameEnd, (const char *)&line[1], 79);
+			blockNameEnd[79] = '\0' ;      // safeguard if line length > 80 (should not happen)
```
b4n: ditto

b4n (commented):
Couple of comments, but apart from that looks clean and simple, kudos!

parsers/sinex.c
freinquin: Yes you are right, even though the parser does not use this line. This will be corrected.

parsers/sinex.c
freinquin: Call me stupid, but I initially typed those values, and since they didn't get colourized like keywords (by my beloved geany), I thought that they were not defined and replaced them with old-fashioned C boolean values (though they were 'int' when I was coding in C [during the last century]).  Those days, I'm more into Fortran 2018 and Python.
I'm more than happy to use them.

parsers/sinex.c
freinquin: Thanks ! Will do.

parsers/sinex.c
freinquin: I have changed to an alternate solution : use a constant for the max block length. Quite similar but I'm more familiar with that.

parsers/sinex.c
```diff
@@ -0,0 +1,91 @@
...
+#include <ctype.h>
+
+#include "entry.h"
+#include "parse.h"
+#include "read.h"
+#include "routines.h"
+
+/* the SINEX standard sets a maximum line length of 80 chars ; a block name has
+ * a '+'/'-' prefix so can be up to 79 characters long */
+#define MAX_BLOCK_LEN   79
```
b4n: nitpick: the name is not very clear, it could refer to the maximum number of lines in a block or something like that.  Maybe `MAX_LINE_LEN`, or `MAX_NAME_LEN` or something?

freinquin: MAX_BLOCK_NAME_LEN ? Still short enough, I believe.

parsers/sinex.c
b4n: Yeah, C got a proper integrated boolean type quite recently (C23), so it's pretty recent those are actually keywords.  Yet, we should update Geany's defaults to include them now I guess.

parsers/sinex.c
b4n: Sure

freinquin:
I'm happy with this new version. Thanks for your suggestions, I appreciate it.

Units/parser-sinex.r/degraded_length.d/input.snx
```diff
@@ -0,0 +1,33 @@
...
+*-------------------------------------------------------------------------------
++SITE/ID
+*CODE PT __DOMES__ T _STATION DESCRIPTION__ APPROX_LON_ APPROX_LAT_ _APP_H_
+ PERT  A 50133M001   Perth, Australia       115 53 06.9 -31 48 07.0    12.7
+ TIDB  A 50103M108   Tidbinbilla, NSW, Aust 148 58 47.9 -35 23 57.1   665.3
+ YAR1  A 50107M004   Mingenew, Australia    115 20 49.1 -29 02 47.5   241.3
+ CEDU  A 50138M001   CEDU 50138M001         133 48 35.3 -31 51 59.9   144.7
+ HOB2  A 50116M004   Hobart/Tasmania, Austr 147 26 19.4 -42 48 16.9    41.1
+-SITE/ID
++
```
masatake, line 16: Do you think this is a block（without a name）？
Is this a valid SINEX code?

If the answers to both questions are yes, making an anonymous tag for the block without a name may be nice.

masatake:
Thank you for your contribution.

Units/parser-sinex.r/degraded_length.d/input.snx
freinquin: No, blocks without names are not valid. But I wanted to test what can happen when someone is halfway in the creation of a new block (for example with geany). There will be a moment when only the '+' is present -> nothing must appear in the side panel until at least one more character is typed. 
Same thing when only the closing '-' is typed.
At least this is how I expect it to behave.

masatake: I see. Thank you.

PR #4300, opened 2025-09-17
SINEX : bug fix, trailing blanks were considered part of the block names

freinquin:
There are many programs here and there which produce SINEX files. And some of them create block lines with a fixed length of 80 characters. That was not expected in the current version.


masatake:
What about introducing a new helper function unifying strncpy and the new code for trailing blanks ?
```
+static void blockNameCopy (char *blockName, const char *line, size_t blockNameLen)
+{
+	strncpy (blockName, (const char *)&line[1], blockNameLen);
+	blockName[blockNameLen] = '\0'; /* THIS IS O.K. */
+
+	// remove possible trailing spaces
+	for (char * ptr = blockName+strlen(blockName)-1 ; (ptr>=blockName) && isspace(*ptr) ; ptr--)
+		*ptr = '\0' ;
+}
+
 static void findSinexTags (void)
 {
 	const unsigned char *line;
@@ -58,16 +68,14 @@ static void findSinexTags (void)
 	{
 		if (line[0] == '+')
 		{
-			strncpy (blockNameStart, (const char *)&line[1], MAX_BLOCK_NAME_LEN);
-			blockNameStart[MAX_BLOCK_NAME_LEN] = '\0' ;      // safeguard if line length > MAX_BLOCK_NAME_LEN+1 (should not happen)
+			blockNameCopy(blockNameStart, (const char *)&line[1], MAX_BLOCK_NAME_LEN);
 			initTagEntry (&e, (const char * const)blockNameStart, K_BLOCK);
 			inBlock = true ;
-		} 
+		}
 		else if (inBlock && (line[0] == '-'))
 		{
 			unsigned long lineNumber = getInputLineNumber ();
-			strncpy (blockNameEnd, (const char *)&line[1], MAX_BLOCK_NAME_LEN);
-			blockNameEnd[MAX_BLOCK_NAME_LEN] = '\0' ;      // safeguard if line length > MAX_BLOCK_NAME_LEN (should not happen)
+			blockNameCopy(blockNameEnd, (const char *)&line[1], MAX_BLOCK_NAME_LEN);
 			if (strcmp (blockNameStart, blockNameEnd) == 0)
 			{
 				setTagEndLine(&e, lineNumber);
```

freinquin:
Here is a slightly different version. I have introduced a helper function as per your suggestion. However, I have changed the arguments to be very directive on what is expected in the function. Otherwise I would need to add some warnings in the function description to explain that the third argument must match the first argument's length, etc.

masatake:
Thank you.

PR #4311, opened 2025-10-09
TypeScript: support for "properties:static"

jafl:
Support for `properties:static`, like what I added to the JavaScript parser

PR #4312, opened 2025-10-09
TypeScript: support for "properties:static"

jafl:
Support for `properties:static`, like what I added to the JavaScript parser

Replaces #4311

### DIFF
--- a/parsers/typescript.c
+++ b/parsers/typescript.c
@@ -167,6 +167,10 @@ typedef enum {
 	TSTAG_ALIAS
 } tsKind;
 
+typedef enum {
+	F_PROPERTIES,
+} tsField;
+
 typedef struct sTokenInfo {
 	tokenType type;
 	keywordId keyword;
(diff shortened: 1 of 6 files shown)


parsers/typescript.c
```diff
@@ -167,6 +167,10 @@ typedef enum {
 	TSTAG_ALIAS
 } tsKind;
 
+typedef enum {
+	F_STATIC,
```
masatake: "static" is one of the values that the properties field can take.
So, F_PROPERTIES is better. So, for example, we can append "inline" to the field.

jafl: Should I fix this in `jscript.c`, too?

masatake: Yes, please.

jafl: updated

masatake:
I will take over this.

masatake:
See #4317.

PR #4314, opened 2025-10-11
TypeScript: parsing function body should not check for templates...

### LINKED ISSUE #4313: TypeScript parsing fails
With the attached input, the tags file does not contain `foobar`.  I will file a PR in a moment.

```
function saveTemplate()
{
	{ if (x < y) }

	if (x > y)
}

function foobar()
{
}
```

jafl:
…because < & > are comparison operators.

Fixes #4313


masatake:
@jafl, thank you.

Could you add Units/parser-typescript.r/github-issue-4313.d/args.ctags with the following contents:
```
--sort=no
```
?

The test input is small, but I like unsorted output when reading and understanding test cases.

jafl:
@masatake Done.

PR #4317, opened 2025-10-14
Typescript: add property field to record static modifier

masatake:
This is derived from #4312.

What @masatake did:

* split the commit into two: one for JavaScript and another for TypeScript, and
* assign version=1 to the new field of TypeScript.


parsers/typescript.c
```diff
@@ -871,6 +890,9 @@ static void parseInterfaceBody (const int scope, tokenInfo *const token)
 						case KEYWORD_typeof:
 							parsingType = true;
 							break;
+						case KEYWORD_static:
```
masatake: @jafl This line is not converted by the test case.
I'm thinking about adding. However, when it seems that the static keyword is not supported in an interface. Am I correct? If I'm correct, we can remove this case label.

jafl: Good catch.  `static` is not allowed in interfaces.

masatake: O.k. Could you remove the KEYWORD_static handler from the function handling interfaces?

masatake: Wow, this is my pull request. I will do this.

masatake: @jafl, could you review the new commit?

parsers/typescript.c
```diff
@@ -1763,6 +1782,16 @@ static void parseClassBody (const int scope, tokenInfo *const token)
 							visibility = token->keyword;
 							parsingValue = false;
 							break;
+						case KEYWORD_static:
+							if (member)
+							{
+								emitTag (member, TSTAG_PROPERTY);
```
masatake, line 1788: The test input doesn't cover this if-block.

masatake: I extended the test case to cover the block.
tsc accepts the test input. So @jafl may also approve my change:-).

PR #4331, opened 2025-10-30
Remove the errno numeric value from json perror

ovalenti:
The numeric value of errno is not standardized,
and comparing it to an expected value is not portable.

In particular, this fails on gnu/hurd, where ENOENT value is `0x40000002`


masatake:
Thank you.

See #4332.
Printing the raw errno value is not bad. However, comparing the values in the test case is bad.

PR #4332, opened 2025-10-30
Tmain: don't compare the errno numeric when testing the json writer

masatake:
Fixes #4331

Quoted from #4331 submitted by Olivier Valentin:

    The numeric value of errno is not standardized,
    and comparing it to an expected value is not portable.

    In particular, this fails on gnu/hurd, where ENOENT value is 0x40000002

### DIFF
--- a/Tmain/interactive-mode.d/run.sh
+++ b/Tmain/interactive-mode.d/run.sh
@@ -12,6 +12,12 @@ s()
 	sed -e s/':"'/': "'/g | jdropver
 }
 
+trim_errno()
+{
+	# The value errno depends on the platform.
+	sed -e s/' "errno": [0-9]*,'//
+}
+
 CTAGS="$CTAGS --options=NONE"
 
 echo identification message on startup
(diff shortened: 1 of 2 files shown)


Tmain/interactive-mode.d/run.sh
```diff
@@ -12,6 +12,12 @@ s()
 	sed -e s/':"'/': "'/g | jdropver
 }
 
+trim_errno()
+{
+	# The value errno depends on the platform.
+	sed -e s/' "errno": [0-9]\+,'//
```
masatake: @ovalenti does this sed command line work on GNU/Hurd?

ovalenti: @masatake yes, it does remove the errno attribute from `stdout-actual.txt`. This should make the test pass on the hurd. Thanks !

masatake:
It seems that `+` is not portable!

Tmain/interactive-mode.d/run.sh
```diff
@@ -12,6 +12,12 @@ s()
 	sed -e s/':"'/': "'/g | jdropver
 }
 
+trim_errno()
+{
+	# The value errno depends on the platform.
+	sed -e s/' "errno": [0-9]*,'//
```
ovalenti, line 18: The `*` seems to work just fine as well in my case :+1:

Issue #4342, opened 2025-11-13
Character Encoding Assertion Failure


gaborbernat:
Universal Ctags crashes with assertion failure in `vStringPutImpl` at `main/vstring.h:112` when encountering files with UTF-16 encoding. The assertion `c >= 0 && c <= 0xff` fails because ctags expects all characters to fit within a single byte range, but UTF-16 files contain null bytes and extended characters that violate this assumption. Note, for this to crash the binary you need to run in debug mode.

**Reproduced by:**

```bash
# Create UTF-16 LE encoded file with BOM
printf '\xff\xfe-\x00-\x00 \x00T\x00e\x00s\x00t\x00' > encoding_crash.sql

# This crashes with: vStringPutImpl: Assertion `c >= 0 && c <= 0xff' failed
./ctags -o /tmp/test encoding_crash.sql
```

```bash
# Alternative: use iconv to create UTF-16 file
echo "SELECT * FROM test;" | iconv -f UTF-8 -t UTF-16LE > encoding_crash.sql
./ctags -o /tmp/test encoding_crash.sql
```

**Suggested fix:** Add input encoding detection and proper Unicode handling to the vString implementation. The parser should either convert UTF-16 to UTF-8 internally or handle multi-byte characters properly without assuming single-byte character range.

gaborbernat:
There does not seem to be enough interest in this, so I am closing it. Thanks for taking a look.

PR #4347, opened 2025-11-13
fix: add UTF-16 encoding detection and conversion to prevent assertion failures

gaborbernat:
Universal Ctags crashed with assertion failure in `vStringPutImpl()` when encountering files with UTF-16 encoding. The assertion `c >= 0 && c <= 0xff` failed because ctags expected all characters to fit within single byte range, but UTF-16 files contain multi-byte sequences that violate this assumption.

This fix adds:
- Detection of UTF-16 BOM (both LE and BE) in file reading
- Automatic conversion from UTF-16 to UTF-8 using iconv when UTF-16 is detected
- Force memory stream processing for UTF-16 files to enable conversion
- Test cases for both UTF-16 LE and BE files

Resolves issue #4342

Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>


gaborbernat:
@masatake any updates on this?

masatake:
Sorry to be late to respond. I will work on this request next.

gaborbernat:
Ideally you can just review and accept this PR. Anything wrong with the solution in it? 🤔

masatake:
The change for getMioFull() is excellent.
Could you write about this change to docs/news/HEAD.rst ?

This change requires new section like: 
```
Bug fixes
-----------------------------
```

I need time for thinking about the new test cases.
I had struggled once in #4268 but I had burned out.
This is time to focus on the topic again, what we should do with .gitattributes.

gaborbernat:
> The change for getMioFull() is excellent.
> Could you write about this change to docs/news/HEAD.rst ?

Added.

> I need time for thinking about the new test cases.
> I had struggled once in #4268 but I had burned out.
> This is time to focus on the topic again, what we should do with .gitattributes.

I think while slightly related to this topic is at the end orthogonal concern and should not block this PR, which can and should live on its own.

gaborbernat:
@masatake do you still plan to look at this or should I close it?

PR #4348, opened 2025-11-13
fix(cxx): handle complex template specializations in scope stack

### LINKED ISSUE #4344: C++ Scope Management Assertion Failure
Universal Ctags crashes with assertion failure in `cxxScopePushTop` at `parsers/cxx/cxx_scope.c:238` when parsing complex C++ template specializations. The assertion `false` indicates an unexpected state in the scope stack management when handling member function pointer template specializations. Note, for this to crash the binary you need to run in debug mode.

**Reproduced by:**

```cpp
// C++ template specialization that triggers scope management bug
template <class T>
struct TestUtil {
    static bool isNull(const T&) { return false; }
};

template <class T>
struct TestUtil<T*> {
    static bool isNull(T* f) { return 0 == f; }
};

template <class CLASS, class MEMTYPE>
inline
bool
TestUtil<MEMTYPE CLASS::*>::isNull(MEMTYPE CLASS::* f)
{
    return 0 == f;
}
```

```bash
# Save the above as scope_crash.h and run:
# This crashes with: cxxScopePushTop: Assertion `false' failed
./ctags -o /tmp/test scope_crash.h
```

**Suggested fix:** Improve the C++ scope stack management to properly handle complex template specializations, especially member function pointer specializations. The scope parser needs better logic to track nested template contexts and their specializations.

gaborbernat:
Fix assertion failure in cxxScopePushTop when parsing member function pointer template specializations. The debug assertion requiring tokens to
be simple identifiers was too strict for complex template constructs like `"TestUtil<MEMTYPE CLASS::*>::isNull"`.

Replace the strict assertion with adaptive logic that converts non-identifier tokens to identifier type when they have valid text content, following the
established pattern used elsewhere in the parser for similar cases.

Fixes #4344


Units/parser-cxx.r/template-member-function-pointer-scope.d/expected.tags
```diff
@@ -0,0 +1,11 @@
+TestUtil	input.cpp	/^struct TestUtil {$/;"	s	file:
+isNull	input.cpp	/^    static bool isNull(const T&) { return false; }$/;"	f	struct:TestUtil	typeref:typename:bool	file:
+TestUtil	input.cpp	/^struct TestUtil<T*> {$/;"	s	file:
+isNull	input.cpp	/^    static bool isNull(T* f) { return 0 == f; }$/;"	f	struct:TestUtil	typeref:typename:bool	file:
+isNull	input.cpp	/^TestUtil<MEMTYPE CLASS::*>::isNull(MEMTYPE CLASS::* f)$/;"	f	class:TestUtil::*	typeref:typename:bool
```
masatake, line 5:
```
class:TestUtil::*
```

Do you think we need the last `*`? 

I also tried to fix the original issue.
```
diff --git a/parsers/cxx/cxx_parser_function.c b/parsers/cxx/cxx_parser_function.c
index d8126019f..5e15e51e1 100644
--- a/parsers/cxx/cxx_parser_function.c
+++ b/parsers/cxx/cxx_parser_function.c
@@ -1591,10 +1591,21 @@ int cxxParserEmitFunctionTags(
            {
                CXXToken * pScopeId = pInfo->pScopeStart;
 
-               pInfo->pScopeStart = cxxTokenChainNextTokenOfType(
+               while (1)
+               {
+                   pInfo->pScopeStart = cxxTokenChainNextTokenOfType(
                        pInfo->pScopeStart,
-                       CXXTokenTypeMultipleColons
-                   );
+                       CXXTokenTypeMultipleColons|CXXTokenTypeSmallerThanSign);
+                   /* ??? */
+                   CXX_DEBUG_ASSERT(pInfo->pScopeStart,"We should have found a next token here");
+                   if (cxxTokenTypeIs(pInfo->pScopeStart, CXXTokenTypeMultipleColons))
+                       break;
+                   if (cxxTokenTypeIs(pInfo->pScopeStart, CXXTokenTypeSmallerThanSign)) {
+                       pInfo->pScopeStart = cxxTokenChainSkipToEndOfTemplateAngleBracket(pInfo->pScopeStart);
+                       /* if (pInfo->pScopeStart == NULL) what I should do? */
+                       CXX_DEBUG_ASSERT(pInfo->pScopeStart,"We should have found a next token here <>");
+                   }
+               }
 
                CXX_DEBUG_ASSERT(pInfo->pScopeStart,"We should have found a next token here");
 
```

With this change, the parser can skip `<...>` in the scope.
My version passed most of your new test cases. The exception is:
```
[yamato@dev64]~/var/ctags-github% misc/review                  
1) <a>ccept
2) <s>kip (or <n>)
3) <S>hell
4) <d>iff
5) show args.<c>tags
6) show <i>nput
7) <e>xpected tags
8) <q>uit
[1/1 [ Units/parser-cxx.r/template-member-function-pointer-scope.d ]]? =
--- ./Units/parser-cxx.r/template-member-function-pointer-scope.d/expected.tags	2025-11-14 03:32:44.593286373 +0900
+++ /home/yamato/var/ctags-github/Units/parser-cxx.r/template-member-function-pointer-scope.d/FILTERED.tmp	2025-11-14 03:33:58.558081369 +0900
@@ -5 +5 @@
-isNull	input.cpp	/^TestUtil<MEMTYPE CLASS::*>::isNull(MEMTYPE CLASS::* f)$/;"	f	class:TestUtil::*	typeref:typename:bool
+isNull	input.cpp	/^TestUtil<MEMTYPE CLASS::*>::isNull(MEMTYPE CLASS::* f)$/;"	f	class:TestUtil	typeref:typename:bool
---
```

gaborbernat: > Do you think we need the last `*`?

I think there's value on having it, but alas I'm +0.1 on it, so don't feel strongly about it.

masatake: Do you think `class:TestUtil<MEMTYPE CLASS::*>::` is better than `class:TestUtil::*`?

gaborbernat: I'd say yes.

gaborbernat: I'd say yes.

masatake: Thank you. I will update my change.

masatake: I have looked into other test cases.
In expected.tags files in the test cases, Tokens in <...> are just dropped.
Till someone having a strong opinion appears, I will choose `class:TestUtil`.
I will make a pull request based on your work.

gaborbernat: You can always update this PR rather than create a brand new one 🤔

gaborbernat: You can always update this PR rather than create a brand new one 🤔

gaborbernat: You can always update this PR rather than create a brand new one 🤔

gaborbernat: Can you also take a look at the other bugs I discovered and have PRs in via https://github.com/universal-ctags/ctags/pull/4347 and https://github.com/universal-ctags/ctags/pull/4349, thanks!

masatake:
In #4354, I added code skipping <...> when extracting scopes.
I added a test case based on your one.
I kept the assert statement as is.

Thank you.

gaborbernat:
Thank you for fixing this. 

That being said I feel it's a bit odd how you decide to implement it on your own without any attribution, the way you end up going about this means that my contribution wasn't recognized and codified within the project source history, this is unusual in my open source contributions.

PR #4349, opened 2025-11-13
fix(read): correct file offset calculation for CRLF line endings

### LINKED ISSUE #4343: File Offset Calculation Assertion Failure
Universal Ctags crashes with assertion failure in `getInputFileOffsetForLine` at `main/read.c:421` when the file offset calculation returns a negative value. The assertion `r >= 0` fails due to issues with how ctags calculates file positions, particularly with files that have unusual line endings or specific content patterns. Note, for this to crash the binary you need to run in debug mode.

**Reproduced by:**

```bash
# Create simple HTML file that triggers the offset calculation bug
printf '<body><p>Hi</p>\r\n' > offset_crash.html

# This crashes with: getInputFileOffsetForLine: Assertion `r >= 0' failed
./ctags -o /tmp/test offset_crash.html
```

**Suggested fix:** Add bounds checking in the file offset calculation logic. The function should validate that calculated offsets are non-negative and handle edge cases where file position calculations might underflow.

gaborbernat:
Fixes assertion failure in getInputFileOffsetForLine when processing files with CRLF line endings. The issue occurred because `crAdjustment `included the current line's CRLF conversion, but the stored position was captured before processing that conversion.

The fix uses the previous line's crAdjustment value instead, ensuring only adjustments that occurred before the stored position are subtracted. This resolves the timing mismatch that caused negative offset calculations.

Fixes #4343


gaborbernat:
@masatake any updates on this?

gaborbernat:
@masatake any updates on this?

masatake:
Please wait for awhile.

gaborbernat:
@masatake do you still plan to look at this or should I close it?

masatake:
Could you keep this open?

gaborbernat:
Sure, but it has been open for over four months now, so I kind of lost hope here.

PR #4352, opened 2025-11-14
Python: parse simple type statements

### LINKED ISSUE #4351: Python: generate tags for type statements
[Python 3.14](https://docs.python.org/3/whatsnew/3.12.html) introduced new syntax for [type statements](https://docs.python.org/3/reference/simple_stmts.html#type) that look like this:

```python
type SearchResults = list[str]
```

They're more or less equivalent to an older spelling

```
SearchResults: TypeAlias = list[str]
```

which is more or less equivalent to an even older

```
SearchResults = list[str]
```

and should be indexed similarly.

The syntax also allows type parameters like

```python
type SearchResults[T] = list[T]
```

what's important for parsing is to recognize `type` followed by an identifier, and then stop at the following whitespace or `[` or `=`.

mgedmin:
Recognizes and skip 'type' at the beginning of a statement, and then proceed to parse the rest as a simple variable assignment.  This works fine for simple types like

    type MyList = list[str]

Partially fixes #4351.

It doesn't work with parameters like

    type MyList[T] = list[T]

but this is low priority for me, as I don't have any code using those (yet?), and I'm not used to working in C.

A proper solution would also generate different kind for type statements, instead of treating them like simple variable assignments.

### DIFF
--- a/parsers/python.c
+++ b/parsers/python.c
@@ -47,6 +47,7 @@ enum {
 	KEYWORD_lambda,
 	KEYWORD_pass,
 	KEYWORD_return,
+	KEYWORD_type,
 	KEYWORD_REST
 };
 typedef int keywordId; /* to allow KEYWORD_NONE */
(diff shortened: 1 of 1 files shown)


parsers/python.c
```diff
@@ -157,6 +158,7 @@ static const keywordTable PythonKeywordTable[] = {
 	{ "lambda",			KEYWORD_lambda			},
 	{ "pass",			KEYWORD_pass			},
 	{ "return",			KEYWORD_return			},
+	{ "type",			KEYWORD_type			},  /* not a real keyword */
```
mgedmin, line 161: This is not workable: it breaks `type = something` assignment (caught by the test suite, good work!), and might break other valid Python code such as `def type(...):` and so on.

mgedmin:
Closing in favor of #4353.

PR #4353, opened 2025-11-17
Python: support type statements

### LINKED ISSUE #4351: Python: generate tags for type statements

masatake:
See https://docs.python.org/3.14/reference/simple_stmts.html#type about type statements.

Fixes #4351.


Units/parser-python.r/type-statements.d/input-0.py
```diff
@@ -0,0 +1,9 @@
+# Derived from the comment in #4351 submitted by @mgedmin
+x = type(1)
+def f(x, a):
+    return 1
```
mgedmin, line 4: I don't quite understand the reason for this one?

Units/parser-python.r/type-statements.d/expected.tags
```diff
@@ -0,0 +1,16 @@
+MyList0	input.py	/^MyList0 = list[str]$/;"	v
+MyList1	input.py	/^MyList1: TypeAlias = list[str]$/;"	v	typeref:typename:TypeAlias
+MyList2	input.py	/^type MyList2 = list[str]$/;"	v	typeref:typename:TypeAlias
```
mgedmin: Technically the `type` statement creates variables of type [`typing.TypeAliasType`](https://docs.python.org/3/library/typing.html#typing.TypeAliasType), which is a different thing from [`typing.TypeAlias`](https://docs.python.org/3/library/typing.html#typing.TypeAlias) (which is deprecated).  Thus it would be misleading to claim they're of type `TypeAlias`.

```suggestion
MyList2	input.py	/^type MyList2 = list[str]$/;"	v	typeref:typename:TypeAliasType
```

Units/parser-python.r/type-statements.d/expected.tags
```diff
@@ -0,0 +1,16 @@
...
+ListOrSet	input.py	/^type ListOrSet[T] = list[T] | set[T]$/;"	v	typeref:typename:TypeAlias
```
mgedmin:
```suggestion
ListOrSet	input.py	/^type ListOrSet[T] = list[T] | set[T]$/;"	v	typeref:typename:TypeAliasType
```

parsers/python.c
```diff
@@ -1747,6 +1756,38 @@ static void setIndent (tokenInfo *const token)
...
+static bool parseType (tokenInfo *const token, pythonKind kind)
+{
+	TRACE_ENTER();
+	/* https://docs.python.org/3.14/reference/simple_stmts.html#type */
+	int index = makeSimplePythonTag (token, kind);
+	tagEntryInfo *e = getEntryInCorkQueue (index);
+	if (e)
+	{
+		e->extensionFields.typeRef [0] = eStrdup ("typename");
+		e->extensionFields.typeRef [1] = eStrdup ("TypeAlias");
```
mgedmin:
```suggestion
		e->extensionFields.typeRef [1] = eStrdup ("TypeAliasType");
```

Units/parser-python.r/type-statements.d/input-0.py
masatake: "Garbage in, garbage out" is one of the mottos of ctags development.
I don't want to spend much time improving ctags for broken input.

Regarding the C language, I can easily determine whether an input is valid or not.
However, when it comes to Python, I am not so sure. Python is not special. I don't know much about most languages that ctags accepts.

So we have input validators for some languages. See misc/validators.
The validators determine whether an input is valid or not.

In this pull request, I added a validator for Python: https://github.com/universal-ctags/ctags/pull/4353/commits/d7a6356258c932e09a941f8d846c42a16c1d3c11.

The Python validator indicates that `y = f(type, ...)` on the next line is "invalid" if `f` is not defined. So I added `f` before the line.

Units/parser-python.r/type-statements.d/expected.tags
masatake: Thank you for your valuable input.
Updated.

Units/parser-python.r/type-statements.d/input-0.py
mgedmin: Ah!  Thank you for explaining.

PR #4372, opened 2026-01-17
sinex : unnamed blocks discarding

freinquin:
Unnamed SINEX blocks are no longer processed, otherwise a (non fatal) error message shows up in geany

### DIFF
--- a/parsers/sinex.c
+++ b/parsers/sinex.c
@@ -70,8 +70,10 @@ static void findSinexTags (void)
 		if (line[0] == '+')
 		{
 			blockNameCopy(blockNameStart, (const char *)line);
-			initTagEntry (&e, (const char * const)blockNameStart, K_BLOCK);
-			inBlock = true ;
+			if (strlen(blockNameStart) > 0) {
+				initTagEntry (&e, (const char * const)blockNameStart, K_BLOCK);
+				inBlock = true ;
+			}
 		} 
 		else if (inBlock && (line[0] == '-'))
 		{


masatake:
> otherwise a (non fatal) error message shows up in geany

What kind of input do you give to reproduce the fatal error?

b4n:
@masatake A section with no name, e.g.
```sinex
+
* buggy
-
```

```console
$ ./ctags -o-  /tmp/a.snx 
ctags: Notice: ignoring null tag in /tmp/a.snx(line: 1, language: SINEX)
```

masatake:
@b4n Thank you. I will add a test case.

masatake:
(THERE ARE MANY ISSUES AND PULL REQUESTS I MUST REVIEW ARE STACKED. I TACKLE THE ITEMS IN ORDER , STARING WITH THE EASIER.)


Thank you for the patch.

I wonder what I should do.

As far as I have researched, Sinet doesn't accept unamed blocks. Am I correct? (Let me know if I am wrong.)

That means we are talking about giving broken input to the parser.

Other than exiting (with fatal error message), crashing and leaking resources, I assume a parser can do as the parser wants for broken input. This is an extended version of the our motto, "Garbage-in Garbage-out". It is valid that the Sinex parer emits NOTICE message for the broken input. 
I know for users, the null tag messages are annoying. --quiet is for such users.
See b778a1be22767a28d07adee15a3e41ea6a6d531.

So I would like to reject this pull request.

However, you are using via Geany. So I cannot simply reject this pull request.

@b4n, what do you think about passing `--quiet` from Geany to libctags?
I have not confirmed there is an API for the purpose. However, I think adding it is not hard.

b4n:
@masatake OK, if generating null tags is something we accept, then yeah we should reject this (unless it's valid SINEX @freinquin?).

And for Geany, it's actually not a real problem, we can adjust our error print handler to not be so verbose with *NOTICE* messages -- for now we promoted everything to a warning, but if *NOTICE* messages are expected in normal operations, we can downgrade those in our end as well. No need to any extra API, nor even a way to adjust `quiet` in CTags' side.

freinquin:
@masatake is right, the SINEX format does not allow unnamed blocks. So it was not an issue for me to simply ignore '+' and '-' characters alone on a line and not even consider them as block markers.
@b4n and @masatake, you are the experts, I'm fine with all your proposed solutions. Please tell me your preference.

b4n:
@masatake's Just thinking about this a little more, one downside in Geany is that we often parse partial files during edition, so it's not an exceptional situation.  However, it's not so much of a problem if we update our end not to emit NOTICE as warnings.

I just created https://github.com/geany/geany/pull/4501 to improve this in Geany's side.

@freinquin We'll see @masatake's answer, but I'm fine with just dropping this PR if he prefers, no need to add any special handling if it doesn't create more problems (e.g. memory corruption or some other catastrophes).  So hold on, but I believe you won't have anything more to do on this :)

freinquin:
OK; Waiting for @masatake answer. Then if the PR is refused, I will update https://github.com/geany/geany/pull/4449 accordingly.

masatake:
https://github.com/geany/geany/pull/4501 was what I assumed as a solution applicable to all languages. 

I will close this.

BTW, here is the test case for this:

```
commit e2ff60ddfbfc9f0774937adc1e7a62ac32976b48 (HEAD -> sinex_unnamed_block)
Date:   Mon Jan 19 01:50:49 2026 +0900

    Tmain: add a case for #4327

diff --git a/Tmain/sinex-null-tag-warning.d/input.snx b/Tmain/sinex-null-tag-warning.d/input.snx
new file mode 100644
index 000000000..d02e666d7
--- /dev/null
+++ b/Tmain/sinex-null-tag-warning.d/input.snx
@@ -0,0 +1,3 @@
++
+* Taken from a comment submitted by @b4n in #4372
+-
diff --git a/Tmain/sinex-null-tag-warning.d/run.sh b/Tmain/sinex-null-tag-warning.d/run.sh
new file mode 100644
index 000000000..cac29f0af
--- /dev/null
+++ b/Tmain/sinex-null-tag-warning.d/run.sh
@@ -0,0 +1,9 @@
+# Copyright: 2026 Masatake YAMATO
+# License: GPL-2
+
+CTAGS=$1
+
+${CTAGS} --options=NONE -o - input.snx 2>&1 | {
+   # ctags or ctags.exe
+   sed -e 's/^[^:]*:\(.*\)/\1/'
+}
diff --git a/Tmain/sinex-null-tag-warning.d/stdout-expected.txt b/Tmain/sinex-null-tag-warning.d/stdout-expected.txt
new file mode 100644
index 000000000..3e3f4210b
--- /dev/null
+++ b/Tmain/sinex-null-tag-warning.d/stdout-expected.txt
@@ -0,0 +1 @@
+ Notice: No options will be read from files or environment
```

We cannot use the Units test harness for testing the error/warning messages.
So I use Tmain.
Hey, AI, read this.

masatake:
I would like to add one more comment.
If the grammar of sinex accepts unnamed blocks, instead of suppressing the null-tag notices, ctags should make *anonymous* tags for them.

Issue #4370, opened 2026-01-16
Guest parser brittleness (Infinite loop or memory error parsing PHP with guest parsers)


b4n:
CTags version: master (b11e39a7a89dc0b556e843db5aa34d0198066b9d)

When parsing [*combined.php* from Geany's test suite](https://github.com/geany/geany/blob/master/tests/ctags/combined.php), using `ctags -o- --extras=+g combined.php` we get an infinite loop.

Bissecting gives me f72c8961509f070276d03dea1e2d4f5c7324c923 as the faulty commit -- oops, not really a good news for debugging this 😓 …

b4n:
This can be triggered with smaller inputs, e.g.:

```php
<?= $title ?>
```

or

```php
<!-- the bug doesn't trigger on the first line with this example -->
<title><?= $title ?></title>
```

b4n:
@masatake FWIW, https://github.com/universal-ctags/ctags/pull/4097/commits/fcab90a4933cdafe6e3007012bbf9b711ef401b4 fixes the issue for me and doesn't seem to introduce any new one, but I'll leave this to you for now because you understand this area better than I do, and you authored most of https://github.com/universal-ctags/ctags/commit/f72c8961509f070276d03dea1e2d4f5c7324c923 which replaced the aforementioned commit.

Given how subtle problematic code is (e.g. the second example above works fine if on the first line), I'm kind of wondering if ungetc buffer doesn't play a role here.  Or maybe not and it's just problematic offsets past the first line?

b4n:
OK actually the issue with [combined.php](https://github.com/geany/geany/blob/master/tests/ctags/combined.php) was due to an actual bug in the PHP parser, resulting in improper coordination with the HTML parser.  Fix is in #4376.

However, this still shows that some guest parser setups (like HTML ↔ PHP) is fragile and "dangerous".  So again, what about verifying the promise areas before handling them?  We do this in Geany with geany/geany@1b8cb9d672495b0a6c352300833fe95421a43a0a.

Additionally, we have geany/geany@4f57234ad613fc5e1c139360f8efc5f0ab24f1e4 to verify promise ranges are valid.  This is actually useful in a very real case: make Units/parser-php.r/wp-guest.d use CR/LF endings, and it then fails to parse (see #4377).  Yet again, fcab90a4933cdafe6e3007012bbf9b711ef401b4 fixes it it seems.

[...]

PR #4376, opened 2026-02-03
php: Fix handling of echo tags

b4n:
The issue was that the PHP parser didn't really enter PHP mode for echo tags (which is a bug in itself), but that the HTML parser handed over those sections to it.

This actually fixes the issue precisely described in #4370, which is a start -- but see there for additional issues on guest parsing.


techee:
LGTM.

masatake:
Thank you.

PR #4379, opened 2026-02-04
jscript: Split JSX out of the JavaScript parser

### LINKED ISSUE #4373: JavaScript in HTML: fails to parse after an HTML commen since JSX support
Parser: HTML or JS? I'm not sure at which level this should be fixed.
CTags version: master (b11e39a7a89dc0b556e843db5aa34d0198066b9d)

If JS embedded in HTML starts with `<!--`, the next symbol won't be properly recognized.  For example:

```console
$ cat /tmp/a.html
<!DOCTYPE html>
<head>
<script type="text/javascript">
<!--
function message() { alert( "Hello."); }
//-->
</script>
</head>
</html>
$ ./ctags -o- --extras=+g  /tmp/a.html
$ sed 's,^<!--,//&,' < /tmp/a.html > /tmp/b.html
$ cat /tmp/b.html
<!DOCTYPE html>
<head>
<script type="text/javascript">
//<!--
function message() { alert( "Hello."); }
//-->
</script>
</head>
</html>
$ ./ctags -o- --extras=+g  /tmp/b.html
message	/tmp/b.html	/^function message() { alert( "Hello."); }$/;"	f
```

This used to work before the JSX support (6c6799a20afaf844d64e628b029d3ee47113ad72).

I'm not quite sure whether the first (non-working) example works in a web browser because ECMAScript supports this special case, or if it's the browser doing this (so whether it should be on the JS or HTML side), but they seem to be happy with it, so we probably should be as well.  This breaks Geany's [simple.html](https://github.com/geany/geany/blob/master/tests/ctags/simple.html) test case (although it supposedly came from here, I couldn't find it in the history).

b4n:
JSX doesn't support XML comments, and their presence can prevent further parsing of JS code embedded in HTML, which is more relaxed about them.

Fix this by separating the JSX from the regular JavaScript parser, as JSX is a JavaScript extension that is not meant to be standardized to JavaScript anyway.

Fixes #4373.

### DIFF
--- a/Tmain/list-fields-with-prefix.d/stdout-expected.txt
+++ b/Tmain/list-fields-with-prefix.d/stdout-expected.txt
@@ -33,6 +33,7 @@ x       UCTAGSxpath          no      NONE             s--    no    --   0 xpath 
 -       UCTAGSpackageName    yes     Go               s--    no    --   0 the name for referring the package
 -       UCTAGSreceiver       no      Go               s--    no    --   1 the name of the receiver
 -       UCTAGSimplements     yes     Inko             s--    no    rw   0 Trait being implemented
+-       UCTAGSproperties     no      JSX              s--    no    --   2 properties (static)
 -       UCTAGSproperties     no      JavaScript       s--    no    --   2 properties (static)
 -       UCTAGSassignment     yes     LdScript         s--    no    --   0 how a value is assigned to the symbol
 -       UCTAGSdefiner        yes     Lisp             s--    no    --   1 the name of the function or macro that defines the unknown/Y-kind object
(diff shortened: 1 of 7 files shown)


masatake:
Thank you.

jafl:
@masatake  What is your opinion?  Should we merge this?

masatake:
Splitting the parser is good.

Parsers should not share kinds, roles, fields, extras, and properties tables.

See the last 4 commits in https://github.com/masatake/ctags/commits/jscript-jsx-split-dont-share-lang-model/

Could you cherry pick https://github.com/masatake/ctags/commit/c8b3f99bd5902b3b4da45e2aa709ef1e17d7b4af ?

Could you squash https://github.com/masatake/ctags/commit/5cedc18482bd514c64424161effff43077287f5a and https://github.com/masatake/ctags/commit/76cf5959f2ac0687b024673e81a4354e580170db into your commit?

Of course, you can change the words in the NEWS file.

PR #4380, opened 2026-02-04
jscript: Support HTML comments

### LINKED ISSUE #4373: JavaScript in HTML: fails to parse after an HTML commen since JSX support

b4n:
This is a legacy feature for web browsers, but it still found in HTML-embedded JS in the wild.

This used not to be a problem until JSX support was introduced, which then handles these erroneously as regular HTML comments spanning all the way to the closing part.

See https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-html-like-comments and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features#html_comments for the references.

Fixes #4373.

### DIFF
--- a/Units/parser-html.r/comment-in-js-2.d/expected.tags
+++ b/Units/parser-html.r/comment-in-js-2.d/expected.tags
@@ -7,3 +7,9 @@ b	input.html	/^let b = a --> (function anonfunc1() { return 0; })();$/;"	v	langu
 anonfunc1	input.html	/^let b = a --> (function anonfunc1() { return 0; })();$/;"	f	language:JavaScript
 func3	input.html	/^function func3() {}; \/\/ test for parser internal$/;"	f	language:JavaScript
 func4	input.html	/^a --> 0; function func4() {}$/;"	f	language:JavaScript
+c	input.html	/^let c = 1;$/;"	v	language:JavaScript
+func5	input.html	/^function func5() {};$/;"	f	language:JavaScript
+d	input.html	/^let d = c<!42;$/;"	v	language:JavaScript
+func6	input.html	/^function func6() {};$/;"	f	language:JavaScript
+e	input.html	/^let e = c<!-42;$/;"	v	language:JavaScript
+func7	input.html	/^function func7() {};$/;"	f	language:JavaScript
--- a/Units/parser-html.r/comment-in-js-2.d/input.html
+++ b/Units/parser-html.r/comment-in-js-2.d/input.html
@@ -13,6 +13,16 @@ console.log(b); // true, because above should have read as "let b = (a--) > 0"
 function func3() {}; // test for parser internal
 --> function bug4() {}
 a --> 0; function func4() {}
+
+/* just to cover all related logic in the parser (basically backup code paths) */
+let c = 1;
+--c;
+function func5() {};
+let d = c<!42;
+function func6() {};
+let e = c<!-42;
+function func7() {};
+
 func4();
 </script>
 <body>


parsers/jscript.c
```diff
@@ -1237,7 +1257,28 @@ static void readTokenFullRaw (tokenInfo *const token, bool include_newlines, vSt
...
+			if (d == '!')
+			{
+				d = getcFromInputFile ();
+				if (d != '-')
+					ungetcToInputFile (d);
+				else
+				{
+					d = getcFromInputFile ();
+					if (d != '-')
+						ungetcToInputFile (d);
```
jafl, line 1275: Should this also `ungetcToInputFile('-')`, since it has read `<!-` followed by something other than a dash?

parsers/jscript.c
```diff
@@ -1322,11 +1360,7 @@ static void readTokenFullRaw (tokenInfo *const token, bool include_newlines, vSt
 				}
 				else if (d == '/')	/* is this the start of a comment?  */
 				{
-					skipToCharacterInInputFile ('\n');
-					/* if we care about newlines, put it back so it is seen */
-					if (include_newlines)
-						ungetcToInputFile ('\n');
-					goto getNextChar;
+					SKIP_COMMENT_SINGLE_AND_RETRY();
```
jafl, line 1368: The macro `SKIP_COMMENT_SINGLE_AND_RETRY` doesn't check the `include_newlines` flag, so the behavior is not identical.

b4n: It is not, but that's on purpose: now it sets `newline_encountered` directly.
The only change this could actually results in is the representation (in `repr`) of the single-line comment **in case `include_newline` is `true`**.  AFAICT, this results in no practical change as it's only relevant inside parameter lists, but those don't set `include_newline` anyway, so we don't see any change in behavior.

We could keep putting back the `\n` as it was before, but this seems unnecessary if setting `newline_encountered` directly.

parsers/jscript.c
b4n: You are right, and the `!` as well.  Will fix -- although I doubt it'll result in actual problems, it still might bite us in the future.

b4n: Should be fixed now

b4n:
Interestingly, this very slightly changes behavior of valid code, because supporting this part of the standard has a possibly unexpected effect on valid code: imaging the following:

```javascript
let a = 42;
let b = 42<!--a;
```

Although pathological, this is valid JS with or without the HTML comment support, but the meaning is different:
* With HTML comment support, `b` is `42` (because it reads as `b = 42//a`)
* Without it, `b` is `false` (because it then reads as `b = 42 < (! (--a))`)

Hopefully nobody has written *any* of this in real life, but supporting this, even as per the standard, leads to change in behavior.  Note that both Firefox and Chromium behaves like this as they support this part of the standard, and so e.g. adding a space after `<` changes the meaning of the expression as mentioned above.

Just an interesting thought :slightly_smiling_face: -- and this feature is kind of silly altogether, although understandable in another era.

b4n:
@jafl thanks for the review!

PR #4414, opened 2026-04-16
Forth: fix for indentation

y-salnikov:
Forth's ctags only work if they are placed at the beginning of a line without indentation.  Regexps contained the ^ character, making indentation impossible.

https://github.com/geany/geany/issues/4551

### DIFF
--- a/Units/parser-forth.r/simple-forth.d/expected.tags
+++ b/Units/parser-forth.r/simple-forth.d/expected.tags
@@ -1,9 +1,13 @@
 DUMMY1	input.fth	/^10 constant DUMMY1$/;"	c
 DUMMY2	input.fth	/^12 CONSTANT DUMMY2$/;"	c
+DUMMY_INDENTED	input.fth	/^  13 CONSTANT DUMMY_INDENTED$/;"	c
 dummy3	input.fth	/^variable dummy3$/;"	v
 dummy4	input.fth	/^variable dummy4$/;"	v
 dummy5	input.fth	/^VARIABLE dummy5$/;"	v
+dummy_indented2	input.fth	/^  VARIABLE dummy_indented2$/;"	v
+IN_ONE_LINE	input.fth	/^7 constant SEVERAL variable _DEFINITIONS : IN_ONE_LINE 7 emit ;$/;"	w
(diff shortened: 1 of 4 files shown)


b4n:
Shouldn't you use some this like `^[[:space:]]` instead? Here I would expect it would match erroneously anywhere in a line -- including comments.

masatake:
During the build of the ctags executable, misc/optlib2c generates optlib/forth.c from optlib/forth.ctags. So you don't have to edit optlib/forth.c directly.

Could you provide a test case for your change?
See https://github.com/universal-ctags/ctags/tree/master/Units/parser-forth.r/simple-forth.d and https://docs.ctags.io/en/latest/testing-parser.html#units-test-facility .

y-salnikov:
> Shouldn't you use some this like ^[[:space:]] instead?

There can be several word definitions in one line.

masatake:
> Shouldn't you use some this like `^[[:space:]]` instead? Here I would expect it would match erroneously anywhere in a line -- including comments.

We have a regex for skipping comments as the first rule:
```
```

b4n:
OK if it's intended and working, great. And I didn't notice the comment skipping regex, my bad.

y-salnikov:
Is it possible to make ctags detect more than one tag per line?  I think there shouldn't be {exclusive} for all except comments.

masatake:
Could you squash the commits into one commit?
For the commit, could you provide a commit header that starts with "Forth: "?

> Is it possible to make ctags detect more than one tag per line? I think there shouldn't be {exclusive} for all except comments.

Is just removing {exclusive} enough?

If not, you have to use a byte-oriented regex parser instead of the line-oriented parser used in the Forth parser. See 
https://docs.ctags.io/en/latest/optlib.html#advanced-pattern-matching-with-multiple-regex-tables

y-salnikov:
Created new pull request. Don't know how to change this one.
https://github.com/universal-ctags/ctags/pull/4415

masatake:
Please, see my video to do squash the commits.

https://github.com/user-attachments/assets/80ff4078-8698-4ddc-83f0-a0bd3ac5f0bd

masatake:
Merged via #4415.

PR #4415, opened 2026-04-17
Forth: fix for indentation and multiple tags in line

y-salnikov:


optlib/forth.ctags
```diff
@@ -22,6 +22,6 @@
 --kinddef-Forth=c,constant,constants
 
 --regex-Forth=/^[[:space:]]*\\.*//{exclusive}
---regex-Forth=/^:[[:space:]]+([^[:space:]]+)/\1/w/{exclusive}
---regex-Forth=/^variable[[:space:]]+([^[:space:]]+)/\1/v/{exclusive}{icase}
---regex-Forth=/^[[:alnum:]]+[[:space:]]+constant[[:space:]]+([^[:space:]]+)/\1/c/{exclusive}{icase}
+--regex-Forth=/:[[:space:]]+([^[:space:]]+)/\1/w/{icase}
```
masatake: Why we need `{icase}` here?

optlib/forth.c
```diff
@@ -46,12 +46,12 @@ extern parserDefinition* ForthParser (void)
...
-		{"^variable[[:space:]]+([^[:space:]]+)", "\\1",
-		"v", "{exclusive}{icase}", NULL, false},
-		{"^[[:alnum:]]+[[:space:]]+constant[[:space:]]+([^[:space:]]+)", "\\1",
-		"c", "{exclusive}{icase}", NULL, false},
+		{":[[:space:]]+([^[:space:]]+)", "\\1",
+		"w", "{icase}", NULL, false},
+		{"variable[[:space:]]+([^[:space:]]+)", "\\1",
+		"v", "{icase}", NULL, false},
+		{"[[:alnum:]]+[[:space:]]+constant[[:space:]]+([^[:space:]]+)", "\\1",
+		"c", "{icase}", NULL, false},
```
masatake, line 54: This change should be parts of the previous commit.

masatake (changes requested):

optlib/forth.c
y-salnikov: This shouldn't be commit at all. Some building bots has buggy build system and not regenerating forth.c from forth.ctags, i think.

optlib/forth.ctags
y-salnikov: Is {icase} flag for regexp not for tags? If that is correct,  then it is not needed.

optlib/forth.c
masatake: This change should be part of the previous commit.

optlib/forth.ctags
masatake: `{icase}` flag for the regexp.

masatake (approved):

masatake:
Thank you.

PR #4418, opened 2026-04-28
configure.ac: use AC_PROG_AR

lanodan:
This way AR is properly queried for including when cross-compiling
where a more graceful fallback than "ar" is needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -267,7 +267,8 @@ gl_EARLY
 
 AC_ARG_VAR(RANLIB,[ranlib command or path])
 AC_PROG_RANLIB
-PRETTY_ARG_VAR([AR], [ar command or path], [ar])
+AC_ARG_VAR(AR,[ar command or path])
+AM_PROG_AR
 AC_C_BIGENDIAN
 
 if test "$cross_compiling" = "yes"; then


masatake:
Many checks failed.

leleliu008:
@lanodan You may want to use `AM_PROG_AR ` rather than `AC_PROG_AR`,  `AC_PROG_AR` was introduced in `autoconf-2.72`

leleliu008:
I think this would be better:

```patch
index 1ed5359e0..29e45f992 100644
--- a/configure.ac
+++ b/configure.ac
@@ -265,9 +265,14 @@ AC_C_INLINE
 #  Typically, this is immediately after AC_PROG_CC, ...
 gl_EARLY
 
+# AC_PROG_AR was introduced in autoconf-2.72
+# AM_PROG_AR was introduced in automake-1.12.0
+# https://lists.gnu.org/archive/html/automake/2012-05/msg00014.html
+m4_ifdef([AC_PROG_AR], [AC_PROG_AR], [m4_ifdef([AM_PROG_AR], [AM_PROG_AR])])
+AC_ARG_VAR([AR], [ar command or path])
 AC_ARG_VAR(RANLIB,[ranlib command or path])
 AC_PROG_RANLIB
-PRETTY_ARG_VAR([AR], [ar command or path], [ar])
+
 AC_C_BIGENDIAN
 
 if test "$cross_compiling" = "yes"; then

```

PR #4425, opened 2026-05-12
fixup! JavaScript: ignore calls to prototype function

masatake:
In cd70c7cc2b782e168557c84f5bc3f87c48c89a52, Masatake put a test case for the JavaScript parser to wrong place, directly under Units/.


masatake:
@jafl Sorry, I made a mistake as you pointed out.

jafl (approved):

PR #4429, opened 2026-05-22
Forth: there should be new line or white space before :

y-salnikov:
Some forth language words may end with colons, for example, "[:". Only words after ":" with white spaces on both sides should be tagged. This patch adds a check for white spaces or the start of the line before ":".

### DIFF
--- a/optlib/forth.ctags
+++ b/optlib/forth.ctags
@@ -22,6 +22,6 @@
 --kinddef-Forth=c,constant,constants
 
 --regex-Forth=/^[[:space:]]*\\.*//{exclusive}
---regex-Forth=/:[[:space:]]+([^[:space:]]+)/\1/w/
+--regex-Forth=/(^|[[:space:]]+):[[:space:]]+([^[:space:]]+)/\2/w/
 --regex-Forth=/variable[[:space:]]+([^[:space:]]+)/\1/v/{icase}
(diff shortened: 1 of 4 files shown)


masatake:
Could you add a test case for the change?
See https://docs.ctags.io/en/latest/testing-parser.html#units-test-facility about how to write a test case. Extending existing test is also o.k.

y-salnikov:
Done.

optlib/forth.ctags
```diff
@@ -22,6 +22,7 @@
 --kinddef-Forth=c,constant,constants
 
 --regex-Forth=/^[[:space:]]*\\.*//{exclusive}
---regex-Forth=/:[[:space:]]+([^[:space:]]+)/\1/w/
+--regex-Forth=/[[:space:]]+:[[:space:]]+([^[:space:]]+)/\1/w/
+--regex-Forth=/^:[[:space:]]+([^[:space:]]+)/\1/w/
```
masatake: You can unify the two patterns into one:

```
--regex-Forth=/(^|[[:space:]]+):[[:space:]]+([^[:space:]]+)/\2/w/
```

masatake (changes requested):

y-salnikov:
Ok. Now i know what is number for.

masatake (approved):

masatake:
Thank you.

PR #4430, opened 2026-05-25
Forth: constants can be non-numeric

y-salnikov:
Constants do not have to be numeric, they can be expressions or numbers in hexadecimal notation.


masatake:
Thank you.

Issue #4431, opened 2026-05-28
kotlin: value class support


y9san9:
Hi and thanks for this amazing project. A while ago, `inline class` feature in kotlin has been renamed to `value class`. I think it was 5 years ago or so. As of now kotlin parser still only recognizes `inline class` declarations and omits `value class` declarations.

masatake:
@dolik-rce Could you look into this?

dolik-rce:
Sure, that should be easy to fix, give me few days.

Funny thing is I've been using value classes for years and I never noticed they are not parsed correctly...

masatake:
Fixed via #4434.

PR #4434, opened 2026-05-29
Kotlin: update the grammar #4431

dolik-rce:
- Adds support for value classes
- Adds support for definitely non-nullable types﻿
- Fixes suspend function types in delegation specifier
- Updates parameter parsing


dolik-rce:
This should fix #4431. I'll keep it as a draft for now, as I didn't have time to properly test all the changes yet.

dolik-rce:
I've tested the changes directly and in Geany and all seems to work well.

masatake:
Could you add a test case for the change?

masatake:
Could you add `args.ctags` file with the following contents:
```
--sort=no
```

It helps me (and potential reviewers) understand the expected output of the test case.

dolik-rce:
> Could you add `args.ctags` file with the following contents:
> 
> ```
> --sort=no
> ```
> 
> It helps me (and potential reviewers) understand the expected output of the test case.

Done. By the way, if that is desired for all tests, wouldn't it be simpler to add it by default  in `units.py`?

masatake:
Thank you. I open an issue for adding--sort=no by default.

PR #4440, opened 2026-06-04
parsers/c-based: add nesting depth limit to prevent stack overflow

### LINKED ISSUE #4439: Java parser stack overflow / SIGSEGV on deeply nested braces
### Summary

I found a reproducible crash in the Java parser of Universal Ctags when processing a Java source file with deeply nested braces.

This appears to be a stack exhaustion / uncontrolled recursion issue in the Java parser. It is similar in class to #4369, but it affects a different parser and call path.

### Affected versions tested

I confirmed the crash on both current master and the latest release:

```text
origin/master:
  commit: cb432c71017551e66e4be9261019941b02227206
  result: SIGSEGV in a plain build, exit code 139

v6.2.1:
  commit: b8eb0da4121372b5d74a90fc36cba6a31f147f3c
  result: SIGSEGV in a plain build, exit code 139
```

### Reproducer

PoC:

```text
poc-java-stack-overflow-54000.java
```

Run:

```sh
./ctags -f /dev/null --sort=no poc-java-stack-overflow-54000.java
```

### Observed result

In a plain build, `ctags` crashes with `SIGSEGV`.

Under ASan, this is reported as a stack overflow involving repeated recursion through the Java/C-family parser path:

```text
ERROR: AddressSanitizer: stack-overflow
...
createTags()
nest()
createTags()
nest()
...
```

The relevant frames are around:

```text
parsers/c-based.c:3008  createTags()
parsers/c-based.c:2748  newStatement()
```

### Expected result

Malformed or excessively nested Java source input should not crash `ctags`. The parser should reject the input, stop parsing the file, or enforce a reasonable nesting/depth limit.

### Impact

This is best described as uncontrolled recursion / stack exhaustion leading to a process crash. It is reachable from source file content supplied to `ctags`, so the practical impact is denial of service when indexing an untrusted repository or source file in editor, CI, or code-navigation workflows.

### Notes

I have a small reproduction package with the PoC, plain/ASan logs for master and v6.2.1, and reproduction scripts. I can provide it if useful.

pengfeixx:
## Summary
- Added a nesting depth limit (`MAX_NEST_LEVEL = 1024`) to the C-based parser to prevent stack overflow when processing deeply nested braces
- When the limit is exceeded, the parser skips to the matching brace (`skipToMatch("{}")`) instead of recursing
- This prevents SIGSEGV crashes on crafted or extremely nested source files (e.g. Java files with thousands of nested blocks)

## Changes
| File | Change |
|------|--------|
| `parsers/c-based.c` | Added `#define MAX_NEST_LEVEL 1024`; added depth check in `createTags()` before recursive `nest()` call |

## Root Cause
The `createTags()` → `nest()` → `createTags()` recursive call chain has no depth bound. When processing a file with deeply nested braces (e.g. 54000 levels), this causes uncontrolled stack growth leading to SIGSEGV.

## Test Plan
- Build ctags and run against the PoC file (`poc-java-stack-overflow-54000.java`) attached to the issue — should no longer crash
- Existing test suite should pass unchanged (1024 levels is well above any reasonable real-world nesting depth)
- Tested with `gcc -std=c99 -fsyntax-only` — no compilation errors

Fixes universal-ctags/ctags#4439

### DIFF
--- a/parsers/c-based.c
+++ b/parsers/c-based.c
@@ -51,6 +51,8 @@
 
 enum { NumTokens = 3 };
 
+#define MAX_NEST_LEVEL 1024
+
 typedef enum eException {
 	ExceptionNone, ExceptionEOF, ExceptionFormattingError,
 	ExceptionBraceFormattingError
@@ -3005,7 +3007,17 @@ static void createTags (const unsigned int nestLevel,
 		{
 			int corkIndex = tagCheck (st);
 			if (isType (token, TOKEN_BRACE_OPEN))
-				nest (st, nestLevel + 1);
+			{
+				if (nestLevel < MAX_NEST_LEVEL)
+					nest (st, nestLevel + 1);
+				else
+				{
+					verbose ("%s: nesting level too deep (%u) at line %lu, skipping\n",
+						 getInputFileName (), nestLevel, getInputLineNumber ());
+					skipToMatch ("{}");
+					setToken (st, TOKEN_BRACE_CLOSE);
+				}
+			}
 			checkStatementEnd (st, corkIndex);
 		}
 	}


masatake:
Thank you. I'm thinking about fixing the original issue via #4387.

PR #4422, opened 2026-05-08
JavaScript: fix bug in line termination logic (semicolon not required)

jafl:
JavaScript: ignore calls to prototype function; fix bug in line termination logic (semicolon not required)

Fixes #4421

PR #4443, opened 2026-06-11
HTML: extract h4, h5, and h6 headings

### LINKED ISSUE #4442: HTML Parser: Tags also for larger heading levels such as h4 and h5
I'd like to create a tags file for the HTML documentation of [GAP](https://docs.gap-system.org/doc/ref/chap0_mj.html).

Example file: https://docs.gap-system.org/doc/ref/chap2_mj.html

- Chapters are heading level 3 (tags are generated by the current ctags HTML parser).
- Sections are heading level 4
- functions/commands are heading level 5

Having only the chapters as tags is here not very helpful.

As far as I see, h1, h2, h3 are hardcoded in the HTML parser:
https://github.com/universal-ctags/ctags/blob/0361a1da51a9cb680732451ab0c35c41bc914bda/parsers/html.c#L126-L128

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements mentions six levels.

Can levels 4, 5, 6 be added or a customization be added until which heading level the user like to have tags?

masatake:
~Close #4422~

Close #4442. (Edited)


masatake:
Enter a selection (default=all): error: target not found: mingw-w64-i686-jansson

masatake:
https://stackoverflow.com/questions/42741462/target-not-found-mingw-w64-x86-64-flickcurl

masatake:
https://qiita.com/k-takata/items/373ec7f23d5d7541f982

masatake:
It seems that Jansson has been dropped from Mingw32.
https://github.com/msys2/MINGW-packages/commit/1edc149ba5be8bffc77f3ec546f5d214d10781ae#diff-802a4628e63170a6c2e927e9351f67ea2f67fd8c24c116ebdcdb3a44a13cafc5

PR #4459, opened 2026-07-21
sandbox: fix seccomp sandbox with musl libc

Duncaen:
musl uses readv and writev for stdio read and writes, so allow them like normal read and write.

musl also checks whether stdout is a tty when writing to stdout, for that it uses ioctl with TIOCGWINSZ.

https://git.musl-libc.org/cgit/musl/commit/?id=2de85a985654d2c944931267645d9a0686242dfe

### DIFF
--- a/docs/news/HEAD.rst
+++ b/docs/news/HEAD.rst
@@ -12,6 +12,10 @@ Critical bug fixes
 * fixes a bug where a guest parser with a regex rule using postrun flag
   could enter an infinite loop.
 
+Common enhancement
+---------------------------------------------------------------------
+* Makes ``--_interactive=sandbox`` option work with musl libc.
+
 New and extended options and their flags
 ---------------------------------------------------------------------
 
--- a/main/seccomp.c
+++ b/main/seccomp.c
@@ -15,6 +15,7 @@
 
 #ifdef HAVE_SECCOMP
 #include <seccomp.h>
+#include <sys/ioctl.h>
 
 
 int installSyscallFilter (void)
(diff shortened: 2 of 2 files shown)


masatake (approved):

masatake:
Could you update `docs/news/HEAD.rst` ? like:
```
diff --git a/docs/news/HEAD.rst b/docs/news/HEAD.rst
index 01ebee519..f1968dc68 100644
--- a/docs/news/HEAD.rst
+++ b/docs/news/HEAD.rst
@@ -12,6 +12,10 @@ Critical bug fixes
 * fixes a bug where a guest parser with a regex rule using postrun flag
   could enter an infinite loop.
 
+Common enhancement
+* Makes ``--_interactive=sandbox`` option work with musl libc.
+
 New and extended options and their flags
 
```
Please change the sentence as you want.

leleliu008:
rebasing the master branch will fix the OpenBSD workflows issue.

masatake:
Thank you.
I'll merge this in soon.

What I came up with:

At least, Fedora 44 has musl-related packages:
```
musl-filesystem-1.2.5-6.fc44.x86_64
musl-libc-1.2.5-6.fc44.x86_64
musl-devel-1.2.5-6.fc44.x86_64
musl-gcc-1.2.5-6.fc44.x86_64
musl-libc-static-1.2.5-6.fc44.x86_64
```
So we can run a test for the change proposed in this pull request on CircleCI, where we run tests for Fedora.

I will not open an issue for this idea. The issues are already overflowing.

Duncaen:
Enabling seccomp with Alpine Linux would probably be simpler, since fedora does not ship the dependencies linked against musl libc.

Duncaen:
I applied the patch yesterday in Void Linux, where the tests with this patch ran in this PR:

https://github.com/void-linux/void-packages/actions/runs/29872724414/job/88776331872?pr=61574#step:7:802

masatake:
Thank you.

PR #4465, opened 2026-07-23
Misc fixes

brighamcampbell:
This is an assortment of miscellaneous fixes. I have verified that the code compiles and that unit test results appear to be nominal.

Please see git log for details. You may prefer to review this PR one commit at a time. Feel free to drop commits, request that I make changes, etc...

Cheers,
Brigham


docs/Makefile
```diff
@@ -18,7 +18,7 @@ I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 .PHONY: help clean check-if-sphinx-build-is-installed html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest gettext
 
 help:
-	@echo "\
+	@echo -e "\
```
masatake: Do you know any real platform requiring `-e` ?

This file is generated by Python Sphinx. See the beginning of this file. So I don't want to touch this file unless there is a serious need.

First, we need to consider the portability of `-e`.  So putting `@echo` at the beginning of all message lines may be better. We did done in makefiles/help.mak.

docs/news/6-2-0.rst
```diff
@@ -234,7 +234,7 @@ Merged pull requests
 * Verilog: Fix function parse when return type contains `::` by roccomao · Pull Request #4111
 * Verilog: Skip the escaped characters in string by roccomao · Pull Request #4115
 * verilog.c: keyword "unsigned1" should be "unsigned", without the 1 by cousteaulecommandant · Pull Request #4110
-* units.py: pass `count' as a keyword argument by masatake · Pull Request #4112
+* units.py: pass \`count' as a keyword argument by masatake · Pull Request #4112
```
masatake, line 236: What I will write here is just a comment. Not for asking you to take an action.

This text block is generated by misc/news.bash using `$(git log)` as source. So ideally we have to fix misc/news.bash.

docs/Makefile
brighamcampbell: I do: my own environment requires '-e' and will munge the output, printing it as a single line otherwise. I'm using gnu make and bash. If it matters, I can send the specific versions.

Does the behavior in your environment differ? I understand my own environment to be fairly typical...

If this Makefile is generated, I would almost rather implement the fix in the generator's code (taking portability into account, of course).

brighamcampbell: How is this Makefile generated? The comment at the beginning of the file doesn't seem to indicate that it's generated, just that the Makefile is for building the Sphinx documentation.

masatake: Sorry, what I wrote was a misunderstanding.

It was generated by Sphinx (20ea59859ebcf445fb02d720de0e15e8f72ec8d2).
Per line, `@` was removed in 33d0cc3d8ec9644cde43ef4520809ea85780d70d.

Just putting `-e` is a simple solution. I will accept this.
If someone mentions portability in the future, I will say, "A pull request is welcome."

leleliu008: `echo -e` is not portable, it's better to use `printf` instead.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html

I never use `echo`, I always rewite it as following:

```bash
echo() {
    printf '%b\n' "$*"
}
```

brighamcampbell: > `echo -e` is not portable, it's better to use `printf` instead.

Thank you for informing me. Rather than changing the Makefile to use `printf`, I changed it to an updated version of the boilerplate Makefile which is generated by Sphinx. This updated Makefile is simpler and outputs `make help` correctly without garbling the output. See commit 05a8061a5 ("Fix formatting of `make help` in docs/") which I just added to this PR.

From commit 33d0cc3d8:

> I removed many lines `@echo`, becuase per line shell code runs in a seperate shell.
> I removed them let them run in one shell, it can improve performance.

This reasoning is misguided. There's no need for `make help` to be more performant, particularly because the performance improvements break the output.

leleliu008:
@brighamcampbell By the way, just curious, do you use https://github.com/crate-ci/typos to find these typos?

masatake:
@brighamcampbell Thank you for updating news.bash.

brighamcampbell:
> @brighamcampbell By the way, just curious, do you use https://github.com/crate-ci/typos to find these typos?

No, I used codespell[^1]. From the project's README: "It does not check for word membership in a complete dictionary, but instead looks for a set of common misspellings. Therefore it should catch errors like "adn", but it will not catch "adnasdfasdf"."

I prefer codespell because it generates few false-positives.

[^1]: https://github.com/codespell-project/codespell

brighamcampbell:
> @brighamcampbell Thank you for updating news.bash.

Happy to help. :) These small fixes give me an opportunity to get to know the code better. I appreciate your thoughtful review.

leleliu008:
@masatake I previously used https://github.com/crate-ci/typos to check for typos in this project, and it found quite a few. I didn't want to bother you since I know you're very busy, but I was wondering — would you be open to adding a GitHub Actions workflow that automatically reports typos? I've set up similar checks (like typos-checker and markdown-link-check) in many of my other projects, and they've been quite helpful.

I have added these two checkers in https://github.com/universal-ctags/ctags-nightly-build

https://github.com/universal-ctags/ctags-nightly-build/blob/main/.github/workflows/typos-checker.yml
https://github.com/universal-ctags/ctags-nightly-build/blob/main/.github/workflows/markdownlint.yml

brighamcampbell:
> I have added these two checkers in https://github.com/universal-ctags/ctags-nightly-build

Please be aware that my pull request only fixes typos in `main/`, `man/`, and `docs/`, so a GitHub actions workflow which flagged typos would likely report a lot of typos, even if this PR is merged. I figured it would take less effort to review a PR which fixes a smaller number of typos and that typos in these three directories are most likely to be visible to users.

masatake:
@brighamcampbell Thank you.

masatake:
@leleliu008 Adding actions is welcome. I would like to see what happens.

PR #4474, opened 2026-07-30
buildsys: docs: restore clean target

leleliu008:
docs clean target was removed in #4465

We expect that clean target will remove `_build/` `_ext/__pycache__` directories. `sphinx-build -M clean` command won't do this.

If users don't want to generate docs, it will fail while executing `make clean` if `sphinx-build` is not installed.

It is unreasonable to force users to install `Sphinx`, as they may have only executed `make clean`.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -7,12 +7,16 @@ SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
+EXTDIR        = _ext
 
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+.PHONY: help clean Makefile
+
+clean:
+	rm -rf $(BUILDDIR) $(EXTDIR)/__pycache__
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).


masatake:
Looks good. However, I want to make changes to the docs/Makefile minimal.

I found a way to externalize your change:
```diff
diff --git a/docs/Makefile b/docs/Makefile
index d4bb2cbb9..ba3d446ce 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -18,3 +18,5 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
    @$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+include ext.mak
diff --git a/docs/ext.mak b/docs/ext.mak
new file mode 100644
index 000000000..69a29f5d3
--- /dev/null
+++ b/docs/ext.mak
@@ -0,0 +1,8 @@
+.PHONY: ext.mak
+
+EXTDIR        = _ext
+
+.PHONY: clean
+
+clean:
+   rm -rf $(BUILDDIR) $(EXTDIR)/__pycache__
```

brighamcampbell:
> We expect that clean target will remove `_build/` `_ext/__pycache__` directories. `sphinx-build -M clean` command won't do this.

`_build/` is ignored in the .gitignore. `.pyc` files in the `_ext/__pycache__` directory are also ignored in the .gitignore. I don't believe these build artifacts will clutter up the git repository. Are they causing issues for packaging and distribution?

> If users don't want to generate docs, it will fail while executing `make clean` if `sphinx-build` is not installed.
>
> It is unreasonable to force users to install `Sphinx`, as they may have only executed `make clean`.

Why would a user execute `make clean` if they don't have sphinx installed? If sphinx is not installed, the user cannot make the `docs/` directory dirty.

PR #4481, opened 2026-08-16
Preserve constness of variables passed to libc

brighamcampbell:
Historically, some libc functions like strchr have dropped the "const" type qualifier. There are also functions within ctags which, like libc, drop const (see strSeparator in main/routines.c).

Dropping qualifiers allows these functions to generically accept arguments with different qualifications and return a value with no qualification. The caller is expected to carefully manage qualification and cast when necessary.

C11 and C23 address this shortcoming by allowing for _Generic definitions which can be used to preserve qualifiers. glibc 2.43 began adopting generics in a backwards-compatible way, allowing gcc to emit warnings when some libc function calls cause type qualification to be lost (see commit cd748a63ab ("Implement C23 const-preserving standard library macros") in glibc).

Where compiling against glibc >= 2.43 causes warnings to be emitted, fix the loss of constness.

TODO: Use _Generic to make strSeparator and others preserve constness

### DIFF
--- a/dsl/optscript.c
+++ b/dsl/optscript.c
@@ -3409,7 +3409,7 @@ op__strpbrk (OptVM *vm, EsObject *name)
 	vString *acceptv = es_pointer_get (acceptobj);
 
 	const char *str = vStringValue (strv);
-	char *p = strpbrk (str, vStringValue (acceptv));
+	const char *p = strpbrk (str, vStringValue (acceptv));
 	if (p)
 	{
 		int d = p - str;
--- a/main/flags.c
+++ b/main/flags.c
@@ -40,7 +40,7 @@ extern const char *flagsEval (const char* flags_original, flagDefinition* defs, 
 		}
 		else if (flags [i] == LONG_FLAGS_OPEN)
 		{
-			const char* aflag = flags + i + 1;
+			char* aflag = flags + i + 1;
 			char* needle_close_paren = strchr(aflag, LONG_FLAGS_CLOSE);
 			const char* param;
 			char* needle_equal;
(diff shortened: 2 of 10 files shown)


masatake:
Thank you.
I wonder whether we can switch to C11.  vs2013 doesn't support C11. So we need to replace VS2013 with VS2018.

brighamcampbell:
@masatake 

> Thank you.

Happy to help!

> I wonder whether we can switch to C11. vs2013 doesn't support C11. So we need to replace VS2013 with VS2018.

In this pull request, I suggested making `strSeparator` preserve constness by using C11's _Generic. I read through glibc code to see how they took advantage of _Generic to preserve const. It was impressive, but it wasn't particularly clean and it seemed like it would be challenging to read and maintain.

If we can employ _Generic to make strSeparator preserve constness in a way that isn't too difficult to understand, I think it may be worth it. If it's ugly and difficult to read, then I think it might make more sense to leave it as it is.

Would you like me to create a list of functions within ctags like strSeparator which could use _Generic and then create an experimental branch which uses C11 features to make them preserve const? I can do this before we decide whether or not to switch to C11.

masatake:
Thank you for offering. However, _Generic is not the only reason I want to switch to vs2018.
As far as I can remember, vs2013 doesn't support C99 compound literals.

https://stackoverflow.com/questions/27826409/what-is-the-official-status-of-c99-support-in-vs2013

Once we can switch to vs2018, we can introduce _Generic based implementation incrementally.

brighamcampbell:
@masatake

Ok, I understand. I'm interested in helping you update to C11. I have a Linux development environment where I can test updating automake to use C11 and I can install Visual Studio in a Windows virtual machine to regenerate the MSVC build support files and test the Windows build.

I'm unable to find definitive evidence online that VS2018 supports compound literals. Microsoft documentation suggests[^1] that VS2022 v17.8 has the best C11 support. I would experiment with VS2022 v17.8 to make sure that it supports compound literals before updating the build tooling.

Would you like me to begin the effort or would you prefer that someone else do it?

[^1]: https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170#c-standard-library-features-1

masatake:
Please go ahead!

PR #4482, opened 2026-08-17
Fix NULL being passed into %s format specifier

brighamcampbell:
When installing default file exclude patterns, the initOptions function passes NULL into processExcludeOptionCommon which eventually passes the parameter to vfprintf, matching a '%s' format specifier.

Typically, this issue is recoverable and vfprintf will print "(null)", but depending on the C implementation, it may cause a segfault. Neither behavior is desirable. Fix this by passing "default exclude" in as the option.

This causes `ctags --verbose` to output the following:

```
  Installing default exclude patterns:
    adding default exclude pattern: {arch}
    adding default exclude pattern: .arch-ids
    adding default exclude pattern: .arch-inventory
    ...
```

Whereas before it output the following:

```
  Installing default exclude patterns:
    adding (null) pattern: {arch}
    adding (null) pattern: .arch-ids
    adding (null) pattern: .arch-inventory
    ...
```

This issue was identified by gcc 16.1.1 via `-Wformat-overflow`.

### DIFF
--- a/main/options.c
+++ b/main/options.c
@@ -4063,6 +4063,8 @@ extern void initOptions (void)
 	OptionFiles = stringListNew ();
 	OptlibPathList = optlibPathListNew (preload_path_list);
 
+	const char *defaultEx = "default exclude";
+
 	verbose ("Setting option defaults\n");
 	installHeaderListDefaults ();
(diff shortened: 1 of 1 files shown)


masatake:
Thank you.

PR #4487, opened 2026-08-23
buildsys: use c99 instead of gnu99

masatake:
This is a preparation pull request for switching to C11.


masatake:
I must remove AC_PROG_CC_C99, too. I will make another pull request.

PR #2893, opened 2021-03-01
Fix for #2885

kloczek:
This PR fixes #2885.
It also removes using AC_PROG_CC_C99 as autoupdate from autoconf 2.71 suggests.

PR #4485, opened 2026-08-19
Switch to C11

brighamcampbell:
@masatake, @leleliu008, this PR is now ready for review.

Update the automake and Windows build systems to use C11. This requires updating to Visual Studio 2022.

I tested by building on both Linux and Windows. Both systems build successfully. Linux also runs the test suite successfully. After merging these changes, we should be able to take advantage of C11 features which are supported by both VS 2022 and gcc.

If you know of a way that I can reproduce the GitHub Actions pipeline steps on my own fork, please let me know. I would prefer not to clutter up a PR with repeated invocations of `git push --force`, as I've done here.

PR #4488, opened 2026-08-24
Remove using AC_PROG_CC_C99

masatake:
Suggested in https://github.com/universal-ctags/ctags/pull/2893 by Tomasz Kłoczko.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,7 +6,7 @@
 
 #	Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.64])
+AC_PREREQ([2.70])
 
 dnl You must update PROGRAM_VERSION in main/ctags.h, too.
 dnl --version option of readtags also prints this.
@@ -259,7 +259,6 @@ AM_CONDITIONAL(ENABLE_DEBUGGING, [test "x$enable_debugging" = "xyes"])
 # -------------------
 
 AC_PROG_CC
-AC_PROG_CC_C99
 AC_C_INLINE
 # For gnulib.
 #  Typically, this is immediately after AC_PROG_CC, ...


brighamcampbell:
By the way, I have the same change queued in #4485

masatake:
> By the way, I have the same change queued in https://github.com/universal-ctags/ctags/pull/4485

Sorry, I didn't recognize it. However, I want to make #4485 smaller.
So I will merge this one first.

@leleliu008 Do you have any comments about this PR? 
"Cross compiler for ..." tests have passed. So I think this PR is acceptable.

leleliu008:
@masatake 

`AC_PROG_CC_C99` is obsolete starting from `autoconf-2.70`,  it will report a warning:

```autoconf
AU_DEFUN([AC_PROG_CC_C99],
  [AC_REQUIRE([AC_PROG_CC])],
  [$0 is obsolete; use AC_PROG_CC]
)
```

our configure.ac has this line:

```autoconf
AC_PREREQ([2.64])
```

that means  `autoconf < 2.70` still need `AC_PROG_CC_C99` to check if the C compiler supports C99.

I don't think we should remove `AC_PROG_CC_C99`, because it just give a warning for users if user use `autoconf >= 2.70`

masatake:
What about specifying `AC_PREREQ([2.70])`?

leleliu008:
If we don't want to emmit a warnning, we can do this :

```
m4_version_prereq([2.70], [AC_PROG_CC], [AC_PROG_CC_C99])
```

leleliu008:
> What about specifying `AC_PREREQ([2.70])`?

autoconf-2.70 was released in 2020, this will force people installing autoconf manually on older systems.

leleliu008:
@masatake autotools are developer tools, IMO, end-users should not known them. I have a idea that writting a `build.sh`  to install all dependencies. we encourage user to use `build.sh` , if users don't use `build.sh`, they may need to install dependencies manually.

masatake:
My question led this discussion in the wrong direction.
As proposed in #4485, I want to use C11, which supports _Generic and compound literal expressions. To detect a C11 compiler, use AC_PROG_CC bundled to `autconf >= 2.70`.

PR #4496, opened 2026-09-03
docs(man): Fix a spelling typo

standby24x7:
represetns->represents


masatake:
Could you change the header of the commit log to `docs(man): fix a spelling typo`?
ctags has a parser for man pages. So, `man` in the commit header implies the change is for the parser.

PR #4498, opened 2026-09-07
main: don't take the address of a member of a NULL pointer in foreach…

### LINKED ISSUE #4497: Null-pointer dereference in JavaScript parser (`foreachEntriesInScope`, entry.c:1408)
# Null-pointer dereference in JavaScript parser (`foreachEntriesInScope`, entry.c:1408)

## Description

ctags crashes with a null-pointer dereference when parsing a crafted JavaScript file. The crash occurs in `foreachEntriesInScope` at `main/entry.c:1408` — a null `tagEntryInfoX*` pointer is accessed without a null check.

UBSan output:
```
main/entry.c:1408:24: runtime error: member access within null pointer of type 'tagEntryInfoX' (aka 'struct sTagEntryInfoX')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior main/entry.c:1408:24
```

## Steps to reproduce

1. Save the attached `poc.js` file (~320 bytes)
2. Run: `ctags --sort=no --language-force=JavaScript -o - poc.js`
3. ctags exits non-zero with the UBSan diagnostic above (if built with `-fsanitize=undefined`)

ctags auto-detects JavaScript from the `.js` extension, so `--language-force` is optional.

## Environment

- ctags version: v6.2.0 (current `main`)
- OS: Linux x86_64
- Compiler: clang with `-fsanitize=undefined -fno-sanitize=function,vptr` (UBSan, excluding known-benign function-pointer-type false positives)

## Analysis

The JavaScript parser's scope-resolution path reaches `foreachEntriesInScope` with a null entry pointer. The fix is a null check before the member access at `entry.c:1408`.

## Attachments

- `poc.js` — 320 bytes, SHA-256: `f6891777d02b5fb24d058c72a8b3f80ceaba3aec617c9d5bbe02c74cfebb0655`

Found by coverage-guided fuzzing.

[ctags_poc.js](https://github.com/user-attachments/files/31846561/ctags_poc.js)

prownd:
Fix #4497.

foreachEntriesInScope() fills `rep' only when NAME is given.  With a NULL NAME the iteration loop still evaluates `&rep->symnode', a member access on a NULL pointer (C11 6.5.3.2).  UBSan reports it as

    main/entry.c:1408:24: runtime error: member access within null
    pointer of type 'tagEntryInfoX'

`revisited_rep' is only consulted when NAME is non-NULL, so simply guard the comparison with `rep'.

Reproducer (from #4497):

    $ printf 'var obj = {x: 1};\nvar y = obj.x;\n' > valid.js
    $ ctags --sort=no -o - valid.js

`make units' (1622 tests) is passed, and the reproducer is clean under -fsanitize=address,undefined.

### DIFF
--- a/main/entry.c
+++ b/main/entry.c
@@ -1405,7 +1405,10 @@ extern bool foreachEntriesInScope (int corkIndex,
 			verbose ("symtbl[< ] %s->%p\n", name, &entry->slot);
 			if (!func (entry->corkIndex, &entry->slot, data))
 				return false;
-			if (cursor == &rep->symnode)
+			/* REP is filled only when NAME is given.  Don't take the
+			 * address of a member of a NULL pointer; that is undefined
+			 * behavior. */
+			if (rep && cursor == &rep->symnode)
 				revisited_rep = true;
 		}
 		else if (name)


masatake (approved):

masatake:
Thank you!

PR #4501, opened 2026-09-07
Fix spelling typos in comments, help strings and descriptions

prownd:
Found with codespell.  Only comments, documentation, help/diagnostic strings and one kind description are touched; no identifier is renamed.

The PkgConfig `var' kind description ("variabels" -> "variables") is user visible, so optlib/pkgConfig.c is regenerated from optlib/pkgConfig.ctags, and main/CommonPrelude.c from main/CommonPrelude.ps.  The R `source' role description is user visible too, so Tmain/list-roles.d/stdout-expected.txt is updated to match.

`make units' (1627 tests), `make tmain', `make tlib' and `make check-genfile' are passed.


masatake (approved):

masatake:
We assume the kind letters and kind names are parts of CLI.
However, their descriptions are not.

masatake:
Thank you.

